Issue #36, opened 2022-05-11
Migrate to an SQL Database


kof:
## Use cases
- As a Webstudio hosted service, I need a scalable and cheap alternative to MongoDB.
- As an OSS contributor to Webstudio, I want an easy way to install the Designer and work with it locally
- As a Webstudio user I want run webstudio locally (offline in some future).

## Options

- SQLite for a local app
- supabase (PostgreSQL) for hosted service, deno workers in the same cluster would be also 🔥
- planetscale (MySQL)
- D1 cloudflare
- Others?

## Todo

- [ ] weigh in pros and cons of each database
- [ ] do we care if its MySQL vs PostgreSQL when using prisma?
- [ ] how hard is it to stay compatible with any of them, including SQLite?
- [ ] who offers the best scalable pricing?
- [ ] who offers hosting at the edge?
- [ ] who works with workers well? Problem with workers is they will create a lot of socket connections. Postgress has pgbouncer what is planetscale using for MySQL to scale it?

SaraVieira:
I would at least for now go with Postgres for a couple of reasons

- Support in Prisma is slightly better 
- there's a lot of options to have these databases
- when using it with supabase we could also use their storage and auth and that would be a super huge deal to get into beta and allow people to upload fonts and images 
- i am not sure about workers but codesandbox used Postgres with live and we didn't have any issues whatsoever with it

kof:
What just dropped https://blog.cloudflare.com/introducing-d1/ 🔥

kof:
Given d1 is based on sqlite, maybe we should start with migrating to sqlite on any type of hosting for now and then start using d1 once available

SaraVieira:
On it! Will start the migation across this repo and the SDK one

SaraVieira:
Can you assign it to me?

SaraVieira:
This is taking longer than expected for two reasons:

- No other database supports JSON as a tye in the database so I am at least for now stringifying everything
- Types in mongo had to be transformed to new models and so small calls like these:
```
await prisma.breakpoints.create({ data });
  ```

Have to be transformed into a massive call like so:

```
  const all = breakpoints.map(
    async (breakpoint: Breakpoint) =>
      await prisma.breakpoints.update({
        where: { treeId },
        data: {
          values: {
            create: breakpoint,
          },
        },
      })
  );
```

[...]

PR #42, opened 2022-05-16
Start the work to move to SQLite

SaraVieira:
Very much a draft

To be merged with https://github.com/webstudio-is/webstudio-sdk/pull/2

closes https://github.com/webstudio-is/webstudio-designer/issues/36


app/shared/db/breakpoints.server.ts
```diff
@@ -28,12 +31,29 @@ export const create = async (
...
 ) => {
-  const data = {
+  await prisma.breakpoints.create({
+    data: {
+      treeId,
+    },
+  });
+
+  const all = breakpoints.map(async (breakpoint: Breakpoint) => {
+    new Promise((resolve) => setTimeout(resolve, 500));
```
SaraVieira, line 41: Without this it breaks on a timeout

From my digging the database locks when you make one request

app/shared/db/tree.server.ts
```diff
@@ -62,14 +62,19 @@ export const createRootInstance = (breakpoints: Array<Breakpoint>) => {
...
-  return (await prisma.tree.findUnique({
+  const data = (await prisma.tree.findUnique({
     where: { id: treeId },
-  })) as Tree;
+  })) as unknown as Tree;
+
+  return {
+    ...data,
+    root: JSON.parse(data.root as unknown as string),
+  };
```
SaraVieira, line 77: Next PR would fix this and give it types and a new migration

SaraVieira:
@kof Would love to have a call with you to see if this how it would look like in your computer and see if it works well there too

app/shared/db/breakpoints.server.ts
kof: makes sense since sqlite is runs with no network between, it expects each query come sequentially, lets try using https://www.prisma.io/docs/concepts/components/prisma-client/transactions

package.json
```diff
@@ -95,11 +102,15 @@
...
+    "relative-deps": "^1.0.7",
     "typescript": "^4.1.2"
   },
   "engines": {
     "node": ">=14"
   },
   "sideEffects": false,
-  "license": "MIT"
+  "license": "MIT",
+  "relativeDependencies": {
```
kof, line 113: is this also a new thing from this PR?

kof:
✅ First pass done

package.json
SaraVieira: Yes! Will add docs about relative dependencies and how it works today and put it in this MR

app/shared/db/breakpoints.server.ts
SaraVieira: this just worked, never used them but yeah it worked perfectly

app/shared/db/props.server.ts
```diff
@@ -40,9 +40,14 @@ export const clone = async ({
...
-    data,
-  });
+  const all = await prisma.$transaction(
+    data.map((prop) =>
+      prisma.instanceProps.create({
+        data: prop,
+      })
+    )
+  );
+  await Promise.all(all);
```
kof, line 50: shouldn't need Promise.all now

app/shared/db/breakpoints.server.ts
```diff
@@ -57,8 +79,18 @@ export const patch = async (
   const breakpoints = await load(treeId);
   if (breakpoints === null) return;
   const nextBreakpoints = applyPatches(breakpoints.values, patches);
-  await prisma.breakpoints.update({
-    where: { treeId },
-    data: { values: nextBreakpoints },
-  });
+  console.log(nextBreakpoints);
```
kof, line 82: log

app/shared/db/breakpoints.server.ts
```diff
@@ -28,12 +32,30 @@ export const create = async (
...
 ) => {
-  const data = {
+  await prisma.breakpoints.create({
+    data: {
+      treeId,
+    },
+  });
+
+  const all = await prisma.$transaction(
+    breakpoints.map((breakpoint: Breakpoint) =>
```
kof, line 42: might be a stupid question because I haven't used sqlite yet, but why  not put breakpoint values in the create call above?

PR #48, opened 2022-05-17
Add Github actions

SaraVieira:
Add GH aactions for:

- test
- lint
- typecsheck

We just need to not allow merges from non passing pipelines on the settings now cc @kof 

closes https://github.com/webstudio-is/webstudio-designer/pull/48


SaraVieira:
@kof can you check this option on the repo settings?

CleanShot 2022-05-17 at 21 59 45@2x

kof:
looks great 🎉

PR #49, opened 2022-05-21
Move database to be postgres

SaraVieira:
This PR does several things

- Updates the schema and mutations to use posgres instead of mongo
- Adds small script to always copy over latest version of schema
- Updates remix packages

closes https://github.com/webstudio-is/webstudio-designer/issues/36

As part of this work a migrations repo was created here: https://github.com/webstudio-is/migrations


api/index.js
```diff
@@ -1,5 +0,0 @@
-const { createRequestHandler } = require("@remix-run/vercel");
-
-module.exports = createRequestHandler({
-  build: require("./_build")
-});
```
kof, line 5: So we won't be able to deploy to vercel any more? Was it necessary?

SaraVieira: That's actually a very valid point, i can keep it. That and the vercel.json

app/shared/db/breakpoints.server.ts
```diff
@@ -13,9 +13,18 @@ export const load = async (treeId?: Tree["id"]) => {
...
     where: { treeId },
   });
+
+  if (breakpoint === null) {
+    throw new Error("Breakpoint not found");
+  }
+
+  return {
+    ...breakpoint,
+    values: JSON.parse(breakpoint?.values),
```
kof, line 26: since you now added the null check
```suggestion
    values: JSON.parse(breakpoint.values),
```

api/index.js
SaraVieira: But it will never work on vercel because of the database

app/shared/db/misc.server.ts
```diff
@@ -32,7 +32,7 @@ export const publish = async ({
     previousTreeId: project.devTreeId,
     nextTreeId: tree.id,
   });
-  const { prodTreeIdHistory } = project;
+  const prodTreeIdHistory = project.prodTreeIdHistory as unknown as string[];
```
kof, line 35: hmm, maybe we should refine the type closer to the source?

app/shared/db/props.server.ts
```diff
@@ -20,9 +20,11 @@ export const loadByProject = async (
 };
 
 export const loadByTreeId = async (treeId: Tree["id"]) => {
-  return await prisma.instanceProps.findMany({
+  const trees = await prisma.instanceProps.findMany({
```
kof, line 23: hmm these are not trees, those are instancePropsEntries or something

app/shared/db/tree.server.ts
```diff
@@ -60,16 +61,24 @@ export const createRootInstance = (breakpoints: Array<Breakpoint>) => {
   return createInstance(rootConfig);
 };
 
-export const create = async (root: Instance): Promise<Tree> => {
-  return (await prisma.tree.create({
-    data: { root },
-  })) as Tree;
+export const create = async (root: Instance): Promise<DbTrree> => {
```
kof, line 64: DbTrrrrrrrree

fly.toml
```diff
@@ -0,0 +1,43 @@
+# fly.toml file generated for webstudio-dev on 2022-05-21T21:31:32+02:00
+
+app = "webstudio-dev"
```
kof, line 3: just realizing that this is basically fixed, so we won't be able to have webstudio-prod or webstudio-staging ... how do we go about that?

package.json
```diff
@@ -4,15 +4,19 @@
...
-    "postinstall": "remix setup node",
+    "dev": "yarn setup && node -r dotenv/config node_modules/.bin/remix dev",
+    "start": "remix-serve build",
     "typecheck": "tsc -b",
     "lint": "eslint . --ext .ts,.tsx --max-warnings 0",
     "test": "jest",
+    "setup": "cp ./node_modules/@webstudio-is/sdk/prisma/schema.prisma prisma/ && yarn migrate && yarn generate",
     "checks": "yarn lint && yarn typecheck && yarn test",
-    "watch:sdk": "relative-deps watch"
+    "migrate": "yarn prisma migrate dev",
```
kof, line 14: did it need yarn prisma or yarn relative-deps? I remember we tried that but I recall inside scripts it can be just the command

kof: oh unless the script is being called from npm run and so this woiuldn't run through yarn :)

remix.config.js
```diff
@@ -1,10 +1,7 @@
 /**
- * @type {import('@remix-run/dev/config').AppConfig}
+ * @type {import('@remix-run/dev').AppConfig}
  */
 module.exports = {
```
kof, line 4: these were for vercel, right?

kof (changes requested):

api/index.js
kof: true, hmmm, that means to finish the migration I need to also update the DNS to fly etc, we should examine a bit if our new setup holds a decent spike of visitors if we release the beta and there will be a huge traffic spike

kof: going down during beta launch is the worst thing that can happen to a startup

kof: I would take a few most database heavy actions and generate workload, I remember from many years ago this tool was great, but maybe there is something better now https://github.com/wg/wrk

app/shared/db/misc.server.ts
SaraVieira: Fixed!

app/shared/db/props.server.ts
SaraVieira: Truuum fixed

fly.toml
SaraVieira: You can pass an `--app` flag to `fly deploy`

https://fly.io/docs/flyctl/deploy/

Renamed it to webstudio aswell

package.json
SaraVieira: It does not you are right, removed!

api/index.js
SaraVieira: But the idea is thatr we migrate now and we don't go into beta now

fly.toml
kof: but how would you create staging and dev apps? Do we need a different app name? Should it become an env variable?

api/index.js
kof: What if we go in beta in 5 month and we have to launch with this?

SaraVieira: But why would we only migrate in 5months? All that's missing is the seeds

kof: Nono, I mean lets assume we migrate now and this thing is incapable is serving enough user to survive beta spike, should we try to test it now?

SaraVieira:
Oh definitely, we should do some testing now

Can try that tool and also ask andras how we does it

kof:
Reminder: we talked about putting fly setup into a separate PR for later in case we want to test it again

app/canvas/canvas.tsx
```diff
@@ -41,6 +40,7 @@ import {
   useHandleBreakpoints,
   useInitializeBreakpoints,
 } from "./shared/breakpoints";
+import { ParsedProject } from "~/shared/db/project.server";
```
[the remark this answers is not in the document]
SaraVieira: Done!

app/canvas/shared/sync/use-sync.ts
```diff
@@ -1,7 +1,7 @@
 import useInterval from "react-use/lib/useInterval";
 import { sync } from "immerhin";
 import { enqueue } from "./queue";
-import { Project } from "@webstudio-is/sdk";
+import { Project } from "~/shared/db/project.server";
```
kof, line 4: same

app/routes/dashboard/index.tsx
```diff
@@ -19,7 +19,7 @@ export const action: ActionFunction = async ({ request }) => {
   const { userId, headers } = await ensureUserCookie(request);
   try {
     const project = await db.project.create({ title, userId });
-    return redirect(`${config.designerPath}/${project.id}`, { headers });
+    return redirect(`${config.designerPath}/${project?.id}`, { headers });
```
kof, line 22: I think project can't be empty here, at least from looking at the code and types I don't see how

kof: `await prisma.project.create(` doesn't return null based on types, not sure why parseProject assumes it can be null, is prisma types wrong there?

kof: maybe its the switch from mongodb to PostgreSQL that changed the types?

app/routes/rest/project/clone.$domain.ts
```diff
@@ -1,16 +1,17 @@
...
+import { Project } from "~/shared/db/project.server";
 
 const ensureProject = async ({
   userId,
   domain,
 }: {
   userId: User["id"];
   domain: string;
-}): Promise<Project> => {
+}): Promise<Project | null> => {
```
kof, line 14: please check this one

app/canvas/canvas.tsx
```diff
@@ -36,6 +35,7 @@ import {
   useHandleBreakpoints,
   useInitializeBreakpoints,
 } from "./shared/breakpoints";
+import { Project } from "~/shared/db/project.server";
```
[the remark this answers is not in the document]
SaraVieira: Done!

app/routes/rest/project/clone.$domain.ts
```diff
@@ -29,7 +30,7 @@ export const loader: LoaderFunction = async ({ request, params }) => {
   const { headers, userId } = await ensureUserCookie(request);
   try {
     const project = await ensureProject({ userId, domain: params.domain });
-    return redirect(`${config.designerPath}/${project.id}`, { headers });
+    return redirect(`${config.designerPath}/${project?.id}`, { headers });
```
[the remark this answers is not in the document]
SaraVieira: Done!

app/routes/dashboard/index.tsx
SaraVieira: It's because of findUnique and that can return null :/

app/routes/rest/project/clone.$domain.ts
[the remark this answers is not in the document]
SaraVieira: Same

app/routes/rest/project/clone.$domain.ts
SaraVieira: Fixed

PR #53, opened 2022-05-27
CSF in Properties

kof:
## Tasks

- [x] Get `control` from `argTypes` which is coming from the story file (using CSF format)
- [x] Support `defaultValue` and `required` attributes for props (always display them)
- [ ] Implement "reset to defaultValue" for props 

## Support for following types of controls

- [x] `text`
- [x] `number`
- [x] `radio`
- [x] `select`
- [ ] `multi-select`
- [x] `boolean`      
- [ ] `object` (maybe just keep it unsupported for now or show a plain textarea)
- [x] `range`
- [x] `file`
- [x] `check`
- [x] `color`
- [x] `inline-radio` - same as radio?
- [x] `inline-check` — same as check?
- [x] `date` — not sure why this is needed in the designer right now
    
Fallback is always a text field.


okonet:
@kof I moved the list to the description in the PR. I don't think it makes sense to support everything in this PR since there might be dependencies like for multi-select but everything else seems doable now.

app/designer/features/props-panel/props-panel.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+      },
+    },
+  },
+};
+
+const meta = componentsMeta["Button"];
+const argTypes = meta?.argTypes as Record<
+  string,
+  { control: { defaultValue?: unknown } }
+>; // TODO: Add type to argTypes
```
kof, line 71: lets use "@todo" for consistency

app/designer/features/props-panel/props-panel.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+      component: "Heading",
+      cssRules: [],
+      browserStyle: {},
+      props: {
+        id: "2",
+        instanceId: "2",
+        treeId: "1",
+        props: Object.entries(argTypes).map(([prop, value]) => {
+          return {
+            id: `${prop}`,
```
kof, line 86:
```suggestion
            id: prop,
```

app/designer/features/props-panel/props-panel.tsx
```diff
@@ -1,26 +1,42 @@
...
+  component,
+  onChange,
+  onDelete,
+}: PropertyProps) => {
+  const meta = componentsMeta[component];
+  const argType = meta?.argTypes?.[prop as keyof typeof meta.argTypes];
+  const type = argType?.control.type;
+  const defaultValue = argType?.control.defaultValue;
+  const options = argType?.options;
+  // TODO: We need to sync defaultValue with value if value is not set and publish this change
```
kof, line 34:
```suggestion
  // @todo: We need to sync defaultValue with value if value is not set and publish this change
```

app/designer/features/props-panel/props-panel.stories.tsx
kof, line 71:
```suggestion
>; // @todo: Add type to argTypes
```

app/designer/features/props-panel/use-props-logic.test.ts
```diff
@@ -0,0 +1,265 @@
...
+    );
+    expect(result.current.userProps.length).toEqual(1);
+    expect(result.current.userProps[0]).toMatchObject({
+      prop: "href",
+      value: "",
+      required: true,
+    });
+  });
+
+  test.only("should return different default props for different instances", () => {
```
kof, line 40: forgotten .only, need to add a lint rule for this

app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -0,0 +1,169 @@
...
+    .filter(([_, value]) => value.required)
+    .map(([prop, _]) => ({
+      id: ObjectId().toString(),
+      prop,
+      value: "",
+      required: true,
+    }));
+};
+
+// TODO: This returns same props for all instances.
```
kof, line 37:
```suggestion
// @todo: This returns same props for all instances.
```

package.json
```diff
@@ -57,6 +57,7 @@
     "hyphenate-style-name": "^1.0.4",
     "immer": "^9.0.12",
     "immerhin": "^0.1.2",
+    "lodash": "^4.17.21",
```
kof, line 60: I have been adding specific functions only so that we have a kind of control over what we want to use from lodash, otherwise its easy to start using the entire lodash in places where we don't need it ... at least that's the thinking

app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -0,0 +1,169 @@
...
+  selectedInstanceData,
+  publish,
+}: UsePropsLogic) => {
+  const props =
+    selectedInstanceData.props === undefined ||
+    selectedInstanceData.props.props.length === 0
+      ? []
+      : selectedInstanceData.props.props;
+
+  const initialState = uniqBy(
```
kof, line 73: the amount of data bassed to this function is not making it a candidate for memoization?

jest.config.js
```diff
@@ -2,7 +2,17 @@ module.exports = {
   testEnvironment: "node",
   testMatch: ["<rootDir>/app/**/*.test.ts"],
   transform: {
-    "^.+\\.tsx?$": "esbuild-jest",
+    "^.+\\.tsx?$": [
+      "esbuild-jest",
+      {
+        // This is needed for inline snapshots to work
```
kof, line 8: ❤️

app/designer/features/props-panel/control.tsx
```diff
@@ -0,0 +1,186 @@
...
+      return <TextControl {...props} type="color" />;
+    case "date":
+      return <TextControl {...props} type="date" />;
+    case "number":
+      return <TextControl {...props} type="number" />;
+    case "range":
+      return <NotImplemented />;
+    case "object":
+      return <TextControl {...props} />;
+    case "radio":
```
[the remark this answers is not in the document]
okonet: Yeah I'm aware but I let co-pilot to write this switch case for me :)

app/designer/features/props-panel/use-props-logic.ts
okonet: No, this is a premature optimization at its best. The amount of data = max dozens of objects.

kof: dozens is fine, I was thinking this could be all of the props, but I guess in this case its only a few indeed

package.json
okonet: I noticed that and I think it's a maintenance nightmare. Instead, I think we should use specific imports or (better) use an optimization like this one: https://github.com/lodash/babel-plugin-lodash

app/designer/features/props-panel/use-props-logic.test.ts
okonet: removed

package.json
kof: does it work with esbuild?

kof: I feel like generally you don't need majority of lodash nowadays, a lot of it is just easier to understand with modern js

kof: importing specific packages gives you a thight control over things that you want to be used from lodash. Sometimes you want to replace a thing that lodash does with a different implementation, so having lodash as a dependency is kindof a box once opened hard to close, contributors may start using anything from it and you would have to tell them to just write a simple [].map or something

kof: but again if you really really insist, we can drop this idea and start using lodash, then wait until what I am describing becomes a problem and fix it by providing our own utils package that exports lodash functions where needed

okonet: I think an eslint rule could help to enforce this. We can explore other options with better tree shaking support but lodash is the most popular choice so I'd stick with it for the sake of contributors.

okonet: We can do a follow up. I'll install a specific package for now.

kof:
What's left to do in this PR? Can it be merged and iterated?

okonet:
I was hoping to finish the combo box work before this can be merged. But we can merge and improve later I think.

okonet:
Yep go for it after some final testing.

kof:
- [ ] Layout with value as an input field seems broken
Screenshot 2022-07-04 at 21 15 56

kof:
- [ ] Wrong props don't show as error or something and also don't let me enter value, one of these needs to be enabled I guess?

Screenshot 2022-07-04 at 21 19 08

kof:
- [ ] data- properties should always be enabled, right? Maybe except of the once we use in the system? e.g. data-component, data-id and a few more .... though maybe we just need to switch to something like data-webstudio-something to avoid conflicts
- [ ] 
Screenshot 2022-07-04 at 21 20 26

kof:
I would say we should make a follow-up issue and merge now to prevent future conflicts

Issue #39, opened 2022-05-11
User authentication


kof:
As a designer persona or as a developer persona I want to be able to log in to get access to my projects

## Tasks

- [ ] research a good way that is i locking us into some service and works well with workers-like environment e.g. https://github.com/sergiodxa/remix-auth

SaraVieira:
Decisions so far:

- Stack: https://github.com/sergiodxa/remix-auth

Order:

- Google Signin: https://github.com/pbteja1998/remix-auth-google
- Github SignIn: https://github.com/sergiodxa/remix-auth-github

SaraVieira:
Other related tasks:

- [ ]  Create database schema for users
- [ ]  Decide what data to get from users
- [ ] Create privacy policy

kof:
At this point we will also need the GDPR thing, right?

kof:
UI-wise 
- [ ] I would create a separate page that would show the login in the middle
- [ ] would automatically forward to /dashboard if user is already logged in
- [ ] /dashboard should automatically forward to /login if user is not logged in
- [ ] Optionally both /dashboard and /login could be the same route so we safe ourselves the redirects and just render conditionally from the same route?

SaraVieira:
Routes should do this server-side correct?

> At this point we will also need the GDPR thing, right?

I don't think so because we don't have tracking and the only cookies we have are for login. If we do tracking I would say https://plausible.io/ is a good option as it's GDPR compliant.

We do need a privacy policy to say what we have in terms of the projects and user data from GitHub

kof:
> Routes should do this server-side correct?

The idea behind Remix router is that it entirely automates this and makes it work with client-side navigation without full page reload if JS is enabled and works serverside if js is disabled or when user navigates directly to a specific page

kof:
> I don't think so because we don't have tracking and the only cookies we have are for login. If we do tracking I would say https://plausible.io/ is a good option as it's GDPR compliant.

We are saving an auth token into the cookie and we can potentially ask user their name, email. So at this point we are tracking the user to even provide them login, right? I could be wrong

SaraVieira:
I am also not sure but I think login cookies do not count for GDPR

kof:
Apparently we don't need the cookie banner as long as we keep the analytics/auth info in our servers, not share it with 3rd party https://github.blog/2020-12-17-no-cookie-for-you/

This is nice!

SaraVieira:
Tasks:

[...]

PR #62, opened 2022-06-08
Auth

SaraVieira:
Keeping track of progress:

- [x] Create user and login with GitHub
- [x] Create pretty login page
- [x] Create logout
- [ ] Login with Google
- [x] Block pages to unauthenticated users
- [x] Transfer old projects

Related #39


app/routes/auth/github/callback.tsx
```diff
@@ -0,0 +1,9 @@
+import { LoaderFunction } from "@remix-run/node";
+import { authenticator } from "~/services/auth.server";
+
+export const loader: LoaderFunction = ({ request }) => {
+  return authenticator.authenticate("github", request, {
+    successRedirect: "/dashboard",
```
kof, line 6: we got a config.ts in the root with paths

SaraVieira: I fixed this :)

SaraVieira:
Needs to be merged with: https://github.com/webstudio-is/webstudio-sdk/pull/5

app/routes/login/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { Login, links } from "~/auth/";
+
+export { links };
+
+export const meta: MetaFunction = () => {
+  return { title: "Webstudio Login" };
+};
+
+const LoginRoute = () => {
```
kof, line 17: I think we should reuse this document between designer

app/routes/logout.tsx
```diff
@@ -0,0 +1,10 @@
+import { ActionFunction } from "@remix-run/node";
+import { authenticator } from "~/services/auth.server";
+
+export default function Login() {
+  return "";
+}
+
+export const loader: ActionFunction = async ({ request }) => {
+  await authenticator.logout(request, { redirectTo: "/login" });
```
kof, line 9: loginPath from config

app/services/auth.server.ts
```diff
@@ -0,0 +1,49 @@
...
+import {
+  createOrLoginWithGithub,
+  createOrLoginWithGoogle,
+} from "~/shared/db/user.server";
+
+const url = `${
+  process.env.VERCEL_URL
+    ? `https://${process.env.VERCEL_URL}`
+    : "http://localhost:3000"
+}/auth`;
```
kof, line 15: path from config?

app/services/auth.server.ts
```diff
@@ -0,0 +1,49 @@
...
+import { sessionStorage } from "~/services/session.server";
+import {
+  createOrLoginWithGithub,
+  createOrLoginWithGoogle,
+} from "~/shared/db/user.server";
+
+const url = `${
+  process.env.VERCEL_URL
+    ? `https://${process.env.VERCEL_URL}`
+    : "http://localhost:3000"
```
kof, line 14: 3000 shouldn't be hardcoded, if we change it for the main server, it will need to automatically change here too

app/services/auth.server.ts
```diff
@@ -0,0 +1,49 @@
...
+}/auth`;
+
+const github = new GitHubStrategy(
+  {
+    clientID: process.env.GH_CLIENT_ID as string,
+    clientSecret: process.env.GH_CLIENT_SECRET as string,
+    callbackURL: `${url}/github/callback`,
+  },
+  async ({ profile, context }) => {
+    const user = createOrLoginWithGithub(profile, context?.userId);
```
kof, line 24: missing await? wonder how did it work if the function is async

app/services/auth.server.ts
```diff
@@ -0,0 +1,49 @@
...
+);
+
+const google = new GoogleStrategy(
+  {
+    clientID: "YOUR_CLIENT_ID",
+    clientSecret: "YOUR_CLIENT_SECRET",
+    callbackURL: `${url}/google/callback`,
+  },
+  async ({ profile }) => {
+    const user = createOrLoginWithGoogle(profile);
```
kof, line 37: same

app/services/auth.server.ts
kof: Oh I guess that works because you return a promise to an async function so it resolves it later, but still better to await here for clarity?

app/shared/db/project.server.ts
```diff
@@ -75,7 +75,16 @@ export const create = async ({
   const tree = await db.tree.create(db.tree.createRootInstance(breakpoints));
   const project = await prisma.project.create({
     data: {
-      userId,
+      User: {
+        connectOrCreate: {
+          create: {
+            id: userId as string,
```
kof, line 81: seems like a wrong thing to do, you are basically saying it can't be null, where the model says it can be null
Why can it be null on the model? If there is a valid reason - then we should bail out before

app/shared/db/project.server.ts
```diff
@@ -96,11 +105,19 @@ export const clone = async (clonableDomain: string): Promise<Project> => {
...
   const domain = generateDomain(clonableProject.title);
-
   const [project] = await Promise.all([
     prisma.project.create({
       data: {
-        userId: clonableProject.userId,
+        User: {
+          connectOrCreate: {
+            create: {
+              id: clonableProject.userId as string,
```
kof, line 114: same

kof: not part of this PR, but I just noticed the same thing below `parseProject(project as BaseProject) as Project;`
Instead we should treat the values as they are assumed by compiler instead of making compiler take it

app/shared/db/user.server.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    });
+
+    return connectedUser;
+  }
+  const user = await prisma.user.findUnique({
+    where: { email: profile._json.email },
+  });
+
+  if (user) {
```
kof, line 32: same

app/shared/db/user.server.ts
```diff
@@ -0,0 +1,69 @@
...
+    return user;
+  }
+
+  const newUser = prisma.user.create({
```
kof, line 36: await?

app/shared/db/user.server.ts
```diff
@@ -0,0 +1,69 @@
...
+): Promise<User> => {
+  const user = await prisma.user.findUnique({
+    where: { email: profile._json.email },
+  });
+
+  if (user) {
+    return user;
+  }
+
+  const newUser = prisma.user.create({
```
kof, line 59: await?

app/shared/db/user.server.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export const createOrLoginWithGoogle = async (
+  profile: GoogleProfile
+): Promise<User> => {
+  const user = await prisma.user.findUnique({
+    where: { email: profile._json.email },
+  });
+
+  if (user) {
```
kof, line 55: same

kof (changes requested):
Code review done 🎉

Going to test now

kof:
Getting this migration error, any idea?

```
Applying migration `20220616143541_add_projects`
Error: P3018

A migration failed to apply. New migrations cannot be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve

Migration name: 20220616143541_add_projects

Database error code: 23503

Database error:
ERROR: insert or update on table "Project" violates foreign key constraint "Project_userId_fkey"
DETAIL: Key (userId)=(629f787d56ac470b2c4fd8dd) is not present in table "User".

DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E23503), message: "insert or update on table \"Project\" violates foreign key constraint \"Project_userId_fkey\"", detail: Some("Key (userId)=(629f787d56ac470b2c4fd8dd) is not present in table \"User\"."), hint: None, position: None, where_: None, schema: Some("public"), table: Some("Project"), column: None, datatype: None, constraint: Some("Project_userId_fkey"), file: Some("ri_triggers.c"), line: Some(2539), routine: Some("ri_ReportViolation") }

```

kof:
started it second time, relinked the sdk, works

kof:
- [ ] can't login: click on login with github > authorize > redirected to login page (then try again and the same happens in a cycle)

app/routes/auth/github/callback.tsx
```diff
@@ -0,0 +1,15 @@
+import { LoaderFunction } from "@remix-run/node";
+import config from "~/config";
+import { authenticator } from "~/services/auth.server";
+import { ensureUserCookie } from "~/shared/session";
+
+export const loader: LoaderFunction = async ({ request }) => {
+  const userId = await ensureUserCookie(request);
```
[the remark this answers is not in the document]
SaraVieira: Got userId from it instead

app/config.ts
```diff
@@ -5,6 +5,8 @@ const config = {
   dashboardPath: "/dashboard",
   loginPath: "/login",
   logoutPath: "/logout",
+  googleCallback: "/google/callback",
```
kof, line 8:
```suggestion
  googleCallbackPath: "/google/callback",
```

app/config.ts
```diff
@@ -5,6 +5,8 @@ const config = {
...
+  githubCallback: "/github/callback",
```
kof, line 9:
```suggestion
  githubCallbackPath: "/github/callback",
```

app/services/auth.server.ts
```diff
@@ -11,17 +12,17 @@ import {
...
-    : "http://localhost:3000"
+    : `http://localhost:${process.env.PORT || 3000}`
 }/auth`;
 
 const github = new GitHubStrategy(
   {
     clientID: process.env.GH_CLIENT_ID as string,
     clientSecret: process.env.GH_CLIENT_SECRET as string,
-    callbackURL: `${url}/github/callback`,
+    callbackURL: `${url}${config.githubCallback}`,
```
kof, line 22: `path.join()`  is great for this prevents having accidental too little or too many slashes

kof (commented):
added a few more nits but looks good, going to test next

kof:
- [ ] the fact that I can click on avatar to logout is super hidden, it needs the chevron icon like github has to be clear
- [ ] when logging out, login page looks unstyled, reload then fixes it (can look into it if you want)

Screenshot 2022-06-18 at 19 51 02

app/routes/auth/github/callback.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return authenticator.authenticate("github", request, {
+    context: {
+      ...userId,
```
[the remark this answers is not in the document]
SaraVieira: done with userID

app/routes/logout.tsx
SaraVieira: done!

app/services/auth.server.ts
SaraVieira: Done!

SaraVieira: Done!

app/services/auth.server.ts
SaraVieira: Fixed!

app/services/auth.server.ts
SaraVieira: yap! Done!

app/shared/db/user.server.ts
```diff
@@ -0,0 +1,69 @@
...
+  profile: GitHubProfile,
+  userId: string | null
+): Promise<User> => {
+  const userData = {
+    email: profile._json.email,
+    username: profile.displayName,
+    image: profile._json.avatar_url,
+    provider: profile.provider,
+  };
+  if (userId) {
```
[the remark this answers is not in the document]
SaraVieira: This has actually been removed and its now different

app/shared/db/user.server.ts
SaraVieira: fixed

app/routes/login/index.tsx
SaraVieira: What do you mean?

app/shared/db/project.server.ts
SaraVieira: I removed it from the model, it now cannot be null on the model

app/shared/db/user.server.ts
SaraVieira: has also been removed

app/shared/db/user.server.ts
SaraVieira: Fixed

app/shared/design-system/components/menu.tsx
```diff
@@ -19,6 +19,7 @@ export const baseItemCss = css({
   whiteSpace: "nowrap",
   height: "$5",
   px: "$5",
+  cursor: "pointer",
```
kof, line 22: hmm, I think normally cursor is default, no?

app/shared/design-system/components/button.tsx
```diff
@@ -160,6 +161,13 @@ export const Button = styled("button", {
             "inset 0 0 0 1px hsla(0,100%,100%,.35), 0 0 0 1px hsla(0,100%,100%,.35)",
         },
       },
+      raw: {
```
kof, line 164: why not just using Avatar component from design system?

kof: it has interactive prop

kof:
Seems like it stopped working, redirects to the login page after login every time

SaraVieira:
You need to run yarn deps and clean up your database again

app/dashboard/dashboard.tsx
```diff
@@ -26,12 +34,18 @@ export const links = () => {
...
 };
 
+export const Avatar = styled("img", {
+  width: "$5",
+  borderRadius: "50%",
+});
+
 type DashboardProps = {
   projects?: Array<{ id: string; title: string }>;
+  user?: User;
```
kof, line 44: at this point user can't be optional or can it?

kof: I mean if there is no user we shouldn't be rendering dashboard

app/routes/auth/github.tsx
```diff
@@ -0,0 +1,15 @@
+import { ActionFunction } from "@remix-run/node";
+import config from "~/config";
+import { authenticator } from "~/services/auth.server";
+import { ensureUserCookie } from "~/shared/session";
+
+export const action: ActionFunction = async ({ request }) => {
+  const { userId } = await ensureUserCookie(request);
+  return authenticator.authenticate("github", request, {
```
kof, line 8:
```suggestion
  return await authenticator.authenticate("github", request, {
```

app/routes/auth/github/callback.tsx
```diff
@@ -0,0 +1,15 @@
+import { LoaderFunction } from "@remix-run/node";
+import config from "~/config";
+import { authenticator } from "~/services/auth.server";
+import { ensureUserCookie } from "~/shared/session";
+
+export const loader: LoaderFunction = async ({ request }) => {
+  const { userId } = await ensureUserCookie(request);
+  return authenticator.authenticate("github", request, {
```
kof, line 8:
```suggestion
  return await authenticator.authenticate("github", request, {
```

app/routes/login/index.tsx
kof: the html document you are sending here, is a copy/paste, we should reuse it, don't see differences

app/routes/logout.tsx
```diff
@@ -0,0 +1,11 @@
+import { ActionFunction } from "@remix-run/node";
+import config from "~/config";
+import { authenticator } from "~/services/auth.server";
+
+export default function Login() {
```
kof, line 5: I am not sure we need this at all (can't try rn), but if we do it would be Logout and return null

app/shared/db/user.server.ts
```diff
@@ -0,0 +1,65 @@
...
+const genericCreateAccount = async (
+  userData: {
+    email: string;
+    username: string;
+    image: string;
+    provider: string;
+  },
+  userId: string
+): Promise<User> => {
+  const userExists = await prisma.user.findUnique({
```
kof, line 15: userExists sounds like a boolean, but but its a user object

app/shared/design-system/components/button.tsx
```diff
@@ -3,6 +3,7 @@ import { styled } from "../stitches.config";
 export const Button = styled("button", {
   // Reset
   all: "unset",
+  cursor: "pointer",
```
kof, line 6: there is a bunch of resources suggesting cursor should be default on a button

kof: e.g. https://ux.stackexchange.com/a/20985

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Webstudio Designer UI as a service",
   "scripts": {
     "build": "remix build",
-    "dev": "yarn setup && node -r dotenv/config node_modules/.bin/remix dev",
+    "dev": "yarn setup && HOST=localhost node -r dotenv/config node_modules/.bin/remix dev",
```
kof, line 7: what's required HOST?

prisma/schema.prisma
```diff
@@ -10,6 +10,16 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
+model User {
+  id        String    @id @default(uuid())
+  email     String?   @unique
```
kof, line 15: can email be optional actually? I think there is always one, no matter if its custom login or auth?

kof (changes requested):
more review done, currently doesn't work for me - stuck in redirect, couldn't find the reason why

app/dashboard/dashboard.tsx
SaraVieira: Mistake, fixed!

app/auth/login.tsx
```diff
@@ -24,6 +25,8 @@ export const meta: MetaFunction = () => {
 };
 
 export const Login = () => {
+  const [devLoginClicked, setDevLoginClicked] = useState(false);
```
kof, line 28: is `isDevLogin` better?

app/routes/logout.tsx
SaraVieira: I think we do but fixed the name and return

app/dashboard/dashboard.tsx
```diff
@@ -73,11 +83,13 @@ export const Dashboard = ({ projects = [], config, user }: DashboardProps) => {
...
                 <Flex gap="1" align="center">
-                  {user?.image ? (
-                    <Avatar src={user?.image} alt="Open menu" />
-                  ) : (
-                    <UserIcon color="white" />
-                  )}
+                  <Avatar.Root>
+                    {user?.image && <Avatar.Image src={user?.image} />}
+                    <StyledFallback delayMs={500}>
+                      {(user?.username || "X").charAt(0).toLocaleUpperCase()}
```
kof, line 89: could use email as fallback for username if username is really optional and can't be enforced

app/services/auth.server.ts
```diff
@@ -48,3 +50,20 @@ if (process.env.GH_CLIENT_ID && process.env.GH_CLIENT_SECRET) {
   authenticator.use(github, "github");
   authenticator.use(google, "google");
 }
+
+if (process.env.NODE_ENV === "development") {
+  authenticator.use(
+    new FormStrategy(async ({ form }) => {
+      const secret = form.get("secret");
+      console.log(secret === process.env.AUTH_SECRET);
```
kof, line 58: ups

app/shared/db/user.server.ts
SaraVieira: truuuu, fixed!

app/routes/auth/github/callback.tsx
SaraVieira: Fixed

app/routes/auth/github.tsx
SaraVieira: Fixed

app/dashboard/dashboard.tsx
```diff
@@ -51,7 +50,7 @@ const StyledFallback = styled(Avatar.Fallback, {
 
 type DashboardProps = {
   projects?: Array<{ id: string; title: string }>;
-  user?: User;
+  user: User;
```
kof, line 53: if that's true, no need for optional checking for user in all accessors?

kof:
> You need to run yarn deps and clean up your database again
> […](#)

what is yarn deps? what's the wrong state we are in with the database? Why am I in and how do we prevent it or automate resolving it? We can't let this happen to others and block them

kof:
> You need to run yarn deps and clean up your database again

just removed all node_modules, reinstalled, run yarn watch:sdk,  yarn dev and now it generated a missing migration, I think it was supposed to be part of the pr, no?

kof:
hahaha, @SaraVieira I do like this picture too, but I think I am not THAT likable
Screenshot 2022-06-19 at 22 38 13

app/shared/design-system/components/button.tsx
SaraVieira: removed

prisma/schema.prisma
SaraVieira: it can because when you use without an account you don't have an email

package.json
SaraVieira: that was a experiemnt I forgot to remove

app/auth/login.tsx
SaraVieira: updated

app/dashboard/dashboard.tsx
SaraVieira: Done! Gets username, email and then does the x fallback

app/services/auth.server.ts
SaraVieira: Fixed

app/dashboard/dashboard.tsx
SaraVieira: hm?

SaraVieira:
Fixed a bunch of comments and also fixed the image issue

SaraVieira:
To login as dev you use whatever you placed as the 5de6e46209e81fd2a77e687b20bced666a1077015043e93d8e89c454f32393c3 in your env

app/routes/login/index.tsx
SaraVieira: Where should it go?

app/shared/design-system/components/button.tsx
SaraVieira: I think this button shuld exist either way but added the avatar component from radix instead

kof:
> To login as dev you use whatever you placed as the

Nice, lets add this to the docs

prisma/schema.prisma
kof: in dev mode and demo mode? maybe we should just have a placehoder and avoid checking everywhere if there is an email? Just an idea

kof:
> Nice, lets add this to the docs

Actually also can put a hint as a placeholder in that input!

app/dashboard/dashboard.tsx
kof: down the line a lot of defensive checking user?.something, which make it look like user is optional

kof: I think I looked at some wrong diff

kof:
- [ ] lets prevent header on dashboard jumping in height when reloading while the data is not there (avatar should be rendered all the time while data is loading, so that it holds the size of the header, alternatively head would need a fixed height, not sure what's better) 
- [ ] avatar component should be imported from the design system folder, we never actually import directly from radix, because design system is what makes radix components styled for us in a reusable form
- [ ] dashboard/dashboard.tsx has grown into a too big component and can be split into logical components to make it readable again

kof (approved):

Issue #52, opened 2022-05-26
CSS input with values/units autocomplete and drag


natlusyht:
The following is the spec for the input component that should have the general features and be able to support the extensions where needed.

# General

Should be able to input any arbitrary value with features that make enhance the experience.

## unit detection

>see chrome dev tools as an example

should show a arrow icon when the mouse is over a unit.

<img width="85" alt="Screenshot 2022-05-26 at 2 03 23 PM" src="https://user-images.githubusercontent.com/105234327/170475940-22ff3930-f5ea-469a-80d8-bcd9d0b486b3.png">

## number detection

>see chrome dev tools as an example

should show a [ew-resize cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) when hovering over a number.

![3-resize](https://user-images.githubusercontent.com/105234327/170476464-86e1b927-db1e-4b69-aa72-8da7e67c8909.gif)

> for both of the above we can either
> 1. destructive implementation: spans/divs within a rich "input". (this is what chrome dev-tools uses)
> 2. non-destructive implementation: use `measureText`, `style.cusror` and `mousemove` on the native input element to detect and decorate the input without the need for a rich "input". essentially using mouse x-coordinates relative to the input detect whether the mouse is over a unit/number from the current value of the input and update `style.cursor` based on that.

## Auto completion

should autocomplete valid property values as specified for that instance of the input.

<img width="241" alt="Screenshot 2022-05-26 at 2 27 34 PM" src="https://user-images.githubusercontent.com/105234327/170479254-a64e8d23-92e0-4c6e-a339-c9d392bcb897.png">

# Extensions

Should be able to support the following add-ons depending on the input instance

## ~~Auto-correct/delimited values~~

<details>
<summary>Archived</summary>
Should be able to add support for multiple delimited values and masked values such that i.e `10px10px` is `10px 10px` and `10px, 10px` is `10px 10px` essentially should be able to detect correct almost valid values to their valid counterparts and mask any denylisted characters to their counter parts i.e `comma` to `space` if specified.

an example of masking [imask.js.org](https://imask.js.org)

<img width="438" alt="Screenshot 2022-05-26 at 2 21 24 PM" src="https://user-images.githubusercontent.com/105234327/170478415-71c620ef-2c9f-4cf2-9cba-d4f706a1e6b2.png">
</details>

## ~~Expansion control~~

<details>
<summary>Archived</summary>
[...]

PR #64, opened 2022-06-12
auto-input component

ghost:
References https://github.com/webstudio-is/webstudio-designer/issues/52

- [x] detects numbers.
- [x] updates numbers through drag interaction or keyboard up/down interaction when the cursor/mouse is on the number.
- [x] shift up/down for 10x increment steps.
- [x] unit selection.
- [x] alt up/down for 0.1x increment steps.
~~- [ ] datalist extension.~~(another PR for a auto-datalist hook).

### DIFF
--- a/app/shared/design-system/components/primitives/auto-input.tsx
+++ b/app/shared/design-system/components/primitives/auto-input.tsx
@@ -14,7 +14,6 @@ export class AutoInput extends EventTarget {
 		// @todo: use mdn data
 		if (/^(%|cm|mm|in|px|pt|pc|em|ex|ch|rem|vw|vh|vmin|vmax|vb|vi|svw|svh|lvw|lvh|dvw|dvh|lh|rlh)$/.test(value)) return 'unit';
 		if (/^(\w+)$/.test(value)) return 'keyword';
-		if (/^(\w+\()/.test(value)) return 'function';
 		return '';
 	}
 	#measureEvent(value, event, target) {
@@ -42,11 +41,19 @@ export class AutoInput extends EventTarget {
 			return {tokenType: this.#typeOfValue(value), tokenValue: value}
 		});
 		for (const [tokenIndex, tokenObject] of valueTokens.entries()) {
-			const {tokenValue} = tokenObject;
+			const {tokenType, tokenValue} = tokenObject;
 			const tokenOffset = this.#canvasContext.measureText(tokenValue).width;
-			if ((eventOffset >= pixelOffset && eventOffset <= pixelOffset + tokenOffset)) {
-				valueOffset = tokenIndex;
-				break;
+			switch (tokenType) {
+				case 'unit': case 'number': {
+					const deltaStart = eventOffset - pixelOffset;
+					const deltaEnd = eventOffset - (pixelOffset + tokenOffset);
+					const rangeStart = -4;
+					const rangeEnd = 1;
+					if (deltaStart >= rangeStart && deltaEnd <= 1) {
+						valueOffset = tokenIndex;
+						break;
+					}
+				}
 			}
 			charsOffset += tokenValue.length;
 			pixelOffset += tokenOffset;
@@ -182,14 +189,10 @@ export class AutoInput extends EventTarget {
 			case 'keydown': {
 				switch (event.code) {
 					case 'ArrowUp': case 'ArrowDown': {
-						let {selectionStart, selectionEnd} = currentTarget;
-						if (selectionStart > selectionEnd) [selectionStart, selectionEnd] = [selectionEnd, selectionStart];
 						const {pixelPadding} = currentTarget.props;
 						const currentTargetValue = currentTarget.value;
-						const currentSelectionValue = currentTargetValue.substring(currentTarget.selectionStart, currentTarget.selectionEnd);
 						let offsetX = this.#canvasContext.measureText(currentTargetValue.substring(0, currentTarget.selectionStart)).width;
 						if (pixelPadding) offsetX += pixelPadding;
-						if (!isNaN(parseFloat(currentSelectionValue))) offsetX += 0.25;
 						let movementY = event.code == 'ArrowUp' ? -1 : 1;
 						if (event.shiftKey) movementY *= 10;
 						if (event.altKey) movementY *= 0.1;


app/shared/design-system/components/auto-input.tsx
```diff
@@ -0,0 +1,129 @@
+export class AutoInput extends EventTarget {
+	connectedCallback() {
+		if (!this.activeInput) this.activeInput = null;
+		if (!this.canvasContext) Object.getPrototypeOf(this).canvasContext = document?.createElement('canvas')?.getContext('2d');
+		for (let name of ['keydown', 'mouseup', 'mousedown', 'mousemove']) globalThis.addEventListener(name, this, false);
```
kof, line 5: afaik mousemove is a very performance sensitive event, because triggered a lot, you want to only bind it after mousedown and unbind it on mouseup

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
+export class AutoInput extends EventTarget {
```
kof, line 1: we need a better name for this, auto input is not very descriptive, maybe PointerInput?

app/designer/features/style-panel/render-property.tsx
```diff
@@ -197,7 +197,7 @@ const ComboboxControl = ({
             setValue(String(value.value), { isEphemeral: true });
           }}
         />
-        <Unit value={value} />
+        {/*<Unit value={value} />*/}
```
kof, line 200: can be removed including the definition?

app/shared/design-system/components/combobox.tsx
```diff
@@ -28,9 +28,10 @@ export const Combobox = ({
   const [isOpen, setIsOpen] = useState(false);
   const [currentValue, setCurrentValue] = useState(value);
 
-  useEffect(() => {
-    setCurrentValue(value);
-  }, [value]);
+  // this has a selection preservation and or double entry bug
```
kof, line 31: todo

app/shared/design-system/components/primitives/auto-datalist.tsx
```diff
@@ -0,0 +1,21 @@
+export class AutoDatalist extends EventTarget {
```
kof, line 1: needs better name as well

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+	#activeEvents = ['keydown', 'focusin', 'focusout', 'pointerover', 'pointerout', 'pointerup', 'pointerdown'];
+	#passiveEvents = ['pointermove'];
+	#resizeCursor = 'ns-resize';
+	#pointerCursor = 'pointer';
+	#pointerCapture = false;
+	#ownerDocument = globalThis.document;
+	#canvasContext = globalThis.document?.createElement('canvas').getContext('2d');
+	#typeOfValue(value) {
+		if (/^([-+]?\d*\.?\d+)$/.test(value)) return 'number';
+		if (/^(%|cm|mm|in|px|pt|pc|em|ex|ch|rem|vw|vh|vmin|vmax|vb|vi|svw|svh|lvw|lvh|dvw|dvh|lh|rlh)$/.test(value)) return 'unit';
```
kof, line 14: we have a units list generated from mdn data in SDK import { units } from "@webstudio-is/sdk";
once we move to a monorepo, we can put mdn-generated stuff into a separate package and let units be imported in isolation, but for now I would use it directly from SDK so we don't spread the duplication, moving later is trivial

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+		if (/^(\w+)$/.test(value)) return 'keyword';
```
kof, line 15: this is probably not gonna work for everything, but may be good enough for now, stuff like CSS functions e.g. var()

kof: would a add a todo at least for now

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+		if (/^(\w+\()/.test(value)) return 'function';
```
kof, line 16: haha just noticed this after commented above, still the question is if we want to parse CSS with regexes, the answer is pretty sure no, but I also understand the desire to keep it small, since we will need a CSS parser for for many places, we may as well start using some good dependency

app/shared/design-system/components/primitives/auto-input.tsx
kof: CSSPointerInput since its already specialized on css, unless you can decouple it

kof: Got an even better name: CSSValueGuestureInput

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+		return '';
+	}
+	#measureEvent(value, event, target) {
+		const {fontSize, fontFamily, fontWeight, textAlign, paddingLeft, paddingRight} = getComputedStyle(target);
+		this.#canvasContext.font = `${fontWeight} ${fontSize} ${fontFamily}`;
+		const eventOffset = event.offsetX;
+		let charsOffset = 0;
+		let pixelPadding = 0
+		switch (textAlign) {
```
kof, line 25: whats this handling? can't guess from reading

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+				break;
+			}
+			case 'right': {
+				pixelPadding += target.clientWidth - this.#canvasContext.measureText(value).width - parseFloat(paddingRight);
+				break;
+			}
+		}
+		let pixelOffset = pixelPadding;
+		let valueOffset = -1;
+		const valueTokens = value.split(/([*/,()\s]?)([-+]?\d*\.?\d+)|([*/,()\s]+)/).filter(value => value).map(value => {
```
kof, line 37: seems like you are still trying to handle multi-value input? Why bother with complexity if we don't have a use for it?

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+		return {valueOffset, valueTokens, charsOffset, pixelOffset, pixelPadding};
+	}
+	// PUBLIC
+	handleEvent(event) {
+		const eventTarget = event.target;
+		const currentTarget = this.#activeTarget || eventTarget;
+		if (currentTarget?.nodeName !== 'INPUT' || currentTarget?.type !== 'text' || currentTarget?.readOnly !== false) return;
+		switch (event.type) {
+			case 'pointerover': case 'focusin': {
+				if (this.#pointerCapture === false) this.#pointerCapture = !this.#passiveEvents.forEach(eventName => this.#ownerDocument.addEventListener(eventName, this, false));
```
kof, line 59: `!this.#passiveEvents.forEach`

am I missing something or is it basically `![]` and will always be false?

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+				break;
+			}
+			case 'pointerout': case 'focusout': {
+				if (this.#pointerCapture === true) this.#pointerCapture = !!this.#passiveEvents.forEach(eventName => this.#ownerDocument.removeEventListener(eventName, this, false));
```
kof, line 63: same question here

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+				const currentTargetCursor = currentTarget.style.cursor;
+				switch (currentTargetCursor) {
+					case this.#resizeCursor: {
+						this.#ownerDocument.documentElement.style.cursor = currentTargetCursor;
+						break;
+					}
+					case this.#pointerCursor: {
+						// dismiss if already active
+						if (this.#selectTarget) return this.#selectTarget.blur();
+						event.preventDefault?.();
```
kof, line 78: there is an `event` where preventDefault isn't a function or why the `?`?

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+						let {activeElement} = this.#ownerDocument;
+						// get cached measure metrics
+						const measureMetrics = currentTarget.props;
+						const {valueOffset, valueTokens, pixelOffset} = measureMetrics;
+						const activeToken = valueTokens[valueOffset];
+						const {tokenValue} = activeToken;
+						// create select dropdown
+						const selectTarget = this.#selectTarget = this.#ownerDocument.documentElement.appendChild(this.#ownerDocument.createElement('select'));
+						// TODO: inline for now
+						const unitList = '%,px,em,rem,ch,cm,mm,in,pt,vw,vh,vmin,vmax,svw,svh,lvw,lvh,dvw,dvh'.split(',');
```
kof, line 90: same as above, should be single source of truth

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,205 @@
...
+						// cache selection state
+						let {selectionStart, selectionEnd} = currentTarget;
+						let {activeElement} = this.#ownerDocument;
+						// get cached measure metrics
+						const measureMetrics = currentTarget.props;
+						const {valueOffset, valueTokens, pixelOffset} = measureMetrics;
+						const activeToken = valueTokens[valueOffset];
+						const {tokenValue} = activeToken;
+						// create select dropdown
+						const selectTarget = this.#selectTarget = this.#ownerDocument.documentElement.appendChild(this.#ownerDocument.createElement('select'));
```
kof, line 88: lets completely decouple select/dropdown from CssValueGuestureInput

app/shared/design-system/components/primitives/auto-input.tsx
kof: I am also wondering if we can get the abstraction right to have just an input with a number, not specific to CSS value and have the drag gesture applied, then use the same logic to create a CSS value put etc, reusable pieces

kof:
You probably want to split the input logic into
1. a thing that only knows how to bind events to any dom element and change a single numeric value based on guestures
2. a thing that only contains CSS-specific parsing logic
3. a thing that implements a datalist 
4. a thing that combines all above to make a CssValueGuestureInput

- All the above pieces should make it possible to use gesture control a plain numeric input.
- Last but not least would need a design-system-specific component wrapper that applies the theme using stitches like other  ds components components

kof (changes requested):
quite a bit of work to go, let me know if you want me to test it

kof:
btw we haven't added prettier to the checks, I can see there will be a lot of corrections by prettier, generally everything so far was autoformatted

kof:
added prettier to the `yarn checks` script so that even in case people don't run prettier locally, it still is getting checked on CI, please merge the main into your branch, then run yarn checks

app/shared/design-system/components/primitives/auto-input.tsx
kof: uhh, a lot of other things can impact the measuring, e.g. letter spacing

app/shared/design-system/components/primitives/auto-input.tsx
kof: ah right , css functions are multi-value already, shit ....

kof: please add inline docs to every such things so one doesn't have to guess what it does :)

app/shared/design-system/components/primitives/auto-input.tsx
kof: right, but why would you write it this way? it looks like its made so that someone can't read it, extra cryptic lol

app/shared/design-system/components/primitives/auto-input.tsx
kof: why not just this.#pointerCaputure = false, really this is what humans expect

app/shared/design-system/components/primitives/auto-input.tsx
kof: yeah, from reading code one doesn't expect the event to be a not event, very unexpected

app/shared/design-system/components/primitives/auto-input.tsx
kof: I mean if user defines custom letter spacing as style, same as font-family, it will impact the size of text, and I haven't seen you considering it, also letter spacing might be not the only one missing

kof:
Tested it a bit today, here is some feedback:

- [ ] drag is vertical, not horizontal?  I would say it should be by default for normal inputs horizontal, but for padding/margin it would depend on if its vertical or horizontal value, so it should be an option to go vertical or horizontal
- [ ] cursor changes right now in the empty space before and after the text, indicating actions, but actions only work above the specific text
- [ ] during the drag you want to render the changes real-time on the canvas (see webflow margin/padding interaction)
- [ ] we need to do something about a numeric value like (e.g. line-height) where unit is legit optional, there are 2 problems: 1 user should be able to remove unit 2. user should be able to select a unit from the list even if it was removed, so we probably need to represent the absence of a unit with something like (-) so that user can click on it or something else?

Can we please cut down the scope of the PR? It is trying to do way too much, maybe even open a separate PR that does just one thing and we can merge it. I strongly suggest to start with a number manipulation gesture  implementation that works standalone, knows nothing about units, CSS, dropdowns etc. So we can merge it.

The scope of this PR is way to big and I am afraid with the current speed we won't be able to merge it if we don't cut the scope, make abstractions small, isolated and mergeable.

app/shared/design-system/components/primitives/auto-input.tsx
```diff
@@ -0,0 +1,217 @@
...
+				switch (event.code) {
+					case 'ArrowUp': case 'ArrowDown': {
+						let {selectionStart, selectionEnd} = currentTarget;
+						if (selectionStart > selectionEnd) [selectionStart, selectionEnd] = [selectionEnd, selectionStart];
+						const {pixelPadding} = currentTarget.props;
+						const currentTargetValue = currentTarget.value;
+						const currentSelectionValue = currentTargetValue.substring(currentTarget.selectionStart, currentTarget.selectionEnd);
+						let offsetX = this.#canvasContext.measureText(currentTargetValue.substring(0, currentTarget.selectionStart)).width;
+						if (pixelPadding) offsetX += pixelPadding;
+						if (!isNaN(parseFloat(currentSelectionValue))) offsetX += 0.25;
```
kof, line 192: Very few people know: `!isNaN` is very hard to read, because of `!i`, instead a much better thing to do for readability is `if (isNaN() === false)`

kof:
we can close this one I guess?

PR #82, opened 2022-06-22
numericGestureControl

ghost:
extracted abstraction from the https://github.com/webstudio-is/webstudio-designer/pull/64 monolith.

### DIFF
--- a/app/shared/design-system/components/primitives/numericGestureControl.js
+++ b/app/shared/design-system/components/primitives/numericGestureControl.js
@@ -31,15 +31,8 @@ export const numericGestureControl = (
       direction === "horizontal" ? "0 1px" : "1px 0"
     } 1.1px rgba(0,0,0,.4))`,
   };
-  const handleEvent = ({
-    type,
-    target,
-    offsetX,
-    offsetY,
-    movementY,
-    movementX,
-    pressure,
-  }) => {
+  const handleEvent = (event) => {
+    const { type, pressure, offsetX, offsetY, movementY, movementX } = event;
     const offset = direction === "horizontal" ? offsetX : offsetY;
     const movement = direction === "horizontal" ? movementX : -movementY;
     switch (type) {
@@ -48,7 +41,7 @@ export const numericGestureControl = (
         break;
       }
       case "pointerdown": {
-        const handlePointerLockEvent = (event) => {
+        const handlePointerLockEvent = () => {
           const { pointerLockElement } = ownerDocument;
           switch (ownerDocument.pointerLockElement) {
             case targetNode: {
@@ -86,7 +79,7 @@ export const numericGestureControl = (
           state.offset += movement * state.velocity;
           state.cursor.style[state.position] = `${state.offset}px`;
           onValueChange({
-            target,
+            target: targetNode,
             value: state.value,
             preventDefault: () => event.preventDefault(),
           });


kof:
This looks good to me, lets add a stories file (setup already done), just add {file}.stories.tsx and we can merge it

kof (approved):
Tested, works great. 

Check out what happens in figma if you go with cursor to the edge of the window and keep going outside :)

Lets create an issue for this as an improvement, I guess its not trivial, or is it?

kof:
niiiiiice cc @okonet

kof:
> Can we add storybook builds to the "preview deploy"?

yeah, though we don't have a storybook published to vercel yet, so we would need to do that and it will get us automatically a preview, right?

app/shared/design-system/components/primitives/numericGestureControl.js
```diff
@@ -0,0 +1,112 @@
...
+ *   onValueChange: (event) => {
+ *     event.preventDefault();
+ *     event.target.value = event.value;
+ *     event.target.select();
+ *   }
+ * });
+ */
+export const numericGestureControl = (
+  targetNode,
+  { initialValue = 0, direction = "horizontal", onValueChange = ({}) => null }
```
okonet, line 18: I have a question why is this written in JS where all other code base is TS?

app/shared/design-system/components/primitives/numericGestureControl.js
```diff
@@ -0,0 +1,112 @@
...
+    value: initialValue,
+    cursor: cursorNode.cloneNode(true),
+    offset: 0,
+    velocity: direction === "horizontal" ? 1 : -1,
+    position: direction === "horizontal" ? "left" : "top",
+    rotation: direction === "horizontal" ? "rotate(0deg)" : "rotate(90deg)",
+    direction: direction === "horizontal" ? "ew-resize" : "ns-resize",
+    dropShadow: `drop-shadow(${
+      direction === "horizontal" ? "0 1px" : "1px 0"
+    } 1.1px rgba(0,0,0,.4))`,
```
okonet, line 32: IMO this the UI concern and should not be mixed with the gesture state itself. That being said, I think we should still calculate and return the styles from the hook but I would scope them under a separate object like `css` or `styles` to make it 1. clear there the separation is and 2. easier to get final styles using our design-system.

app/shared/design-system/components/primitives/numericGestureControl.stories.jsx
```diff
@@ -0,0 +1,39 @@
...
+      direction: direction,
+      onValueChange: (event) => {
+        event.preventDefault();
+        event.target.value = event.value;
+        event.target.select();
+      }
+    });
+    return () => disconnectedCallback();
+  }, [direction]);
+}
```
okonet, line 17: IMO this should be the export from the implementation. Why do we want to require developers to use a custom hook every time they want to use this gesture control?

By converting the `numericGestureControl` to a hook, you'll also solve potential issues with mutations and state updates, because ultimately that's what you're trying to express in the end.

app/shared/design-system/components/primitives/numericGestureControl.js
```diff
@@ -0,0 +1,112 @@
...
+        const handlePointerLockEvent = () => {
+          const { pointerLockElement } = ownerDocument;
+          switch (ownerDocument.pointerLockElement) {
+            case targetNode: {
+              state.offset = offset;
+              state.cursor.style.position = "absolute";
+              state.cursor.style.top = `${offsetY}px`;
+              state.cursor.style.left = `${offsetX}px`;
+              state.cursor.style.transform = state.rotation;
+              state.cursor.style.setProperty("--drop-shadow", state.dropShadow);
```
okonet, line 53: I feel uneasy looking at those state mutations here. If you wanted to expose styles, then it's okay to put them into the state but I would separate styles and state somehow. See my above comments.

But from what I'm reading here it doesn't even need to be exposed yet (which I'm not sure yet, since I can totally imagine how we'd like to customize styles later when we'll be implementing flex gap etc controls).

okonet (commented):
I know it's already merged but since I didn't have a chance to review, I still wanted to leave a few comments. We can address them in a follow up PR, so no rush.

app/shared/design-system/components/primitives/numericGestureControl.js
kof: agree

app/shared/design-system/components/primitives/numericGestureControl.stories.jsx
kof: we can have both, a hook and a function, might be useful

app/shared/design-system/components/primitives/numericGestureControl.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+const cursorNode = new Range().createContextualFragment(`
+  <svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="46" height="15" style="filter:var(--drop-shadow);">
+    <g transform="translate(2 3)">
+      <path d="M 15 4.5L 15 2L 11.5 5.5L 15 9L 15 6.5L 31 6.5L 31 9L 34.5 5.5L 31 2L 31 4.5Z" fill="#111" fill-rule="evenodd" stroke="#FFF" stroke-width="2"></path>
+      <path d="M 15 4.5L 15 2L 11.5 5.5L 15 9L 15 6.5L 31 6.5L 31 9L 34.5 5.5L 31 2L 31 4.5Z" fill="#111" fill-rule="evenodd"></path>
+    </g>
+  </svg>
+`).firstElementChild;
```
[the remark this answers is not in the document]
kof: I guess this is for simulated cursor when your real cursor goes outside of the window and you show a fake one coming from the other side

app/shared/design-system/components/primitives/numericGestureControl.js
kof: yeh, its only possible to write like that if you never expose this state to the outside, maybe worth adding a comment?

kof:
@natlusyht I don't know how we managed to have it not in ts :) and I missed that during review. We need to agree to use ts always

app/shared/design-system/components/primitives/numericGestureControl.js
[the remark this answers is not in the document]
kof: interesting, worth a comment!

PR #86, opened 2022-06-25
Add errors on auth & prisma fixes

SaraVieira:
This PR adds errors when somthing goes wrong in the login process and it also fixes the "There are already 10 instances of Prisma Client actively running" warning

Followed https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices

related #39

### DIFF
--- a/app/auth/login.tsx
+++ b/app/auth/login.tsx
@@ -1,7 +1,14 @@
 import { LinksFunction, MetaFunction } from "@remix-run/node";
 import { Form, useLoaderData } from "@remix-run/react";
 import { useState } from "react";
-import { Button, Card, Flex, Heading, TextField } from "~/shared/design-system";
+import {
+  Button,
+  Card,
+  Flex,
+  Heading,
+  Text,
+  TextField,
+} from "~/shared/design-system";
 import interStyles from "~/shared/font-faces/inter.css";
 
 import { GoogleIcon, GithubIcon, CommitIcon } from "~/shared/icons";
@@ -24,7 +31,7 @@ export const meta: MetaFunction = () => {
   return { title: "Webstudio Login" };
 };
 
-export const Login = () => {
+export const Login = ({ errorMessage }: { errorMessage: string | null }) => {
   const [isDevLogin, setIsDevLogin] = useState(false);
   const loaderData = useLoaderData();
 
@@ -35,10 +42,14 @@ export const Login = () => {
       align="center"
       justify="center"
     >
-      <Card css={{ width: "$10", padding: "$5", zoom: 1.4 }} variant="active">
+      <Card css={{ width: 200, padding: "$5", zoom: 1.4 }} variant="active">
         <Flex direction="column" gap="2" align="center">
           <Heading>Login</Heading>
-
+          {errorMessage && (
+            <Text css={{ textAlign: "center" }} variant="red">
+              {errorMessage}
+            </Text>
+          )}
           <Flex gap="2" direction="column" align="center">
             <Form action="/auth/github" method="post">
               <Button type="submit">
--- a/app/auth/login.tsx
+++ b/app/auth/login.tsx
@@ -67,7 +67,7 @@ export const Login = () => {
                       minLength={2}
                       required
                       autoFocus
-                      placeholder="Place your AUTH_SECRET"
+                      placeholder="Place your auth secret here"
                     />
                   </Form>
                 ) : (
--- a/app/auth/login.tsx
+++ b/app/auth/login.tsx
@@ -31,7 +31,7 @@ export const meta: MetaFunction = () => {
   return { title: "Webstudio Login" };
 };
 
-export const Login = ({ errorMessage }: { errorMessage: string | null }) => {
+export const Login = ({ errorMessage }: { errorMessage: string }) => {
   const [isDevLogin, setIsDevLogin] = useState(false);
   const loaderData = useLoaderData();
 
@@ -45,11 +45,11 @@ export const Login = ({ errorMessage }: { errorMessage: string | null }) => {
       <Card css={{ width: 200, padding: "$5", zoom: 1.4 }} variant="active">
         <Flex direction="column" gap="2" align="center">
           <Heading>Login</Heading>
-          {errorMessage && (
+          {errorMessage.length ? (
             <Text css={{ textAlign: "center" }} variant="red">
               {errorMessage}
             </Text>
-          )}
+          ) : null}
           <Flex gap="2" direction="column" align="center">
             <Form action="/auth/github" method="post">
               <Button type="submit">
--- a/app/routes/auth/github.tsx
+++ b/app/routes/auth/github.tsx
@@ -1,6 +1,7 @@
 import { ActionFunction, redirect } from "@remix-run/node";
 import config from "~/config";
 import { authenticator } from "~/services/auth.server";
+import { sentryException } from "~/shared/sentry";
 import { ensureUserCookie } from "~/shared/session";
 import { AUTH_PROVIDERS } from "~/shared/session/useLoginErrors";
 
@@ -22,7 +23,12 @@ export const action: ActionFunction = async ({ request }) => {
     // all redirects are basically errors and in that case we don't want to catch it
     if (error instanceof Response) return error;
     if (error instanceof Error) {
-      console.log(error);
+      sentryException({
+        message: error.message,
+        extra: {
+          loginMethod: "github",
+        },
+      });
       return redirect(
         `${config.loginPath}?error=${AUTH_PROVIDERS.LOGIN_GITHUB}&message=${
           error?.message || ""
--- a/app/routes/auth/github.tsx
+++ b/app/routes/auth/github.tsx
@@ -2,7 +2,7 @@ import { ActionFunction, redirect } from "@remix-run/node";
 import config from "~/config";
 import { authenticator } from "~/services/auth.server";
 import { ensureUserCookie } from "~/shared/session";
-import { LOGIN_ERROR_CODES } from "~/shared/session/useLoginErrors";
+import { AUTH_PROVIDERS } from "~/shared/session/useLoginErrors";
 
 export default function GH() {
   return null;
@@ -24,7 +24,7 @@ export const action: ActionFunction = async ({ request }) => {
     if (error instanceof Error) {
       console.log(error);
       return redirect(
-        `${config.loginPath}?error=${LOGIN_ERROR_CODES.LOGIN_GITHUB}&message=${
+        `${config.loginPath}?error=${AUTH_PROVIDERS.LOGIN_GITHUB}&message=${
           error?.message || ""
         }`
       );
--- a/app/routes/auth/github.tsx
+++ b/app/routes/auth/github.tsx
@@ -18,14 +18,16 @@ export const action: ActionFunction = async ({ request }) => {
       successRedirect: config.dashboardPath,
       throwOnError: true,
     });
-  } catch (error: any) {
+  } catch (error: unknown) {
     // all redirects are basically errors and in that case we don't want to catch it
     if (error instanceof Response) return error;
-    console.log(error);
-    return redirect(
-      `${config.loginPath}?error=${LOGIN_ERROR_CODES.LOGIN_GITHUB}&message=${
-        error?.message || ""
-      }`
-    );
+    if (error instanceof Error) {
+      console.log(error);
+      return redirect(
+        `${config.loginPath}?error=${LOGIN_ERROR_CODES.LOGIN_GITHUB}&message=${
+          error?.message || ""
+        }`
+      );
+    }
   }
 };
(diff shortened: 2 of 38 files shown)


app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,49 @@
+import { useEffect } from "react";
+import { useSearchParams } from "react-router-dom";
+
+export const LOGIN_ERROR_CODES = {
```
kof, line 4: those seem more like auth providers than error codes, because usually error codes map to a specific error

app/services/auth.server.ts
```diff
@@ -60,8 +60,10 @@ if (process.env.DEV_LOGIN === "true") {
         try {
           const user = await createOrLoginWithDev(secret);
           return user;
-        } catch (e) {
-          console.log(e);
+        } catch (error) {
+          if (error instanceof Error) {
+            throw new Error(error?.message);
```
kof, line 65: - seems not obvious why you are rethrowing an error
- why not just throw error when rethrowing

kof:
```
      throw new Error("Wrong code");
```
Could be more descriptive

app/shared/db/prisma.server.ts
```diff
@@ -1,16 +1,14 @@
...
 
-export const prisma = new PrismaClient();
-export { Prisma, PrismaClientKnownRequestError };
+declare global {
+  // allow global `var` declarations
+  // eslint-disable-next-line no-var
+  var prisma: PrismaClient | undefined;
+}
+
+export const prisma = global.prisma || new PrismaClient();
```
kof, line 10: is this fixing the dev issue with hot reload?

kof: needs at least an explanation

kof: cc @okonet seems like @SaraVieira fixed that problem you also mentioned

app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,49 @@
...
+    "There has been an issue logging you in with Google",
+};
+
+export const useLoginErrors = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const error = searchParams.get("error");
+  const message = searchParams.get("message");
+  useEffect(() => {
+    if (error !== null && !!message) {
+      console.log(message);
```
kof, line 25: @todo

app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,49 @@
...
+  [LOGIN_ERROR_CODES.LOGIN_GOOGLE]:
+    "There has been an issue logging you in with Google",
+};
+
+export const useLoginErrors = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const error = searchParams.get("error");
+  const message = searchParams.get("message");
+  useEffect(() => {
+    if (error !== null && !!message) {
```
kof, line 24: please don't do double negation thing, its always harder to read and less explicit in this case than Boolean(message) or message !== ''

prisma/schema.prisma
```diff
@@ -10,25 +10,33 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
+model Team {
+  id    String @id @default(uuid())
+  users User[]
+
```
kof, line 16: nl

app/routes/auth/github.tsx
```diff
@@ -1,19 +1,33 @@
...
+        userId,
+      },
+      successRedirect: config.dashboardPath,
+      throwOnError: true,
+    });
+  } catch (error: unknown) {
+    // all redirects are basically errors and in that case we don't want to catch it
+    if (error instanceof Response) return error;
+    if (error instanceof Error) {
+      console.log(error);
```
kof, line 25: @todo

kof: I wonder if we should start logging to sentry right away

kof: we definitely shouldn't use console.* api because we need to lint it to prevent keeping console.* left overs from debugging by accident, so we will need to use a different logging functions generally, but maybe we should just start using sentry

SaraVieira: Do we have sentry?

kof: nope, but let me create it in a sec

kof: done, you should have received an invite

kof: one important thing is to make sure we can distinguish logs in sentry from production and always pass userid,projected etc, whatever is available to reproduce the error later

kof: check it out, if adding sentry is not trivial, we can  make a separate PR

kof: specifically we need to make sure that sentry is using fetch API to send the events, not node apis, because otherwise that won't work in a CF worker

kof: oh there is a client for sentry that works in CF worker, I bet it works in node too, but need to check https://github.com/robertcepa/toucan-js

docs/test-cases.md
```diff
@@ -57,3 +57,13 @@
    - Click on the canvas closes breakpoints selector
    - Styles added on each breakpoint are only applied to each breakpoint, verify that by resizing the canvas in preview mode
    - Resizing canvas width in preview mode doesn't limit resize slider to each breakpoint
+
+1. Auth
```
kof, line 61: the format of those should be step by step: click this click that, you know, like you are explaining to a person who knows nothing about development, because this might be actually the case once we ask someone to help test it

kof (changes requested):
first pass done

app/shared/session/useLoginErrors.ts
SaraVieira: renamed

app/services/auth.server.ts
SaraVieira: I am very confused by these questions

This errors is when somethiing bad happens in the create account but the code is correct

app/shared/db/prisma.server.ts
SaraVieira: Yes, the explanation is here: https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices

Its linked in the PR, want me to add it to the code?

app/shared/session/useLoginErrors.ts
SaraVieira: updatedc to boolean

prisma/schema.prisma
SaraVieira: hm?

app/routes/auth/github.tsx
SaraVieira: Made a new PR for sentry

docs/test-cases.md
SaraVieira: But then this is not a testing case because a testing case is for a QA and what needs to be tested

app/shared/db/prisma.server.ts
kof: ah, yes please link it inline, because not obvious to the reader

prisma/schema.prisma
kof: there is a newline that doesn't seem like should be there, doesn't matter though

docs/test-cases.md
kof: you mean we need to rename testing-cases into something else? Sure lets do it. I just think we should be clear about how things are reproduced because they may be not trivial, especially when you have less context than you do right now

app/shared/db/prisma.server.ts
SaraVieira: Done!

prisma/schema.prisma
SaraVieira: OHHH, next person that updates the sdk can change it, I can't change it here, it does nothing

app/services/auth.server.ts
kof: I think we could something like that in the error, "Wrong code" can mean so meany things haha

SaraVieira: OHHHH, updated the message

docs/test-cases.md
SaraVieira: oh, I think I missunderstood what you said above. Made things more clear

app/auth/login.tsx
```diff
@@ -35,10 +42,14 @@ export const Login = () => {
       align="center"
       justify="center"
     >
-      <Card css={{ width: "$10", padding: "$5", zoom: 1.4 }} variant="active">
```
SaraVieira, line 38: This does not exist

SaraVieira: The `width: $10`

app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const useLoginErrors = (): string => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const [messageToReturn, setMessageToReturn] = useState("");
+  const error = searchParams.get("error");
+  const message = searchParams.get("message");
+
+  useEffect(() => {
+    const hasMessageToShow =
+      error !== null && Boolean(message) && typeof message === "string";
```
kof, line 27: Why Boolean check, are you checking if message is empty?

kof: otherwise typeof check should be enough

app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,57 @@
...
+  [AUTH_PROVIDERS.LOGIN_GITHUB]:
+    "There has been an issue logging you in with Github",
+  [AUTH_PROVIDERS.LOGIN_GOOGLE]:
+    "There has been an issue logging you in with Google",
+};
+
+export const useLoginErrors = (): string => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const [messageToReturn, setMessageToReturn] = useState("");
+  const error = searchParams.get("error");
```
kof, line 22: error and message are only used inside the useEffect hook, so they could be defined there

app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const LOGIN_ERROR_MESSAGES = {
+  [AUTH_PROVIDERS.LOGIN_DEV]: "There has been an issue logging you in with dev",
+  [AUTH_PROVIDERS.LOGIN_GITHUB]:
+    "There has been an issue logging you in with Github",
+  [AUTH_PROVIDERS.LOGIN_GOOGLE]:
+    "There has been an issue logging you in with Google",
+};
+
+export const useLoginErrors = (): string => {
```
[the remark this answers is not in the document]
kof: I would even say `useLoginErrorMessage` because its not the error thing you got in search params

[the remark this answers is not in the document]
SaraVieira: Done!

app/shared/session/useLoginErrors.ts
SaraVieira: Done!

app/shared/session/useLoginErrors.ts
kof: so if you really checking for empty, why not `message !== ''` that's literally saying what you want to do, otherwise its quite obfuscated

app/shared/session/useLoginErrors.ts
```diff
@@ -0,0 +1,57 @@
...
+        break;
+      case AUTH_PROVIDERS.LOGIN_GOOGLE:
+        setMessageToReturn(LOGIN_ERROR_MESSAGES[AUTH_PROVIDERS.LOGIN_GOOGLE]);
+        break;
+
+      default:
+        break;
+    }
+
+    setSearchParams({});
```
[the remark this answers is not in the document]
kof: setting state from effect is a typical way to get an infinite loop, its like a generally bad thing to have, even if it works for some reason here, its kindof dangerous, let me know if you need help finding a better way to handle it, don't have a solution out of top of my head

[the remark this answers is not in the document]
kof: I was just discussing today this problem its a nasty problem with useEffect and state, but we have to deal with it somehow, one way is to use useRef so that you have access to that value without having a dependency on the ref and setting ref can be done directly inside the component, its  kind ugly as well but the best I know so far:

something like this:

```ts
  const [searchParams, setSearchParams] = useSearchParams();
 const searchParamsRef = useRef()
searchParamsRef.current = searchParams

useEffect(() => {
  const error = searchParams.current.get("error");
  const message = searchParams.current.get("message");
}, [setSearchParams])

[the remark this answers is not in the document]
SaraVieira: I can tell you exactly why this does not rerender and its because the search params in the URL are not reactive, they don't work like state so it Wong cause a run of the use effect again 

I was gonna either keep it because i know it doesn't re render , i worked with URLs a lot at remote or make the use effect only run on start since that's all we cafe about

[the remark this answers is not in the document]
kof: ignore me then

app/shared/session/useLoginErrors.ts
SaraVieira: updated

PR #87, opened 2022-06-25
createContentController

ghost:
extracted abstraction from the https://github.com/webstudio-is/webstudio-designer/pull/64 monolith.

- [x] stories

### DIFF
--- a/app/shared/design-system/components/primitives/create-content-controller.stories.tsx
+++ b/app/shared/design-system/components/primitives/create-content-controller.stories.tsx
@@ -61,12 +61,8 @@ const ExampleContentController = () => {
   );
 };
 
-export const DetectInput = Object.assign(ExampleContentController.bind({}), {
-  args: {},
-});
+export const DetectInput = Object.assign(ExampleContentController.bind({}));
 
 export default {
-  title: "createContentController",
   component: ExampleContentController,
-  argTypes: {},
 };


app/shared/design-system/components/primitives/createContentController.js
```diff
@@ -0,0 +1,106 @@
+/*
+	@example
+
```
kof, line 3: lets format comment with 2 spaces too

kof: would be also great to add a few sentences what this this is designed to do, might be not trivial to understand, same for others

app/shared/design-system/components/primitives/createContentController.js
```diff
@@ -1,16 +1,23 @@
...
-				{ name: 'number', match: (value) => isNaN(parseFloat(value)) === false },
-				{ name: 'unknown', match: (value) => true },
-			],
-			onMouseMove: ({name, value, position}) => { },
-			onCaretMove: ({name, value, position}) => { },
-		});
-*/
+/**
+ * @description
+ * - detects whether the pointer/keyboard is on any given token as specified by the {contents: []} contract.
```
kof, line 3: btw maybe we should call it tokens instead of contents?

app/shared/design-system/components/primitives/createContentController.js
```diff
@@ -0,0 +1,113 @@
...
+    .map((value) => {
+      const width = measureTextWidth(value);
+      const head = index - offset;
+      const tail = index - (offset + width);
+      const props = { head, tail, value, width, length, offset };
+      length += value.length;
+      offset += width;
+      return props;
+    });
+};
```
okonet, line 57: I wonder if you have considered a more simplistic implementation for this taking into account that we should be using monospaced fonts or rely on OpenType feature `font-variant-numeric: tabular-nums;` for those controls to avoid layout shifts. This would simplify the problem to `const textWidth = numOfChars * CHAR_WIDTH`. For now we could even hard-code the constant and later we could get the value from font metrics or calculated styles once on the app init. WDYT?

app/shared/design-system/components/primitives/createContentController.js
```diff
@@ -0,0 +1,113 @@
...
+    read = (node) => node.value,
+    write = (value) => {},
+    contents = [],
+    onMouseMove = () => {},
+    onCaretMove = () => {},
+  }
+) => {
+  const eventNames = ["keyup", "pointermove"];
+  const handleEvent = ({ type, offsetX }) => {
+    const targetValue = read(targetNode);
```
okonet, line 70: I could not find where `read` was defined. Can you point me to it please?

okonet:
It feels like what we're trying to do here is related to https://github.com/webstudio-is/webstudio-designer/pull/89 and IMO we should not use multiple implementations for parsing / working with DOM. I would rather leave the caret positioning etc to a library maintainers rather than maintaining it ourselves.

kof:
@okonet a content-editable with react components inside is quite a different problem, including problems with caret. Input is a super simple thing to do this for compared to content-editable. Can explain in detail if you want.

okonet:
I'm not sure why we should care about content-editable if we want to enter a CSS unit TBH. I would like to advocate for staying pragmatic and not try to solve problems we do t have yet :)

kof:
@okonet exactly, #89 is unrelated

app/shared/design-system/components/primitives/createContentController.js
kof: This would have been so much simpler and avoided bugs, lets add this as a comment inline in code, if this measuring is going to be buggy, we just switch to this simplified solution? Just need to have it here to remember there is a different way, more limited to monospaced fonts, but reliable

app/shared/text-editor/example-text-editor.stories.tsx
```diff
@@ -11,7 +11,7 @@ const onChange = (editorState: EditorState) => {
     // Read the contents of the EditorState here.
     const root = $getRoot();
     const selection = $getSelection();
-
+    // eslint-disable-next-line
```
kof, line 14: fixed on main, need to merge main into this branch

kof:
Is this one mergreable?

app/shared/design-system/components/primitives/create-content-controller.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </fieldset>
+      <fieldset>
+        <h1>Currently on: {content.name}</h1>
+        <h1>With a value of: {content.value}</h1>
+      </fieldset>
+    </form>
+  );
+};
+
+export const DetectInput = Object.assign(ExampleContentController.bind({}), {
```
kof, line 64: check out select.stories.tsx for how to write down the story better, its a newer api

app/shared/design-system/components/primitives/create-content-controller.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  args: {},
+});
+
+export default {
+  title: "createContentController",
+  component: ExampleContentController,
+  argTypes: {},
```
kof, line 71: no need for title and argTypes

kof (approved):
please update the stories format and merge

PR #88, opened 2022-06-25
Add sentry

SaraVieira:


app/services/auth.server.ts
```diff
@@ -60,8 +61,15 @@ if (process.env.DEV_LOGIN === "true") {
...
           const user = await createOrLoginWithDev(secret);
           return user;
-        } catch (e) {
-          console.log(e);
+        } catch (error: unknown) {
+          if (error instanceof Error) {
+            sentryException({
+              message: error.message,
+              extra: {
+                loginMethod: "dev",
```
kof, line 69: does this need to change for prod?

app/shared/sentry.ts
```diff
@@ -0,0 +1,49 @@
+import * as Sentry from "@sentry/browser";
+import { Extras } from "@sentry/types";
+
+import { BrowserTracing } from "@sentry/tracing";
+
+export const initSentry = () =>
+  Sentry.init({
+    dsn: "https://e4b439185ee94fca9599b768a75d211e@o1298958.ingest.sentry.io/6530661",
```
kof, line 8: I think you want to get this over env config

kof: and make sentry  optional if its not provided?

kof (changes requested):
Looking good, a few things commented
Question: what should we do about server-side logging? new issue? Is it solving the needs for logging for auth? If yes we can create a separate issue for server-side sentry

app/services/auth.server.ts
SaraVieira: Nop, this is the method the user used to login, not the environment, that's set on set up

app/shared/sentry.ts
SaraVieira: sooooo, I don't have access to env variables here

SaraVieira:
> Question: what should we do about server-side logging? new issue? Is it solving the needs for logging for auth? If yes we can create a separate issue for server-side sentry

Server-side it currently works if we specifically send the message to sentry but it won't catch anything we don't manually send.

I would say we create a new issue and tag the remix SDK support they are working on. This works great for auth

app/shared/sentry.ts
kof: sure, it has to be passed when you call init

app/services/auth.server.ts
kof: hmm, but should it not change depending on method?

SaraVieira: Yeah but this is on the dev method

app/shared/sentry.ts
SaraVieira: Yap :(

kof: so we need to move it to env, read it probably into config and pass it to initSentry call from a place where we have a config, I guess

app/services/auth.server.ts
kof: I haven't seen the if in the diff github showed me lol

app/env.server.ts
```diff
@@ -0,0 +1,9 @@
+// Environment variables we want to send to the UI inlined in the document.
+// Never use a private key here, because it will become public.
+const env = {
+  SENTRY_DSN: process.env.SENTRY_DSN,
```
kof, line 4: - [ ] needs documentation

app/shared/sentry.ts
```diff
@@ -0,0 +1,59 @@
...
+export const initSentry = () => {
+  SENTRY_DSN
+    ? Sentry.init({
+        dsn: SENTRY_DSN,
+        integrations: [new BrowserTracing()],
+        tracesSampleRate: 1.0,
+        environment: VERCEL_ENV || "development",
+      })
+    : () => null;
+};
```
SaraVieira, line 19: 1 - Without the window check I got a window is not defined so we need both  :/
2- VERCEL_ENV is a reserved keyword env by vercel that allows us to know if we are in preview or production

https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables

kof: here it is supposed to just work, with no window checks

SaraVieira: Moved to the getter

kof (approved):

app/shared/env/env-getter.ts
```diff
@@ -12,7 +12,7 @@ export default new Proxy(
       if (typeof window === "undefined") {
         return process.env;
       }
-      const env = (window[namespace as any] ?? {}) as unknown as Env;
+      const env = (window[namespace as never] ?? {}) as unknown as Env;
```
kof, line 15: that wroked? it was screaming at me, maybe ts version thing

kof: try `git merge main && yarn && yarn typecheck --watch`

kof:
I upgraded typescript on main to the latest version that still works - 4.6, for 4.7 we have migration to do

kof (approved):
🎉

PR #89, opened 2022-06-26
Create a POC with Lexical WYSIWYG

kof:
Start storybook to test


app/shared/text-editor/config.ts
```diff
@@ -0,0 +1,11 @@
+import { InstanceNode } from "./nodes/node-instance";
+import { darkTheme } from "./themes/dark";
+
+export const config = {
+  namespace: "example",
+  theme: darkTheme,
```
okonet, line 6: Not sure it should be hard-coded here like that but rather depends on the user preference. Can this be abstracted away?

app/shared/text-editor/example-text-editor.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+export default {
+  title: "Text Editor",
+  component: ExampleTextEditor,
+} as ComponentMeta<typeof ExampleTextEditor>;
+
+const Template: ComponentStory<typeof ExampleTextEditor> = (args) => {
+  return <ExampleTextEditor {...args} />;
+};
+
+export const TextEditor = Template.bind({});
```
okonet, line 28:
```suggestion
export const TextEditor = (args) => <ExampleTextEditor {...args} />
```

app/shared/text-editor/example-text-editor.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+type ExampleTextEditorProps = {
+  onChange: (state: EditorState) => void;
+};
+
+const Menu = () => {
+  const [, setAllUserProps] = useAllUserProps();
+  return (
+    <div className="editor-menu">
+      <button
```
okonet, line 23: Can our design-system components be used here?

app/shared/text-editor/nodes/node-instance.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  updateDOM(prevNode: TextNode, dom: HTMLElement, config: EditorConfig) {
+    const inner = dom.firstChild as HTMLElement;
+    if (inner === null) {
+      return true;
+    }
+    super.updateDOM(prevNode, inner, config);
+
+    return false;
+  }
```
okonet, line 40: How well is this playing with ReactDOM?

app/shared/text-editor/plugins/plugin-instance.tsx
```diff
@@ -0,0 +1,96 @@
...
+}: {
+  instance: Instance;
+  children: string;
+}) => {
+  const [breakpoints] = useBreakpoints();
+  const css = useMemo(
+    () => toCss(instance.cssRules, breakpoints),
+    [instance, breakpoints]
+  );
+  const className = useCss({ instance, css });
```
okonet, line 83: This feel counter-intuitive. the rest of SDK handles this internally, but this component does it here. Is this intentional?

app/shared/text-editor/plugins/plugin-instance.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const userProps = useUserProps(instance.id);
```
okonet, line 84: What if `userProps` already have a `className` set?

app/shared/text-editor/plugins/plugin-instance.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const { Component } = primitives[instance.component];
```
okonet, line 85: We need some sort of safe meter I think. This will blow up sooner than later when the requested component doesn't exist in primitives.

app/shared/text-editor/styles.css
```diff
@@ -0,0 +1,120 @@
+body {
```
okonet, line 1: Any reason we can't use design-system instead?

okonet:
@kof it would help a lot if each PR would have "How to test" instructions and a short description of the work (ideally linked to an issue).

okonet:
I could not test in action due to https://github.com/webstudio-is/webstudio-designer/issues/95

kof:
agree, I should have mentioned you only need to start storybook here, because this doesn't do any integration yet

app/shared/text-editor/config.ts
```diff
@@ -0,0 +1,11 @@
+import { InstanceNode } from "./nodes/node-instance";
+import { darkTheme } from "./themes/dark";
+
+export const config = {
+  namespace: "example",
```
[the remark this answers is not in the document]
kof: I have yet to see why it was required at all haha

app/shared/text-editor/config.ts
kof: need to figure this out yet, theme is basically just a map to class names

app/shared/text-editor/example-text-editor.tsx
kof: yep

app/shared/text-editor/nodes/node-instance.tsx
kof: short answer - not well.
basically manual process, we will let lexical handle the editable area, then where text is inside of webstudio instances, we will manually pass it to the component and rerender it

app/shared/text-editor/plugins/plugin-instance.tsx
kof: wrapper component is part of the designer, not SDK, I had to create a new one for inline components inside the instance, because you don't want there the same wrapper component that renders it from outside, basically this plugin is all about integrating lexical with designer/canvas, not sure where is a better place for this, because its only ever going to be used inside the text editor.

I will see if we can reuse majority of this logic  as a single hook  so that there is less duplication

app/shared/text-editor/plugins/plugin-instance.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  return (
+    <Component
+      {...rest}
+      {...userProps}
+      key={instance.id}
+      id={instance.id}
```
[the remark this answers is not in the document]
kof: yes, there are a few places on canvas where you actually want to get the element using some selector over dom query, but we should switch to data-id from id, to free it up for the user to use

app/shared/text-editor/styles.css
kof: these are demo styles, but let me try and figure out how we can have the entire text editor get the same styles we will have on canvas

kof:
@okonet I will address all comments in this PR and merge it as-is, https://github.com/webstudio-is/webstudio-designer/pull/93

I started already working on the new PR which already has changed somethings and will change more I as I do the integration work, but I will go over all comments here and make sure they are addressed there

app/shared/text-editor/config.ts
kof: removed

app/shared/text-editor/config.ts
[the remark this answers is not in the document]
kof: changed

app/shared/text-editor/styles.css
kof: we neeed to setup global styles for all stories, but otherwise this file is gone

app/shared/text-editor/example-text-editor.stories.tsx
okonet: Doesn't look like anything has been updated here?

kof: all the CR changes are in https://github.com/webstudio-is/webstudio-designer/pull/93

kof: I had to merge this PR as-is and apply changes in the follow-up PR

PR #90, opened 2022-06-27
Env config inlined for the UI before any other script runs

kof:
This is needed for sentry.

Essentially Remix should have its native way of doing it, so this is a temp solution for us

https://github.com/remix-run/remix/discussions/2769

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -14434,7 +14434,7 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@^4.1.2:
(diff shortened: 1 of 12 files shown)


yarn.lock
```diff
@@ -14434,7 +14434,7 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@^4.1.2:
+typescript@4.5.5:
```
kof, line 14437: had to make sure we have a fixed version, because of weird bugs with different once

kof:
@SaraVieira feel free to merge it into your branch and add the env variables to all entry routes where sentry is going to run.

yarn.lock
SaraVieira: perfect!

SaraVieira (approved):

PR #91, opened 2022-06-27
createDropdown

ghost:
createDropdown abstraction, an abstraction born from work on the https://github.com/webstudio-is/webstudio-designer/pull/64 monolith.

- [x] stories.
- [x] make text field optional.
- [ ] callback for resolved value.
- [ ] convert logic to react/radix component.

### DIFF
--- a/app/shared/design-system/components/primitives/createDropdown.js
+++ b/app/shared/design-system/components/primitives/createDropdown.js
@@ -19,9 +19,8 @@
  *   });
  * })
  * @todo calculate if clientX + clientWidth is greater than viewport clientWidth and adjust as needed
- * @todo stories file
  * @todo api to recieve the resolved value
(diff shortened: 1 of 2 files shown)


app/shared/design-system/components/primitives/createDropdown.js
```diff
@@ -0,0 +1,151 @@
...
+    this.selectNode.focus();
+  }
+  constructor() {
+    super();
+    this.innerHTML = `
+		<dialog>
+			<form method="dialog">
+				<fieldset>
+					<input autofocus placeholder="enter a value" value="${
+            this.querySelector("option[selected]")?.value || ""
```
kof, line 43: Idk, querying the thing ... 
it feels like we are back in time, first rendering templates then querying ...

kof: I feel like this should be a react component

PR #103, opened 2022-07-01
chore(sentry) - Use new @remix/sentry

SaraVieira:


kof:
pushed a workaround

app/shared/documents/canvas.tsx
```diff
@@ -6,35 +6,40 @@ import {
...
+          <Env />
+          <Scripts />
+          {process.env.NODE_ENV === "development" && <LiveReload />}
+        </body>
+      </html>
+    </ErrorBoundary>
   );
 };
+
+export const Canvas = withSentryRouteTracing(CanvasRoot);
```
SaraVieira, line 45: Error happens here, seems to be `@sentry/remix` types not allowing for props but app seems to work the same

kof: withSentryRouteTracing has a wrong props type expectation, it should not be that opinionated I think, we should create an issue there

kof (approved):
haven't tested

SaraVieira:
@kof updated the sentry version and no more hack needed


Also tested it

PR #104, opened 2022-07-01
chore(env) - Add env check before starting app

SaraVieira:
This file will make sure any variables we add in there will be checked on start so the user has no surprises

CleanShot 2022-07-01 at 23 09 25
CleanShot 2022-07-01 at 23 09 33


app/shared/env/env-check.js
```diff
@@ -0,0 +1,24 @@
+// @ts-check
```
kof, line 1: lets put it in the root

remix.config.js
```diff
@@ -1,3 +1,7 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const { env } = require("./app/shared/env/env-check");
```
kof, line 3: why do you import the variable at all?

kof:
can we do anything about that error formatter easily?

kof:
haven't used zod, but I think .format() is for structured error data, by default its human readable

app/shared/env/env-check.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * This file is included in `/remix.config.js` which ensures the app isn't built with invalid env vars.
+ * It has to be a `.js`-file to be imported there.
+ */
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { z } = require("zod");
+
+const envSchema = z.object({
+  DATABASE_URL: z.string().url(),
+  AUTH_SECRET: z.string().min(10),
```
kof, line 11: is this always required or only in dev mode?

SaraVieira: This will also work on build

SaraVieira: If you don't have the vars the build will break

app/shared/env/env-check.js
SaraVieira: Done

remix.config.js
SaraVieira: removed

app/shared/env/env-check.js
SaraVieira: Sooo, researched this and in production remix doesn't read the .env file so this is only for dev

SaraVieira:
Removed ZOD since we don't need anything that fancy for what we are doing and formatted the errors better

CleanShot 2022-07-02 at 12 34 17
CleanShot 2022-07-02 at 12 38 04

env-check.js
```diff
@@ -0,0 +1,25 @@
+// @ts-check
+/**
+ * This file is included in `/remix.config.js` which ensures the app isn't built with invalid env vars.
+ * It has to be a `.js`-file to be imported there.
+ */
+
+// Remix does not read .env files when building for production
+if (process.env.NODE_ENV !== "production") {
+  const REQUIRED_ENVS = ["DATABASE_URL", "AUTH_SECRET"];
```
kof, line 9: database url is not required in production?

kof: I think this logic is fundamentally flawed, you want to check first variables that are required in all environment, then to check the once required in specific environments, dev or prod

app/shared/env/env-check.js
kof: that means that we need a place that is going to be executed for example during deployment to production, because you want a good error message if the env wasn't configured correctly and it would run in both dev and prod

kof: I guess in prod you want to catch it before its in prod so that deployment is canceled

env-check.js
SaraVieira: Yeah it is, and that's honestly why i am having so many issues with remix, they make these decisions that make no sense and just make everyone go around in circles to fix issues for their decisions 

It's their decision to not read environment files in remix build, so the issue i can probably support this in prod by still checking BUT then it will break you running yarn build on your machine because remix will not check your env file.

It seems like all options lead to at least one problem

SaraVieira: So what I figured could work would be to add `dotenv` at the start of that file since remix already uses it and that works so now it works locally and in deployment

Should not be this hard

app/shared/env/env-check.js
SaraVieira: has been fixed below

kof (approved):

PR #105, opened 2022-07-01
fix(numeric-gesture-control): abstract PointerLock API

### LINKED ISSUE #99: NumericInput doesn't work in Safari
Currently, the `pointerLock` API which is used in the NumericInput causes Safari to show a special banner

https://user-images.githubusercontent.com/11071/176870343-9077ca47-7a47-487e-b5c7-3e337ae28eb3.mp4

I think we should not use this API since it's not clear how it's going to behave in different browsers. Also [from the docs](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) I read it's for games etc. which isn't our use case IMO.

ghost:
This abstracts the Pointer Lock browser API used in numeric gesture control, enhancing the experience of it on browsers that make it easy, along the same lines as what Figma does.

Fixes #99, more details on the issue.

### DIFF
--- a/app/shared/design-system/components/primitives/numeric-gesture-control.ts
+++ b/app/shared/design-system/components/primitives/numeric-gesture-control.ts
@@ -30,7 +30,7 @@ type Options = {
 
 type State = {
   value: number;
-  cursor: SVGElement | undefined;
+  cursor?: SVGElement;
   offset: number;
   velocity: number;
   direction: string;


app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -28,6 +28,14 @@ type Options = {
   }) => void;
 };
 
+type State = {
+  value: number;
+  cursor: SVGElement | null;
```
kof, line 33: Nit: I generally prefer undefined when possible, because it works with optional arguments in js

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -108,25 +98,72 @@ export const numericGestureControl = (
...
+          <path d="M 15 4.5L 15 2L 11.5 5.5L 15 9L 15 6.5L 31 6.5L 31 9L 34.5 5.5L 31 2L 31 4.5Z" fill="#111" fill-rule="evenodd"></path>
+        </g>
+      </svg>`).firstElementChild as SVGElement;
+    cursorNode.style.position = "absolute";
+    cursorNode.style.left = `${event.offsetX}px`;
+    cursorNode.style.top = `${event.offsetY}px`;
+    cursorNode.style.transform =
+      state.direction === "horizontal" ? "rotate(0deg)" : "rotate(90deg)";
+    cursorNode.style.setProperty(
+      "--drop-shadow",
```
kof, line 138: why do you use a css var here vs setting it directly?

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -108,25 +98,72 @@ export const numericGestureControl = (
...
+      `drop-shadow(${
+        state.direction === "horizontal" ? "0 1px" : "1px 0"
+      } 1.1px rgba(0,0,0,.4))`
+    );
+    state.cursor = cursorNode;
+    targetNode.ownerDocument.documentElement.append(state.cursor as Node);
```
kof, line 144: is `as` used because it cursor could be undefined/null? If yes then better way is to have an "if" check before you use the variable, because it can actually be unset potentially

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -108,25 +98,72 @@ export const numericGestureControl = (
...
+    targetNode.onpointermove = () => {
+      (state.cursor as SVGElement).style[
```
kof, line 146: same here

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -108,25 +98,72 @@ export const numericGestureControl = (
...
+
+const exitPointerLock = (
+  state: State,
+  event: PointerEvent,
+  targetNode: HTMLElement
+) => {
+  targetNode.releasePointerCapture(event.pointerId);
+
+  if (shouldUsePointerLock()) {
+    (state.cursor as SVGElement).remove();
```
kof, line 161: same here

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -108,25 +98,72 @@ export const numericGestureControl = (
...
+    targetNode.onpointermove = null;
+    targetNode.ownerDocument.exitPointerLock();
+  }
+};
+
+const shouldUsePointerLock = () => {
```
kof, line 167: why is it a function, it could be just a value, right?

kof: when we are in node, this should be automatically fasle

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -108,25 +98,72 @@ export const numericGestureControl = (
...
 
-const cursorNode = new Range().createContextualFragment(`
-  <svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="46" height="15" style="filter:var(--drop-shadow);">
-    <g transform="translate(2 3)">
-      <path d="M 15 4.5L 15 2L 11.5 5.5L 15 9L 15 6.5L 31 6.5L 31 9L 34.5 5.5L 31 2L 31 4.5Z" fill="#111" fill-rule="evenodd" stroke="#FFF" stroke-width="2"></path>
-      <path d="M 15 4.5L 15 2L 11.5 5.5L 15 9L 15 6.5L 31 6.5L 31 9L 34.5 5.5L 31 2L 31 4.5Z" fill="#111" fill-rule="evenodd"></path>
-    </g>
-  </svg>
-`).firstElementChild as HTMLElement;
+const requestPointerLock = (
```
kof, line 116: I think we need a good comment here how this works, nobody would understand it without context

kof (changes requested):
first review done

app/shared/design-system/components/primitives/numeric-gesture-control.ts
```diff
@@ -28,6 +28,14 @@ type Options = {
   }) => void;
 };
 
+type State = {
+  value: number;
+  cursor: SVGElement | undefined;
```
kof, line 33: same as
```suggestion
  cursor?: SVGElement;
```

kof (approved):
LGTM 🎉

PR #119, opened 2022-07-06
Full-stack DEBUG variable and FEATURE flags

kof:
Now we can 

```ts
import {isFeatureEnabled} from 'feature-flags'
const isSomethingEnabled = isFeatureEnabled('something')
```

Returns true/false if the feature is turned on.
A feature can be turned on:
- by default directly in ./flags
- by providing an environment variable server-side (locally or on the server): FEATURE="something1, something2" yarn dev
- by setting it in the browser console: localStorage.feature = 'something1, something2', browser defined flag will override server-side flag


kof:
Should we separate the DEBUG and FEATURES env variables completely?

SaraVieira:
There are some conflicts on the yarn lock but this looks good to go!

Issue #108, opened 2022-07-05
Dark/light mode theme switcher


kof:
Radix already supports "automatic" colors for the modes, we need

- [x] menu switcher
- [x] respect system settings by default
- [x] save the setting in cookie

PR #120, opened 2022-07-07
Theme switcher

kof:
https://github.com/webstudio-is/webstudio-designer/issues/108

Theme switch menu is behind a feature flag, to enable run `localStorage.feature="theme"`


apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,13 +1,18 @@
...
 import { Designer, links } from "~/designer";
 import * as db from "~/shared/db";
 import config from "~/config";
 import env from "~/env.server";
+import { authenticator } from "~/services/auth.server";
 export { links };
 
-export const loader: LoaderFunction = async ({ params }) => {
+export const loader: LoaderFunction = async ({ request, params }) => {
+  const user = await authenticator.isAuthenticated(request);
```
kof, line 12: this is a drive-by fix, we haven't checked in the designer if user is authenticated @SaraVieira

apps/designer/app/designer/shared/client-settings/settings.ts
```diff
@@ -0,0 +1,51 @@
+import * as values from "./values";
```
kof, line 1: is is currently unused, I built it before I realized we can't save theme setting into local storage, but I think we will have a plenty of use cases for local user settings in the local storage as an option to  user settings in the database

apps/designer/app/designer/features/topbar/menu/menu.tsx
```diff
@@ -26,6 +34,41 @@ const menuItemCss = {
...
+            css={menuItemCss}
+            onSelect={() => {
+              setThemeSetting(setting);
+            }}
+          >
+            {labels[setting]}
+          </DropdownMenuCheckboxItem>
+        ))}
+      </DropdownMenuContent>
+    </DropdownMenu>
```
okonet, line 68: Consider using `Select` as it should simplify the usage a lot.

apps/designer/app/designer/shared/client-settings/settings.ts
```diff
@@ -0,0 +1,51 @@
...
+const read = (): Settings => {
+  try {
+    const settings = localStorage.getItem(namespace);
+    if (settings === null) return {};
+    return JSON.parse(settings);
+  } catch (error) {
+    // We don't need to handle this one.
+  }
+  return {};
+};
```
okonet, line 19: A test would be great. Also, just swallowing errors doesn't seem to be a good idea. Should those at least be logged and sent to Sentry?

apps/designer/app/shared/feature-flags/feature.ts
```diff
@@ -28,5 +28,5 @@ export const isFeatureEnabled = (name: Name): boolean => {
   const defaultValue = flags[name];
   const envValue = parse(env.FEATURE).includes(name);
   const localValue = readLocal().includes(name);
-  return localValue ?? envValue ?? defaultValue;
+  return localValue || envValue || defaultValue;
```
okonet, line 31: What is the difference?

apps/designer/app/shared/theme/theme.ts
```diff
@@ -0,0 +1,84 @@
...
+  const query = matchMedia("(prefers-color-scheme: light)");
+  const queryColorScheme = (query: MediaQueryList | MediaQueryListEvent) =>
+    query.matches ? "light" : "dark";
+
+  system = queryColorScheme(query);
+
+  query.addEventListener("change", (queryEvent) => {
+    system = queryColorScheme(queryEvent);
+    renderThemeProps();
+  });
```
okonet, line 24: I could not find clean up code. Probably means there is a memory leak. Also, I think this should go into a hook.

apps/designer/app/designer/features/topbar/menu/menu.tsx
kof: That doesn't work, its an actual menu with submenus

apps/designer/app/shared/feature-flags/feature.ts
kof: includes `''`

apps/designer/app/shared/theme/theme.ts
kof: this is a single event listened binding, its done once and never again

apps/designer/app/designer/shared/client-settings/settings.ts
kof: That test would be useless, because localStorage would be mocked or polyfilled and otherwise there is nothing in this function to test.

kof: My strategy for tests is when there is complex logic or a bug already happened to as a regression test. We are not writing a library and testing every function is not viable or useful. Once we get to e2e tests, they will be able to cover real user flows.

okonet: > This will only throw if basic localStorage is not available

It will also throw if JSON.parse won't work.

apps/designer/app/shared/feature-flags/feature.ts
okonet: Ouch. I wonder if there is a less ambiguous syntax or a helper function we could use for such cases.

apps/designer/app/shared/theme/theme.ts
okonet: I see a function that can be called multiple times potentially :)

apps/designer/app/designer/features/topbar/menu/menu.tsx
okonet: Ah I see, it's a sub-menu in another menu.

okonet (approved):
I think it's fine for the current status of the project. We can refactor later.

apps/designer/app/designer/shared/client-settings/settings.ts
kof: right, but that can only mean the data in settings is corrupt, invalid json, what can we do about it except of ignoring it as if it doesn't exist?

okonet: I guess reporting it to us so we know about this?

apps/designer/app/shared/feature-flags/feature.ts
okonet: Because it's confusing even to me (an experienced dev) :)

kof: the reason I supported empty string is just because some may set that flag to en empty string in local storage to disable feature, its not really important, just thought would be more practical

kof: I know remembered why I added those, its not because of empty string, its because of `false`
we return false from .includes, which makes the first value always take precedence with ??

apps/designer/app/designer/shared/client-settings/settings.ts
kof: done

apps/designer/app/shared/feature-flags/feature.ts
okonet: I get the logic but I think the amount of gotcha with `||` and `??` what makes me a bit uncomfortable reading it.

okonet (approved):

PR #122, opened 2022-07-08
Weird margin fix

kof:
This margin above the root container (white) is somehow caused by the margin on the H1.

I don't know if there is a better fix.

Screenshot 2022-07-08 at 23 35 17


okonet:
I'd need to investigate this one. Looks like it could come from H1 or something like this. I can take a look next week.

kof:
@okonet the margin is definitely coming from h1 and somehow makes it look like its outside of the root div

kof:
it looks like margin bleeding lol

kof:
@okonet any idea what that bug is?

okonet:
This is caused by [margin collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) behavior and it's expected to work like this. As we discussed in the chat and after some thoughts I think the right solution would be to apple global styles to the `body` instead of a `div` inside of it. This way it won't only fix the problem but also make the additional element with `100vh` obsolete which is aligned with user's expectations IMO. also it's better to set background styles on the body on mobile devices because of the over scroll behavior. So the right solution to me should be:

1. Introduce a `Root` component to the SDK
2. Render the root component by default

kof:
@okonet I guess when you say Root component you mean Body component, because based on the above text you want the root to be body tag, right?

kof:
Made the workaround using display flex as an intermediate solution, will create an issue for the right one once I fully understand what we want to do:

- Add a Body component to SDK, render it in remix, 
- Let user access body fully, like any other component
- Make it to be the root always

PR #124, opened 2022-07-09
Decouple storybook from yarn dev

kof:
Storybook no longer runs along with yarn dev, so now we need to start it separately with `yarn storybook`

### DIFF
--- a/apps/designer/.storybook/main.js
+++ b/apps/designer/.storybook/main.js
@@ -10,15 +10,15 @@ module.exports = {
   features: {
     storyStoreV7: true,
   },
-  core: {
-    builder: {
-      name: "webpack5",
-      options: {
-        lazyCompilation: true,
-        fsCache: true,
-      },
-    },
-  },
+  //core: {
+  //  builder: {
+  //    name: "webpack5",
+  //    options: {
+  //      lazyCompilation: true,
+  //      fsCache: true,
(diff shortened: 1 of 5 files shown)


SaraVieira (approved):

kof:
Another option is to remove storybook from yarn dev and make it separate again

okonet:
Let's decouple dev and storybook

kof:
@okonet done

turbo.json
```diff
@@ -9,6 +9,9 @@
     },
     "dev": {
       "cache": false
+    },
+    "storybook:run": {
```
okonet, line 13: This should be his `storybook`

apps/designer/.storybook/main.js
```diff
@@ -10,15 +10,15 @@ module.exports = {
...
-  },
+  //core: {
+  //  builder: {
+  //    name: "webpack5",
+  //    options: {
+  //      lazyCompilation: true,
+  //      fsCache: true,
+  //    },
+  //  },
+  //},
```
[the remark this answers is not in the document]
kof: this change was reverted, so those options are on

turbo.json
kof: can you send a PR? not sure what you mean

Issue #37, opened 2022-05-11
Assets manager


kof:
As a user I need a way to upload a picture and use it as a background image or image component source.

## Tasks

- [ ] upload UI - Only allow images
- [ ] management UI
	- [ ] Remove assets
	- [ ] List assets by type
	- [ ] Add image to website
- [ ] file system upload for usage locally or in a docker
- [ ] upload to S3/R2 for hosted version

SaraVieira:
Reference code: https://gist.github.com/SaraVieira/8cdb29104016ad5b9a164b19cf96e25e

Library: https://www.npmjs.com/package/multer-s3 & https://www.npmjs.com/package/multer

SaraVieira:
Publishable parts:

- [ ] Upload single image locally(https://github.com/webstudio-is/webstudio-designer/pull/136)
- [ ] Add multi file support
- [ ] Fix image layouts with background fit and make them all same size
- [ ] Allow user to delete image
- [ ] Allow user to upload to s3

v1.1:

- [ ] Have nice uploading animation

SaraVieira:
Things to think:

- Do we need the path in the DB?

PR #136, opened 2022-07-12
Assets - Allow local upload

SaraVieira:
Adds support for local uploads

Related to https://github.com/webstudio-is/webstudio-designer/issues/37


apps/designer/app/designer/designer.tsx
```diff
@@ -37,6 +37,7 @@ import { usePublishShortcuts } from "./shared/shortcuts";
 import { type SyncStatus } from "~/shared/sync";
 import { useIsPreviewMode, useRootInstance } from "~/shared/nano-states";
 import { useSubscribeClientSetting } from "./shared/client-settings";
+import { Asset } from "@prisma/client";
```
kof, line 40: we don't import prisma directly anywhere, to have a single source of truth for data related stuff, we import from webstudio-react-sdk and will move data stuff into a separate package soon, but for now its there

SaraVieira: Will fix that then

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <Form method="post" encType="multipart/form-data">
+          <input
+            accept="image/*"
+            type="file"
+            name="image"
+            ref={inputRef}
+            onChange={onChange}
+            style={{ display: "none" }}
+          />
+          <button
```
kof, line 35: hmm why do you need 2 buttons?

kof: ah got it, Button is for selecting files, button is for initiating submit event

apps/designer/app/designer/designer.tsx
SaraVieira: Fixed

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
SaraVieira: yap!

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
```diff
@@ -0,0 +1,58 @@
...
+            ref={submitButtonRef}
+            type="submit"
+            style={{ display: "none" }}
+          />
+          <Button onClick={() => inputRef?.current && inputRef.current.click()}>
```
kof, line 40:
```suggestion
          <Button onClick={() => inputRef?.current?.click()}>
```

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
```diff
@@ -8,8 +8,10 @@ import {
   SidebarTabsTrigger,
 } from "~/shared/design-system";
 import { useSelectedInstanceData } from "../../shared/nano-states";
-import * as panels from "./panels";
+import { panels } from "./panels";
 import type { TabName } from "./types";
+import { Asset } from "@prisma/client";
```
kof, line 13: same

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,69 @@
...
   const project = await db.project.loadById(params.id);
+  const assets = await db.assets.loadByProject(params.id);
   if (project === null) {
     return { errors: `Project "${params.id}" not found` };
   }
-  return { config, project, env };
+  return { config, assets, project, env };
 };
 
+const directory = "./public/uploads";
```
kof, line 26: should be a config or even env variable? and also need to resolve the relative path

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,69 @@
...
+  if (params.id === undefined) throw new Error("Project id undefined");
+  const formData = await unstable_parseMultipartFormData(
+    request,
+    unstable_createFileUploadHandler({
+      maxPartSize: 10_000_000,
+      directory,
+      file: ({ filename }) => filename,
+    })
+  );
+  const imageInfo = formData.get("image") as ImageUpload;
```
kof, line 53: I made this mistake too, instead of telling typescript it is ImageUpload as a truth, we should be actually asserting it is the case with code, because this line is already wrong, because it can be null, but also it can be something else, better check would be:

```ts
if (imageInfo instanceof ImageUpload === false) return {errors: 'what happened'}
```

As we speak there must be a way to handle an error in the UI for this

kof: I got a reccomendation lately for this https://github.com/esamattis/react-zorm

apps/designer/prisma/schema.prisma
```diff
@@ -13,7 +13,17 @@ datasource db {
...
 
+model Asset {
+  id        String   @id @default(uuid())
+  project   Project  @relation(fields: [projectId], references: [id])
+  projectId String
+  type      String
+  name      String
+  path      String
+  alt       String?
+  createdAt DateTime @default(now())
```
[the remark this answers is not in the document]
SaraVieira: Like what?

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,69 @@
...
+  if (imageInfo) {
+    const data = {
+      type: imageInfo.type,
+      name: imageInfo.name,
+      path: `/uploads/${imageInfo.name}`,
```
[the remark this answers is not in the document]
kof: I think we should have fileName that would be `name + uuid` to prevent having clashes even if all files of all users end up in a single s3 storage folder some day

kof (changes requested):
almost there

apps/designer/app/routes/designer/$id.tsx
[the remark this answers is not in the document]
kof: have you logged imageInfo, whats inside?

apps/designer/prisma/schema.prisma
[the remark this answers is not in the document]
kof: there should be size and type

apps/designer/app/routes/designer/$id.tsx
SaraVieira: I pushed something, let me know how to improve it

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,73 @@
...
+    unstable_createFileUploadHandler({
+      maxPartSize: 10_000_000,
+      directory,
+      file: ({ filename }) => filename,
+    })
+  );
+  const imageInfo = formData.get("image");
+  ImageUpload.parse(imageInfo);
+
+  if (imageInfo) {
```
kof, line 57: at this point this check should not be needed unless ts doesn't know zod would throw an error

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,73 @@
...
 
 type Error = {
   errors: "string";
 };
 
+const ImageUpload = z.object({
+  name: z.string(),
+  type: z.string(),
+});
+type ImageUpload = z.infer<typeof ImageUpload>;
```
kof, line 42: having same name works? I was always calling the first one imageUploadSchema etc to avoid such conflicts, wonder why it worked

apps/designer/app/routes/designer/$id.tsx
kof, line 57:
```suggestion
  const imageInfo = ImageUpload.parse(formData.get("image"));
```

apps/designer/app/routes/designer/$id.tsx
SaraVieira: They have it on the docs

https://zod.dev/?id=basic-usage

apps/designer/app/routes/designer/$id.tsx
kof: that's how you validate and make sure TS knows imageInfo type in one step

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,73 @@
...
+      maxPartSize: 10_000_000,
+      directory,
+      file: ({ filename }) => filename,
+    })
+  );
+  const imageInfo = formData.get("image");
+  ImageUpload.parse(imageInfo);
+
+  if (imageInfo) {
+    const info = imageInfo as ImageUpload;
```
kof, line 55: can be removed

apps/designer/app/routes/designer/$id.tsx
kof: Interesting, I must have failed seeing why that worked, need to rename everything if that works

kof: it must be some typescript feature when they are used next to each other that you can augment like this

SaraVieira: Yeah, also thought it was weird but it works great!

apps/designer/prisma/schema.prisma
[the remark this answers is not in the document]
SaraVieira: i pass the type, the size is not in node, only in the frontend :/

I will after this PR take a look on how to add that

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Done! Let me know your thoughts

apps/designer/prisma/schema.prisma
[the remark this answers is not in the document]
SaraVieira: Added to the list for after this is merged

apps/designer/app/routes/designer/$id.tsx
SaraVieira: You removed :D

apps/designer/app/routes/designer/$id.tsx
SaraVieira: done!

apps/designer/.env.example
```diff
@@ -2,4 +2,8 @@ DATABASE_URL=postgresql://user:pass@localhost/webstudio
 AUTH_SECRET="# Linux: `openssl rand -hex 32` or go to https://generate-secret.now.sh/32"
 GH_CLIENT_SECRET=
 GH_CLIENT_ID=
-DEV_LOGIN=true
\ No newline at end of file
+DEV_LOGIN=true
+
+
+# relative to the public directory
+FILE_UPLOAD_PATH="uploads"
```
kof, line 9: should we allow upload anywhere? without contraining to public directory? idk

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
```diff
@@ -0,0 +1,58 @@
...
+    if (submitButtonRef.current) {
+      submitButtonRef.current.click();
+    }
+  };
+
+  return (
+    <Flex gap="3" direction="column" css={{ padding: "$1" }}>
+      <Flex
+        css={{
+          justifyContent: "space-between",
```
kof, line 22: there is a justify prop on Flex

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,82 @@
...
 };
 
+const ImageUpload = z.object({
+  name: z.string(),
+  type: z.string(),
+});
+type ImageUpload = z.infer<typeof ImageUpload>;
+
+export const action: ActionFunction = async ({ request, params }) => {
+  const uploads = path.join(__dirname, "../public");
```
kof, line 44: we shouldn't use __dirname (logic where to upload shouldn't depend on where this code is, if we mode code it breaks), also I think this is still the wrong path to public

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,30 +1,82 @@
...
+        file: ({ filename }) => filename,
+      })
+    );
+
+    const imageInfo = ImageUpload.parse(formData.get("image"));
+
+    const data = {
+      type: imageInfo.type,
+      name: imageInfo.name,
+      path: `/${path.join(folderInPublic, imageInfo.name)}`,
```
kof, line 64: why do you wanna put path into database? we can always have a path based on the file name

kof: is imageInfo.name having a unique part in it?

apps/designer/app/routes/designer/$id.tsx
SaraVieira: It's not, it works

But what should I use?

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Yes, that is unique but that's mostly because then to show it in the frontend you don't have to check if the user is using S3 or wtv you can just have the path as src

apps/designer/.env.example
SaraVieira: Nope, it won't work outside of public

apps/designer/app/routes/designer/$id.tsx
kof: is there anything in env that we can use? kinda need to have a stable reference to a root somewhere

apps/designer/app/routes/designer/$id.tsx
kof: In the frontend you want to show the image name, not the path, right?

apps/designer/.env.example
kof: enforced by remix?

apps/designer/app/routes/designer/$id.tsx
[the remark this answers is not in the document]
SaraVieira: oh, remix does that by default, we do not need to do it, if an image exists they append the timestamp

apps/designer/.env.example
SaraVieira: yap

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Messaed this on discord bt this dirname is the build and thats stable

apps/designer/app/routes/designer/$id.tsx
SaraVieira: yes

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
SaraVieira: Fixed

apps/designer/app/routes/designer/$id.tsx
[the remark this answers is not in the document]
SaraVieira: You can see that here: https://remix.run/docs/en/v1/api/remix#unstable_createfileuploadhandler-node

Issue #127, opened 2022-07-09
Style panel changes are not reflected on canvas on the landing site


kof:
## Steps
1. Click "try" button from landing so that it clones landing project
2. try changing any properties on style panel
3. it saves the changes but they don't apply on canvas

kof:
@SaraVieira 


turned out this is data corruption during migration, breakpoints on existing projects don't have an idea like it is according the schema


```
type Breakpoint = {
    id: string;
    label: string;
    minWidth: number;
};
```

Now I am not sure how easy it is to fix, because that id has to correspond to breakpoint in the CSS rule



```
CssRule = {
    style: Style;
    breakpoint: Breakpoint["id"];
}

```

kof:
Also realized we would have noticed it earlier if we had data validation at runtime, I think its time to add zod

PR #137, opened 2022-07-12
Make sure we have validated at runtime data that isn't part of Prisma

kof:
This is trying to prevent cases like [this](https://github.com/webstudio-is/webstudio-designer/issues/127) in the future. We need to add zod to all ingoing/outgoing data in all routes

I haven't published SDK package, so if you want to test, will have to use yarn watch:sdk and check out [this](https://github.com/webstudio-is/webstudio-sdk/pull/8) on SDK repo. This workflow is soon going to be a thing of the past.

### DIFF
--- a/apps/designer/app/routes/canvas/$projectId.tsx
+++ b/apps/designer/app/routes/canvas/$projectId.tsx
@@ -4,6 +4,7 @@ import { Canvas } from "~/canvas";
 import { loadCanvasData, type ErrorData, type CanvasData } from "~/shared/db";
 import env, { Env } from "~/env.server";
 import { ErrorMessage } from "~/shared/error";
+import { sentryException } from "~/shared/sentry";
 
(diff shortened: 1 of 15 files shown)


SaraVieira:
This looks like a really good upgrade that will help memove the breakpoints and other things in prisma to JSON and tables

SaraVieira (approved):

SaraVieira:
My main concern here is that the sdk is still not in the monorepo, it would make it way easier to make changes without even needing to make a new version

apps/designer/app/routes/canvas/$projectId.tsx
```diff
@@ -21,7 +20,7 @@ export const loader: LoaderFunction = async ({ params }): LoaderReturnTypes => {
   } catch (error) {
     if (error instanceof Error) {
```
kof, line 21: we should probably report to sentry here

Issue #134, opened 2022-07-11
Move SDK into monorepo & setup automatic publishing


kof:
- [x] move SDK into packages
- [x] rename it to sdk-react (or react-sdk) 
- [x] everything that SDK depends on needs to be automatically published, right?

PR #140, opened 2022-07-13
Merge @webstudio-is/sdk to monorepo

andrasbacsai:
- Merge SDK to the repository.
- Rename sdk to `react-sdk`, based on [this](https://github.com/webstudio-is/webstudio-designer/issues/134).


packages/webstudio-sdk/.storybook/main.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
kof, line 1: @okonet should we have the same config? how would you do it, move into a separate package?

packages/webstudio-sdk/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
kof, line 1: please only add here whats not already in base.json

packages/webstudio-sdk/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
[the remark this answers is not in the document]
andrasbacsai: It should.

kof:
- [ ] relativeDeps in root package.json can be removed now

packages/webstudio-sdk/.storybook/main.js
okonet: Depends: if we want one storybook for all packages, then we should create a separate package that would include everything. Or we do local storybooks in each package. I think both is fine right now and we can decide later.

packages/webstudio-sdk/package.json
```diff
@@ -0,0 +1,77 @@
...
+  "name": "@webstudio-is/sdk",
+  "version": "0.5.0",
+  "description": "Webstudio JavaScript / TypeScript API",
+  "author": "Webstudio <github@webstudio.is>",
+  "repository": "git@github.com:webstudio-is/webstudio-sdk.git",
+  "homepage": "https://webstudio.is",
+  "scripts": {
+    "build": "yarn build:mdn-data && yarn build:args && yarn build:prisma && yarn build:lib",
+    "build:mdn-data": "node --experimental-loader esbuild-node-loader ./bin/mdn-data.ts ./src/css",
+    "build:args": "node --experimental-loader esbuild-node-loader ./bin/generate-arg-types.ts './src/components/*.tsx !./src/**/*.stories.tsx'",
```
kof, line 11: lets check if we actually need those flags

packages/webstudio-sdk/.storybook/main.js
okonet: I'm gravitating towards separate storybooks rn since configs can be different and it will be hard to merge them IMO.

packages/webstudio-sdk/package.json
andrasbacsai: If I change the shebang from `#!/usr/bin/env` to `#!/usr/bin/env -S` it works.

packages/webstudio-sdk/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@webstudio-is/sdk",
+  "version": "0.5.0",
+  "description": "Webstudio JavaScript / TypeScript API",
+  "author": "Webstudio <github@webstudio.is>",
+  "repository": "git@github.com:webstudio-is/webstudio-sdk.git",
+  "homepage": "https://webstudio.is",
+  "scripts": {
+    "build": "yarn build:mdn-data && yarn build:args && yarn build:prisma && yarn build:lib",
```
kof, line 9: we don't need to call many of those every time actually, it was already slowing things down because we called it a lot, e.g.
- build:mdn-data only needs to be built when we need something new from the mdn-data package, which will only happen if we update the dependency
- build:args - will be needed when ./src/components/** changes, but I am not sure how we can prevent doing it too many times
- build:prisma - only when schema.prisma changes, think we need to only do it on postinstall automatically and otherwise call manually, when we change schema, it doesn't change often

packages/webstudio-sdk/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "build:mdn-data": "node --experimental-loader esbuild-node-loader ./bin/mdn-data.ts ./src/css",
+    "build:args": "node --experimental-loader esbuild-node-loader ./bin/generate-arg-types.ts './src/components/*.tsx !./src/**/*.stories.tsx'",
+    "build:prisma": "prisma format && prisma generate",
+    "typecheck": "tsc --noEmit",
+    "build:lib": "rm -fr lib && tsc",
+    "postinstall": "prisma generate",
+    "test": "jest",
+    "checks": "yarn typecheck && yarn test",
+    "prepublishOnly": "yarn typecheck && yarn build",
```
kof, line 18: I would change it to
```suggestion
    "prepublishOnly": "yarn checks && yarn build:lib",
```

packages/webstudio-sdk/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "storybook:run": "NODE_OPTIONS=--openssl-legacy-provider start-storybook -p 6006",
```
kof, line 19: these NODE_OPTIONS were used for node 17-18, we rolled back to 16, now they are useless

turbo.json
```diff
@@ -8,6 +8,7 @@
       "outputs": []
     },
     "dev": {
+      "dependsOn": ["@webstudio-is/sdk#build"],
```
kof, line 11: I think dev should not depend on build, everything that was supposed to be built in SDK, should probably be built just once upon installation, after that, SDK is just a typescript code

kof:
Made a first round of review @okonet you probably want to review this too

packages/webstudio-sdk/src/css/categories.ts
```diff
@@ -0,0 +1,253 @@
+import type { StyleProperty } from "./types";
```
okonet, line 1: It feels like this should be a separate package. @kof does this exist in designer somewhere?

packages/webstudio-sdk/src/css/index.ts
```diff
@@ -0,0 +1,8 @@
+export * from "./properties";
+export * from "./units";
+export * from "./types";
+export * from "./get-browser-style";
+export * from "./categories";
+export * from "./keyword-values";
+export * from "./breakpoints";
+export * from "./schema";
```
okonet, line 8: What parts of this already exist in designer and can be re-used? @kof ?

turbo.json
andrasbacsai: In that case the `main` and `types` should be changed in `package.json` of `webstudio-is` package to:
```json
  "main": "src/index.ts",
  "types": "src/index.ts",
```

packages/webstudio-sdk/src/user-props/use-user-props.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (propsData == undefined) return;
+    const nextProps: UserProps = {};
+    for (const prop of propsData.props) {
+      nextProps[prop.prop] = prop.value;
+    }
+    setProps(nextProps);
+  }, [propsData]);
+
+  return props;
+};
```
okonet, line 35: What is this doing in SDK? Feels similar to designer's code but I'm not sure.

packages/webstudio-sdk/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+  "extends": "../../node_modules/tsconfig/base.json",
```
okonet, line 2: Please add the same code comment as here: https://github.com/webstudio-is/webstudio-designer/blob/46d0331671bd030f7a0519bf9d02bf59fa5491ea/apps/designer/tsconfig.json#L2

turbo.json
okonet: I think it shouldn't pre-build it in dev. Instead, it should work with the source code.

packages/webstudio-sdk/.github/workflow/main.yml
```diff
@@ -0,0 +1,26 @@
+name: Main workflow
```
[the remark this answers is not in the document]
kof: I think we want to automate publishing to npm for SDK package though

turbo.json
okonet: @andrasbacsai that sounds good to me or is there a problem with doing so?

packages/webstudio-sdk/src/css/categories.ts
kof: designer is using it from here, yeah, we will split a lot of things into separate packages

packages/webstudio-sdk/src/css/index.ts
kof: designer uses it from sdk,  we will have to split those in different packages and different packages will come with their own schema

packages/webstudio-sdk/.github/workflow/main.yml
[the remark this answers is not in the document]
kof: Actually we will have to also publish every monorepo package SDK will depend on and since we want to be able to shift things easily around, we probably need to be publishing every package to npm

okonet (approved):

kof (approved):

Issue #128, opened 2022-07-09
Tree flickers when selecting different instances


kof:
## Steps

1. create some nested tree or clone landing
2. open a nested item with siblings
3. click on one, than on another

![tree flicker](https://user-images.githubusercontent.com/52824/178117160-8147fce4-c3be-4e39-8cd0-17e52796df10.gif)

PR #145, opened 2022-07-14
Fix tree flicker

rpominov:
Related issue: #128

The flicker is caused by broken open/close animation ( https://www.radix-ui.com/docs/primitives/components/collapsible#animating-content-size )

The animation relies on `--radix-collapsible-content-height` CSS variable. In order for animation to work correctly the variable must be present during the animation.

There were two problems:

1. For the nodes that have children we used to remove the children as soon as the node is collapsed, which caused `--radix-collapsible-content-height` to disappear and animation to use incorrect height value.
2. For the nodes without children, the variable `--radix-collapsible-content-height` is never present, so animation was always broken.


kof (approved):
🎉

Issue #131, opened 2022-07-11
Enable drag&drop in the tree navigator


kof:
We need to be able to reorder items in the tree navigator because that's often easier than doing it directly on the canvas

## Todo
- [x] enable drag&drop into instances
- [x] ensure user can't drag an instance into an instance that doesn't support instances as children: e.g. paragraph or headline should not support dragging in anything (see canAcceptChild())
- [x] enable dragging horizontally to change the parent but stay at the same level
- [x] disable dragging inline instances out of their parent (e.g. Paragraph can have a Bold), but Bold can't exist top level, only as part of paragraph
- [x] enable scrolling when list is big
- [x] enable hovering over a collapsed tree item, when holding a bit - it opens the item and you can continue dragging inside of it
- [x] animation starts when value changes in style panel (https://share.descript.com/view/qsbbe2UQlVi)

kof:
Definitely check out how Webflow handles this interaction, it works very well!

kof:
cc @kkga @taylornowotny anything else you know we need to be aware regarding drag and drop in tree navigator?

kof:
@taylornowotny do you think we can have some design proposal next week for this? We can start building with "any" design for sure, but I think it would be good to have at least a semi-final version

taylornowotny:
More suggestions:

- [ ] accordion-style opening/closing of parent elements to show/hide the children
- [ ] a collapse-all button
- [ ] ability to pin the navigator panel so it's always open on the left side (I have this permanently enabled in Webflow)

There's also the issue of naming elements.
In Figma you can name layers directly (and most users don't). In Webflow the element name in the navigator is the name of the class.
What do you think our approach should be here?

taylornowotny:
Another thing, how do you want to handle the body HTML tag? Can users change the styles on the tag?

In Webflow the body is always the top level parent. It can be given a class but doesn't need to.

It's useful to be able to select the body element for things like setting the typography defaults for the project, or changing the body to overflow:hidden to prevent page scrolling when a modal is open, for example.

kof:
>  accordion-style opening/closing of parent elements to show/hide the children

Can you clarify more specifically what you think should happen?

kof:
>  ability to pin the navigator panel so it's always open on the left side (I have this permanently enabled in Webflow)

[...]

PR #146, opened 2022-07-15
Drag&drop in the tree (WIP)

rpominov:
Related issue: #131

For my own reference:

- [react-sortable-tree.js](https://github.com/frontend-collective/react-sortable-tree/blob/master/src/react-sortable-tree.js), [Storybook](https://frontend-collective.github.io/react-sortable-tree/?path=/story/basics--minimal-implementation)
- https://react-dnd.github.io/react-dnd/examples/sortable/simple, [src](https://github.com/react-dnd/react-dnd/tree/gh-pages/examples_ts/04-sortable/simple/src)
- [dnd-kit / SortableTree.tsx](https://github.com/clauderic/dnd-kit/blob/master/stories/3%20-%20Examples/Tree/SortableTree.tsx), [Storybook](https://master--5fc05e08a4a65d0021ae0bf2.chromatic.com/?path=/story/examples-tree-sortable--basic-setup)

### DIFF
--- a/apps/designer/app/shared/tree-utils/move-instance.ts
+++ b/apps/designer/app/shared/tree-utils/move-instance.ts
@@ -1,0 +1,59 @@
+import { type Instance } from "@webstudio-is/react-sdk";
+import { insertInstanceMutable } from "./insert-instance";
+
+type InstanceMoveSpec = {
+  id: Instance["id"];
+  newParentId: Instance["id"];
+  newIndex: number;
+};
+
+const extractInstance = (
+  treeNode: Instance,
+  id: Instance["id"]
+): Instance | null => {
+  const index = treeNode.children.findIndex(
+    (child) => typeof child !== "string" && child.id === id
+  );
+
+  if (index !== -1) {
+    const instance = treeNode.children[index];
+
+    // for TypeScript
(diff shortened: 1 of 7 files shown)


apps/designer/app/shared/tree-utils/move-instance.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (extracted !== null) {
+        return extracted;
+      }
+    }
+  }
+
+  return null;
+};
+
+export const moveInstanceMutable = (
```
kof, line 42: this looks like what is usually called "parenting", which we already do in useSubscribeReparentInstance: basically its 2 calls: deleteInstanceMutable() and insertInstanceMutable()

kof: also note that we use mutating functions because we call them inside a createTransaction that uses immer.js data structures, so the mutable functions don't actually mutate a datastructure that will be passed anywhere else, they mutate so that immer can record the changes and create a new immutable data structure

kof:
do we still need this oen?

rpominov:
No, I'll start another one when I get to it

PR #147, opened 2022-07-15
Contraints for dropping instances only in parents that accept children

kof:


kof:
Closing, will be replaced by a new approach

PR #148, opened 2022-07-16
Add more image info to DB

SaraVieira:
Related to #37

Adds:

- File size
- Format
- Width
- Height


apps/designer/app/routes/designer/$id.tsx
```diff
@@ -54,14 +52,15 @@ export const action: ActionFunction = async ({ request, params }) => {
...
     imagesInfo.forEach((image) => {
       const data = {
-        type: image.type,
         name: image.name,
         path: `${path.join("/", folderInPublic, image.name)}`,
       };
+      const absolutePath = path.join(directory, image.name);
       const projectId = params.id as string;
-      db.assets.create(projectId, data);
+      db.assets.create(projectId, data, absolutePath);
```
SaraVieira, line 63: Anything uploaded will not send an absolute path since we can fetch this new image and get it's size

kof, line 63: this is an async function and you want to create those in parallel while the following return should return after they finished

apps/designer/app/shared/db/assets.server.ts
```diff
@@ -15,11 +18,30 @@ export const loadByProject = async (projectId?: Project["id"]) => {
 
 export const create = async (
   projectId: Project["id"],
-  values: { type: string; name: string; path: string }
+  values: { name: string; path: string },
+  absolutePath?: string
 ) => {
+  let size = 0;
+  const image = await sharp(absolutePath);
+  const metadata = await image.metadata();
```
kof, line 26: what are  perf characteristics of those calls?

apps/designer/app/shared/db/assets.server.ts
```diff
@@ -15,11 +18,30 @@ export const loadByProject = async (projectId?: Project["id"]) => {
...
+
+  // if there is no absolute path then it means it's a remote image and we can use fetch
+  if (absolutePath) {
+    size = fs.statSync(absolutePath).size;
```
kof, line 30: never use sync calls on a server after the bootstrap phase, that's very important

kof: also didn't we receive size from the browser as we were uploading? I am sure we did

kof (changes requested):
first round done

apps/designer/app/shared/db/assets.server.ts
kof: we can make a naive bench by running this in a for loop 1000 times and measure average time spent there

SaraVieira: https://sharp.pixelplumbing.com/performance#the-task

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -54,14 +52,15 @@ export const action: ActionFunction = async ({ request, params }) => {
     );
 
     const imagesInfo = ImagesUpload.parse(formData.getAll("image"));
+
     imagesInfo.forEach((image) => {
       const data = {
-        type: image.type,
         name: image.name,
         path: `${path.join("/", folderInPublic, image.name)}`,
```
kof, line 59:
```suggestion
        path: path.join("/", folderInPublic, image.name),
```

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
```diff
@@ -1,37 +1,29 @@
...
 
   return (
     <Flex gap="3" direction="column" css={{ padding: "$1" }}>
       <Flex justify="between">
         <Heading>Assets</Heading>
-        <Form method="post" encType="multipart/form-data">
+        <Form
+          method="post"
+          encType="multipart/form-data"
+          onChange={(event) => submit(event.currentTarget)}
```
kof, line 19: oh yeah, much better

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Done!

apps/designer/app/shared/db/assets.server.ts
SaraVieira: Fixed!

apps/designer/app/shared/db/assets.server.ts
kof: also what if upload was to s3, will those download again to get meta?

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Fixed!

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
SaraVieira: Waaaay cleaner

apps/designer/app/shared/db/assets.server.ts
kof: > https://sharp.pixelplumbing.com/performance#the-task

that's a bit too generic, what's the cost of those specific calls is the question

kof: and also how can you do it if you don't have a local file

SaraVieira: yeah because idk why remix doesn't send the file size or the actual file after doing the file upload, at least for now.

Will see what happens with the other one, I think I have to make my own handler so I can make sure to still return the image array buffer and I can pass that to sharp

SaraVieira:
weird I have no errors locally

SaraVieira:
Fixed by making sure we run generate on prisma

apps/designer/app/shared/db/assets.server.ts
```diff
@@ -15,11 +18,30 @@ export const loadByProject = async (projectId?: Project["id"]) => {
...
+  // if there is no absolute path then it means it's a remote image and we can use fetch
+  if (absolutePath) {
+    size = await (await fsPromises.stat(absolutePath)).size;
+  } else {
+    const arrayBuffer = await fetch(values.path).then((rsp) =>
+      rsp.arrayBuffer()
+    );
+    size = arrayBuffer.byteLength;
+  }
+
```
kof, line 37: I am sure we don't have to do it, its all available during upload, because browser sends that info, just need to figure out how to get that

apps/designer/app/shared/db/assets.server.ts
kof: there might be a way to get this out of remix, check out their implementation

kof: at least the size

kof: Regarding metadata, if this lib can only make like 30 images per second, its a nogo, way too slow to be in the main server process. Also if we have to upload to S3, you would have to redownload. 

Instead I would try looking into approaches which can take a few bytes and read exif data from there, without downloading the full image or needing access directly to the file system, this way we can get those bytes from any storage and parse the exif data

kof: For context: image exif data is always first few bytes of the image, but majority of graphic tools read the entire image because they are preparing to do more stuff with the image. I used a tiny parser many years ago, it will operate in microseconds area.

SaraVieira:
turns out I didn't know zod remove the parts of the object that I didn't declare on the types and that's why I didn't have access to the size and function to turn the file into an  array buffer

apps/designer/app/shared/db/assets.server.ts
SaraVieira: Updated everything

way better now

apps/designer/app/shared/db/assets.server.ts
SaraVieira: ignore this, it uses the array buffer now

SaraVieira:
We will not be downloading the image anymore, i send the array buffer over

apps/designer/app/shared/db/assets.server.ts
```diff
@@ -15,11 +16,18 @@ export const loadByProject = async (projectId?: Project["id"]) => {
 
 export const create = async (
   projectId: Project["id"],
-  values: { type: string; name: string; path: string }
+  values: { name: string; path: string; size: number; arrayBuffer: Uint8Array }
 ) => {
+  const image = sharp(values.arrayBuffer);
+  const metadata = await image.metadata();
```
kof, line 22: still neeed to measure this, I don't trust it until I see some numbers

SaraVieira: https://sharp.pixelplumbing.com/api-input

It actually only uses the header of the image for the metadata

SaraVieira: The 30 per second is opening an image, resizing it and compressing it

SaraVieira: https://sharp.pixelplumbing.com/performance#running-the-benchmark-test

I guess you can run it for metadata

kof:
- [ ] bug: click on upload button > select image > upload > click on upload button > it has already uploaded the last image again

kof:
- [ ] sort order - last uploaded image should come first

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
```diff
@@ -1,37 +1,29 @@
 import { ImageIcon } from "~/shared/icons";
```
kof, line 1: image-upload.tsx should probabbly be asset-manager.tsx

kof:
- [x] did some primitive benchmarking of `sharp` calls, it gets meta data of an image (3mb) 1000 times within 200ms, then I tried to do the same with a 200kb image, perf is similar, that tells me they really only read a few bytes. So this is really fast, I am happy.

apps/designer/app/shared/db/assets.server.ts
```diff
@@ -15,11 +16,19 @@ export const loadByProject = async (projectId?: Project["id"]) => {
...
+  const image = sharp(values.arrayBuffer as Uint8Array);
+  const metadata = await image.metadata();
   const newAsset = await prisma.asset.create({
     data: {
-      ...values,
+      name: values.name,
+      path: values.path,
+      size: values.size,
+      format: metadata.format,
+      ...(metadata.width ? { width: Math.round(metadata.width) } : {}),
```
kof, line 30: should we round? idk

kof (approved):
🎉 see a few todos

kof:
> Fixed by making sure we run generate on prisma

it needed prisma migrate, which we currently don't run automatically, not sure what's the best practice when to run it

apps/designer/app/designer/features/sidebar-left/panels/image-upload.tsx
SaraVieira: Done!

apps/designer/app/shared/db/assets.server.ts
SaraVieira: yeah, it's incredibly rare that an image is like 1980.2 px but if it was we would break the db

kof: yeah, I wonder why an image can be floating point width and if we round, where will we have potential other issues, as an option we could have a db field type that supports float numbers

SaraVieira: So basically I changed the logic to always return a float so if the number is 10 it always returns 10.0 and the db type is float, this way it will never break

SaraVieira:
Can this be merged?

Issue #31, opened 2022-05-04
Show a drop indicator


kof:
To better understand where the instance is going to drop, we should show a line similar to https://craft.js.org/

smotched:
Do you think a tooltip similar to mentioned project is needed. I think it could be useful to distinguish between dragging vs highlighting text. Webstudio currently does both which could be confusing.

rpominov:
Done in #176

Issue #7, opened 2022-03-24
Restrict dropping instances into certain instance types


kof:
Some instances can't accept children at all (Input), some only accept specific components, e.g. div can't be inside paragraph

## Todo
- [x] prevent inserting into components which don't accept an instance
- [x] outline during drag the instance that will accept the insteance instead of the one that can't
- [x] when single-click adding an instance while the target instance can't accept it automatically select the closest parent instance that can ([duplicate issue ](https://github.com/webstudio-is/webstudio-designer/issues/109))

kof:
Todo: check if we can create a function that inserts an element into the target element before it gets inserted to validate if there is a thrown "validateDOMNesting" before it actually happens and then prevents inserting and then show an error to the user also  make the target show it doesn't accept the node

kof:
This is where this warning is coming from, its not DOM, its react https://github.com/facebook/react/blob/main/packages/react-dom/src/client/validateDOMNesting.js

we should probably just copy that and use as pre-validation, but also this is only logging in dev mode and not affecting any production

kof:
Deprioritizing this issue for now

kof:
Actually found this library that can be introduced easily https://github.com/MananTank/validate-html-nesting

kof:
We should at least disable for now basic things: e.g. h1 should get into Text/Paragraph etc. and respect the .canAcceptChild function on primitives

rpominov:
Can we close this now, or do we still want to use something like [validate-html-nesting](https://github.com/MananTank/validate-html-nesting)?

kof:
I would say lets close, we will potentially have to come back to it at some point once we see more concrete use cases.
Originally I thought of a generic validator that is based on HTML spec is a good idea because it would work for any components, but then I realized that you don't want certain components be dropped into certain components: e.g. div into h1, even though it would technically still work, so a generic HTML based linter is not quite solving the problem. We need to think more about this problem to enable something that would work for all components and for custom sdk's.

Maybe we can come up with a higher level API that would have HTML constraints backed in and on top of that allow caller to specify additional constraints ... adk

kof:
Actually, I just recall the original reason is that React has its own built-in validator and warns in dev mode when nodes are nested illegally. It's a much smaller subset of HTML it seems and I don't know how correct it actually is.

PR #149, opened 2022-07-17
Drag & drop prototype

kof:
## Issues

- https://github.com/webstudio-is/webstudio-designer/issues/131
- https://github.com/webstudio-is/webstudio-designer/pull/147
- https://github.com/webstudio-is/webstudio-designer/issues/7
- https://github.com/webstudio-is/webstudio-designer/issues/31

## Terminology

- drop target - item accepting draggable element for drop
- drag item - item that has originally received pointerdown event
- drop index - specific index at which item is going to drop relative to children
- drop target outline - frame that outlines a drop target
- placement indicator - a line that shows where the item is going to be placed
- edge detection - when pointer is near (~5px) an edge of a drop target

## Principles

- you can drop anywhere. At any point of time, there is a drop target and an index that would define the best possible position based on the current pointer position. There will never be an error state. If the best possible position is the original position - so be it.
- at any point of time we can calculate drop target outline rect and placement indicator rect

## Architecture

- 3 levels of abstraction will be needed: primitives, design system, application
- primitives level will handle events and other logic that knows nothing about rendering specific concerns
- design system level are components that actually do rendering (only tree in our case), but they know nothing about application-level specific concerns: (e.g. specific icon/text of each instance is still application concern, tree renders tree with some basic markup and styles specific to our design system, similar to other DS components) e.g. tree also only knows the a subset of the instance data structure that can be used for any tree `({id, children: [{id., children}] })`
- application level knows full data structure (Instance), it uses the ds component for the tree, on canvas we have to use primitives directly

## Todo

- [x] handle drag events only if mouse down happened on draggable element (onStart({element, stop}): true navigator and canvas can have different attributes, on canvas it's data-component and data-id right now, so consumer will decide by calling `stop()`
- [x] `startThreashold` - only start drag when user dragged for ~5px to avoid clicks with slight drag trigger drag guesture
- [x] when dragging over an element, we need to run a callback `onHover({element})`, which will check in tree data if the current element can't accept the drag item see [findClosestAcceptingParent](https://github.com/webstudio-is/webstudio-designer/pull/147/files#diff-a54ce4ea3a1280bcb113c7a13ac7743d97d83fb760b334d683bec391ae2489bcR10) and will inform drop target outline as well as placement indicator
- [x] Edge detection - when sorting inside a list, nearing the edge of the hovered element should define position for placeholder indicator and drop index after/before that element
- [x] stop searching for parent as an acceptable drop target at the currentTarget (the element where we added useMove props)
- [x] handle release outside of the window
- [x] onEnd() drag is over, placement indicator and target outline can be removed here (target outline is external to this library since we only use it on canvas, but not in navigator)
- [x] holding over a drop target for holdThreashold=1s should call `onHoldOver` in the primitive. Tree component will use it  to open the item and enable finding new drop target inside
- [x] detect horizontal drag (20px horizontal max 5px vertical) `onShift({direction: "left"})` (no need to define this callback on canvas)
- [x] Placement Indicator component should render a horizontal or vertical line, it can be "dumb" and have top, left, width, height as props. When horizontal, the width should correspond the width of the drop target, height 1-2px. When vertical - the other way around.
- [x] enable scrolling when mouse is the edge of scrollContainer, which is dom node defined by consumer
- [x] To draw the placement indicator, we need some smart way of understanding where the drag item will be placed once dropped. Depending on layout, it can be vertical (flex-direction: vertical) or it can be horizontal. Maybe we need to render an invisible probe and actually test it out? Because calculating different layout situations, ourselves will be very hard to do and potentially always buggy
- [x] Edge of a drop target (~5px) means user wants to drop __after__ the current drop target, meaning we have to redefine the drop target to it's parent and drop after the previous drop target
- [x] test with touch device (pointer events will work there anyways)


rpominov:
Makes sense overall.

The following two items seem to describe the same:

> Edge detection - when sorting inside a list, nearing the edge of the hovered element should define position for placeholder indicator and drop index after/before that element

> Edge of a drop target (~5px) means user wants to drop after the current drop target, meaning we have to redefine the drop target to it's parent and drop after the previous drop target

-------------

> holding over a drop target for holdThreashold=1s should call onHoldOver in the primitive. Tree component will use it to open the item and enable finding new drop target inside

Not sure how to do this. Do you mean generate an even on a DOM node, which then can be picked up by the component that have rendered that node?

apps/designer/app/shared/design-system/components/primitives/dnd/use-drop-target.ts
```diff
@@ -2,15 +2,29 @@ import { useRef } from "react";
...
   const areaRef = useRef<Area>("center");
+  const holdTimeoutRef = useRef<NodeJS.Timeout>();
+
+  const checkHold = (target: HTMLElement) => {
+    clearTimeout(holdTimeoutRef.current);
+    holdTimeoutRef.current = setTimeout(() => {
+      if (target === targetRef.current) {
+        onHold({ target });
+      }
+    }, holdTimeThreshold);
```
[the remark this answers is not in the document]
kof: good catch

apps/designer/app/shared/design-system/components/primitives/dnd/handlers.stories.tsx
```diff
@@ -40,32 +41,59 @@ export const Playground = () => {
...
   });
 
   return (
-    <Box {...dragProps} ref={rootRef} css={{ display: "flex" }}>
+    <Box
+      {...dragProps}
+      ref={(el) => {
+        dropTargetHandlers.rootRef(el);
+
+        // TODO: do same as above for consistency
```
kof, line 80: ref containers are bad for composition because you can't react to them as an effect, better to always pass a function like above, so it basically becomes reactive to ref chamges

apps/designer/app/shared/design-system/components/primitives/dnd/use-move.ts
```diff
@@ -0,0 +1,332 @@
...
+      }
+
+      // FIXME: this is a temporary fix
+      // we need to figure out how to properly support keyboard and touch
+      if (
+        !(originalEvent.target instanceof HTMLElement) ||
+        !("pageX" in originalEvent)
+      ) {
+        return;
+      }
```
rpominov, line 127: `useMove` uses touch events only as a fallback when Pointer events are not supported. I think we should remove this fallback as it will simplify things for us. https://caniuse.com/pointer

I would also remove the keyboard support for now, but not sure. Do we need to make it work?

apps/designer/app/shared/design-system/components/primitives/dnd/canvas.stories.tsx
```diff
@@ -0,0 +1,308 @@
...
+  return (
+    <>
+      <Box
+        css={{
+          background: "white",
+          padding: 10,
+          width: 500,
+          height: 500,
+          overflow: "auto",
+          touchAction: dragItemId === undefined ? "none" : "auto",
```
rpominov, line 284: On touch devices DnD will always clash with scroll. Not sure it's possible to make it work well.

This seem to work on the demos, although I only tested in the Chrome's responsive mode.

rpominov:
I think this is good enough for prototype. I'd merge this, and start to work on the integration.

apps/designer/app/shared/design-system/components/primitives/dnd/canvas.stories.tsx
rpominov: Sorry I messed up the condition. I was going to try `dragItemId === undefined ? "auto" : "none"`, which doesn't work.

For now the only thing that works for sure is simply `touchAction: "none"`, but it disables the scroll.

`useMove` does the same in their demo: https://react-spectrum.adobe.com/react-aria/useMove.html

andrasbacsai:
After #170 is merged, you can merge main to this, and it will work.

apps/designer/app/shared/design-system/components/primitives/dnd/use-move.ts
kof: agree, please make a todo

PR #150, opened 2022-07-17
Allow image upload to S3

SaraVieira:
related https://github.com/webstudio-is/webstudio-designer/issues/37


apps/designer/app/routes/designer/$id.tsx
```diff
@@ -12,6 +12,10 @@ import * as db from "~/shared/db";
 import config from "~/config";
 import env from "~/env.server";
 import { ImagesUpload } from "~/designer/features/sidebar-left/types";
+import { s3UploadHandler } from "~/shared/uploads/s3-upload-handler";
+import { sentryException } from "~/shared/sentry";
+import { uploadToS3 } from "~/shared/uploads/upload-to-s3";
+import { uploadToDisk } from "~/shared/uploads/upload-to-disk";
```
kof, line 18: we should have all s3 specific logic one place and only use single interface everywhere else that is universal

kof: actually local upload is also one of the cases that should be behind a universal upload interface

apps/designer/app/shared/db/assets.server.ts
```diff
@@ -17,6 +17,10 @@ export const loadByProject = async (projectId?: Project["id"]) => {
   return assets;
 };
 
+const forceFloat = (number: number) => {
+  return parseFloat(Number(number).toFixed(1));
```
kof, line 21: Why Number(), can it be something else?

SaraVieira:
For some reason TS thinks this without it will return a string

apps/designer/app/shared/uploads/upload-to-disk.ts
```diff
@@ -0,0 +1,30 @@
...
+}) => {
+  const imagesInfo = ImagesUpload.parse(formData.getAll("image"));
+  const allInfo = imagesInfo.map(async (image) => {
+    const arrayBuffer = await image.arrayBuffer();
+    const data = {
+      name: image.name,
+      path: path.join("/", folderInPublic, image.name),
+      size: image.size,
+      arrayBuffer,
+    };
```
kof, line 22: isn't data supposed to be the same for both uploaders?

apps/designer/app/shared/uploads/array-buffer-helpers.ts
```diff
@@ -0,0 +1,25 @@
+export const mergeUint8Arrays = (a: Uint8Array, b: Uint8Array) => {
+  const c = new Uint8Array(a.length + b.length);
+  c.set(a);
+  c.set(b, a.length);
+
+  return c;
+};
+
+export const getFilenameAndExtension = (filename: string) => {
```
kof, line 9: for this there are node built-in functions

apps/designer/app/shared/uploads/upload-to-disk.ts
```diff
@@ -0,0 +1,30 @@
+import { ImagesUpload } from "~/designer/features/sidebar-left/types";
```
kof, line 1:
```suggestion
import type { ImagesUpload } from "~/designer/features/sidebar-left/types";
```

apps/designer/app/shared/uploads/upload-to-disk.ts
```diff
@@ -0,0 +1,30 @@
...
+import path from "path";
+
+export const uploadToDisk = async ({
+  formData,
+  projectId,
+  folderInPublic,
+}: {
+  formData: FormData;
+  projectId: string;
+  folderInPublic: string;
```
kof, line 12: lets finally discuss this

apps/designer/app/shared/db/assets.server.ts
kof: see no error, fixed

SaraVieira: Hm?

packages/asset-uploader/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "module": "commonjs",
+    "target": "ES2019",
+    "strict": true,
+    "baseUrl": ".",
+    "paths": {
+      "~/*": ["./src/*"]
+    },
+    "outDir": "./lib"
+  }
+}
```
kof, line 16: you don't need everything that has to do with browser, because that's package for node only

kof: we probably need a default tsconfig/node.json  for all pure node packages

packages/asset-uploader/jest.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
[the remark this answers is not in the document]
kof: please reuse something for both sdk and this one, I think it makes sense to have 2 different once, e.g. one for react, one for node, the node one doesn't need majority of stuff

packages/asset-uploader/tsconfig.json
SaraVieira: Updated it but doesn't fix anything :/

packages/asset-uploader/jest.config.js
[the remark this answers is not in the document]
kof: I thought this is pure node code, so all the jsx jazz etc is not needed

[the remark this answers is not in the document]
kof: oh wait you don't have jsx, its just ts and we need to run ts through esbuild, I think your jest config is perfectly fine if that works, lets just move it into a separate package so that we can reuse it for other node packages

apps/designer/app/shared/uploads/array-buffer-helpers.ts
SaraVieira: there is for extension, changed it

packages/asset-uploader/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const uploadAsset = async ({
+  request,
+  projectId,
+  db,
+  dirname,
+}: {
+  request: Request;
+  projectId: string;
+  db: unknown;
```
kof, line 19: this should probably have a @todo

packages/asset-uploader/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  dirname: string;
+}) => {
+  const IS_S3_UPLOAD =
```
kof, line 22: if that's a boolean it should be camel case and Boolean(....)

packages/asset-uploader/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+        db,
+      });
+    }
+
+    return {
+      ok: true,
+    };
+  } catch (error) {
+    if (error instanceof Error) {
+      throw new Error(error.message);
```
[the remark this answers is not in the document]
kof: if you are handling it at the consumer, no need to catch here, you catch to handle or to add context to the erorr

packages/asset-uploader/src/targets/s3/uploader.ts
```diff
@@ -0,0 +1,23 @@
...
+    const imagesInfo = formData.getAll("image") as Array<string>;
+    const allInfo = imagesInfo.map(async (image) => {
+      const uploadedImage = JSON.parse(image);
+      const newAsset = await db.assets.create(projectId, uploadedImage);
+      return newAsset;
+    });
+    await Promise.all(allInfo);
+  } catch (error) {
+    if (error instanceof Error) {
+      throw new Error(error.message);
```
kof, line 20: same

kof (changes requested):
almost there 😅

packages/asset-uploader/src/index.ts
kof: also maybe if some of those env variables are not present, we could show an error at the validator in the env, because if all 3 only work together as one thing, it should be obvious that all 3 need to be defined in the env

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -37,48 +31,17 @@ type Error = {
...
-  const folderInPublic =
-    process.env.FILE_UPLOAD_PATH || config.defaultUploadPath;
-  const directory = path.join(uploads, folderInPublic);
-  try {
-    const formData = await unstable_parseMultipartFormData(
+  if (request.method === "POST") {
+    await uploadAssets({
       request,
-      unstable_createFileUploadHandler({
-        maxPartSize: 10_000_000,
```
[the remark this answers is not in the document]
SaraVieira: We do but its bad
image

Added a todo

packages/asset-uploader/src/index.ts
SaraVieira: Done!

packages/asset-uploader/src/index.ts
SaraVieira: Done and done!

It was actually four so I am glad I checked

packages/asset-uploader/src/index.ts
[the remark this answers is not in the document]
SaraVieira: valid point, fixed

packages/asset-uploader/src/targets/disk/upload.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      const newAsset = await db.assets.create(projectId, data);
+
+      return newAsset;
+    });
+
+    await Promise.all(allInfo);
+  } catch (error) {
+    if (error instanceof Error) {
+      throw new Error(error.message);
```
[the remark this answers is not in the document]
SaraVieira: also fixed

packages/asset-uploader/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      await uploadToDisk({
+        projectId,
+        formData,
+        folderInPublic,
+        db,
+      });
+    }
+
+    return {
+      ok: true,
```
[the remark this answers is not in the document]
SaraVieira: Fixed and now the route returns all the uploaded files in an array in case we want to show a tooltip or something

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    // these upload handlers only allow strings to be returned
+    return JSON.stringify({
+      metadata,
+      name: key,
+      path: newFile.Location,
+    });
+  } catch (error) {
+    if (error instanceof Error) {
+      throw new Error(error.message);
```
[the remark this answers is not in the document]
SaraVieira: removed

packages/asset-uploader/src/targets/s3/uploader.ts
SaraVieira: also removed

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -37,47 +31,21 @@ type Error = {
...
+        request,
+        projectId: params.id,
+        db,
+        dirname: __dirname,
+      });
       return {
-        errors: error.message,
+        assets,
       };
+    } catch (error) {
```
kof, line 45: you are missing the point of using try/catch, its used to handle the errors, but you just rethrow which is the same as not handling them

handling in this case is returning a response that is parsed by the client to show user an error, that's why we used in other cases `return {errors: 'something'}`

apps/designer/package.json
```diff
@@ -126,6 +126,7 @@
     "esbuild-node-externals": "^1.4.1",
     "esbuild-node-loader": "^0.8.0",
     "eslint-config-custom": "*",
+    "@webstudio-is/asset-uploader": "*",
```
kof, line 129: why is it a dev dependency?

kof: shouldn't it be dependency?

packages/asset-uploader/src/helpers/array-buffer-helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+export const mergeUint8Arrays = (a: Uint8Array, b: Uint8Array) => {
+  const c = new Uint8Array(a.length + b.length);
+  c.set(a);
+  c.set(b, a.length);
+
+  return c;
+};
+
+export const getFilenameAndExtension = (filename: string) => {
+  return [
```
kof, line 12: check out path.basename() and path.extname(), just use it where you need this, no need for custom utility, at least I don't see a point

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -0,0 +1,57 @@
...
+    Key: key,
+    Body: uint8Array,
+    ContentType: contentType,
+    Metadata: {
+      filename,
+    },
+  };
+
+  const client = new S3Client({
+    endpoint: process.env.S3_ENDPOINT ?? "",
```
kof, line 37: I think if any of those variables are missing, it should throw, fallback to empty strings will only obfuscate the error potentially, in all of those cases, potentially you could reuse a zod schema here and in env checks to make sure all env variables are present or its a hard failure

packages/asset-uploader/src/helpers/array-buffer-helpers.ts
SaraVieira: Path.basename returns the whole file name and i am already using path.extname

packages/asset-uploader/src/targets/s3/handler.ts
SaraVieira: It already is in the env check

kof: yes, but the logic suggests it can be empty when I read it I may not know it, also assume its a package with a public API the checks should happen here as well at the beginning and the logic should be clear that those are not optional

SaraVieira: I fixed it, let me know if its okay

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -44,7 +44,9 @@ export const action: ActionFunction = async ({ request, params }) => {
       };
     } catch (error) {
       if (error instanceof Error) {
-        throw new Error(error.message);
+        return {
+          error: error.message,
```
kof, line 48: have you tried rendering that error? I am sure its not going to work because its `errors`

packages/asset-uploader/src/targets/s3/handler.ts
SaraVieira: even after parsing, TS thinks that the env can be undefined I don't know why

apps/designer/app/routes/designer/$id.tsx
SaraVieira: I didn't know that remix had a default for that honestly, I assumed I would have to get it manually

SaraVieira: done

SaraVieira: Showing the error now too

kof: nono, its just us, I am not even sure that's the best way, just got it from some examples

kof: all it does it returning that object into outlet component below where we handle it

SaraVieira: But yeah it looks good, i just didn't know we had a standard

apps/designer/app/shared/db/misc.server.ts
```diff
@@ -1,5 +1,8 @@
+import { uploadAsset } from "@webstudio-is/asset-uploader";
 import * as db from "./";
 
+export const uploadAssets = uploadAsset;
```
kof, line 4: why reexporting though instead of using directly? maybe I miss something but I don't see why

packages/asset-uploader/src/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const DEFAULT_UPLPOAD_PATH = "uploads";
+export const S3_ENV_KEYS = [
```
kof, line 2: looks like those are not all keys, but only the once that are required, would call it then 
```suggestion
export const S3_REQUIRED_ENV_KEYS = [
```

kof: the entire thing with the keys looks like something that will be quite a bit to manage, I would really make a zod schema with optional keys and defaults in one central place, then call it once and export everything as a single config object where we have everything validated and in one place for the s3 uploader

packages/asset-uploader/src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export const ImagesUploadedSuccess = z.object({
+  Location: z.string(),
+});
+export type ImagesUploadedSuccess = z.infer<typeof ImagesUploadedSuccess>;
+
+export const s3EnvVariables = z.object({
+  S3_ENDPOINT: z.string(),
+  S3_REGION: z.string(),
+  S3_ACCESS_KEY_ID: z.string(),
+  S3_SECRET_ACCESS_KEY: z.string(),
```
kof, line 20: Oh I missed this one, so the thing is its not all variables, there are some optional, you can use z.optional or something for this and then you can use just the return from schema.parse and never have to deal with accessing process.env again, have guaranteed values

packages/asset-uploader/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  db,
+  dirname,
+}: {
+  request: Request;
+  projectId: string;
+  db: unknown;
+  dirname: string;
+}) => {
+  const uploads = path.join(dirname, "../public");
+  const folderInPublic = process.env.FILE_UPLOAD_PATH || DEFAULT_UPLPOAD_PATH;
```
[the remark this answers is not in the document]
SaraVieira: updated

packages/asset-uploader/src/types.ts
SaraVieira: all done!

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -37,56 +31,36 @@ type Error = {
...
 
 const DesignerRoute = () => {
+  const actionData = useActionData();
   const data = useLoaderData<Data | Error>();
   if ("errors" in data) {
     return <p>{data.errors}</p>;
   }
+  if (actionData && "errors" in actionData) {
+    return <p>{actionData.errors}</p>;
+  }
```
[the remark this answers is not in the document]
SaraVieira: Yes, updated it to use that

apps/designer/app/shared/db/misc.server.ts
kof: oh that's because remix makes sure to not load serverside code on the client, I think if you put the logic inside asset-uploader into a .server.ts file you won't need this

kof: but also tbh, maybe its better to do it this way, because a standalone package donesn't have to be concerned with remix specific stuff, but please name uploadAsset(s) consistently

apps/designer/env-check.js
```diff
@@ -14,6 +14,13 @@
 require("dotenv").config();
 
 const REQUIRED_ENVS = ["DATABASE_URL", "AUTH_SECRET"];
+const S3_KEYS = [
```
kof, line 17: hmm, maybe I am missing something but why is this one different than S3_REQUIRED_ENV_KEYS?

apps/designer/app/shared/db/misc.server.ts
SaraVieira: it also doesn't work lol

changed the names

apps/designer/app/shared/db/misc.server.ts
```diff
@@ -1,5 +1,8 @@
+import { uploadAssets as assetUploaderAssetsUploader } from "@webstudio-is/asset-uploader";
 import * as db from "./";
 
+export const uploadAssets = assetUploaderAssetsUploader;
+
```
kof, line 5:
```suggestion
export { uploadAssets  } from "@webstudio-is/asset-uploader";
import * as db from "./";
```

kof: you know you can directly export from a package?

packages/asset-uploader/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  dirname: string;
+}) => {
+  const uploads = path.join(dirname, "../public");
+  const directory = path.join(uploads, fsUploadVars.FILE_UPLOAD_PATH);
+  const formData = await unstable_parseMultipartFormData(
+    request,
+    isS3Upload
+      ? (file) => s3UploadHandler(file)
+      : unstable_createFileUploadHandler({
+          maxPartSize: 10_000_000,
```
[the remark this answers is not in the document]
SaraVieira: I actually set it in a common thing for both and now I also check the max file size when uploading to s3

apps/designer/env-check.js
SaraVieira: You can't import from that module because its an es module

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -37,55 +32,35 @@ type Error = {
...
-    };
-  } catch (error) {
-    if (error instanceof Error) {
+  if (request.method === "POST") {
+    try {
+      const assets = await uploadAssets({
+        request,
+        projectId: params.id,
+        db,
+        dirname: __dirname,
```
kof, line 41: __dirname is still a red flag, we need a good way to have a root folder, maybe at least use it from a central place so that  we never have to use it inside the codebase.

in newer node versions its deprecated even

apps/designer/env-check.js
kof: oh I see, put them in a js file or json file?

kof: I actually see that the fact this is a js and not compiled is going to be a problem in the future, we probably need to have a ts file that compiles to a js file to be usable where it is currently used

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -0,0 +1,67 @@
...
+  getFilenameAndExtension,
+} from "../../helpers/array-buffer-helpers";
+
+export const s3UploadHandler: S3UploadHandler = async ({
+  file: { data, filename: baseFileName, contentType },
+  maxPartSize,
+}) => {
+  const s3Envs = s3EnvVariables.parse(process.env);
+  const { MAX_UPLOAD_SIZE } = assetEnvVariables.parse(process.env);
+  if (!data) return;
```
kof, line 23: when is data empty? is that normal or should we throw an error?

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (uint8Array.byteLength > maxPartSize) {
+    throw new Error(`Your asset cannot be bigger than ${MAX_UPLOAD_SIZE}mb`);
+  }
+
+  const [name, extension] = getFilenameAndExtension(filename);
+  const key = `${name}_${Date.now()}.${extension}`;
+
+  const params: PutObjectCommandInput = {
+    Bucket: s3Envs.S3_BUCKET,
+    ACL: s3Envs.S3_ACL ?? "public-read",
```
kof, line 37: this default can be in zod .default()

kof (approved):
I found a few more things, you can make it 💪

Approved

packages/asset-uploader/src/schema.ts
```diff
@@ -0,0 +1,41 @@
...
+  S3_BUCKET: z.string(),
+  S3_ACL: z.string().optional(),
+});
+
+export const fsEnvVariables = z.object({
+  FILE_UPLOAD_PATH: z.string().optional().default(DEFAULT_UPLPOAD_PATH),
+});
+
+export const assetEnvVariables = z.object({
+  MAX_UPLOAD_SIZE: z.number().optional().default(10),
```
kof, line 32: yeah, should be in bytes

SaraVieira: That is super confusing if the person has to write it in bytes, I would hate that experrience

kof: actually since its coming from env, you won't be able to pass this validation because its going to be a string and zod expects a number, unless z.number does the magical casting somebody hast to cast

kof: yeah, I agree regarding writing in bytes, because that would come from env without a nicer js syntax like `10_000`

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -0,0 +1,77 @@
...
+  assetEnvVariables,
+  ImagesUploadedSuccess,
+  s3EnvVariables,
+} from "../../schema";
+import {
+  getArrayBufferFromIterable,
+  getFilenameAndExtension,
+} from "../../helpers/array-buffer-helpers";
+
+export type S3UploadHandler = ({
```
kof, line 17:
```suggestion
type S3UploadHandler = ({
```

kof (approved):

packages/asset-uploader/src/schema.ts
```diff
@@ -0,0 +1,41 @@
...
+});
+
+export type S3UploadHandler = ({
+  file,
+  maxPartSize,
+}: {
+  file: UploadHandlerPart;
+  maxPartSize: number;
+}) => Promise<File | string | null | undefined>;
```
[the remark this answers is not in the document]
SaraVieira: moved and remove everything but striong

packages/asset-uploader/src/schema.ts
SaraVieira: truuue, let me parseInt

SaraVieira: done!

apps/designer/app/routes/designer/$id.tsx
SaraVieira: lets make an issue fdor that

PR #154, opened 2022-07-18
Fix follow up issues from CSF branch

### LINKED ISSUE #151: Layout with value as an input field seems broken
- [ ] Layout with value as an input field seems broken
Screenshot 2022-07-04 at 21 15 56

_Originally posted by @kof in https://github.com/webstudio-is/webstudio-designer/issues/53#issuecomment-1174129080_

### LINKED ISSUE #152: Wrong props don't show as error or something and also don't let me enter value, one of these needs to be enabled I guess?
- [ ] Wrong props don't show as error or something and also don't let me enter value, one of these needs to be enabled I guess?

Screenshot 2022-07-04 at 21 19 08

_Originally posted by @kof in https://github.com/webstudio-is/webstudio-designer/issues/53#issuecomment-1174136273_

### LINKED ISSUE #153: data- properties should always be enabled, right? Maybe except of the once we use in the system? e.g. data-component, data-id and a few more .... though maybe we just need to switch to something like data-webstudio-something to avoid conflicts
- [ ] data- properties should always be enabled, right? Maybe except of the once we use in the system? e.g. data-component, data-id and a few more .... though maybe we just need to switch to something like data-webstudio-something to avoid conflicts

Screenshot 2022-07-04 at 21 20 26

_Originally posted by @kof in https://github.com/webstudio-is/webstudio-designer/issues/53#issuecomment-1174140453_

okonet:
Closes #151
Closes #152
Closes #153

### DIFF
--- a/apps/designer/app/designer/features/props-panel/props-panel.tsx
+++ b/apps/designer/app/designer/features/props-panel/props-panel.tsx
@@ -7,8 +7,8 @@ import {
 import { Control } from "~/designer/features/props-panel/control";
 import { CollapsibleSection, ComponentInfo } from "~/designer/shared/inspector";
 import type { SelectedInstanceData } from "~/shared/canvas-components";
-import { Box, Button, Grid, TextField } from "~/shared/design-system";
-import { PlusIcon, TrashIcon } from "~/shared/icons";
+import { Box, Button, Grid, TextField, Tooltip } from "~/shared/design-system";
+import { PlusIcon, TrashIcon, ExclamationTriangleIcon } from "~/shared/icons";
 import { handleChangePropType, usePropsLogic } from "./use-props-logic";
 
 type PropertyProps = UserProp & {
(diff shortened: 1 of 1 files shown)


kof:
Looking much better now, a few more things:

- [ ] visual bugs with the tooltip's design system:

    Screenshot 2022-07-18 at 20 06 10
    Screenshot 2022-07-18 at 20 07 11

okonet:
This is a good follow up IMO which isn't relevant to this Pr. Feel free to create a new issue and I'll take a look.

kof:
Another one: we should focus the input field when clicking + button, making a follow up issue

kof:
And the last one: we need some better logic to identify the properties we show right away, because user may not know what they need and we still need some baseline for each component, e.g. form always needs target etc, I am not sure which logic to apply, we can go as well with a hardcoded list for now for each component, just the important basics

PR #164, opened 2022-07-20
Autodeploy SDK

andrasbacsai:


apps/designer/remix.config.js
```diff
@@ -17,4 +17,7 @@ module.exports = {
   // assetsBuildDirectory: "public/build",
   serverBuildPath: "api/index.js",
   // publicPath: "/build/",
+  watchPaths: () => {
```
andrasbacsai, line 20: This is the same as [here](https://github.com/webstudio-is/webstudio-designer/pull/163/files#diff-c6a539315bf0a259f65d6b284bc4d692549ea4a5287e04b73d6ca0daad96a503).

packages/react-sdk/package.json
```diff
@@ -62,9 +63,10 @@
     "react-nano-state": "^0.4.0",
     "zod": "^3.17.3"
   },
-  "main": "src/index.ts",
```
andrasbacsai, line 65: This is the same as [here](https://github.com/webstudio-is/webstudio-designer/pull/163/files#diff-38b40d8cc858fc034768ee58216c9fdb6f15f82e4c1b38d3e1114ccab19e1c17).

apps/designer/remix.config.js
```diff
@@ -17,4 +17,7 @@ module.exports = {
...
+    return ["../../packages/react-sdk/lib/**/*"];
```
kof, line 21: the problem is that its not only react-sdk, basically we need to watch all of them or otherwise we will have to add each time we add a new package and we will have many more packages there

packages/react-sdk/package.json
```diff
@@ -5,6 +5,7 @@
   "author": "Webstudio <github@webstudio.is>",
   "homepage": "https://webstudio.is",
   "scripts": {
+    "dev": "tsc -w",
```
kof, line 8: using tsc for compiling is slow, that's why I suggested using tsup, because it uses esbuild, which is just stripping off the types

andrasbacsai: Ah, I see.  Let me move it to `tsup` then.

kof: imagine you have 20 packages and start yarn dev in the root, so you start watching 20 packages and compiling them all with typescript ... its a nightmare

andrasbacsai: Yeah, right!

apps/designer/remix.config.js
andrasbacsai: Okay, will change it to watch the whole packages folder.

packages/react-sdk/package.json
```diff
@@ -5,6 +5,7 @@
   "author": "Webstudio <github@webstudio.is>",
   "homepage": "https://webstudio.is",
   "scripts": {
+    "dev": "tsup src/index.ts --format esm,cjs --watch --out-dir=lib",
```
kof, line 8: we already have the same thing in asset-uploader cc @SaraVieira added recently

kof: basically we need to add it in every package that is using ts, right?

kof (approved):
I haven't tested, please make sure running designer and changing things in SDK reflects the changes

kof:
added NPM_TOKEN to the workflows in development

packages/react-sdk/package.json
SaraVieira: Will update this PR to do the same for the asset uploader and test

SaraVieira:
Merging this to a branch in the repo

PR #168, opened 2022-07-20
Combobox

kof:
- see combobox in design system in storybook
- see usage of it in style panel and props panel


apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -35,20 +42,34 @@ const Property = ({
...
+            inputProps={{
+              ...inputProps,
+              readOnly: required,
+              placeholder: "Property",
+            }}
+          />
+        )}
+        contentProps={{
+          align: "start",
+          sideOffset: 5,
```
okonet, line 72: Generally I consider this kind of props a red flag and a sign of the need for compound components. Why do you want to expose them?

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -58,22 +79,23 @@ const Property = ({
...
-        onClick={() => {
-          onDelete(id);
-        }}
-      >
-        <TrashIcon />
-      </Button>
+      {required !== true && (
+        <Button
+          ghost
+          disabled={required}
```
okonet, line 91: Can be removed

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -35,20 +42,34 @@ const Property = ({
...
-        onChange={(event) => {
-          onChange(id, "prop", event.target.value);
+        itemToString={(item) => item ?? ""}
+        onItemSelect={(value) => {
+          onChange(id, "prop", value);
+        }}
+        disclosure={({ inputProps, toggleProps }) => (
+          <ComboboxTextField
+            toggleProps={toggleProps}
+            inputProps={{
```
okonet, line 63: Same as my next comment.

apps/designer/app/designer/features/style-panel/render-property.tsx
```diff
@@ -177,27 +184,51 @@ const ComboboxControl = ({
...
-          css={{
-            // @todo drag&drop cursor to adjust numeric value
-            // const cursorUrl = data:image/svg+xml;base64,${btoa(svgCursor)}
-            //cursor: `url(${cursorUrl}), text`,
-            textAlign: "right",
+          contentProps={{ sideOffset: 5 }}
+          value={{
+            name: String(currentValue.value),
+            label: String(currentValue.value),
           }}
```
okonet, line 193: I removed this since it introduces a bug when the selected item is never selected in the dropdown.

apps/designer/app/designer/features/props-panel/props-panel.tsx
kof: I felt like it looked weird when the tooltip is aligned horizontally in the middle relative to that input, it feels in the sidebar as if it should be in the middle of the sidebar, need to discuss this

kof: we could expose those props on the first level though, instead of contentProps, it was just simpler to forward them like this

kof: problem is that PopperContent is a component, if it was a hook, it would be simpler

kof: actually we can just export that PopperContent as a Content and let consumer pass those things directly to it

kof: yeah, let me do this right now

apps/designer/app/designer/features/style-panel/render-property.tsx
kof: can you reword? I don't understand the situation

apps/designer/app/designer/features/props-panel/props-panel.tsx
kof: going to rename disclosure to renderContent, renderTrigger and renderList

kof: wait, you meant this is related to some bug

apps/designer/app/designer/features/props-panel/props-panel.tsx
kof: done

apps/designer/app/designer/features/props-panel/props-panel.tsx
kof: done

apps/designer/app/designer/features/style-panel/render-property.tsx
kof: let me know once you read it what todo here, lets fix it in a separate PR, because need to merge this now

PR #172, opened 2022-07-21
Make prisma its own package

SaraVieira:
This PR makes Prisma its own package so we can use it and import it anywhere
Also updates both Prisma and Prisma client to the latest version

follow up https://github.com/webstudio-is/webstudio-designer/pull/174


packages/prisma-client/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@webstudio-is/prisma-client",
+  "version": "0.0.1",
+  "description": "Webstudio Prisma layer",
+  "author": "Webstudio <github@webstudio.is>",
+  "homepage": "https://webstudio.is",
+  "scripts": {
+    "checks": "tsc --noEmit",
+    "build:prisma": "prisma format && prisma generate",
+    "migrate": "dotenv -e ../../apps/designer/.env -- prisma migrate"
```
SaraVieira, line 10: this makes it so prisma will read the env from the designer and anyone just needs to have one env file

packages/prisma-client/.gitignore
```diff
@@ -0,0 +1,2 @@
+lib
+generated
```
kof, line 2: both should probably be or maybe already are globally ignored, because they sound like every package could use them

packages/prisma-client/package.json
kof, line 10: we should probably figure out a way to avoid having designer here

packages/prisma-client/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@webstudio-is/prisma-client",
+  "version": "0.0.1",
+  "description": "Webstudio Prisma layer",
+  "author": "Webstudio <github@webstudio.is>",
+  "homepage": "https://webstudio.is",
+  "scripts": {
```
kof, line 7: is it supposed to build generate/migrate on postinstall?

packages/prisma-client/package.json
kof: yeah, the problem is that prisma package ideally is not supposed to know anything about designer from abstraction perspective

why does it need to read the .env, can't it be already read before one runs prisma migrate?

SaraVieira: The two options are:

- We have two .env files, one here and one in the designer
- We do this

What do you prefer?

packages/prisma-client/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "checks": "tsc --noEmit",
```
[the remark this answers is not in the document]
SaraVieira: Done!

packages/prisma-client/package.json
SaraVieira: If we want yes

packages/prisma-client/.gitignore
SaraVieira: Done!

packages/prisma-client/package.json
SaraVieira: Here is the prisma page about it: https://www.prisma.io/docs/guides/development-environment/environment-variables/managing-env-files-and-setting-variables

kof: e.g. if we call this from designer, .env was already read, so when do we really call migrate in this package without first having the env?

packages/prisma-client/package.json
SaraVieira: done!

andrasbacsai (approved):
Tested and looks good!

packages/react-sdk/package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Webstudio <github@webstudio.is>",
   "homepage": "https://webstudio.is",
   "scripts": {
-    "build": "yarn build:mdn-data && yarn build:args",
+    "build": "yarn build:mdn-data && yarn build:args && yarn build:prisma",
```
andrasbacsai, line 8: This fixes the previous error with prisma, but now we have another one.

andrasbacsai:
`--ignore-scripts` in Vercel build command makes `sharp` module unavailable. Why do we change it?

andrasbacsai (approved):

andrasbacsai (approved):

kof (approved):
one more comment

packages/asset-uploader/package.json
```diff
@@ -22,6 +22,7 @@
     "jest-config": "*",
     "tsup": "^6.1.3",
     "typescript": "^4.6.2",
+    "@webstudio-is/prisma-client": "*",
```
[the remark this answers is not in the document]
SaraVieira: updated

PR #174, opened 2022-07-21
Move asset db code into asset package

SaraVieira:
…o asset-db-inpackage## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

##  Test cases

## Before merging

- [ ] added tests


apps/designer/app/routes/designer/$id.tsx
```diff
@@ -7,13 +7,16 @@ import config from "~/config";
 import env from "~/env.server";
 import { uploadAssets } from "~/shared/db/misc.server";
 import { ErrorMessage } from "~/shared/error";
+// if this file does not end in .server remix will not build
+// since it only allows node code in those files
+import { loadByProject } from "@webstudio-is/asset-uploader/src/db/index.server";
```
kof, line 12: the biggest problem with this is that one can't have nice names in the from statement, maybe we could at least have an index.server in the root of asset-uploader?

kof (approved):

apps/designer/app/routes/designer/$id.tsx
SaraVieira: I also forgot to try make the index.setver the main file so I'll try that and if it doesn't work I'll do what you are saying

SaraVieira: didn't work, made what you said

Issue #175, opened 2022-07-22
DnD primitive improvements


rpominov:
We need to make some improvements to the DnD primitive developed in #149

 - [x] Remove touch and keyboard listeners from useMove https://github.com/webstudio-is/webstudio-designer/pull/149#discussion_r926713532
 - [x] Add a demo with horizontal shift
 - [x] Fix spacing in the list demos
 - [ ] Highlight current drop target somehow in demos
 - [x] Add a demo when a drop target expands on hold
 - [x] Add unit tests
 - [x] Move technical descriptions from the PR to a readme file in the primitive
 - [ ] Add a demo where a copy of a drag item is being dragged visually rather than moving an empty cursor
 - [x] If the user releases the mouse button while outside of the browser window, the UI shouldn't get stuck in the "dragging" state
 - [x] Pressing "esc" should cancel the active drag
 - [x] Make sure test-cases.md is up to date and complete regarding DnD

rpominov:
While I'm trying to figure out what should live in the iframe, I came up with this table:

| Needs access to: |  Drop targets DOM | Drag items DOM | Scroll events |
|-|-|-|-|
| `useDrag` | | V | |
| `useDropTarget` | V | | V |
| `usePlacement` | V | | V |
| `useAutoScroll` | V (assuming scroll container is in targets DOM) | | |

Which made me realise that `useDropTarget` and `usePlacement` have very similar concerns. Might make sense to merge them to a single hook.

rpominov:
Another though: we probably shouldn't used the probe geometry directly for the placement indicator. This doesn't work well. Instead, use probe only to determine children direction when there're less than 2 children.

And to determine placement indicator geometry rely on target and its children rects. This will allow to, for example, always show the placement indicator neatly between two children at the same distance from both of them.

kof:
> And to determine placement indicator geometry, rely on the target and its children's rects. This will allow to, for example, always show the placement indicator neatly between two children at the same distance from each of them.

it does sound like its more complexity, but if there is no other way, I guess ...

rpominov:
> it does sound like its more complexity, but if there is no other way, I guess ...

Need to try, but it doesn't seem that complex.

I actually worry more about weird behaviour of the probe in edge cases, and workarounds we'd have to add to account for that.

rpominov:
> Add a demo with horizontal shift
> Add a demo when a drop target expands on hold

After tree is moved to design system it will act as a demo for these features

PR #176, opened 2022-07-22
New drag and drop implementation in canvas

rpominov:
Related issues / PRs:

- https://github.com/webstudio-is/webstudio-designer/pull/149
- https://github.com/webstudio-is/webstudio-designer/pull/147
- https://github.com/webstudio-is/webstudio-designer/issues/7
- https://github.com/webstudio-is/webstudio-designer/issues/31
- https://github.com/webstudio-is/webstudio-designer/issues/175 (isn't done as a part of this PR, but linked)

TODO:

- [x] We can't use `area` and `rect` that `onDropTargetChange` produces, because we want to change the target element based on data. Need to fix this in the primitive.

When drag starts on the canvas:

- [x] Monitor drag in canvas using `useDrag`
- [x] Monitor information about the current drop target using `useDropTarget` and `usePlacement`
- [x] Pass the information about the current drop target outside of the canvas and render drop target outline and placement indicator
- [x] Add auto scroll using `useAutoScroll`
- [x] Make changes to the data on drop end
- [x] Make sure tree-preview works

When drag starts on the panel:

- [x] Monitor drag in the panel using `useDrag`
- [x] Show a copy of drag item that follows the cursor
- [x] When drag starts, disable pointerEvents on the canvas (the logic already exists)
- [x] Send `start`, `move`, `end` events to the canvas
- [x] In the canvas handle the `start`, `move`, `end` events similarly to how events from the internal `useDrag` are handled
- [x] The rest hopefully should work the same as if the drag started on the canvas

--

- [x] Hooks refactoring
- [x] Cleanup / testing / bugfixes
---------------------------------------

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

##  Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/canvas.tsx
```diff
@@ -56,6 +53,101 @@ import {
   config,
 } from "~/canvas/features/wrapper-component/text-editor";
 
+// TODO: move this somewhere else
```
kof, line 56: we use @todo for consistency
```suggestion
// @todo move this somewhere else
```

apps/designer/app/canvas/canvas.tsx
```diff
@@ -56,6 +53,101 @@ import {
...
+import { useDrag } from "~/shared/design-system/components/primitives/dnd/use-drag";
```
kof, line 57: lets export those from components/index so we don't have to reach into the inner folders

apps/designer/app/canvas/canvas.tsx
```diff
@@ -56,6 +53,101 @@ import {
...
+import { useDropTarget } from "~/shared/design-system/components/primitives/dnd/use-drop-target";
+import { findInstanceById, getInstancePath } from "~/shared/tree-utils";
+import { primitives } from "~/shared/canvas-components";
+const useDragAndDrop = () => {
+  const [rootInstance] = useRootInstance();
+
+  const dropTargetHandlers = useDropTarget({
+    isDropTarget(element) {
+      return (
```
kof, line 66: why not doing the canAcceptChild check here instead of onDropTargetChange?

apps/designer/app/canvas/canvas.tsx
```diff
@@ -56,6 +53,101 @@ import {
...
+  // We want to use <body> as a root for drag items.
+  // The DnD hooks weren't designed for that.
+  // This is a temporary solution.
+  //
+  // NOTE: maybe use root instance's element as a root?
+  useEffect(() => {
+    const handlePointerDown = (event: PointerEvent) => {
+      dragProps.onPointerDown?.(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        event as any as React.PointerEvent<HTMLElement>
```
kof, line 127: this is wrong because its not a react event

apps/designer/app/canvas/canvas.tsx
```diff
@@ -56,6 +53,101 @@ import {
...
+      );
+    };
+
+    const handleScroll = () => {
+      dropTargetHandlers.handleScroll();
+    };
+
+    const rootElement = document.body;
+
+    rootElement.addEventListener("pointerdown", handlePointerDown);
```
kof, line 137: wasn't this supposed to be part of the primitive?

apps/designer/app/canvas/canvas.tsx
kof: here is a good location for it: app/canvas/shared/use-drag-drop.ts

apps/designer/app/shared/canvas-components/draggable.ts
```diff
@@ -1,6 +1,10 @@
-import type { XYCoord } from "react-dnd";
 import { type Instance } from "@webstudio-is/react-sdk";
 
+interface XYCoord {
```
kof, line 3: I think this should go to use-drag-drop.ts

kof: I mean this entire file

apps/designer/app/canvas/canvas.tsx
rpominov: The way useMove uses the event it doesn't matter. But I won't leave this `any` like this, just trying to make it work for now

kof: please don't add eslint-disable if its temporarily, because that's what prevents you from forgetting and merging as-is

apps/designer/app/canvas/canvas.tsx
rpominov: We designed it with an assumption that the root will be available as a react element to which we can pass the props. This is not the case if we want to use `body` as root.

I'm thinking of using the root instance's node instead of body, it'll probably be more correct in any case. But even then there might not be a convenient way to add listeners to it via React. So, I'll probably change the primitive to only take refs and then add all listeners internally.

apps/designer/app/canvas/canvas.tsx
rpominov: We'll still need to be able to swap the target based on the area. I want to do all the swapping in one place. And also want to rely more on the data tree instead of DOM tree when choosing the target. I've built a new `useDropTarget` in the primitive that will make this easier. Will try to use it today here.

kof: right, forgot

apps/designer/app/canvas/features/wrapper-component/wrapper-component.tsx
```diff
@@ -35,28 +34,16 @@ export const WrapperComponentDev = ({
   const [, setSelectedElement] = useSelectedElement();
   const isEditing = editingInstanceId === instance.id;
   const [editableRefCallback, editableProps] = useContentEditable(isEditing);
-  const { dragRefCallback, ...draggableProps } = useDraggable({
-    instance,
-    // We can't drag if we are editing text.
-    isDisabled: isEditing === true,
-  });
```
rpominov, line 42: Note to self: make sure this still works

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -8,7 +8,7 @@ import {
   usePlacement,
   useDrag,
   useDropTarget,
-} from "@webstudio-is/design-system";
+} from "@webstudio-is/design-system/src/components/primitives/dnd";
```
kof, line 11: needs to be exported from index in design-system, can't go inside any folder that has index, we need a linter for this

rpominov: Thought it would be easier for naming if DnD wouldn't share namespace with everything else in design-system. Maybe we should do something like this? 

```js
import { dnd } from "@webstudio-is/design-system";
const { useDrag } = dnd;
```

This will likely break tree-shaking though.

kof: I don't think we will have a namespaces problem even if we export useDrag straight, because there won't be another drag&drop solution there, but if there will be, we can always do what you just said, not sure regarding tree shaking there, I mean if you import dnd you will likely use all its hooks, so ...

rpominov:
@kof This is ready for review. I want to do some improvements on how target / placement selection works, but I'll do them in a new PR.

> tested locally and on preview environment

Deploy fails after I merged `main`, so I tested only locally.

rpominov:
Ready for review!

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+  dragItem: undefined as Instance | undefined,
+};
+
+export const useDragAndDrop = () => {
+  const [rootInstance] = useRootInstance();
+  const [textEditingInstanceId] = useTextEditingInstanceId();
+
+  const state = useRef({ ...initialState });
+
+  const autoScrollHandlers = useAutoScroll({ fullScreen: true });
```
kof, line 48:
```suggestion
  const autoScrollHandlers = useAutoScroll({ fullscreen: true });
```

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      const path = getInstancePath(rootInstance, dropTarget.data.id);
+      path.reverse();
+
+      if (dropTarget.nearEdge) {
+        path.shift();
+      }
+
+      // Don't allow to dpop inside drag item or any of its children
+      const dragItemIndex = path.findIndex((x) => x.id === dragItem.id);
```
kof, line 91: we have a rule to never use non-descriptive names 
```suggestion
      const dragItemIndex = path.findIndex((child) => child.id === dragItem.id);
```

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      if (!data) {
+        return getDefaultDropTarget();
+      }
+
+      if (data.id === dropTarget.data.id) {
+        return dropTarget;
+      }
+
+      const element = data && document.getElementById(data.id);
```
kof, line 108: we probably need to move all getElementById calls into a function, because I think long-term we should stop using attributes which could be useful to the user,

I think best place is dom-utils/getInstanceElementById

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      if (!element) {
```
kof, line 110: NIT: would be better to be more explicit in all such cases, but its my personal thing, e.g. if element can only be null - === null, if undefined === undefined, if null or undefined == null, problem with !x:
1. includes empty strings, 0 etc
2. sometimes if it proceeds an "i": `!isSomething` it looks unreadable

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+    onDropTargetChange(dropTarget) {
+      state.current.dropTarget = dropTarget;
+      publish<"dropTargetChange", DropTargetChangePayload>({
+        type: "dropTargetChange",
+        payload: {
+          rect: dropTarget.rect,
+          placement: dropTarget.placement,
+          position: dropTarget.indexWithinChildren,
+          instanceId: dropTarget.data.id,
+          instanceComponent: dropTarget.data.component,
```
kof, line 126:
```suggestion
          component: dropTarget.data.component,
```

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      publish<"dragStart", DragStartPayload>({
+        type: "dragStart",
+        payload: {
+          origin: "canvas",
+          dragItem: instance,
+        },
+      });
+    },
+    onMove: (poiterCoordinate) => {
```
kof, line 169: there must be a better name than pointerCoordinate, also there is a typo, maybe just `point`?

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,273 @@
...
+      }
+
+      state.current = { ...initialState };
+    },
+  });
+
+  // We have to use useLayoutEffect to setup the refs
+  // because we want to use <body> as a root.
+  // We prefer useLayoutEffect over useEffect
+  // because it's closer in the life cycle to when React noramlly calls the "ref" callbacks.
```
kof, line 210: this literally doesn't matter, the point of using useLayoutEffect over useEffect is to be called after the dom node is rendered but synchronously, so that you can read layout from dom without causing a reflow, in this case it literally doesn't matter afaik (I haven't tested), I would just use an effect

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
```diff
@@ -1,90 +1,146 @@
...
-  onDragChange,
-  onClick,
-}: DraggableThumbProps) => {
-  const dragRef = useDraggable({ component, onDragChange });
-  return (
-    <ComponentThumb component={component} ref={dragRef} onClick={onClick} />
+  pointerPosition,
+}: {
+  component: Instance["component"];
+  pointerPosition: { x: number; y: number };
```
kof, line 45: would this be better?
```suggestion
  point: { x: number; y: number };
```

apps/designer/app/canvas/shared/use-drag-drop.ts
rpominov: It does matter in our case.

If I use `useEffect` in the following place `rootRef.current` isn't initialised in time:

https://github.com/webstudio-is/webstudio-designer/blob/c8f5b8a341865e9d357b6ca11af3fbc23e869a72/packages/design-system/src/components/primitives/dnd/use-drag.ts#L138-L146

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
```diff
@@ -1,90 +1,146 @@
...
-export const TabContent = ({
-  onDragChange,
-  publish,
-  onSetActiveTab,
-}: TabContentProps) => {
-  const componentNames = (
-    Object.keys(components) as Array<Instance["component"]>
-  ).filter((component) => components[component].isInlineOnly === false);
+const elementToComponentName = (element: Element) => {
+  const parentWithData = element.closest("[data-drag-component]");
```
kof, line 77: hmmm, are we inside the side panel here? wondering why we do this if we are

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
```diff
@@ -1,90 +1,146 @@
...
+  if (parentWithData === null || !(parentWithData instanceof HTMLElement)) {
```
kof, line 78: if its not an instance of element, it is automatically not null? Or was typescript too stupid?

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
```diff
@@ -1,90 +1,146 @@
...
+    return;
+  }
+  const { dragComponent } = parentWithData.dataset;
+  return componentNames.find((c) => c === dragComponent);
```
kof, line 82:
```suggestion
  return componentNames.find((component) => component === dragComponent);
```

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
```diff
@@ -1,90 +1,146 @@
...
+};
+
+export const TabContent = ({ publish, onSetActiveTab }: TabContentProps) => {
+  const [dragComponent, setDragComponent] = useState<Instance["component"]>();
+  const [pointerPosition, setPointerPosition] = useState<{
+    x: number;
+    y: number;
+  }>({ x: 0, y: 0 });
```
kof, line 90: we should have a global type for {x, y} that we can use everywhere

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
rpominov: This is to make drag work when you start to drag on an icon inside the button. I that case element points to an SVGElement that does not have the data.

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
```diff
@@ -91,7 +73,10 @@ export const SidebarLeft = ({
             ...sidebarTabsContentStyle,
             // We need the node to be rendered but hidden
             // to keep receiving the drag events.
-            visibility: isDragging ? "hidden" : "visible",
+            visibility:
+              dragAndDropState.isDragging && dragAndDropState.origin === "panel"
```
kof, line 77: is that parent getting isDragging true when origin is not panel? Just wondering

apps/designer/app/designer/features/workspace/canvas-tools/outline/outline.tsx
```diff
@@ -1,6 +1,8 @@
 import { useMemo } from "react";
 import { styled } from "@webstudio-is/design-system";
 
+type Rect = Pick<DOMRect, "top" | "left" | "width" | "height">;
```
kof, line 4: I think we should have this inside some reusable place, because we seem to need it everywhere

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
rpominov: I couldn't decide where to put it and thought it's small enough to just inline everywhere. Should I put it in the design-system/dnd ?

apps/designer/app/shared/tree-utils/reparent-instance.ts
```diff
@@ -0,0 +1,36 @@
...
+    return;
+  }
+
+  let newPositionAdjusted = newPosition;
+
+  // If parent is the same, we need to adjust the position to account for the removal of the instance.
+  const currentParent = findParentInstance(rootInstance, instance.id);
+  if (currentParent !== undefined && currentParent.id === newParentId) {
+    const currentPosition = currentParent.children.findIndex(
+      (x) => typeof x !== "string" && x.id === instance.id
```
kof, line 24:
```suggestion
      (child) => typeof child !== "string" && child.id === instance.id
```

apps/designer/package.json
```diff
@@ -49,9 +49,6 @@
     "nanoid": "^3.2.0",
     "react": "^17.0.2",
     "react-color": "^2.19.3",
-    "react-dnd": "^14.0.5",
```
kof, line 52: yay

packages/design-system/jest.config.js
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line
+const base = require("@webstudio-is/jest-config");
+
+module.exports = {
+  ...base,
+};
```
kof, line 6:
```suggestion
module.exports = require("@webstudio-is/jest-config");
```

packages/design-system/src/components/primitives/dnd/canvas.stories.tsx
```diff
@@ -127,99 +128,147 @@ export const Canvas = () => {
...
-          }
-        }
+      const path = findItemPath(data, id) ?? [];
+
+      if (nearEdge) {
+        path.shift();
       }
 
-      setDropTarget(ROOT_ID, rootElement);
+      // Don't allow to dpop inside drag item or any of its children
```
kof, line 226:
```suggestion
      // Don't allow to drop inside drag item or any of its children
```

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
rpominov: Not sure what parent you're referring to, but basically we need the visibility hack only when the drag is started on the panel. We could apply the hack always though, wouldn't hurt.

packages/design-system/src/components/primitives/dnd/canvas.stories.tsx
```diff
@@ -127,99 +128,147 @@ export const Canvas = () => {
...
+        return getDefaultDropTarget();
+      }
+
+      const element = idToElement(rootElement, newItem.id);
+
+      if (element === undefined) {
+        return getDefaultDropTarget();
+      }
+
+      return { data: newItem.id, element };
```
kof, line 244: I wonder if we could reuse this somehow with the actual logic,  seems like this stuff may break between storybook and designer

apps/designer/app/designer/features/workspace/canvas-tools/outline/outline.tsx
rpominov: Also, couldn't decide on the place :)

packages/design-system/src/components/primitives/dnd/dom-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  getRectsOrientation,
+} from "./geometry-utils";
+
+export type UsableElement = HTMLElement | SVGElement;
+
+// Many DOM APIs produce values of type Element,
+// which lacks certain methods that we often need.
+// It's also usually safe to assume that the value is in fact HTMLElement or SVGElement.
+// We use this helper to convince TypeScript that the value is a UsableElement.
+export const toUseableElement = (
```
kof, line 13: if its only for typescript, there should be a way to do this in types only, without generating runtime code, wdyt? or do we actually need the runtime checks?

rpominov: I think everywhere where I use this helper, the check always passes. But I'm not 100% sure. Also, not sure how to do this using only types.

packages/design-system/src/components/primitives/dnd/geometry-utils.ts
```diff
@@ -0,0 +1,305 @@
...
+  third: Rect | undefined
+): ChildrenOrientation => {
+  const orientations = [
+    first && getTwoRectsOrientation(first, second),
+    third && getTwoRectsOrientation(second, third),
+  ];
+
+  // @todo: Maybe we should check that at least one is reversed.
+  // Need to test what works better, but keep an eye on false positives.
+  const allReverse = orientations.every((x) => !x || x.reverse !== false);
```
kof, line 246:
```suggestion
  const allReverse = orientations.every((orientation) => !orientation || orientation.reverse !== false);
```

packages/design-system/src/components/primitives/dnd/geometry-utils.ts
```diff
@@ -0,0 +1,305 @@
...
+
+  const types = orientations.map((x) => x && x.type);
```
kof, line 248:
```suggestion
  const types = orientations.map((orientation) => orientation && orientation.type);
```

packages/design-system/src/components/primitives/dnd/canvas.stories.tsx
rpominov: Do you mean reuse the logic inside swapDropTarget? Even between the 3 places where we use the hook it's kinda unique in each place. I'd kept it in the app code for now rather than trying to move into the dnd hooks.

packages/design-system/src/components/primitives/dnd/sortable-list.stories.tsx
```diff
@@ -63,43 +65,47 @@ export const SortableList = ({
...
-  const dropTargetHandlers = useDropTarget({
-    isDropTarget(element: HTMLElement) {
+  const useDropHandlers = useDrop<true>({
+    isDropTarget(element) {
       return element instanceof HTMLUListElement;
     },
-    onDropTargetChange(event) {
-      placementHandlers.handleTargetChange(event.target);
+    swapDropTarget(dropTarget) {
+      if (!dropTarget) {
```
kof, line 77: use early return
```suggestion
      if (dropTarget)  return dropTarget;
```

packages/design-system/src/components/primitives/dnd/use-auto-scroll.ts
```diff
@@ -38,25 +38,33 @@ const getSpeed = (
   return 0;
 };
 
-export type Parameters = {
+export type UseAutoScrollProps = {
   edgeDistanceThreshold?: number;
 
+  // If set to true the entire document will be scrolled.
+  // No need to set targetRef in this case.
+  fullScreen?: boolean;
```
kof, line 46:
```suggestion
  fullscreen?: boolean;
```

packages/design-system/src/components/primitives/dnd/sortable-list.stories.tsx
rpominov: This will not be the same as the current logic

rpominov: Ah, actually here it's the same, yes.

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -0,0 +1,293 @@
...
+};
+
+export const useDrop = <Data>(props: UseDropProps<Data>): UseDropHandlers => {
+  // We want to use fresh props every time we use them,
+  // but we don't need to react to updates.
+  // So we can put them in a ref and make useMemo below very efficient.
+  const latestProps = useRef<UseDropProps<Data>>(props);
+  latestProps.current = props;
+
+  const rootRef = useRef<UsableElement>();
```
kof, line 78: Is UsableElement in the end always an instance element? Mybe we call it InstanceElement then?

kof: or WsElement

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -0,0 +1,293 @@
...
+      );
+
+      const indexWithinChildren = closestChildIndex + indexAdjustment;
+
+      const current = state.current.dropTarget;
+      if (
+        current === undefined ||
+        current.element !== partialDropTarget.element ||
+        current.indexWithinChildren !== indexWithinChildren ||
+        !isEqualRect(current.rect, parentRect)
```
kof, line 135: this is exactly the thing I was talking about, hard to read with some fonts
```suggestion
        isEqualRect(current.rect, parentRect) === false
```

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -0,0 +1,293 @@
...
+        isDropTarget,
+        swapDropTarget,
+      } = latestProps.current;
+
+      const { pointerCoordinates } = state.current;
+      const root = rootRef.current;
+
+      // @todo: Cache this?
+      // Not expensive by itself, but it may call isDropTarget multiple times.
+      let candidate =
```
kof, line 186: ideally we would test it with a something that has 1000+ instances and measure the time we spend there at least just simply using console.time() to get a rough picture of what will happen soon

packages/design-system/src/components/primitives/dnd/use-drop.ts
rpominov: Here it's used for `rootRef` and we don't assume that the root itself is necessarily a drop-target. So, no it doesn't necessarily corresponds to an instance.

Also, it probably would be wrong to use the term Instance here in the dnd hooks?

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -0,0 +1,293 @@
...
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
+        const swappedTo = swapDropTarget(
+          candidate &&
+            pointerCoordinates && {
+              ...candidate,
+              nearEdge: isNearEdge(
+                pointerCoordinates,
+                edgeDistanceThreshold,
+                candidate.element.getBoundingClientRect()
```
kof, line 233: this will be potentially an expensive call since its going to trigger reflow, how often do you think this is going to be called in a large tree?

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -0,0 +1,293 @@
...
+      if (!isNewCandidate && !isNewIsNearEdge && state.current.dropTarget) {
+        // Still need to call setDropTarget to update rect and/or placement.
+        // Because indexWithinChildren might have changed,
+        // or parent coordinates might have moved in case of a scroll
+        setDropTarget(state.current.dropTarget);
+        return;
+      }
+
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
```
kof, line 225: this makes my heart stop every time I read it, I know in theory it should be fine, but maybe there is a better way to express it, e.g. with a condition of found candidate or something

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -0,0 +1,293 @@
...
+  root,
+  initialElement,
+  isDropTarget,
+}: {
+  root: UsableElement;
+  initialElement: UsableElement | undefined;
+  isDropTarget: (target: UsableElement) => Data | false;
+}): PartialDropTarget<Data> | undefined => {
+  // The element we get from elementFromPoint() might not be inside the root
+  if (!initialElement || !root.contains(initialElement)) {
```
kof, line 278: this one

packages/design-system/src/components/primitives/dnd/use-move.ts
```diff
@@ -168,112 +159,44 @@ export function useMove(props: MoveEvents): MoveResult {
...
-        if (e.button === 0) {
-          start();
-          e.stopPropagation();
-          e.preventDefault();
-          state.current.lastPosition = { pageX: e.pageX, pageY: e.pageY };
-          addGlobalListener(window, "mousemove", onMouseMove, false);
-          addGlobalListener(window, "mouseup", onMouseUp, false);
-        }
-      };
+    let onPointerMove = (e: PointerEvent) => {
```
kof, line 162:
```suggestion
    let onPointerMove = (event: PointerEvent) => {
```

kof (changes requested):
first round of conceptual review done 🎉

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
kof: yeah, I think so, thought the same thing at one point, but now I can see as a  reviewer its annoying, kindof

packages/design-system/src/components/primitives/dnd/use-drop.ts
rpominov: You tested it with a loop in a console back when we paired. I'll keep an eye on this when I'll do the test with 1000+ instances.

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
kof: what I mean is, do we get notified when the drag started not on the panel? I mean we are inside the panel, that's why I am wondering, why do we know about the drag started outside the panel, just a question

packages/design-system/src/components/primitives/dnd/use-drop.ts
rpominov: This wasn't the first choice for me either :)
But with `while(true)` the code is simplest compared to all other options

packages/design-system/src/components/primitives/dnd/use-drop.ts
rpominov: I don't understand :)

kof:
Line positioning is super accurate and smooth, I think I found a small bug  https://share.descript.com/view/q5mfqn8reAT

kof:
- [ ] Esc during the drag should cancel the drag

apps/designer/app/designer/features/sidebar-left/panels/components/components.tsx
kof: Oh I see, lets add a comment

apps/designer/app/designer/features/workspace/canvas-tools/outline/outline.tsx
kof: same?

packages/design-system/src/components/primitives/dnd/dom-utils.ts
kof: I guess you could use an assertion like `as HTMLElement` if you only need type check

packages/design-system/src/components/primitives/dnd/use-drop.ts
kof: what is rootRef? not a root instance?

kof: we are migrating towards having body also as an instance, so I am not sure how you can have anything that isn't an instance

packages/design-system/src/components/primitives/dnd/use-drop.ts
kof: I think I was testing get element from point performance

packages/design-system/src/components/primitives/dnd/use-drop.ts
kof: I bet we can have an extra variable like foundSomething and while (foundSomething === undefined) or similar, would be one more line but 100% better

packages/design-system/src/components/primitives/dnd/use-drop.ts
kof: `!i`

packages/design-system/src/components/primitives/dnd/use-drop.ts
kof: true, design system knows nothing about instances ...

kof: well lets keep it, I don't have a better name, but Usable sounds weird because usable for what? There should be a better name for it, what is actually an example of an unusable element?

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
rpominov: Sure, the dragAndDropState gets updated no matter where the drag has started

packages/design-system/src/components/primitives/dnd/dom-utils.ts
rpominov: Ah, right. I'm used to Flow that wouldn't allow that because it's unsafe:

https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqCmEAu0DeBfAUJ0kYOQGEALASxABNp4APJeAOwv0RVU2mgDcwQBXeAC5oDPgFsARvABOAbg7RgAewbJpfYEiXSAFAEo0CzhD4AHGfvmdOSMhAB0PfvGgBeaAEYr0LFhwrkaHpA9wZ4AHc4VktsZVUUJwFhUUkZN2gdYJRIKKIySj1HXgEgA

But for our case it's probably better to use `as HTMLElement` indeed.

packages/design-system/src/components/primitives/dnd/use-drop.ts
rpominov: I'll try to `as HTMLElement` instead of the helper

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,289 @@
...
+          placement: dropTarget.placement,
+          position: dropTarget.indexWithinChildren,
+          instance: toBaseInstance(dropTarget.data),
+        },
+      });
+    },
+  });
+
+  const dragHandlers = useDrag<Instance>({
+    isDragItem(element) {
```
kof, line 142: this naming suggests it will return a boolean, but we return an instance in the end

apps/designer/app/canvas/shared/use-shortcuts.ts
```diff
@@ -75,6 +79,7 @@ export const useShortcuts = () => {
     breakpointsMenu: publishOpenBreakpointsMenu,
     breakpoint: publishSelectBreakpoint,
     zoom: publishZoom,
+    cancel: publishCancelCurrentDrag,
```
kof, line 82: cancelDrag
```suggestion
    cancelDrag: publishCancelCurrentDrag,
```

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -117,7 +117,7 @@ export const usePropsLogic = ({
       const isPropRequired = draft[index].required;
       switch (field) {
         case "prop":
-          if (!isPropRequired) {
+          if (isPropRequired !== true) {
```
kof, line 120: could have been === false 😅

apps/designer/app/canvas/shared/use-shortcuts.ts
kof: or alternatively just send a generic "esc" and let those subscribe who need to know ?

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -0,0 +1,289 @@
...
+  type Point,
+  useAutoScroll,
+  useDrag,
+  useDrop,
+} from "@webstudio-is/design-system";
+import {
+  publish,
+  useSubscribe,
+  type Instance,
+  components as primitives,
```
kof, line 22: lets avoid this mapping for consistency with the rest

packages/design-system/src/components/primitives/dnd/geometry-utils.ts
```diff
@@ -0,0 +1,306 @@
+export type Rect = Pick<DOMRect, "top" | "left" | "width" | "height">;
```
kof, line 1: I don't know why I actually extended DOMRect, could be just a numeric thing, probably won't have compatibility issues

packages/design-system/src/components/primitives/dnd/placement-indicator.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const placementStyle = {
+  boxSizing: "content-box",
+  position: "absolute",
+  background: "#f531b3",
+  pointerEvents: "none",
+};
+
+export const PlacementIndicator = ({ placement }: { placement: Placement }) => {
+  const style =
```
kof, line 12: would probably put this into a function and use early return to avoid multiline ternaries and also to avoid cluttering the component

packages/design-system/src/components/primitives/dnd/sortable-list.stories.tsx
```diff
@@ -63,43 +65,49 @@ export const SortableList = ({
...
     },
-    onDropTargetChange(event) {
-      placementHandlers.handleTargetChange(event.target);
+    swapDropTarget(dropTarget) {
+      if (dropTarget) {
+        return dropTarget;
+      }
+
+      if (rootRef.current === null) {
+        throw new Error("should not happen");
```
kof, line 82: not sure if there is a better message

kof (approved):
a few minor comments

apps/designer/app/designer/features/props-panel/use-props-logic.ts
rpominov: `!== true` is needed to handle `undefined`. isPropRequired is `boolean | undefined`

packages/design-system/src/components/primitives/dnd/geometry-utils.ts
rpominov: Do you mean to just define it as an object instead of using Pick?

kof: yeah, if that works in terms of compatibility with DOMRect so one can pass both DOMRect and a React

PR #180, opened 2022-07-22
Add support for cloud flare R2 and clean path in db

SaraVieira:
## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [x] added tests

Related https://github.com/webstudio-is/webstudio-designer/issues/37

### DIFF
--- a/packages/asset-uploader/src/helpers/get-asset-path.test.ts
+++ b/packages/asset-uploader/src/helpers/get-asset-path.test.ts
@@ -5,7 +5,7 @@ const commonAsset = {
   id: "sa-546",
   width: new Decimal(200),
   height: new Decimal(200),
-  projectId: "jdhajk",
+  projectId: "id",
   size: 2135,
   format: "png",
   createdAt: new Date(),
@@ -16,16 +16,20 @@ describe("getAssetPath", () => {
   const OLD_ENV = process.env;
 
   beforeEach(() => {
-    jest.resetModules(); // Most important - it clears the cache
-    process.env = { ...OLD_ENV }; // Make a copy
+    jest.resetModules();
+    process.env = {
+      ...OLD_ENV,
+      S3_ACCESS_KEY_ID: "testid",
+      S3_SECRET_ACCESS_KEY: "testkey",
+    };
   });
 
   afterAll(() => {
-    process.env = OLD_ENV; // Restore old environment
+    process.env = OLD_ENV;
   });
 
   describe("local", () => {
-    test("return local path with no custom path", () => {
+    test("returns local path with no custom path", () => {
       const { getAssetPath } = require("./get-asset-path");
       expect(
         getAssetPath({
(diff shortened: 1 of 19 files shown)


packages/asset-uploader/src/targets/disk/upload.ts
```diff
@@ -21,9 +18,9 @@ export const uploadToDisk = async ({
 
     const data = {
       name: image.name,
-      path: path.join("/", folderInPublic, image.name),
       size: image.size,
       metadata,
+      location: "FS" as "FS" | "REMOTE",
```
SaraVieira, line 23: How can I fix this? It's an enum on prisma

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -41,7 +41,7 @@ export const s3UploadHandler: S3UploadHandler = async ({
   }
 
   const [name, extension] = getFilenameAndExtension(filename);
-  const key = `${name}_${Date.now()}.${extension}`;
+  const key = encodeURI(`${name}_${Date.now()}.${extension}`);
```
andrasbacsai, line 44: To prevent space in URI.

andrasbacsai (approved):
Tested and LGTM!

packages/asset-uploader/src/targets/s3/handler.ts
SaraVieira: Danke

packages/asset-uploader/src/helpers/get-asset-path.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        })
+      ).toBe("/assets/test.png");
+    });
+  });
+  describe("s3", () => {
+    test("returns s3 url", () => {
+      process.env.S3_ENDPOINT = "https://fr1.s3.com";
+      process.env.S3_BUCKET = "bucket";
+      process.env.S3_REGION = "fr1";
+      const { getAssetPath } = require("./get-asset-path");
```
kof, line 62: I would instead just move the functions which access env to the call time, so that you can decide when they access env variables

packages/asset-uploader/src/helpers/get-asset-path.ts
```diff
@@ -0,0 +1,27 @@
+import { DbAsset, Location } from "@webstudio-is/prisma-client";
+import { fsEnvVariables, s3EnvVariables } from "../schema";
+import path from "path";
+
+const s3Envs = s3EnvVariables.safeParse(process.env);
```
kof, line 5: I think this is so fast that you can just do it inside the function and have the trouble with testing

packages/asset-uploader/src/targets/s3/handler.ts
```diff
@@ -64,14 +68,14 @@ export const s3UploadHandler: S3UploadHandler = async ({
...
+  ImagesUploadedSuccess.parse(await upload.done());
   const image = sharp(uint8Array);
   const metadata = await image.metadata();
 
   // these upload handlers only allow strings to be returned
   return JSON.stringify({
     metadata,
     name: key,
-    path: newFile.Location,
+    location: Location.REMOTE,
```
kof, line 79: get over it)

packages/prisma-client/prisma/schema.prisma
```diff
@@ -15,6 +15,11 @@ model Team {
   users User[]
 }
 
+enum Location {
```
kof, line 18: can we please not use enums?

just use a `type Location = 'FS' | 'REMOTE'`

SaraVieira: That doesn't work in Prisma :(

You either do enums or a string

kof: you always pass a string in the end to a consumer

kof: type is only there to make sure you pass the right thing

SaraVieira: So you would prefer prisma to have a string and then we just create the location type?

PR #183, opened 2022-07-23
feature(style panel): layout - flex

natlusyht:
## Tasks:

- [x] Refactor `render-property` logic to support rendering categories in 2d space.
- [x] Layout section > select.
- [x] Layout section > Flex > icon control.
- [x] Layout section > Flex > grid control.
- [x] Layout section > Flex > dropdown.
- [x] Layout section > Flex > row/column-gap control.
- [x] add row/column gap input icons.
- [x] Layout section > Flex > icon control |=> grid control (unidirectional)
- [x] Layout section > Flex > grid control <=> icon control (bidirectional)
> bidirectional: make grid control sync bi-directionally(changes to the grid should affect the icons just as changes to the icons currently affect the grid, the plumbing is already there).

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] tested locally and on preview environment
- [ ] added tests


apps/designer/app/designer/features/style-panel/lib/configs.ts
```diff
@@ -4,9 +4,24 @@ import {
   keywordValues,
   type Category,
   type StyleProperty,
+  type AppliesTo,
 } from "@webstudio-is/react-sdk";
-import type { StyleConfig, Control } from "./types";
-import { humanizeString } from "../string-utils";
+import { humanizeString } from "../../../../shared/string-utils";
```
kof, line 9:
```suggestion
import { humanizeString } from "~/shared/string-utils";
```

apps/designer/app/designer/shared/inspector/collapsible-section.tsx
```diff
@@ -51,14 +51,17 @@ export const CollapsibleSection = ({
               px: "$3",
               color: "$hiContrast",
               cursor: "default",
+              ...(isOpenFinal && { paddingBottom: "0px" }),
```
kof, line 54:
```suggestion
              ...(isOpenFinal && { paddingBottom: 0 }),
```

apps/designer/app/designer/shared/inspector/collapsible-section.tsx
```diff
@@ -51,14 +51,17 @@ export const CollapsibleSection = ({
...
             }}
           >
             <Text size="1" css={{ fontWeight: "500", flexGrow: 1 }}>
               {label}
             </Text>
             {/* @todo: props panel shows double plus icon when rightSlot is set */}
-            {isOpenFinal ? <MinusIcon /> : <PlusIcon />}
             {rightSlot}
+            <Box css={{ marginRight: "-6px" }}>
```
kof, line 62:
```suggestion
            <Box css={{ marginRight: -6 }}>
```

apps/designer/app/shared/design-system/components/combobox.tsx
```diff
@@ -40,7 +40,7 @@ export const Combobox = ({
           value={currentValue}
           autoComplete="off"
           // @todo avoid hardcoding padding
-          css={{ ...css, paddingRight: 30 }}
+          css={{ ...css, paddingRight: 30 - 4 }}
```
kof, line 43: this will be gone once I merge the new combobox

apps/designer/app/shared/design-system/components/comboicon.tsx
```diff
@@ -0,0 +1,138 @@
+import React from "react";
```
kof, line 1: a lot of arbitrary values used in this file that should be design system value

kof: lets add a story for comboicon, btw is that a real name for such a component?

apps/designer/app/shared/design-system/components/comboicon.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { styled } from "@stitches/react";
```
kof, line 2: generally we can use numeric values where its px, looks nicer I think

apps/designer/app/shared/icons/align-content.tsx
```diff
@@ -0,0 +1,157 @@
...
+        clipRule="evenodd"
+        d="M19 5H5V6H19V5ZM19 18H5V19H19V18ZM8 8H11V11H8V8ZM16 8H13V11H16V8ZM13 13H16V16H13V13ZM11 13H8V16H11V13Z"
+        fill="#11181C"
+      />
+    </svg>
+  );
+});
+AlignContentSpaceBetween.displayName = "AlignContentSpaceBetween";
+
+export const alignContent = {
```
kof, line 149: this seems like a map you need in the consumer space

apps/designer/app/shared/icons/align-items.tsx
```diff
@@ -0,0 +1,157 @@
...
+          d="M16 16H19.2V17.6H4.80005V16H8.00005V7.99999H4.80005V6.39999H19.2V7.99999H16V16ZM12.8 16V7.99999H11.2V16H12.8Z"
+          fill="#11181C"
+        />
+      </svg>
+    );
+  }
+);
+AlignItemsStretch.displayName = "AlignItemsStretch";
+
+export const alignItems = {
```
kof, line 150: same

apps/designer/package.json
```diff
@@ -114,6 +114,7 @@
     "@types/lodash.merge": "^4.6.6",
     "@types/lodash.noop": "^3.0.6",
     "@types/lodash.snakecase": "^4.1.6",
+    "@types/lodash.camelcase": "^4.3.0",
```
kof, line 117: we already have a camelcase package

kof (commented):
✅ first round of review, seems on a good path!

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -436,7 +463,13 @@ const GridControl = ({
...
             }}
             onClick={() => {
-              // setPosition(index)
+              const [alignItems, justifyContent] = cells[index]
+                .slice(1)
+                .split("_")
+                .map((value) => ["start", "center", "end"][parseFloat(value)]);
+              // @todo: these two should be set in one instead of two seperate updates
+              setAlignItems(alignItems);
+              setJustifyContent(justifyContent);
```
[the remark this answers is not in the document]
kof: we need to change setProperty interface, publishUpdates inside it takes an array of updates, so its just about setProperty function, let me prototype a better interface in a sec

packages/webstudio-react-sdk/src/css/properties.ts
```diff
@@ -16,7 +16,7 @@ export const properties = {
       value: "normal",
     },
     popularity: 0.2430101,
-    appliesTo: "multilineFlexContainers",
+    appliesTo: "flexContainers",
```
kof, line 19: did you generate this change or changed manually? Bc this file is generated

kof:
@thysultan merged that pr for batching in here

apps/designer/app/designer/features/footer/breadcrumbs.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <>
+      <Button
+        ghost
+        onClick={onClick}
+        css={{
+          color: "$hiContrast",
+          px: "$2",
+          borderRadius: "100vh",
+          height: "100%",
+          lineHeight: "100%",
```
kof, line 23: same thing
```suggestion
          lineHeight: 1,
```

apps/designer/app/designer/shared/inspector/collapsible-section.tsx
```diff
@@ -51,12 +51,19 @@ export const CollapsibleSection = ({
...
             }}
           >
-            {isOpenFinal ? <ChevronDownIcon /> : <ChevronRightIcon />}
             <Text size="1" css={{ fontWeight: "500", flexGrow: 1 }}>
               {label}
             </Text>
+            <Flex
+              align="center"
+              justify="center"
+              css={{ marginRight: -5, opacity: ".4" }}
```
kof, line 63: opacity can be a number too
```suggestion
              css={{ marginRight: -5, opacity: 0.4 }}
 ```

Comment (kof) [reply apps/designer/app/designer/shared/inspector/collapsible-section.tsx:63]: btw why arbitrary margin number, can be design system scale?

File: packages/design-system/src/components/combobox.tsx [hunk 14]
```diff
@@ -115,7 +115,8 @@ export const List = <Item extends BaseItem>({
 export const ComboboxPopperContent = PopperContent;
 
 type ComboboxProps<Item> = {
-  name: string;
+  name?: string;
```
kof, line 118: hmm, why would name ever be optional?

packages/design-system/src/components/comboicon.tsx
```diff
@@ -0,0 +1,146 @@
...
+  },
+  "&[data-state=open]": {
+    backgroundColor: "$colors$blue10",
+    "& path": {
+      fill: "$colors$blue1",
+    },
+  },
+});
+
+export const Comboicon = ({
```
kof, line 87: Do we need this anywhere else in this exact setup? Maybe this is really a flex panel or style panel specific component, not design system wide?

kof: name could be IconButtonWithMenu

packages/design-system/src/components/comboicon.tsx
```diff
@@ -0,0 +1,146 @@
...
+        >
+          {items
+            .map(({ name, label }: { name: string; label: string }) => ({
+              icon: icons?.[name],
+              name,
+              label,
+            }))
+            .filter(({ icon }) => Boolean(icon))
+            .map(({ name, label, icon }) => {
+              const ItemIcon = icon as (props: unknown) => JSX.Element;
```
kof, line 128: you don't have a correct type here because you mapped to icon component in the first .map but didn't provide the type there and now have to cast

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -14,21 +14,23 @@
  * });
  */
 
-export type Direction = "horizontal" | "vertical";
+export type NumericGestureDirection = "horizontal" | "vertical";
 
-export type Value = number;
+export type NumericGestureValue = number;
```
kof, line 19: Why not just a NumericValue ?

packages/design-system/src/components/tabs.tsx
```diff
@@ -30,8 +30,8 @@ export const TabsTrigger = styled(TabsPrimitive.Trigger, {
   },
   [`& > ${Text}`]: {
     color: "$gray9",
-    fontSize: "11px",
-    fontWeight: "600",
+    fontSize: "calc($fontSizes$1 - 1px)",
+    fontWeight: "500",
```
kof, line 34:
```suggestion
    fontWeight: 500,
```

packages/design-system/src/components/tabs.tsx
```diff
@@ -30,8 +30,8 @@ export const TabsTrigger = styled(TabsPrimitive.Trigger, {
   },
   [`& > ${Text}`]: {
     color: "$gray9",
-    fontSize: "11px",
-    fontWeight: "600",
+    fontSize: "calc($fontSizes$1 - 1px)",
```
kof, line 33: I wonder why we would need that -1px thing, it should probably just use one of the predefined fontSizes

packages/design-system/src/stitches.config.ts
```diff
@@ -225,11 +226,13 @@ export const {
       2: "10px",
       3: "15px",
       4: "20px",
-      5: "25px",
+      5: "28px",
```
kof, line 229: should @taylornowotny confirm the grid here?

packages/design-system/src/stitches.config.ts
```diff
@@ -225,11 +226,13 @@ export const {
...
       6: "35px",
       7: "45px",
       8: "65px",
       9: "80px",
+      sidebarLeft: "45px",
+      sidebarRight: "240px",
```
kof, line 235: all of these could be a nummeric value, I copied this from radix's design system example

packages/icons/src/align-content.tsx
```diff
@@ -0,0 +1,157 @@
+import * as React from "react";
+import { IconProps } from "./types";
+
+export const AlignContentStart = React.forwardRef<SVGSVGElement, IconProps>(
```
kof, line 4: maybe we should have a consistent icon per file structure, wdyt? the rest is like this

kof (changes requested):
first pass done

kof:
Screenshot 2022-08-01 at 23 14 44

All I see on deployment preview

packages/design-system/src/components/primitives/numeric-gesture-control.ts
kof: btw we discovered recently a better name for that thing: "scrub", its an actual thing for this interaction, so this component could be NumericScrubControl

kof: or simply ScrubControl if its always numeric

kof: same for the rest of the related primitives

kof:
Oh I was confused by seeing those buttons in case of block, they shouldn't be there, right?

kof:
> Align-items, Align-Content, Row/Col-Gap.
> 
> According to MDN data they should(it says those properties apply to `allElements`), but realistically they shouldn't we need to filter them in the panel(I have a todo in the code for this), all the display sections should filter for what they want to display like the flex-panel does rather really on MDN data.

we will need to sort this out before merging because it looks confusing

packages/icons/src/align-content.tsx
kof: I think they are in one category like many other icons would be, but they are separate icons

kof:
When I switch from block to flex, this is what I see on preview
Screenshot 2022-08-04 at 10 14 14

apps/designer/app/designer/features/style-panel/shared/configs.ts
```diff
@@ -4,9 +4,30 @@ import {
...
+
+export type Control =
+  | "Spacing"
+  | "Combobox"
+  | "Color"
+  | "IconButtonWithMenu"
+  | "Select"
+  | "Empty";
+
+type StyleConfigWithItems = BaseStyleConfig & {
```
kof, line 25: hmm, why do we need an alias? seems like it can be 

```suggestion
export type StyleConfig = BaseStyleConfig & {
```

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,780 @@
...
+  currentStyle,
+  category,
+  styleConfigsByCategory,
+  moreStyleConfigsByCategory,
+}: RenderCategoryProps) => {
+  switch (category) {
+    case "layout": {
+      const css = {
+        alignItems: "center",
+        gap: "8px",
```
kof, line 673: this didn't fit a design system value?

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,780 @@
...
+        css={{
+          gridArea: cells[position],
+          alignItems:
+            `start center end start center end start center end`.split(" ")[
+              position
+            ],
+          flexDirection: "column",
+          width: "100%",
+          height: "100%",
+          gap: "3px",
```
kof, line 552: design system?

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,780 @@
...
+        }}
+      >
+        {["60%", "100%", "60%"].map((value, index) => (
+          <Flex
+            key={index}
+            css={{
+              blockSize: "33.3333%",
+              inlineSize: value,
+              background: "currentColor",
+              borderRadius: "2px",
```
kof, line 563: guess we could add this one to radii design system tokens?

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,780 @@
...
+            }}
+          ></Flex>
+        ))}
+      </Flex>
+    </Grid>
+  );
+};
+
+const LockControl = ({
+  css,
```
kof, line 573: instead of passing css, it seems like you would want to pass gridArea

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,780 @@
...
+                {styleConfigsByCategory}
+                <LockControl
+                  data-property="lock"
+                  css={{ gridArea: "lock" }}
+                  currentStyle={currentStyle}
+                  setProperty={setProperty}
+                />
+                <GridControl
+                  data-property="grid"
+                  css={{ gridArea: "grid" }}
```
kof, line 714: same here

kof: in fact you could get rid of both, CSS and data-property by passing one name and rendering it same way

packages/design-system/src/components/icon-button-with-menu.tsx
```diff
@@ -0,0 +1,140 @@
+import React from "react";
```
kof, line 1: a lot of arbitrary style values instead of using the design system in this file

kof (changes requested):
✅ second round of review

packages/design-system/src/components/icon-button-with-menu.tsx
kof: Also, does this component belong into design system or into the style panel specifically?

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,777 @@
...
+          fontWeight: "500",
+          textTransform: "capitalize",
+          "&:hover": { background: "none" },
+        }}
+      />
+    </Flex>
+  );
+};
+
+const IconButtonWithMenuControl = ({
```
kof, line 372: this could be a module in style-panel/lib, same like color picker and spacing

kof: sorry. I meant IconButtonWithMenu and its already there

kof: btw I just noticed you renamed lib to components, was the problem with lib infra related? I recall there was something ... maybe we should fix the lib to work as expected, because components folder is really meaningless, a lot of things are components, lib was specifically saying that those were things not coupled with the application state etc, so they are reusable even though still a local lib means they are designed for this feature specifically

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,777 @@
...
+          icons as unknown as Record<
+            string,
+            Record<string, (props: unknown) => JSX.Element>
+          >
+        )[styleConfig.property]
+      }
+      css={{
+        ...(styleConfig.property !== "flexDirection" && {
+          svg: {
+            transform: `rotate(${
```
kof, line 420: I think you want to put the transform into the icon code, basically create the icon name you need and use the other icon + transform to fake a specific icon

kof: all those icon states seem like belong into the lib/IconButtonWithMenu, because IconButonWithMenuControl should only contain the glue code between the render component and the state

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,777 @@
...
+          "&:not([data-state=open]):hover": {
+            backgroundColor: "$colors$blue4",
+          },
+        }),
+      }}
+    ></IconButtonWithMenu>
+  );
+};
+
+const GridControl = ({
```
kof, line 439: same thing, gridControl should be just the glue code between state and rendering

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,777 @@
...
+  )[justifyContent];
+  const position = row * 3 + column;
+  const batch = createBatchUpdate();
+  const setAlignItems = batch.setProperty("alignItems");
+  const setJustifyContent = batch.setProperty("justifyContent");
+
+  return (
+    <Grid
+      data-property={name}
+      css={{
```
kof, line 487: any arbitrary values here can come from design system?

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,777 @@
...
+        ))}
+      </Flex>
+    </Grid>
+  );
+};
+
+const LockControl = ({
+  name,
+}: // currentStyle,
+// setProperty,
```
kof, line 575: was this temp?

apps/designer/app/designer/features/style-panel/style-section.tsx
```diff
@@ -0,0 +1,777 @@
...
+          alignItems: "center",
+          height: "100%",
+        },
+      };
+      switch (currentStyle.display?.value) {
+        case "inline-flex":
+        case "flex": {
+          return (
+            <>
+              <Grid
```
kof, line 685: this def. looks like it doesn't belong here

apps/designer/app/designer/features/style-panel/style-section.tsx
kof: oh yeah, we should probably just use dist in others and stop ignoring lib

kof: its kindof wrong to use lib as a dist/build anyways

kof: can you send a separate pr for lib> dist renaming?

apps/designer/app/designer/features/style-panel/style-section.tsx
kof: sure, but we can move it now to make it make sense today, doesn't cost much work

kof:
- [ ] lock button seems to not have a tooltip like the other fields
- [ ] delay before first showing the tooltip seems too high, I think it should come much faster

apps/designer/app/designer/features/style-panel/controls/grid-control.tsx
```diff
@@ -0,0 +1,154 @@
...
+              borderRadius: "calc($radii$1 / 2)",
+            }}
+          ></Flex>
+        ))}
+      </Flex>
+    </Grid>
+  );
+};
+
+export { GridControl };
```
kof, line 154: Nit: would rather export right where the component is defined to see right away that its exported without scrolling

apps/designer/app/designer/features/style-panel/sections/layout.tsx
```diff
@@ -0,0 +1,101 @@
+import { Grid } from "@webstudio-is/design-system";
+import type { RenderCategoryProps } from "../style-sections";
+import { GridControl, LockControl, ShowMore } from "../controls";
+
+const CSSLayoutSection = {
+  alignItems: "center",
+  gap: "$space$styleSection",
+  "& > [data-type=iconbuttonwithmenu]": {
```
kof, line 8: I am not sure why layout knows about that specific component, seems like they should be fully decoupled

kof: sorry, I just now understood that the layout is the name of the section, not a "layout component"

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
```diff
@@ -0,0 +1,145 @@
...
+  Combobox,
+  IconButton,
+  TextField,
+  Tooltip,
+} from "@webstudio-is/design-system";
+import { PropertyName } from "./property-name";
+import { getFinalValue } from "../shared/get-final-value";
+import { useIsFromCurrentBreakpoint } from "../shared/use-is-from-current-breakpoint";
+import { ControlProps } from "../style-sections";
+import { RowGap, ColumnGap } from "@webstudio-is/icons";
```
kof, line 16: I just noticed all other icons have Icon in name, e.g. BrushIcon, should we stay consistent?

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
```diff
@@ -0,0 +1,145 @@
...
+    currentStyle,
+    inheritedStyle,
+    property: styleConfig.property,
+  });
+
+  if (value === undefined) return null;
+
+  const setValue = setProperty(styleConfig.property);
+
+  switch (styleConfig.property) {
```
kof, line 51: we need a code organization that scales to support all properties

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
```diff
@@ -0,0 +1,145 @@
...
+                    },
+                  }),
+                }}
+              >
+                <Icon />
+              </IconButton>
+            </Tooltip>
+            <TextField
+              type="number"
+              defaultValue={parseFloat(String(value.value)) || 0}
```
kof, line 98: this looks very suspicious, what kind of value is this all for?

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "build": "turbo run build",
-    "checks": "turbo run checks",
+    "checks": "turbo run build && turbo run checks",
```
kof, line 13: what is it fixing? build is kinda expensive, you don't want to have to run it every time you need checks

packages/design-system/src/components/icon-button-with-menu.tsx
```diff
@@ -0,0 +1,148 @@
...
```
kof, line 1: Do you think we will ever need this component beyond style panel? If no it should not be in design-system

kof (commented):
another round done

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
kof: I guess I don't understand what is being specifically fixed here. for columnGap or rowGap, but it definitely looks like it belongs somewhere else

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
kof: I don't understand that todo or what problem this parsing and stringification is fixing

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
```diff
@@ -48,11 +48,13 @@ export const ComboboxControl = ({
 
   const setValue = setProperty(styleConfig.property);
 
+  // @todo abstract to a variant of the combobox component
+  // this whole switch block is a stop-gap measure, because i worked on flex-controls before the combobox update
+  // the combobox should be updated to supporting this rendered production
```
kof, line 53: Is ds combobox is missing something that is required for this?

package.json
kof: that's the problem we need to fix differently I think cc @andrasbacsai

apps/designer/app/designer/features/style-panel/controls/combobox-control.tsx
kof: it should allow for any children, this should be easy to change right away, no?

kof: looking at it now its weird that we made items render strings instead of any jsx

apps/designer/app/designer/shared/inspector/collapsible-section.tsx
kof: why are you not using spacing from design system for the margin?

package.json
kof: please remove the build here, we will soon have a solution for typescript (probably this week)

packages/design-system/src/components/icon-button-with-menu.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+const itemStyles = {
+  all: "unset",
+  fontSize: "$2",
+  lineHeight: 1,
+  color: "$colors$slate12",
+  borderRadius: "$radii$1",
+  display: "flex",
+  alignItems: "center",
+  height: "$sizes$5",
```
kof, line 24: why $sizes$5 and not $5? just wondering maybe you don't know or maybe I don't know why this is better?

packages/design-system/src/components/icon-button-with-menu.tsx
```diff
@@ -0,0 +1,148 @@
...
+});
+
+const StyledArrow = styled(DropdownMenuPrimitive.Arrow, {
+  "& *": {
+    fill: "$colors$gray4",
+    stroke: "$colors$gray8",
+  },
+});
+
+const IconButton = styled("button", {
```
kof, line 57: Was IconButton component we already have not usable for this?

packages/icons/src/align-content-center.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+import { IconProps } from "./types";
+
+export const AlignContentCenterIcon = React.forwardRef<
+  SVGSVGElement,
+  IconProps
+>(({ color = "currentColor", ...props }, forwardedRef) => {
+  return (
+    <svg
+      width="24"
```
kof, line 10: The new grid for icons is 16px, can we get them in that size now or you want to create a separate task? If separate please create the task now.

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,128 @@
...
+  Tooltip,
+} from "@webstudio-is/design-system";
+import { PropertyName } from "../../shared/property-name";
+import { getFinalValue } from "../../shared/get-final-value";
+import { useIsFromCurrentBreakpoint } from "../../shared/use-is-from-current-breakpoint";
+import { ControlProps } from "../../style-sections";
+import { RowGapIcon, ColumnGapIcon } from "@webstudio-is/icons";
+import { Unit } from "../../shared/unit";
+
+export const TextControl = ({
```
kof, line 17: I don't understand what this control is for from its semantics. It also seems like it could be 2 separate controls, because renders an entirely different interface depending on property and it shouldn't care about the property at all since we extracted the mapping to outside

apps/designer/app/designer/features/style-panel/shared/flex-grid.tsx
```diff
@@ -0,0 +1,152 @@
+import { Flex, Grid, IconButton } from "@webstudio-is/design-system";
+import { DotFilledIcon } from "@webstudio-is/icons";
+import { useIsFromCurrentBreakpoint } from "./use-is-from-current-breakpoint";
+import type { Style } from "@webstudio-is/react-sdk";
+import type { CreateBatchUpdate } from "./use-style-data";
+
+export const FlexGrid = ({
```
kof, line 7: this component is in shared folder, looks like it would be shared, but afaik this FlexGrid UI is ever only going to be used once where it is currently used, right? Then why putting it to shared? 

please ask yourself the same question for other things from shared

kof:
@natlusyht  Just tested, seems like setting width/height doesn't work

apps/designer/app/designer/features/footer/breadcrumbs.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const Breadcrumbs = ({ publish }: BreadcrumbsProps) => {
+  const [selectedInstanceData] = useSelectedInstanceData();
+  const selectedInstancePath = useSelectedInstancePath(
+    selectedInstanceData?.id
+  );
+  return (
+    <Flex align="center" css={{ height: "100%" }}>
+      {selectedInstancePath.length === 0 ? (
+        <Breadcrumb>
+          <__DEPRECATED__Text size="1">No instance selected</__DEPRECATED__Text>
```
kof, line 44: please migrate to Text right away

kof:
Is it just me or the difference between lock closed and lock open is very hard to see on the icon?

Screenshot 2022-08-28 at 12 42 35
Screenshot 2022-08-28 at 12 42 41

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,228 @@
...
+              : String(currentValue);
+          setValue(currentValueAsString);
+        }}
+      />
+      {value.type === "unit" ? (
+        <Units
+          value={value.unit}
+          items={units
+            .filter((v) =>
+              [
```
kof, line 150: why do we need this shortlist instead of just using units entirely?

kof: Oh its because you don't want them all in the dropdown. I think it makes sense, but we still need to make other units possible, no?

kof:
I just noticed we don't allow any more to type the unit with keyboard, we should allow though, because 

1. its an alternative way to enter a unit with keyboard
2. it would allow using a unit that is not in the menu, which only displays a subset of all units

From UX standpoint it was working this way: you type a unit, as soon as it gets recognized - it sets the unit into the unit area

kof:
Screenshot 2022-08-28 at 13 42 48

what do we do with lack of space to fully display "normal"? also cc @taylornowotny

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,228 @@
...
+                  Math.abs(currentValue).toString().indexOf(".") + 2
+                )
+              : String(currentValue);
+          setValue(currentValueAsString);
+        }}
+      />
+      {value.type === "unit" ? (
+        <Units
+          value={value.unit}
+          items={units
```
kof, line 148: we can extract that this entire thing where you map reduce and map the units into a static array above, so it happens just once

kof: and calling it something like menutUnitsSubset would explain why we have a shortlist

kof:
background of a set value goes over the entire row, instead of property only, this doesn't seem right, but we also need a design proposal for sizes 

Screenshot 2022-08-28 at 16 30 29

kof:
> Unit'ed props like that shouldn't have the default value "normal" as the MDN data suggests more user ergonomic if it's 0px or something.

Would that be valid, do you have a separate todo for that?

kof:
> Yes, noticed, since we changed to update on onChange instead of onEnter since intermediate states of unit inputs are recognised as invalid i.e typing p after 200 -> 200p p, removed recognised as invalid: exemplified by the fact that if you pasted the whole unit, it does as it should.

Ok, but we need a fix for this, because this is breaking this behavior, is it so hard that we need a separate todo for it?

kof:
@taylornowotny did we arrive at a decision on the icon for open/collapsed state of style panel sections?

We seem to be changing back and forth, I changed a while ago to arrow before, Sultan just changed in this PR to arrow after

Screenshot 2022-08-28 at 23 41 43

kof:
@thysultan @taylornowotny  the field column gap is so narrow that even 2 digits are already cut off a bit, we need to figure  out more space, at least 3 digits or something
Screenshot 2022-08-28 at 23 43 51

apps/designer/app/designer/features/style-panel/shared/flex-grid.tsx
kof: shared should be only ever used when you need to share that functionality, local location is always preferred if no sharing necessary

reason is: co-location principle (high cohesion) + shared should be considered generally more fragile, because used in multiple places, so needs higher quality, more testing etc., so shared is the last resort. 

We can have separate files and still keep it local by creating a folder and an index.ts that only exports public api

kof:
We still need to fix the perf problem in this PR, because it makes it almost unusable, it is a regression, current prod version doesn't have this problem

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,228 @@
...
+  numericScrubControl,
+} from "@webstudio-is/design-system";
+import { getFinalValue } from "../../shared/get-final-value";
+import { useIsFromCurrentBreakpoint } from "../../shared/use-is-from-current-breakpoint";
+import { ControlProps } from "../../style-sections";
+import { iconConfigs, StyleConfig } from "../../shared/configs";
+import { type StyleValue, type Unit, units } from "@webstudio-is/react-sdk";
+import { ChevronDownIcon } from "@webstudio-is/icons";
+
+const CSSGridLeft = {
```
kof, line 28: these shouldn't start from upper case, even in case of abbreviation its still cssGridLeft

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,228 @@
...
+  height: "calc($sizes$6 - 6px)",
+  width: "calc($sizes$6 - 6px)",
+  transform: "translate(3px, 3px)",
+  "&:focus": {
+    boxShadow: "none",
+  },
+};
+
+const CSSTextUnit = {
+  fontSize: "$1",
```
kof, line 54: you should not need to set the fontSize manually, $1 is 12 px in design system, also I just realized that I used px values for some reason in the Text component, so I don't know what went through my head when I did this (I used the values straight from figma) but they should have been $1 $2 etc from the theme

kof: I will change those px values to theme values in a bit directly on main, but we should not need to set this here in any case

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,228 @@
...
+        ...CSSGridRight,
+        button: { ...CSSIconButton },
+      }}
+    >
+      <IconButtonWithMenu
+        icon={<Text css={CSSTextUnit}>{value === "number" ? "—" : value}</Text>}
+        items={items}
+        value={String(value)}
+        onChange={onChange}
+      ></IconButtonWithMenu>
```
kof, line 221:
```suggestion
      />
```

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,228 @@
...
+}: {
+  value: Unit;
+  items: StyleConfig["items"];
+  onChange?: (value: string) => void;
+}) => {
+  return (
+    <Box
+      css={{
+        ...CSSGridRight,
+        button: { ...CSSIconButton },
```
kof, line 213: I am confused how this supposed to work, button is not a CSS property ... what do I miss?

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
kof: yeah I think this is a good option too, @taylornowotny what do you think? we got a bunch of other units in CSS and even though they are not popular, they exist and not showing them entirely is also strange cherry picking

We can sort these by popularity and they will end up in the list further down and you don't see them if you don't scroll but at least they are there, wdyt?

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
kof: btw I used style suffix everywhere else: gridLeftStyle

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,251 @@
...
+  .slice(0)
+  .sort((v) =>
+    ["%", "px", "rem", "em", "ch", "vh", "vw", "hv", "vmin", "vmax"].includes(v)
+      ? -1
+      : 1
+  );
+
+const CSSGridLeft = {
+  "& ~ input": {
+    paddingLeft: "calc($sizes$6 + 6px)",
```
kof, line 30: why do we have 6px, you sure we couldn't have used one of the design system values?

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,251 @@
...
+  let Icon;
+
+  const { label, property } = styleConfig;
+
+  // @todo this silo's the flex section
+  let multiColumn = true;
+  if (
+    category === "layout" &&
+    property.includes("Gap") &&
+    (currentStyle.display?.value as string).includes("flex")
```
kof, line 96: why is this having specific property checks? It shouldn't , its not a text control any more ... instead you need a property that does the same thing you need but semantically is not attached to the specific property, same in cases below

packages/icons/src/align-content-center.tsx
kof: ping

packages/design-system/src/components/icon-button-with-menu.tsx
kof: ping

packages/design-system/src/components/icon-button-with-menu.tsx
kof: ping

package.json
kof: ping

apps/designer/app/designer/shared/inspector/collapsible-section.tsx
kof: ping

apps/designer/app/designer/features/footer/breadcrumbs.tsx
kof: ping

apps/designer/app/designer/features/style-panel/shared/flex-grid.tsx
kof: I would normally be not that serious about a local shared folder inside a feature, since that shared folder is not used outside of the style panel, but I am afraid that style panel is going to be a very large portion of the code base, so we need to be strict about it here

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
kof: that's weird because it doesn't use "& button", not sure if stitches made & optional for nesting like this, because button is in place of a property name, how would stitches know you are describing a nested selector

packages/icons/src/align-content-center.tsx
kof: do we need a todo with a link to an issue?

apps/designer/app/designer/features/inspector/inspector.tsx
```diff
@@ -21,6 +21,7 @@ const contentStyle = {
   display: "flex",
   flexDirection: "column",
   overflow: "auto",
+  // boxShadow: "0 -1px 0 $colors$panelOutline",
```
kof, line 24: can be removed?

apps/designer/app/designer/features/style-panel/controls/select/select-control.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const setValue = setProperty(styleConfig.property);
+
+  return (
+    <Flex align="center">
+      <PropertyName
+        property={styleConfig.property}
+        label={styleConfig.label}
+        css={{
+          fontWeight: "500",
```
kof, line 29: I haven't looked up in the design, but why is this property name different?

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,254 @@
...
+      : 1
+  );
+
+const gridTemplateStyle = {
+  gridTemplateColumns: "$sizes$6 auto $sizes$6",
+  gridTemplateRows: "repeat(1, 1fr)",
+};
+
+const gridLeftStyle = {
+  "& ~ input": {
```
kof, line 34: I have been just discussing the kind of selectors in https://github.com/webstudio-is/webstudio-designer/pull/251

The problem is that you are overriding using cascade from parent into the child, where the child is not "your" component. It's a component that knows nothing about what you are doing from here.

This has the potential for specificity issues and accidental visual regressions.

Instead, we could use some explicit way of passing down CSS values to the child: using a prop or using a CSS variable.

Please review the other PR, there are still some open options there and then we can apply the solution here.

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,254 @@
...
+};
+
+const iconButtonGridLeftStyle = {
+  ...gridLeftStyle,
+  ...iconButtonStyle,
+};
+
+const iconButtonGridRightStyle = {
+  ...gridRightStyle,
+  "& button": { ...iconButtonStyle },
```
kof, line 72: same here

apps/designer/app/designer/features/style-panel/sections/layout/shared/lock.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }, [aKey, bKey, aVal, bVal, isPaired, batchUpdate]);
+
+  return (
+    <Tooltip
+      content={isPaired ? "Unlock" : "Lock"}
+      delayDuration={400}
+      disableHoverableContent={true}
+    >
+      <IconButton
+        data-property={name}
```
kof, line 43: why is lock having this attribute as if its a separate control?

kof:
@taylornowotny please try this branch out, we fixed the preview deployments, the biggest changes here are:
1. flex UI
2. scrub interaction for inputs/autocompletes

kof:
@andrasbacsai publish-test fails because with this error:

```
Run npm publish -w ./packages/http-client --access public --dry-run
npm ERR! No workspaces found:
npm ERR!   --workspace=./packages/http-client

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2022-09-05T18_45_44_279Z-debug-0.log
```

kof:
Screenshot 2022-09-05 at 21 44 05
We don't have enough space in the inputs "normal" is cut-off

kof:
- [ ] when I scroll while pointer is over the Select - it closes

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -0,0 +1,264 @@
...
+          />
+        ) : (
+          <Items
+            value={value.value}
+            items={styleConfig.items}
+            onChange={(item) => setValue(item)}
+            onHover={(item) => setValue(item, { isEphemeral: true })}
+          />
+        )}
+        <TextField
```
kof, line 169: this was supposed to be a Combobox instead of this + IconButtonWithMenu

kof:
- [ ] turns out text-control doesn't work as a name, because we also need a plain text field in the future, I suggest to call it css-value-control or something more specific than text

andrasbacsai:
> @andrasbacsai publish-test fails because with this error:
> 
> ```
> Run npm publish -w ./packages/http-client --access public --dry-run
> npm ERR! No workspaces found:
> npm ERR!   --workspace=./packages/http-client
> 
> npm ERR! A complete log of this run can be found in:
> npm ERR!     /home/runner/.npm/_logs/2022-09-05T18_45_44_279Z-debug-0.log
> ```

Fixing in #263, please take a look.

kof:
@natlusyht please create a follow-up issue with all todos we have talked about here that are still unsolved + those which you already know, like the combobox usage for inputs with units etc, lets a good clear overview of next todos, I am merging this now because we can't sustain keeping this branch with so many changes

PR #185, opened 2022-07-24
Added createBatchUpdate to apply multiple properties updates at once

kof:
Hey, I just wrote this function, I didn't test it and it needs to be passed down to the components like setProperty, the idea is this:

```ts
const batch = createBatchUpdate()

const setDisplayValue = batch.setProperty('display')
setDisplayValue('block')

const setSomeOtherValue = batch.setProperty('something')
setSomeOtherValue('somevalue')

batch.publish()
```


kof:
just merge it into your branch if you think its usable

PR #190, opened 2022-07-26
Allow deleting assets

SaraVieira:
This will delete an image from both teh DB and the file system/s3

related https://github.com/webstudio-is/webstudio-designer/issues/37

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [x] added tests


apps/designer/app/routes/designer/$id.tsx
```diff
@@ -35,17 +38,42 @@ type Error = {
...
+  try {
+    if (request.method === "DELETE") {
+      const formData = await request.formData();
+      if (formData.get("assetId")) {
+        const id = formData.get("assetId") as string;
+        const name = formData.get("assetName") as string;
+        const deletedAsset = await deleteAsset({
+          id,
+          name,
+          dirname: __dirname,
```
kof, line 50: please don't use it, we need something importable from a single location for __dirname, I feel like I mentioned it before and wonder now if we have multiple places now

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -35,17 +38,42 @@ type Error = {
 
 export const action: ActionFunction = async ({ request, params }) => {
   if (params.id === undefined) throw new Error("Project id undefined");
+  try {
+    if (request.method === "DELETE") {
+      const formData = await request.formData();
+      if (formData.get("assetId")) {
+        const id = formData.get("assetId") as string;
+        const name = formData.get("assetName") as string;
```
kof, line 46: instead of asserting with typescript, validate mit zod

apps/designer/app/shared/design-system/stitches.config.ts
```diff
@@ -219,6 +219,7 @@ export const {
       7: "45px",
       8: "65px",
       9: "80px",
+      10: "105px",
```
kof, line 222: please check with @taylornowotny and/or @thysultan regarding the grid we have in the design system

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Then i can't delete the file locally. Do you have a better option?

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Will do

apps/designer/app/shared/design-system/stitches.config.ts
SaraVieira: Will check the grid and ask if i don't find it

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Mit zot, went all German there ahahah

packages/asset-uploader/src/delete.ts
```diff
@@ -0,0 +1,41 @@
...
+
+      return await deleteAssetInDb(id);
+    } else {
+      const directory = await getImageLocalDirectory(dirname);
+      await unlink(path.join(directory, name));
+
+      return await deleteAssetInDb(id);
+    }
+  }
+  throw new Error("Asset does not exist");
```
kof, line 40: can we throw early? like early return, easier to skip code I don't need to read

kof (changes requested):
found a few things, will test soon

apps/designer/app/routes/designer/$id.tsx
SaraVieira: Done with zod-form-data, updated other places to use it

apps/designer/app/shared/design-system/stitches.config.ts
SaraVieira: According to the design system everything should actually be in increments of 8, I updated it

packages/asset-uploader/src/delete.ts
SaraVieira: done!|

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
```diff
@@ -2,48 +2,53 @@ import { ImageIcon } from "~/shared/icons";
...
+export const useAssetsState = (baseAssets: Array<Asset>) => {
+  const imageChanges = useActionData();
 
-  const [assets, setAssets] = useState(baseAssets);
+  const [assets, setAssets] = useState<Asset[]>(baseAssets);
 
   useEffect(() => {
-    if (newImages?.length) {
+    if (imageChanges?.uploadedAssets?.length) {
       setAssets((currentAssets) => [
```
kof, line 17: I wonder if there was an actual case for using setState callback here

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
```diff
@@ -2,48 +2,53 @@ import { ImageIcon } from "~/shared/icons";
...
       </Flex>
       <Grid columns={2} gap={2}>
         {assets.map((asset) => (
-          <AssetManagerImage
-            key={asset.id}
-            path={asset.path}
-            alt={asset.alt || asset.name}
-            status={asset.status}
-          />
+          <AssetManagerImage key={asset.id} {...asset} />
```
kof, line 51: This component is so much better off with a "Thumbnail" as a name :)

apps/designer/app/designer/features/sidebar-left/panels/components/image.tsx
```diff
@@ -18,18 +26,17 @@ const useImageWithFallback = ({ path }: { path: string }) => {
...
+  const submit = useSubmit();
   const isUploading = status === "uploading";
   const src = useImageWithFallback({ path });
+  const [isToolTipOpen, setTolltipOpen] = useState(false);
   const [progressBarPercentage, setProgressBarPercentage] = useState(0);
+  const tooltipRef = useRef(null);
+  useClickAway(tooltipRef, () => {
+    setTolltipOpen(false);
+  });
+  useKey("Escape", () => setTolltipOpen(false));
```
kof, line 39: I wonder if we should use consistently useHotkeys for this

apps/designer/app/designer/features/sidebar-left/panels/components/image.tsx
```diff
@@ -67,6 +81,37 @@ export const AssetManagerImage = ({
           ...(isUploading ? { filter: "blur(1px)", opacity: 0.7 } : {}),
         }}
       ></Box>
+      {!isUploading && (
+        <Box ref={tooltipRef}>
+          <Tooltip
```
kof, line 86: is this potentially how our default tooltip inside design system should always be styled?

kof:
I wanted to test on preview this PR, is this broken like this on main or only this branch?

https://webstudio-designer-g0pu1k7il-webstudio-is.vercel.app/login
Screenshot 2022-07-26 at 20 01 16

kof:
What's the keyboard accessibility story here with asset manager?

kof:
Screenshot 2022-07-26 at 20 05 52

kof:
cc @taylornowotny 

Screenshot 2022-07-26 at 20 14 09

kof:
@SaraVieira delete button click doesn't do anything

SaraVieira:
@kof the issue with login was because I cahnged the spacing everywhere. Needs to be in a spearate PR

Reverted and added the width in px and added a todo to the tooltip

apps/designer/app/designer/features/sidebar-left/panels/components/image.tsx
SaraVieira: yeah, I agree, gonna change it in the tooltip

apps/designer/app/designer/features/sidebar-left/panels/components/image.tsx
SaraVieira: definitly, updated

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
SaraVieira: hm?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
SaraVieira: updated!

SaraVieira:
@kof fixed the button, it was the Box that was around the tooltip was not catching anything.

We can now pass a ref to the tooltip and I also added the loading progress bar to the deleting to be more obvious what is happening

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
```diff
@@ -0,0 +1,135 @@
...
+  path,
+  alt,
+  status,
+  name,
+  id,
+}: Asset) => {
+  const submit = useSubmit();
+  const isUploading = status === "uploading";
+  const src = useImageWithFallback({ path });
+  const [isToolTipOpen, setTolltipOpen] = useState(false);
```
kof, line 41:
```suggestion
  const [isTooltipOpen, setTolltipOpen] = useState(false);
```

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const [isDeleting, setIsDeleting] = useState(false);
+  const [progressBarPercentage, setProgressBarPercentage] = useState(0);
+
+  const closeTooltip = () => setTolltipOpen(false);
+
+  const tooltipRef = useRef(null);
+  useClickAway(tooltipRef, closeTooltip);
+  useHotkeys(shortcuts["escape"], closeTooltip);
```
kof, line 49: radix tooltip is supposed to handle both, esc and click outside https://www.radix-ui.com/docs/primitives/components/tooltip#content

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
SaraVieira: Fixed

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
SaraVieira: I see what happened, it had to be updated. Updated it to v1 and it now supports everything

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
```diff
@@ -18,29 +26,41 @@ const useImageWithFallback = ({ path }: { path: string }) => {
...
+
   // @todo rewrite this fake indication to show real progress
   useInterval(
     () => {
       setProgressBarPercentage((percentage) =>
         percentage < 60 ? percentage + 1 : percentage
       );
     },
-    isUploading ? 100 : null
+    isUploading || isDeleting ? 100 : null
```
kof, line 52: why is it even taking null, would have expected 0 or undefined there

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
```diff
@@ -18,29 +26,41 @@ const useImageWithFallback = ({ path }: { path: string }) => {
...
   const isUploading = status === "uploading";
   const src = useImageWithFallback({ path });
+  const [isTooltipOpen, setTolltipOpen] = useState(false);
+  const [isDeleting, setIsDeleting] = useState(false);
   const [progressBarPercentage, setProgressBarPercentage] = useState(0);
 
+  const closeTooltip = () => setTolltipOpen(false);
+
   // @todo rewrite this fake indication to show real progress
   useInterval(
```
kof, line 46: would move it into useFakeProgress

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -33,19 +37,46 @@ type Error = {
...
   if (params.id === undefined) throw new Error("Project id undefined");
+  try {
+    if (request.method === "DELETE") {
+      const { assetId, assetName } = deleteAssetSchema.parse(
+        await request.formData()
+      );
+      const deletedAsset = await deleteAsset({
+        id: assetId,
+        name: assetName,
+        dirname: __dirname,
```
kof, line 55: tbh I don't even understand why we need __dirname, which is the routes/designer in this case ... I would need to follow this code through but on the surface it doesn't make sense, because that folder where we handle the route should not be used at all

apps/designer/package.json
```diff
@@ -43,7 +43,7 @@
     "@radix-ui/react-tabs": "^0.1.5",
     "@radix-ui/react-toggle": "^0.1.4",
     "@radix-ui/react-toggle-group": "^0.1.5",
-    "@radix-ui/react-tooltip": "^0.1.7",
+    "@radix-ui/react-tooltip": "^1.0.0",
```
kof, line 46: we should probably try to upgrade them all unless it breaks something) also feel free to separate the upgrade

SaraVieira: yeah, should definitely be another PR

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
SaraVieira: Done!

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const newImage = new Image();
+    newImage.onload = () => setSrc(path);
+    newImage.onerror = () => setSrc(brokenImage);
+    newImage.src = path;
+  }, [path]);
+
+  return src;
+};
+
+export const useFakeProgress = ({
```
kof, line 29: why do you export

apps/designer/package.json
SaraVieira: tried it and unfortunatly we need to see why it all errors out

image

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
SaraVieira: Removed, its habit

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
kof: setAsset(callback) vs assets

packages/asset-uploader/src/constants.ts
```diff
@@ -1 +1 @@
-export const DEFAULT_UPLPOAD_PATH = "uploads";
+export const DEFAULT_UPLPOAD_PATH = "pub lic/uploads";
```
kof, line 1: `pub lic`???

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
```diff
@@ -0,0 +1,142 @@
...
+    formData.append("assetId", id);
+    formData.append("assetName", name);
+    closeTooltip();
+    setIsDeleting(true);
+    submit(formData, { method: "delete" });
+  };
+
+  return (
+    <Box
+      title={alt || ""}
```
kof, line 78: some better fallback title maybe?

kof: like a meaningful one

packages/asset-uploader/src/delete.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getS3Client } from "./targets/s3/client";
+
+export const deleteAsset = async ({
+  id,
+  name,
+}: {
+  id: string;
+  name: string;
+}): Promise<Asset> => {
+  const currentAsset = await prisma?.asset.findUnique({
```
kof, line 17: how is prisma potentially not defined?
```suggestion
  const currentAsset = await prisma.asset.findUnique({
```

packages/asset-uploader/src/delete.ts
```diff
@@ -0,0 +1,38 @@
...
+    await getS3Client().send(
+      new DeleteObjectCommand({
+        Bucket: s3Envs.S3_BUCKET,
+        Key: name,
+      })
+    );
+
+    return await deleteAssetInDb(id);
+  } else {
+    const directory = await imageFSDirectory();
```
kof, line 33:
```suggestion
    const directory = await imageFsDirectory();
```

packages/asset-uploader/src/helpers/image-fs-path.ts
```diff
@@ -0,0 +1,18 @@
+import path from "path";
+import appRoot from "app-root-path";
+import { fsEnvVariables } from "../schema";
+const fsUploadVars = fsEnvVariables.parse(process.env);
+
+export const imageFSDirectory = () => {
+  const basePath =
+    process.env.NODE_ENV === "development"
+      ? path.join(
```
kof, line 9: please avoid ternaries when you can, this is a perfect example where you could have returned early instead

kof (approved):
see a bunch of nit picks and lets go

packages/asset-uploader/src/helpers/image-fs-path.ts
SaraVieira: I am not returning early, I am deploying this to add the other part of the if

kof: but you can

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
SaraVieira: I always tend to use the callback, its way safer

apps/designer/app/designer/features/sidebar-left/panels/components/thumbnail.tsx
SaraVieira: added the name

packages/asset-uploader/src/delete.ts
SaraVieira: done

packages/asset-uploader/src/helpers/image-fs-path.ts
SaraVieira: i was still testing, it's changed now

I needed to add the version for prod

packages/asset-uploader/src/delete.ts
SaraVieira: done

apps/designer/app/designer/features/sidebar-left/panels/asset-manager.tsx
kof: by doing so you are waiting for the next update cycle, this is only useful when you know there can be an update that came from somewhere else, its not a thing we need to use as a general pattern, only in special cases

SaraVieira: removing it actually broke it, this makes sure we are using the latest version

kof: because you didn't set the dependency?

SaraVieira: I did,updated it to include baseAssets

PR #191, opened 2022-07-27
Fix sync status icon jumping around in width when in loading state

kof:
## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [x] added tests


apps/designer/app/designer/features/topbar/sync-status.tsx
```diff
@@ -1,45 +1,48 @@
...
+  AccessibleIcon,
+  styled,
+  Box,
+} from "~/shared/design-system";
 import { CheckIcon, DotsHorizontalIcon } from "~/shared/icons";
 import { useSyncStatus } from "../../shared/nano-states";
 
 const iconSize = 15;
 
+const StyledCheckIcon = styled(CheckIcon, {
```
kof, line 13: cc @thysultan instead of nested selector

kof: alternative is to make all icons accept a css prop so we don't have to use styled

PR #201, opened 2022-07-27
Edge published sites

andrasbacsai:
## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/routes/rest/publish.ts
```diff
@@ -9,42 +9,25 @@ export const action: ActionFunction = async ({ request }) => {
...
-          headers,
-          body: JSON.stringify({
-            updated: new Date().getTime(),
-            tree,
-            breakpoints,
-            props,
-          }),
-        }
-      );
+      const response = await fetch(process.env.EDGE_PUBLISHER_ENDPOINT, {
```
kof, line 17: I wouldn't even call it edge, because we can use the same endpoint to publish to anywhere else

andrasbacsai: Good point!

apps/designer/app/routes/rest/publish.ts
```diff
@@ -9,9 +9,29 @@ const schema = zfd.formData({
 
 export const action: ActionFunction = async ({ request }) => {
   const { domain, projectId } = schema.parse(await request.formData());
+  if (!domain || !projectId) {
```
kof, line 12: you can move empty string checks to zod schema

apps/designer/app/designer/features/topbar/publish.tsx
```diff
@@ -15,15 +15,20 @@ import {
...
 type PublishButtonProps = { project: Project };
 
-const host =
-  typeof location === "object"
-    ? location.host.includes("webstudio.is")
-      ? "wstd.io"
-      : location.host
-    : "";
+const host = () => {
+  if (env.PUBLISHER_ENDPOINT && env.PUBLISHER_DOMAIN !== "") {
```
kof, line 23: both should be not empty strings and non null/undefined? or is it validated against null/undefined somewhere before?

apps/designer/app/designer/features/topbar/publish.tsx
```diff
@@ -15,15 +15,20 @@ import {
...
+    return env.PUBLISHER_DOMAIN;
+  }
+  if (typeof location === "object") {
+    if (location.host.includes("webstudio.is")) return "wstd.io";
```
kof, line 27: can we avoid hardcoding this check, can't it be based on env?

apps/designer/app/designer/features/topbar/publish.tsx
andrasbacsai: They are validated in the latest commit.

apps/designer/app/designer/features/topbar/publish.tsx
andrasbacsai: It could. Should we rename the `PUBLISHER_DOMAIN` to `SITES_DOMAIN` so it can be used application-wide to identify the domain?

kof: Oh we definitely should have one and the same name if its about naming the same thing, I guess publisher domain makes sense as a name

The other webstudio.is could be the DESIGNER_DOMAIN

andrasbacsai: Could you please check if this is what you thinking of? Thanks in advance.

apps/designer/env-check.js
```diff
@@ -13,7 +13,7 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 require("dotenv").config();
 
-const REQUIRED_ENVS = ["DATABASE_URL", "AUTH_SECRET"];
+const REQUIRED_ENVS = ["DATABASE_URL", "AUTH_SECRET", "DESIGNER_DOMAIN"];
```
kof, line 16: does this mean we need to set it locally too? Should that be required for production only?

andrasbacsai: Yes, you need to set it locally as well, as I removed the hardcoded domain from the codebase.

andrasbacsai:
LMK, if I can ship it to prod.

apps/designer/app/designer/features/topbar/publish.tsx
```diff
@@ -15,15 +15,16 @@ import {
...
-  typeof location === "object"
-    ? location.host.includes("webstudio.is")
-      ? "wstd.io"
-      : location.host
-    : "";
+const host = () => {
+  if (env.PUBLISHER_ENDPOINT && env.PUBLISHER_DOMAIN) {
+    return env.PUBLISHER_DOMAIN;
+  }
+  return env.DESIGNER_DOMAIN || location.host;
```
kof, line 26: please add the comment why we use location.host here

kof (approved):

Issue #165, opened 2022-07-20
Create toast component


SaraVieira:
This issue is to track the creation of a new component for toasts.

It should support:
- Sucess
- Error
- Promises

As decided in https://github.com/webstudio-is/webstudio-designer/issues/85 we will use:

- https://react-hot-toast.com/ for their logic
- https://www.radix-ui.com/docs/primitives/components/toast for all the UI

PR #208, opened 2022-07-29
Toasts v1

SaraVieira:
## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

image
image

closes https://github.com/webstudio-is/webstudio-designer/issues/165


packages/design-system/src/components/toast.tsx
```diff
@@ -0,0 +1,115 @@
...
+  margin: 0,
+  listStyle: "none",
+  zIndex: "$max",
+  outline: "none",
+});
+
+const StyledToast = styled(ToastPrimitive.Root, {
+  borderRadius: 6,
+  boxShadow:
+    "hsl(206 22% 7% / 35%) 0 $space$2 $space$6 -$space$2, hsl(206 22% 7% / 20%) 0 $space$2 $space$4 -$space$3",
```
SaraVieira, line 44: Is there a way to add our colors with transparency?

kof: hmm not sure how to deal with it, we can  define colors in hsl and then have a function that changes opacity parameter, but we need to discuss with @taylornowotny when would you use that and why

SaraVieira: I was thinking RGBA, but that's an implementation detail.

For shadows and stuff we should definitely have something that allows to pass transparency

kof: the question is when do you need to pass a non-static, predefined transparency

SaraVieira: I do agree, at remote we had these predefined so for example $highContrast_20 would be 20% opacity

packages/design-system/src/components/toast.tsx
```diff
@@ -0,0 +1,115 @@
...
+          </Box>
+          <StyledTitle>{toast.message}</StyledTitle>
+        </StyledToast>
+      ))}
+      <StyledViewport />
+    </ToastPrimitive.ToastProvider>
+  );
+};
+
+export const toast = BaseToast;
```
kof, line 115: the mapping is weird, if its just a function, not a constructor and we are calling it as such, why not import it as `toast as well`
```suggestion
export {toast}
```

packages/design-system/src/index.ts
```diff
@@ -11,6 +11,7 @@ export {
   theme,
   type CSS,
 } from "./stitches.config";
+export { toast, Toaster } from "./components/toast";
```
kof, line 14: I started exporting specific components but its tedious, would only do that if there is something to hide I a component

```suggestion
export * from "./components/toast";
```

kof (approved):
a few nits, looking good, lets merge it, then wait for @taylornowotny's suggestion for design and update, need some todo or a new issue for that, a new issue assigned to Taylor is better

packages/design-system/src/components/toast.tsx
SaraVieira: Done!

packages/design-system/src/index.ts
SaraVieira: Also done!

PR #215, opened 2022-08-05
Find a valid parent when inserting into selected instance

rpominov:
- This fixes the remaining todo item in #7
- Also fixed a bug with DnD when `<body>` is a drop target

### step by step interaction description and what is expected to happen

When inserting an instance by clicking on a button in the panel instead of dragging it, the instance should be inserted in the current selected instance.

But if the current selected instance cannot accept a child, for example if it's a Heading, then the new instance should be inserted next to the selected instance.

-------------------------------

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [x] added tests

### DIFF
--- a/apps/designer/app/canvas/shared/instance.ts
+++ b/apps/designer/app/canvas/shared/instance.ts
@@ -4,6 +4,7 @@ import {
   type InstanceProps,
   type Instance,
   type Tree,
+  components,
   allUserPropsContainer,
   getBrowserStyle,
   useAllUserProps,
@@ -18,6 +19,7 @@ import {
   insertInstanceMutable,
   findInstanceById,
   reparentInstanceMutable,
+  getInstancePath,
 } from "~/shared/tree-utils";
 import store from "immerhin";
 import {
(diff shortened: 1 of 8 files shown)


apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -136,6 +136,18 @@ export const useDragAndDrop = () => {
         },
       });
     },
+
+    getValidChildren: (parent) => {
+      if (!(parent instanceof HTMLBodyElement)) {
```
kof, line 141: why is body children not valid? is it possible to be more generic? e.g. not have assumptions about body, just have a check if an element is a ws instance?

apps/designer/app/canvas/shared/instance.ts
```diff
@@ -48,14 +51,40 @@ export const usePopulateRootInstance = (tree: Tree) => {
   }
 };
 
+const findInsertLocation = (
```
kof, line 54: we could potentially have test for this? or is this tested through something else/

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -81,7 +85,11 @@ export const useDrop = <Data>(props: UseDropProps<Data>): UseDropHandlers => {
       if (fromCache !== undefined) {
         return fromCache;
       }
-      const result = getChildrenRects(parent);
+      const children =
+        latestProps.current.getValidChildren !== undefined
+          ? latestProps.current.getValidChildren(parent)
+          : parent.children;
```
kof, line 91: could simplify to this, right?
```suggestion
      const children = latestProps.current?.getValidChildren(parent) ?? parent.children
```

kof: ideally though getValidChildren is always defined and always returns children

apps/designer/app/canvas/shared/use-drag-drop.ts
rpominov: I'll use `getInstanceIdFromElement`. It would work well after we switch to a unique `data` attribute. And currently should be fine too as the `<script>` and `<link>` elements that are inserted don't have `id`s.

kof (approved):
looking great!

PR #216, opened 2022-08-05
Navigator panel

kof:
## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [x] added tests


kof:
@taylornowotny @SaraVieira I just finished the logic for attached/detached navigator panel, but there is one problem:

Since the setting is only saved on the client and the default is "detached", when you switch to "attached" and reload you will see the menu jumping once the client loads and reads the setting.

Let's brainstorm potential solutions to this:

1. Animate the transition so that it looks like its not a bug (not a fan of this, since this is going to happen on each page load if setting is set to attached)
2. Move the setting to the database instead of the client so that it can be shown correctly upfront - not a huge fan, because this need for this setting depends on the screen size, same user may want to have it on or off depending on the screen they are working with, so it can be different from computer to computer for the same user.

kof:
Got another idea is to wait with showing the workspace until we read the navigation layout setting, to synchronize the rendering, trying this out now

kof:
Found a solution technically I am done, I am still not quite sure the Attached/Detached is going to be understood. What I noticed is that its not clear if Attached means to be Attached to the layout or to the tabs, so it can be understood either way.

Right now attached means attached to the tabs ... do we have better names?

kof:
Ok how about "✔ Show navigator"?

SaraVieira:
Why was the header added to the add but not to the asset manager?

kof:
To avoid merge conflicts, thought you are changing it heavily, please add
it in your pr

SaraVieira:
About the text what about "Pin layout"? Show sounds weird because you can always show it

kof:
Yeah thats an option.

What about "Always show navigator" or "Undock navigator"?

We were trying to get away from pin, because when we pin, we remove icon
from tabs

SaraVieira:
Unblock feels weird that's why I said pin but honestly it's all just semantics we can fix later

SaraVieira (approved):

kof:
@SaraVieira not "Unblock", but "Undock", did you misspelled in the answer or misread my wording?

kof:
I think thinking and communicating of the tabs on the left as of dock of an operating system like on macos could be our wany to go, wdyt @webstudio-is/core?

SaraVieira:
I missred sorry, undock sounds good!

SaraVieira (approved):

apps/designer/app/designer/designer.tsx
```diff
@@ -67,11 +73,18 @@ const useSubscribeSyncStatus = () => {
   useSubscribe<"syncStatus", SyncStatus>("syncStatus", setValue);
 };
 
+const useNavigatorLayout = () => {
+  // We need to render the detached state only once the setting was actually loaded from local storage.
+  // Otherwise we may show the detached state because its the default and then hide it immediately.
```
rpominov, line 78: I think if we'd store settings in cookies instead of local storage this would be solvable as cookies available at the server as well.

Also, maybe we should consider not doing SSR at all for the editor? Usually such highly dynamic UIs don't benefit much from SSR.

kof: I was considering going over cookie, but then you loose the setting on logout, which is not quite what we need here. We assume people will choose the setting for the screen and not change it for a long time. 

I am seriously considering to stop doing majority of SSR for the designer and only render a more static part of it. For now this has been solved though.

rpominov (approved):

apps/designer/app/designer/designer.tsx
rpominov: Each setting could be a separate cookie which we wouldn't delete on logout. But yeah, removing SSR sounds like a better solution

PR #217, opened 2022-08-08
Drag & drop in the navigator tree & design update

rpominov:
Related issues: 
- #131 
- #221

Progress:

- [x] D&D works when inserting between nodes
- [x] D&D works when inserting into a node without expanding
- [x] Customise placement indicator
- [x] Expand on hold
- [x] Horizontal shift
- [x] Autoscroll
- [x] Cancel on esc
- [x] Design update
- [x] Accessibility
- [x] Testing / bugfixes / cleanup

---------------------------------------

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [x] added tests


kof:
You probably already know but just in case 

Screenshot 2022-08-13 at 17 16 29

kof:
When dragging into a collapsed item, I think we should open it much faster than right now, too much delay makes it feel slow

kof:
- [ ] cursor during drag could be "grabbing"

rpominov:
@kof I'm still working on tests, but otherwise this is ready for review!

apps/designer/app/canvas/features/wrapper-component/wrapper-component.tsx
```diff
@@ -34,16 +33,22 @@ export const WrapperComponentDev = ({
   const [, setSelectedElement] = useSelectedElement();
   const isEditing = editingInstanceId === instance.id;
   const [editableRefCallback, editableProps] = useContentEditable(isEditing);
-  const focusRefCallback = useEnsureFocus();
 
   const refCallback = useCallback(
     (element) => {
-      if (isEditing) editableRefCallback(element);
-      focusRefCallback(element);
+      if (isEditing) {
```
kof, line 39: this is the wrong place to do it, wrapper-component is decoupled from editable using useContentEditable hook. if the only thing that needs focus is content editable, then focusing should be done inside editableRefCallback

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -167,12 +167,29 @@ export const useDragAndDrop = () => {
         return false;
       }
 
+      // When trying to drag an inline instance, drag its parent instead
+      if (components[instance.component].isInlineOnly) {
```
kof, line 171: I am realizing just now that isInlineOnly is a potentially confusing name, how would you call it?

kof, line 171: Do you think its worth extracting this into a function and add a unit test?

apps/designer/app/designer/shared/tree/tree.tsx
```diff
@@ -1,132 +1,670 @@
...
+
+type ShiftedDropTarget = {
+  instance: Instance;
+  position: number | "end";
+  placement:
+    | { type: "rect"; rect: Rect }
+    | { type: "line"; placement: Placement };
+};
+
+export const useHorizontalShift = ({
```
kof, line 503: this export isn't used outisde of tree, but is exported, I think you want to move this hook into a separate file and make sure that you only export the things that are actually public. Look at index.ts files like they represent the public interface, please check all index files in this PR

kof:
- [ ] lets move the shared/tree into design system
- [ ] tree-item.tsx can be now removed, anything else?

apps/designer/app/shared/dom-hooks/use-measure.ts
```diff
@@ -28,6 +28,18 @@ export const useMeasure = <
     onScrollEnd: handleChange,
   });
 
+  // Detect movement of the element without remounting.
+  useEffect(() => {
```
kof, line 32: I think this will not disconnect the observer if element ref changes, that's why for resize observer, observer is memorized separately, so we create only one observer and then separately call .observe(

packages/design-system/src/components/primitives/dnd/use-drag.ts
```diff
@@ -75,67 +73,64 @@ export const useDrag = <DragItemData>({
...
+    onMoveEnd({ isCanceled }) {
+      // A drag is basically a very slow click.
+      // But we don't want it to register as a click.
+      if (isCanceled === false && state.current.status === "dragging") {
+        const handleClick = (event: MouseEvent) => {
+          event.preventDefault();
+          event.stopPropagation();
+        };
+        window.addEventListener("click", handleClick, true);
+        setTimeout(() => {
```
kof, line 119: why not removing the listener inside the handleClick instead? there must be a more deterministic way of handling it

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -248,21 +282,41 @@ export const useDrop = <Data>(props: UseDropProps<Data>): UseDropHandlers => {
 
     return {
       handleMove(pointerCoordinates) {
-        state.current.pointerCoordinates = pointerCoordinates;
+        if (latestProps.current.emulatePointerAlwaysInRootBounds === true) {
+          const rect = (rootRef.current as Element).getBoundingClientRect();
+          const padding = 2;
```
kof, line 287: where is this number coming from? does it depend on CSS?

packages/design-system/src/components/primitives/dnd/use-move.ts
```diff
@@ -1,20 +1,5 @@
-/* eslint-disable prefer-const */
-/* eslint-disable func-style */
-/*
- * Copyright 2020 Adobe. All rights reserved.
- * This file is licensed to you under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License. You may obtain a copy
- * of the License at http://www.apache.org/licenses/LICENSE-2.0
```
kof, line 7: we can't remove the copyright like that! this is a serious violation of the license

kof: I will ask around, we may need to also add a second license file to the root

kof (changes requested):
first round is done, overall great work! 
will do the testing during the second round

packages/design-system/src/components/primitives/dnd/use-move.ts
kof: we also need to: `You must cause any modified files to carry prominent notices
          stating that You changed the files;`
          
https://github.com/adobe/react-spectrum/blob/main/LICENSE#L97

yeah and we need to put that LICENSE file in the root too

apps/designer/app/canvas/shared/use-drag-drop.ts
rpominov: Depends on what you find confusing. I think there're two things. 

1. The word "only" suggest there can be components that can be used both for formatting and as standalone components.

2. "Inline" suggest this is about block / inline distinction, but this is more about "used for formatting" / "standalone" distinction.

Maybe `isFormatting`?

packages/design-system/src/components/primitives/dnd/use-drag.ts
rpominov: I did that at first, but I'm not 100% sure a click event will always occur. If it doesn't, a random future click will be canceled. I think `setTimeout` is more robust in this case even though a bit hacky.

packages/design-system/src/components/primitives/dnd/use-drop.ts
rpominov: It's defined in a comment above:

```
// The pointer will be always at least 2px away from any edge of the root.
emulatePointerAlwaysInRootBounds?: boolean;
```

Needed simply to account for literal "edge"-cases.

Maybe I should add another threshold option for this.

apps/designer/app/canvas/shared/use-drag-drop.ts
kof: maybe we should be just more explicit, e.g. since those are used to identify draggability, we can say: isDraggable false instead of isInline:true

kof: I was trying to describe what that component IS, but its hard, because this is more about how its being used

rpominov: isReparentable?

packages/design-system/src/components/primitives/dnd/use-move.ts
kof: I was linked on twitter to how properly attribute  https://infra.apache.org/licensing-howto.html

rpominov: Makes me want to refactor use-drag to use browser events directly 😄

kof: yeah its not too far off, there isn't much logic

packages/design-system/src/components/primitives/dnd/use-drag.ts
kof: yeah, though its very indeterministic with javascript when this setTimeout is going to run, what if js runtime decides to run it in 1ms after you called it and the click actually occurs after that for some reason?

kof: it can also cause issues with e2e testing in the future

kof: I think the actual problem here is that the event that handles pointer/mouse up doesn't preventDefault where it should, so you are now upstream trying to workaround the situation

rpominov: I thought that too. But I tried adding it to our `pointerup` and `pointerdown` handlers, and it didn't work

rpominov: The case when `click` might not happen that I was thinking of is when you release the button outside of browser window / iframe. Need to check what happens in that case.

If it still generates a click, then it's probably ok to unsubscribe in the event handler as I'm not aware of other cases.

If it doesn't, then we probably need to both unsubscribe in the handler and after something like 300ms as a fallback.

rpominov: Actually there's a better option: unsubscribe in the handler, prevent default only if less than 300ms have passed since we subscribed.

kof: timeout based logic is flawed in any case, since we know when drag interaction started, we can bind there an event like mouseup and prevent a click there

kof: jfyi click is mousedown+mouseup, so you should be able to prevent click at both mousedown and mouseup, haven't tested mouseup, just assuming it works based on the mental model I have

packages/design-system/src/components/primitives/dnd/use-drop.ts
kof: oh lets make it a constant with a comment, I think I missed it because I was looking at a diff

packages/design-system/src/components/primitives/dnd/use-drag.ts
rpominov: Not sure it's because of pointer event, here they do similar thing with `mouseup`: https://stackoverflow.com/a/20290312/478603

But I will try

taylornowotny:
Looking good! There are a few style differences from the design though:

- Please use blue for the position indicator. I know Oleg instructed to use magenta, but we are going with blue at least for now.
- The position indicator in the design has a white outline, and the circle in the middle is filled with white, not transparent.
Screen Shot 2022-08-19 at 2 10 42 PM
- When nesting, the parent should have an outline as shown in the design.
Screen Shot 2022-08-19 at 2 15 40 PM
- When hovering the chevron, the item hover effect goes away

Missing features (if these are coming later then ignore):
- drag right side edge to re-size panel
- double-click to re-name items
- hide UI

rpominov:
Thanks, will fix

> - When hovering the chevron, the item hover effect goes away

This should be already the case unless I misunderstood

> double-click to re-name items

We have an issue for this https://github.com/webstudio-is/webstudio-designer/issues/222

> drag right side edge to re-size panel
> show/hide items

- [ ] Will create issues

rpominov:
> When hovering the chevron, the item hover effect goes away

Ah, you mean this is a bug.

I did this intentionally because I wanted _the area that has the hover effect_ to exactly match _the area where click makes item selected_. Hovering to the left of the chevron also doesn't have a hover effect because of that.

Should I change this and in what way:

- Should clicking on the chevron make item selected in addition to expanding/collapsing?
  - If no, should there still be the hover effect?
- Should clicking to the left of the chevron (where the nesting lines are) make item selected?
  - If no, should there still be the hover effect?

taylornowotny:
> > When hovering the chevron, the item hover effect goes away
> 
> Ah, you mean this is a bug.
> 
> I did this intentionally because I wanted _the area that has the hover effect_ to exactly match _the area where click makes item selected_. Hovering to the left of the chevron also doesn't have a hover effect because of that.
> 
> Should I change this and in what way:
> 
>     * Should clicking on the chevron make item selected in addition to expanding/collapsing?
>       
>       * If no, should there still be the hover effect?
> 
>     * Should clicking to the left of the chevron (where the nesting lines are) make item selected?
>       
>       * If no, should there still be the hover effect?

Very good questions!

Yes the hover effect should be present any time the cursor hovers any part of the item. Including the chevron, the nesting lines, and the show/hide button. 

Clicking anywhere on the item **except** for the chevron and the show/hide button should select that item.

I see what you were thinking, but the item hover effect is not necessarily meant to communicate that clicking will select the item. The "selected" state of an item doesn't really affect it's function in the Navigator panel – it mostly affects the canvas and the style/properties panel. A user should be able to use pretty much all functions of the Navigator without changing which item is selected. 
So having the hover effect show even when the chevron and the show/hide button is hovered is preferable because it says "you're affecting this item right now".

I hope I've explained that well enough

(btw I know there is no "focus" state for the selected item. I'm working on a better focus indicator that will be visible in all situations)

apps/designer/app/canvas/shared/use-drag-drop.ts
rpominov: Something drag&drop related like `isReparentable` will feel wrong here: https://github.com/webstudio-is/webstudio-designer/blob/94a827b9f460c18f50f00073552084d21e4f8000/apps/designer/app/canvas/shared/use-track-selected-element.ts#L102-L104

packages/design-system/src/components/primitives/dnd/use-drag.ts
rpominov: `setTimeout` removed.

Preventing default in `mouseup` doesn't prevent the click.
Click does happen even when releasing outside of browser window, but I still added a check for 300ms just in case.

rpominov:
@kof @taylornowotny this is ready for review

apps/designer/app/shared/tree-utils/get-instance-path.ts
```diff
@@ -38,14 +38,14 @@ export const getInstancePathWithPositions = (
       if (typeof child === "string") continue;
       const found = find(child);
       if (found) {
-        path.push({ instance: child, position: i });
+        path.push({ item: child, position: i });
```
kof, line 41: please don't use single char variables like `i`

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -0,0 +1,306 @@
...
+              />
+            ))
+          }
+        </CollapsibleContent>
+      )}
+    </Collapsible.Root>
+  );
+};
+
+const Label = styled(__DEPRECATED__Text, {
```
kof, line 273: lets use the new Text component right away?

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -0,0 +1,306 @@
...
+  whiteSpace: "nowrap",
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+  lineHeight: 1.4,
```
kof, line 277: why do we need line height here? isn't it a single line text and should be 1? I assume you are using it for padding and for that we should use padding, line-height is for distance within multiline text

packages/design-system/src/stitches.config.ts
```diff
@@ -257,6 +258,10 @@ export const {
       4: "400",
       max: "999",
     },
+    borderWidths: {
```
kof, line 261: it probably doesn't make sense to have borderWidths a part of the design system if they are not simplifying anything and basically translate 1:1, please check with @taylornowotny

packages/jest-config/index.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   testEnvironment: "node",
   testMatch: ["<rootDir>/src/**/*.test.ts"],
   transform: {
-    "^.+\\.ts?$": "esbuild-jest",
+    "^.+\\.tsx?$": "esbuild-jest",
```
kof, line 5: it looks like it will now only match .tsx files but not .ts, but we may have .ts too, no?

packages/react-sdk/src/components/index.test.ts
```diff
@@ -0,0 +1,9 @@
+import * as components from "./index";
+import { WsComponentMetaSchema } from "./component-type";
+
+test.each(Object.entries(components))("validating: %s", (name, component) => {
```
kof, line 4:
```suggestion
test.each(Object.entries(components))("validating meta definition of %s", (name, component) => {
```

kof (approved):
almost there, a few minor things

packages/jest-config/index.js
rpominov: It matches both. The `?` means the `x` can be present or not

packages/design-system/src/components/tree/tree-node.tsx
kof: hmm, it shouldn't, lets check

packages/design-system/src/components/tree/tree-node.tsx
rpominov: I use `variant={isSelected ? "loContrast" : "contrast"}` from the old Text, and the new one doesn't support this.

I'd kept `__DEPRECATED__Text` here for now and replace it with `Text` when it will be replaced everywhere in the codebase.

packages/design-system/src/stitches.config.ts
rpominov: Yeah, makes sense. I don't think we'll ever want to change all borders from `1px` to `2px`. Will revert.

packages/design-system/src/components/tree/tree-node.tsx
rpominov: Added a code comment

packages/design-system/src/components/tree/tree-node.tsx
kof: lets see what we need to add to support it, its easy,  we should not use anything that has deprecated in it as a rule

kof: I think this would work, adding to variants: 

```ts
```

taylornowotny (approved):
Looks and works great! Excellent work

PR #229, opened 2022-08-17
fix project connectiuon

SaraVieira:
## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
+++ b/apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
@@ -19,7 +19,7 @@ const none = { TabContent: () => null };
 
 type SidebarLeftProps = {
   publish: Publish;
-  assets: Array<Asset>;
+  assets?: Array<Asset>;
 };
 
 export const SidebarLeft = ({ publish, assets }: SidebarLeftProps) => {
@@ -78,7 +78,7 @@ export const SidebarLeft = ({ publish, assets }: SidebarLeftProps) => {
           }}
         >
           <TabContent
-            assets={assets}
+            assets={assets || []}
             publish={publish}
             onSetActiveTab={setActiveTab}
           />
--- a/apps/designer/app/shared/db/project.server.ts
+++ b/apps/designer/app/shared/db/project.server.ts
@@ -19,14 +19,13 @@ export type Project = Omit<BaseProject, "prodTreeIdHistory"> & {
 const parseProject = (project: BaseProject): Project => {
   const prodTreeIdHistory = JSON.parse(project.prodTreeIdHistory);
   TreeHistorySchema.parse(prodTreeIdHistory);
-
   if (project?.assets) {
     return {
+      ...project,
       assets: project.assets.map((asset) => ({
         ...asset,
         path: getAssetPath(asset),
       })),
-      ...project,
       prodTreeIdHistory,
     };
   }
@@ -44,18 +43,29 @@ export const loadById = async (projectId?: Project["id"]) => {
 
   const project = await prisma.project.findUnique({
     where: { id: projectId },
-    include: { assets: true },
+    include: {
+      assets: {
+        orderBy: {
+          createdAt: "desc",
+        },
+      },
+    },
   });
 
   if (!project) return null;
-
   return parseProject(project);
 };
 
 export const loadByDomain = async (domain: string): Promise<Project | null> => {
   const project = await prisma.project.findUnique({
     where: { domain },
-    include: { assets: true },
+    include: {
+      assets: {
+        orderBy: {
+          createdAt: "desc",
+        },
+      },
+    },
   });
   if (!project) return null;
 
@@ -138,7 +148,13 @@ export const clone = async (clonableDomain: string, userId: string) => {
         domain,
         devTreeId: tree.id,
       },
-      include: { assets: true },
+      include: {
+        assets: {
+          orderBy: {
+            createdAt: "desc",
+          },
+        },
+      },
     }),
     db.props.clone({
       previousTreeId: clonableProject.prodTreeId,
@@ -183,7 +199,13 @@ export const update = async ({
         prodTreeIdHistory: JSON.stringify(data.prodTreeIdHistory),
       },
       where: { id },
-      include: { assets: true },
+      include: {
+        assets: {
+          orderBy: {
+            createdAt: "desc",
+          },
+        },
+      },
     });
     return project;
   } catch (error) {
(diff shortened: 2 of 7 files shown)


packages/prisma-client/src/types.ts
```diff
@@ -1,9 +1,10 @@
...
+import type { Asset as DbAsset, Project as BaseProject } from "@prisma/client";
 export { Location } from "@prisma/client";
-export type { InstanceProps, Project, User, Breakpoints } from "@prisma/client";
+export type { InstanceProps, User, Breakpoints } from "@prisma/client";
 
 export type Asset = DbAsset & {
   path: string;
   status?: "uploading" | "uploaded";
 };
+export type Project = BaseProject & { Asset: DbAsset[] };
```
kof, line 9: that type is probably automatically provided by prisma, no? also its not Asset, its assets now

SaraVieira: this has been fixed

apps/designer/app/shared/db/project.server.ts
```diff
@@ -6,41 +6,55 @@ import {
...
+  const newAssets =
+    "assets" in project
+      ? project.assets.map((asset) => ({
+          ...asset,
+          path: getAssetPath(asset),
+        }))
+      : {};
   return {
     ...project,
+    ...newAssets,
```
Andarist, line 35: shouldnt this be something like this?
```suggestion
    assets: newAssets,
```

Andarist: Or rather actually - you probably have meant to create `{ assets: projects.assets.map(() => {...}) }` a couple of lines ago

SaraVieira: They are both the same idea, just committed your first one :)

apps/designer/app/shared/db/project.server.ts
```diff
@@ -6,41 +6,55 @@ import {
...
 const TreeHistorySchema = z.array(z.string());
 
 export type Project = Omit<BaseProject, "prodTreeIdHistory"> & {
   prodTreeIdHistory: z.infer<typeof TreeHistorySchema>;
 };
 
-const parseProject = (project: BaseProject | null): Project | null => {
+const parseProject = (
+  project: BaseProject | null
+): Project | Omit<BaseProject, "assets"> | null => {
```
Andarist, line 22: This return type looks a little bit fishy - I would expect a unified type to come up from this parsing function (so probably smth like `Project | null`).

packages/prisma-client/src/types.ts
```diff
@@ -1,9 +1,10 @@
-import type { Asset as DbAsset } from "@prisma/client";
+import type { Asset as DbAsset, Project as BaseProject } from "@prisma/client";
```
Andarist, line 1: It's a little bit confusing to me that things like `BaseProject` are pointing to different types in different files. It makes it harder to figure out the relationships between those things.

apps/designer/app/shared/db/project.server.ts
Andarist: In here there is a subtle difference and that's also why I've incidentally suggested that first patch at first 😅 It's on the verge of a typo. The problem is that with this patch your `newAssets` are still either an array or an object, the other suggested patch ensures that we spread an object with an optional `assets` property.

packages/prisma-client/src/types.ts
kof: half-migrated state, we have extracted prisma client into a separate package and now need to move all the augmented types to where they are are used, to the vertical/feature code

apps/designer/app/shared/db/project.server.ts
```diff
@@ -8,39 +8,66 @@ import {
   Project as BaseProject,
 } from "@webstudio-is/prisma-client";
 import * as db from ".";
+import { getAssetPath } from "@webstudio-is/asset-uploader/src/helpers/get-asset-path";
```
kof, line 11: this was probably auto-added

packages/prisma-client/src/types.ts
```diff
@@ -1,9 +1,10 @@
...
 export { Location } from "@prisma/client";
-export type { InstanceProps, Project, User, Breakpoints } from "@prisma/client";
+export type { InstanceProps, User, Breakpoints } from "@prisma/client";
 
 export type Asset = DbAsset & {
-  path: string;
+  path?: string;
   status?: "uploading" | "uploaded";
 };
+export type Project = BaseProject & { assets?: Asset[] };
```
kof, line 9: Isn't prisma supposed to know that assets are part of the Project type and give it to us automatically?

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
```diff
@@ -78,7 +78,7 @@ export const SidebarLeft = ({ publish, assets }: SidebarLeftProps) => {
           }}
         >
           <TabContent
-            assets={assets}
+            assets={assets || []}
```
kof, line 81: should probably be a default inside tab content

kof:
I was just thinknig: do we actually need assets loaded upfront when designer loads? Images on canvas have a URL already built-in, no? 

If we only need images when user opens images tab, should we not fetch them then? I just imagined if a project has hundreds of images, we would fetch a substantial payload even if user doesn't go into images tab.

We could even think of using the remix's router for fetching each tab's data and scripts, so we don't load even the logic for handling contents of each tab. I think long-term there will be quite a lot of logic in there.

rpominov:
I see this locally:

Screenshot 2022-08-19 at 11 24 37

Probably because we changed `schema.prisma`, but didn't create a migration file. I'll make a PR to fix this.

kof:
@SaraVieira said there is no need for migration because changes were on models only, how did this even work on dev?

kof:
Just checked, preview has failed as well

rpominov:
Sorry, no migration is needed actually. I just needed to run `yarn build:prisma` locally.

rpominov:
> Just checked, preview has failed as well

Hm is `yarn build:prisma` run during deployment of a preview?

kof:
in vercel config we run `yarn prisma migrate deploy && yarn build`

kof:
yarn build:prisma inside prisma-client package calls prisma generate and we don't do this on CI, the question is why and how did we survive this before?  Was this broken as we moved prisma client to a package?

kof:
@SaraVieira @andrasbacsai what was the reason we called it  build:prisma instead of "build" inside the prisma-client package? Otherwise it would be running automatically on ci and when we pull we have yo run one simple command from the root.

I remember there was something about it ....

andrasbacsai:
I do not remember. I try to fix it.

PR #230, opened 2022-08-17
Redesign asset info

SaraVieira:
## Before requesting a review

- [x] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

image
image


apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,108 @@
...
+              {width} x {height}
+            </Text>
+          </Flex>{" "}
+          <Flex align="center" css={{ gap: "$1" }}>
+            <AspectRatioIcon />
+            <Text size="1">
+              {getAspectRatio(
+                width as unknown as number,
+                height as unknown as number
+              )}
```
SaraVieira, line 93: This is because of the Decimal prisma type

kof: we should be converting it at  data fetching layer, where we augment prisma schema anyways

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,108 @@
...
+  setIsDeleting: (isDeleting: boolean) => void;
+};
+
+export const AssetInfo = ({
+  size,
+  width,
+  height,
+  id,
+  name,
+  closeInfo,
```
kof, line 36: callbacks should always have an onSomething, in this case onClose?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,108 @@
...
+  setIsDeleting,
```
kof, line 37: onDelete

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <SizeIcon />
+            <Text size="1">
+              {width} x {height}
+            </Text>
+          </Flex>{" "}
+          <Flex align="center" css={{ gap: "$1" }}>
+            <AspectRatioIcon />
+            <Text size="1">
+              {getAspectRatio(
+                width as unknown as number,
```
kof, line 91: if we don't adjust the type early, we will have to do this in all consumers

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,108 @@
...
+        css={{ height: 40, paddingLeft: "$3" }}
+        align="center"
+        justify="between"
+      >
+        <Text size="1" weight="700">
+          Asset Details
+        </Text>
+
+        <IconButton
+          onClick={closeInfo}
```
kof, line 61: standard convention is handleSomething, e.g. handleClose in this case, but since this one is just calling the future onClose we would be just passing onClose if not one problem: onClose has semantically no knowledge of the click event, so it should not receive the event object, otherwise its a leaking abstraction, so the best thing in such situations is to have this
```suggestion
          onClick={() => onClose()}
```

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,108 @@
...
+  size,
+  width,
+  height,
+  id,
+  name,
+  closeInfo,
+  setIsDeleting,
+}: AssetInfoProps) => {
+  const submit = useSubmit();
+  const deleteAsset = () => {
```
kof, line 40:
```suggestion
  const handleDeleteAsset = () => {
```

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
```diff
@@ -0,0 +1,22 @@
+export const getHumanReadableFileSize = (size: number) => {
```
kof, line 1: I bet there are battle tested libs for this, is there a small and a really good one?

kof: or maybe I am overestimating the complexity there?

kof: e.g. this one https://github.com/sindresorhus/pretty-bytes/

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  const i = Math.floor(Math.log(size) / Math.log(1024));
+  const types = ["B", "kB", "MB", "GB", "TB"];
+  return `${(size / Math.pow(1024, i)).toFixed(2)} ${types[i]}`;
+};
+
+export const getStartAndEndOfString = (str: string) => {
+  if (str.length > 20) {
```
kof, line 8: 20 is arbitrary number, this won't work depending on the size of the container or font. Usually this is done using CSS, am I right to assume you wanted the ... in the middle of the string? There is a better solution to this that doesn't use arbitrary string length or otherwise detects it

kof: One thing I can think of is to create a component that would split a string in two parts, and make the left part truncate using CSS ... this is just what comes into mind right away, maybe there is an even better way

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    return str.substr(0, 10) + "..." + str.split(".").pop();
+  }
+  return str;
+};
+
+export const gcd = (a: number, b: number): number => {
+  return b ? gcd(b, a % b) : a;
+};
+
+export const getAspectRatio = (width: number, height: number) => {
```
kof, line 18: naming idea: getFormattedAspectRatio, because this is about rendering a representation, not about getting an actual aspect ratio

packages/design-system/src/components/button.tsx
```diff
@@ -129,12 +130,13 @@ export const Button = styled("button", {
...
-            boxShadow: "inset 0 0 0 1px $colors$slate8",
+            background: "$red11",
+            color: "$loContrast",
           },
         },
         "&:active": {
-          backgroundColor: "$red3",
-          boxShadow: "inset 0 0 0 1px $colors$red8",
+          backgroundColor: "$red11",
+          color: "$loContrast",
```
kof, line 139: is this aligned with the button spec from figma for all buttons?

packages/design-system/src/components/text.tsx
```diff
@@ -47,6 +47,20 @@ export const Text = styled("div", {
         textIndent: "-.025em",
       },
     },
+    weight: {
```
kof, line 50: would it make sense to have semantic names?

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping

I would even define those in stitches config and use them everywhere, we need @taylornowotny to define all font weights we are going to need with the corresponding semantic names, maybe he already did that?

packages/icons/src/aspect-ratio.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+import { IconProps } from "./types";
+
+export const AspectRatioIcon = React.forwardRef<SVGSVGElement, IconProps>(
```
kof, line 4: where is it coming from? @taylornowotny said he designed all icons in 16 size and wanted to redo all radix icons because they are 15, so I am wondering why is this one 15.
maybe Taylor just used the radix icon, but in that case you can just import it from the package, why adding it?

packages/icons/src/cloud.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+import { IconProps } from "./types";
+
+export const CloudIcon = React.forwardRef<SVGSVGElement, IconProps>(
```
kof, line 4: this one looks like made by @taylornowotny since its 16, but lets check if its not one of https://icons.radix-ui.com/ 

maybe we need to have a way to communicate which icon is from where for now

packages/design-system/src/components/text.tsx
taylornowotny: in the design system ✅

packages/icons/src/aspect-ratio.tsx
taylornowotny: Correct, all icons are 16px and part of a standardized component set. See the Icons page of the design system.

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
kof: I think aiming to truncate before an extension specifically isn't great because if the file doesn't have an extension for some reason, you end up truncating at the end ... macos seems to be truncating in the middle

kof: quick googling gave an interesting css truncation appraoch https://www.codeply.com/go/UBmrMAiRsG/css-truncate-ellipsis-in-middle-of-text

kof: https://stackoverflow.com/questions/27746218/css-crop-string-in-the-middle

packages/icons/src/aspect-ratio.tsx
kof: hmm, not sure where @SaraVieira got the 15px icon from then, maybe our figma is inconsistent somewhere

kof (changes requested):
first round done

packages/design-system/src/components/text.tsx
kof: we just discussed with taylor, the plan is to use semantic variants for this, so we need to actually rewrite this  entire component to use semantic definitions from the Typography section in DS

we won't need sizes,  or weights

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const getStartAndEndOfString = (str: string) => {
+  if (str.length > 20) {
+    return str.substr(0, 10) + "..." + str.split(".").pop();
+  }
+  return str;
+};
+
+export const gcd = (a: number, b: number): number => {
```
kof, line 14: used onlly in this file
```suggestion 
const gcd = (a: number, b: number): number => {
```

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: done!

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: done

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: done!

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
SaraVieira: fixed

packages/icons/src/aspect-ratio.tsx
SaraVieira: I got it from figma, copied from there, it was 15px idk why

Fixed

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
SaraVieira: done!

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
SaraVieira: changed to it, its en esm so I added it to the dependencies to bundle

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/utils.ts
SaraVieira: done!

packages/icons/src/aspect-ratio.tsx
kof: it may be wrong icon there, like the path could be wrong

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,130 @@
...
+const truncatedText = {
+  position: "relative",
+  maxWidth: 80,
+
+  div: {
+    whiteSpace: "nowrap",
+    textOverflow: "ellipsis",
+    overflow: "hidden",
+
+    "&:after": {
```
kof, line 31: single colon is for pseudos
```suggestion
    "&::after": {
```

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,130 @@
...
+        >
+          <Cross1Icon />
+        </IconButton>
+      </Flex>
+      <Separator />
+      <Box css={{ p: "$2 $3" }}>
+        <Grid columns={2} align="center">
+          <Box css={{ width: 100 }}>
+            <Box css={truncatedText}>
+              <Text size="1" data-extension={name.split(".").pop()}>
```
kof, line 90: could this go sideways if I have the same chats as the extension inside the name anywhere?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,130 @@
...
+  Cross1Icon,
+  SizeIcon,
+  TrashIcon,
+} from "@webstudio-is/icons";
+import { useSubmit } from "@remix-run/react";
+import prettyBytes from "pretty-bytes";
+
+const truncatedText = {
+  position: "relative",
+  maxWidth: 80,
```
kof, line 24: there is no way to avoid hardcoding this? its better than chars but still is gonna break if we change layout, right?

kof: alternative may be to use the fixed box width we already have and calculate from that, so at least if it changes there, it changes here as well, unless we can use layout somehow to enforce this

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: This splits by . and gets the last one, the last one will always be the extension doesn't matter how many `.` you have

kof: I meant if the extension is inside the word but maybe I misunderstood, then ignore me

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  return (
+    <>
+      <Flex
+        css={{ height: 40, paddingLeft: "$3" }}
+        align="center"
+        justify="between"
+      >
+        <Text size="1" weight="700">
```
kof, line 72: is it possible that we could reuse the Header that we currently have in sidebar-left/lib/header? I haven't checked if there are any differences in styling

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: The tooltip already has a width of 240

packages/icons/src/aspect-ratio.tsx
SaraVieira: I scaled and copied

packages/design-system/src/components/text.tsx
SaraVieira: Should I change this to semantic names?

kof: lets do this in a separate PR, because we need to migrate to an entirely new Text component all usages of it. By semantic I mean variants defined in figma: regular, label, title etc

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
kof: you saying we can't be not hardcoding the max width?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
kof: nope?

packages/design-system/src/components/text.tsx
kof: there isn't much point though adding the weight variant to Text right now if we are replacing it entirely

SaraVieira: oki, gonna resolve this then

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: I don't know, from what I tried no, that was my first attempt

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: Extension is inside the word? What?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: my bad, done now!

packages/design-system/src/components/button.tsx
SaraVieira: For the red one yes

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: It does work, changed to it

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: I was wrong, I tried again and it worked, we now use 70%

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
kof: logo-as-png.png

SaraVieira: Nope, that's not an issue because it will split at the . And still get the last png

kof:
@taylornowotny  what's the correct light/dark mode state?  Here is how it looks like for the delete button.
 

Screenshot 2022-08-21 at 19 02 57
Screenshot 2022-08-21 at 19 03 03
Screenshot 2022-08-21 at 19 03 26
Screenshot 2022-08-21 at 19 03 34

kof:
@SaraVieira upload button is out of scope of this PR?

SaraVieira:
Ill make a PR right now just for the upload button

This has too many changes already

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
```diff
@@ -0,0 +1,121 @@
...
+    .split("")
+    .pop();
+  return (
+    <Box
+      css={{
+        position: "relative",
+        maxWidth: "70%",
+      }}
+    >
+      <__DEPRECATED__Text
```
kof, line 33: can we please switch to the new Text component? I just made it so you don't have to in this  PR :)

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+};
+
+export const AssetThumbnail = (asset: Asset | UploadingAsset) => {
+  const { path, alt, status, name } = asset;
+  const [isDeleting, setIsDeleting] = useState(false);
+  const isUploading = status === "uploading";
+  const isUploadedAsset = isUploading === false && "size" in asset;
+
+  const lol = {
```
kof, line 61: lol? :)

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
```diff
@@ -0,0 +1,82 @@
...
+    aspectRatio: "1/1",
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+    padding: "0 $2",
+    position: "relative",
+
+    "&:hover button": {
```
kof, line 69: hmm, this should be somewhere else I think, because this is breaching encapsulation

kof: Imagine you change the button to something else, how would you know to update this line?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+        variant="raw"
+        title="Options"
+        onClick={() => setTooltipOpen(true)}
+        css={{
+          opacity: 0,
+          position: "absolute",
+          top: "$1",
+          right: "$1",
+          cursor: "pointer",
+          color: "$highContrast",
```
kof, line 42: this can't be right

packages/react-sdk/src/components/index.test.ts
```diff
@@ -1,9 +1,12 @@
 import * as components from "./index";
 import { WsComponentMetaSchema } from "./component-type";
 
-test.each(Object.entries(components))("validating meta definition of %s", (name, component) => {
-  if (name === "default") {
-    return;
+test.each(Object.entries(components))(
```
kof, line 4: What happened here, either your editor or @rpominov's isn't formatting using the same prettier settings

kof:
For some reason preview env doesn't work, same is on main right now, do you know what happened? It is clearly again db

kof:
OMG is it possible that we use the same db for all preview environments and the current [work in progress](https://github.com/webstudio-is/webstudio-designer/pull/231) has just migrated the db to the new state, while the rest of the PR's assume the old db state?

packages/react-sdk/src/components/index.test.ts
rpominov: I think we don't have a prettier config at all :)
So everyone runs with their own settings probably.

rpominov: Ah, this might actually be because I've accepted a suggested change in the PR comment: https://github.com/webstudio-is/webstudio-designer/commit/31e710efca594527b5d644da70b95ef4dcc11dae

kof: that makes sense

kof:
Yep we lost preview for now, me and @andrasbacsai will figure something out, flightcontrol is now the most viable option

kof:
Screenshot 2022-08-26 at 22 04 35

I opened locally, is this what I supposed to see? if yes, then the button is incorrect

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
SaraVieira: LOLLLLL

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
SaraVieira: Honestly I agree with you, do you have a better option?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
SaraVieira: Its not, ill fix that in another PR but this one is too big

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info.tsx
SaraVieira: All done!

SaraVieira:
Why is the button incorrect?

kof:
> Why is the button incorrect?

design in figma, where did you look? different border, probably different hover bg


Screenshot 2022-08-28 at 10 00 34

SaraVieira:
Thinsg to in new PR:

- [ ] Move header to components shared
- [ ] Add alt textarea
- [ ] update to be floating panel
- [ ] Upload button needs redesign

packages/prisma-client/src/types.ts
```diff
@@ -2,9 +2,11 @@ import type { Asset as DbAsset, Project as BaseProject } from "@prisma/client";
 export { Location } from "@prisma/client";
 export type { InstanceProps, User, Breakpoints } from "@prisma/client";
 
-export type Asset = DbAsset & {
+export type Asset = Omit<DbAsset, "width" | "height"> & {
```
kof, line 5: We discussed to move all augmentation to entity-specific folders and live prisma client as-is

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
kof: huh, you just need to use $hiContrast, not $high, its a typo

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const [isTooltipOpen, setTooltipOpen] = useState(false);
+  const closeTooltip = () => setTooltipOpen(false);
+
+  return (
+    <Tooltip
+      open={isTooltipOpen}
+      multiline
+      onEscapeKeyDown={closeTooltip}
+      onPointerDownOutside={closeTooltip}
+      css={{ width: 240, maxWidth: 240, padding: 0, paddingBottom: "$2" }}
```
kof, line 23: I think you wanna use PANEL_WIDTH constant

apps/designer/app/routes/designer/$id.tsx
```diff
@@ -1,9 +1,9 @@
 import { useActionData, useLoaderData } from "@remix-run/react";
 import { ActionFunction, LoaderFunction } from "@remix-run/node";
-import type { Project, Asset } from "@webstudio-is/prisma-client";
 import { toast } from "@webstudio-is/design-system";
 import { Designer, links } from "~/designer";
 import * as db from "~/shared/db";
+import type { Project } from "@webstudio-is/prisma-client";
```
kof, line 6: this will be a problem, because we rely on the augmented type, which is db.projects.Project

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
SaraVieira: Done!

packages/prisma-client/src/types.ts
SaraVieira: Moved this to another place in the uploader pckage

apps/designer/app/routes/designer/$id.tsx
SaraVieira: I don't understand this

kof: you are being fooled by the generic, use the type that you return from loadById

kof: also use it here https://github.com/webstudio-is/webstudio-designer/blob/3400b2d33e7df374ce57ac1357023d41e5ff0fd2/apps/designer/app/routes/designer/%24id.tsx#L30

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
kof: ping

SaraVieira: oh sorru, fixec

Issue #54, opened 2022-05-29
Multi-page sites and routing


kof:
As a user I want to be able to create multiple pages and leverage the full power of Remix's router, which doesn't need to do the full page reload between pages once loaded and makes it super fast

## Todo

- [x] Create prisma model (discussed: {id, projected, data: JSON})
- [x] Create a migration for the production database (using [snaplet](https://twitter.com/appfactory/status/1559505547526897667)?)
- [x] Create rest API
- [x] Create data zod schema
- [x] Create  db helpers
- [x] Render the panel in the left sidebar
- [ ] Render the pages tree using tree component we use in navigator
- [x] Page item has 2 main functions: open the page and open the page settings
- [ ] Page item can have a parent
- [ ] Page settings should have a few main key/values predefined, the rest should be added by the user with "+"  and selected from predefined known list + any key/value user wants that will be rendered `<meta content="" property=""/>`
- [ ] Write a Page component for SDK that will render title/meta and replace the current `shared/documents/Canvas.tsx`
- [x] Add new page
- [ ] Set a page as a home page (only one page can be home and should have path as empty, that's how we can indicate in the db the home page)
- [ ] Apply final design 

![Pages Panel](https://user-images.githubusercontent.com/825702/193991016-f05b4bf5-f798-498a-ac41-5640fbabcdac.jpg)

kof:
Currently known questions:

1. Do we want a folder or do we want pages to be nestable?
2. Should folder add a prefix to the path? always? optionally? which one would be by default?

Things to consider:
- if we use folder to auto-add a path prefix, we enforce a more rigid structure for URLs for everything that goes in the folder, so this makes it easy to have a consistent URLs structure like /blog/some-post for all the posts but at the same time if folder is not something meaningful you want to have in the URL, it will just add to the URL useless stuff
- if we have pages instead of folders it would allow have an overview page that links to details like for. /blog it would list a list of posts and /blog/some-post would be the actual post
- in both cases we could make path on folder/page optional and simply disable it or let it be removable, in this case a page without a path is just a folder

rpominov:
I've created a bunch of separate issues for everything remaining to do: https://github.com/webstudio-is/webstudio-designer/labels/area%3Apages

Will close this one

PR #231, opened 2022-08-17
Preparation for pages

kof:
Preparing the database to build pages feature #54 

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/shared/db/build.server.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const Page = z.object({
+  id: z.string(),
+  name: z.string(),
+  path: z.string(),
+  title: z.string(),
+  meta: z.record(z.string(), z.string()),
+});
+const Pages = z.array(Page);
+type Pages = z.infer<typeof Pages>;
```
rpominov, line 12: In Webflow they have Folder as separate type from Page. It has no content, and only `name` and `slug` as properties. Should we do the same? Or do we want folders to also act as pages?

As a user it feels kinda like I'm forced to create some content for pages that I don't need. For example, if I'm building a blog I might want following pages:

 - `/` — list of posts
 - `/posts/first-post`
 - `/posts/second-post`

And I don't want anything at `/posts/`

rpominov: Might be a good idea to have `Homepage` separate from other pages:

```ts
type Pages = {
  homepage: Page;
  pages: array<Folder | Page>;
}
```

This will enforce 2 requirements:

 - Homepage must exist
 - Homepage must not be nested in any folder

kof: good idea

kof: > In Webflow they have Folder as separate type from Page. It has no content, and only name and slug as properties. Should we do the same? Or do we want folders to also act as pages?


@taylornowotny do you think this would be useful? I was also thinking about it and couldn't think of a use case

kof: > As a user it feels kinda like I'm forced to create some content for pages that I don't need. For example, if I'm building a blog I might want following pages:
> 
> / — list of posts
> /posts/first-post
> /posts/second-post
> And I don't want anything at /posts/

We failed to mention the "path" we let user enter is not a slug, user can enter a full path, folder doesn't create a path in our case, that's why having folders in our case is actually like folders: If you want a list you can still use /posts in the posts page

kof: One thing that is a little weird is that you would have a page called "posts" on the same level as the pages for each specific post.

rpominov: > We failed to mention the "path" we let user enter is not a slug, user can enter a full path

Hm, why do we need folders at all then? What if we implement pages as a flat list, at least for the first release? Would safe us a lot of time.

rpominov: Although if we later decide to control pages' URLs via nesting in folders, it will be hard to migrate to that from arbitrary paths.

kof: I have been back and forth on this too. Folders adding to the path force a certain URL structure you might not want to have, this is the feedback I got from @taylornowotny .

At the same time folders are good for organizing pages into categories of some sort when you have a lot of them. 

I was wondering if we could go with flat list of pages for beta too and wasn't sure if its going to save us a lot of time if we go with the same datastructure in the database.

The thing is if we want to leverage the patches, we would still need to update a single object. Potentially we could be updating one page at a time, but never the entire pages object. Then we have to ensure the right sort order. Having a single data structure for all pages ensures sort order automatically.

If we go with a flat list now, how hard will be adding folders in the future?

Thoughts?

kof: I don't think we truly need folders for beta release, but I was thinking its not gonna save us much time if we don't have them

kof: Regarding folder enforcing a path: I guess we could even make it an option. By default folder would have no path, but if added, it gets added automatically to all pages inside that folder, wdyt @taylornowotny ?

kof: Another thought: what if pages and folders are the same? If user doesn't want to use folder as a page, they will simply not use it. Nothing forces them to use that page, they are linking the pages anyways manually. So if its not linked its not used ... ? @taylornowotny

rpominov: > The thing is if we want to leverage the patches, we would still need to update a single object. Potentially we could be updating one page at a time, but never the entire pages object. Then we have to ensure the right sort order. Having a single data structure for all pages ensures sort order automatically.

Yeah, I'd still kept them in one JSON object event if they a flat list.

> If we go with a flat list now, how hard will be adding folders in the future?

A bit hard to tell, but I feel like if folders are used only for organizing, then there won't be much additional work compared to doing it now: `work to implement now with folders ~= now without folders + folders later`.

And this will allow us to think more about this. Maybe we'll decide to have tags instead of folders, or something like that.

kof: yeah, the question is if it saves us much time, in that case, but yeah, thinking through all those questions properly would be highly benefitial

rpominov: >  the question is if it saves us much time

Hard to tell. I think from a day to a week

kof: Ok lets first figure out without pages and then see if we can bring them in within a few days right away. We would still use the same tree component for drag&drop etc, so supporting dragging between folders should be easy

apps/designer/app/designer/designer.tsx
```diff
@@ -219,7 +219,7 @@ const NavigatorPanel = ({ publish, isPreviewMode }: NavigatorPanelProps) => {
 
 type DesignerProps = {
   config: Config;
-  project: Project;
+  project: db.project.Project;
```
kof, line 222: btw this is a total workaround, we need to have a way to get the augmented Project type, which is currently not importable in a good way, we need a project package I think

apps/designer/app/shared/db/build.server.ts
taylornowotny: Re: folders determining the path - I don't think is something we need to include as an option. Maybe some users would want to use folders to organize paths (like "/folder/page") but since they have full control over the path on every page anyway, they can set that up manually.

taylornowotny: Folders are meant to be purely organizational

taylornowotny: We could use tags instead of folders, but I think tags would be an inferior organizational tool in this case for the following reasons:

- folders are better for visual organization in the pages list because of nesting. Users can nest pages to hide them from view while they focus on other pages. With tags there is no nesting
- both tags and folders are good for grouping pages, but there's not currently a UI for tags so that would have to be designed and built
- tags are best used with sorting and filtering controls, which are best used for large sets. An average website might have a dozen pages or less, so the full utility of tags is wasted on such a small list.

apps/designer/app/designer/designer.tsx
rpominov: Might be better to have a single package that works with database. Its API would be based on the augmented types, and as a user of this package I would know nothing about `postgres` or maybe even `prisma`. 

But it should handle all database objects. Otherwise I feel like we'll end up in a dependency hell and might even have to deal with circular NPM dependencies which I don't think possible.

kof: The problem with this centralized approach is that it will end up depending on a LOT of other types across the application.

There is a another way - shared functionality between features always goes into a shared folder that is named based on its function/scope.

So for example there would be a shared/project folder that has everything about the project including augmented database types and utilities.

kof: The way it is right now is not great, because we tried initially to centralize database stuff in a shared/db folder and this is currently the result.

kof: Ultimately if we have it this way no feature will have to access prisma client directly, they will all work with shared/project shared/instance etc, which in return will call into prisma and augment data and types

apps/designer/app/shared/db/build.server.ts
```diff
@@ -0,0 +1,42 @@
...
+  path: z.string(),
+  title: z.string(),
+  meta: z.record(z.string(), z.string()),
+  treeId: z.string(),
+});
+
+type Page = z.infer<typeof Page>;
+
+const Pages: z.ZodType<{ homePage: Page; pages: Array<Page> }> = z.object({
+  homePage: Page,
```
kof, line 16: why not make it an id? 
```suggestion
  home: Page['id'],
```

kof: ah pages won't have the home page? I thought pages will have all the pages, and one of them is home

rpominov: I'm a bit obsessed with the idea of creating data-structures that make invalid states impossible to represent https://www.youtube.com/watch?v=IcgmSRJHu_8

Here, if we go with `id`, it may happen that the pages array will not have a page with that `id`.

But I think maybe we should just make it a simple array after all, with a `boolean` property `Page.isHome`. Need to see how a separate `homePage` complicates or doesn't complicate the code.

rpominov: Will probably go with a simple array + checks for all requirements for home page in the zod schema.

kof: > I'm a bit obsessed with the idea of creating data-structures that make invalid states impossible to represent

me too, definitely

> Here, if we go with id, it may happen that the pages array will not have a page with that id.

could be ensured by the logic, but yeah its less robust

kof: if you go with an isHome you have to ensure only one has it, if you go with a home: id - you only need to ensure the id still exists in the tree, so user can't delete home page without setting to home some other page

rpominov:
Something we decided in a chat about relationships of build an projects:

 - Project always refers to the current dev build
 - Project may refer to a prod build 
 - Build refers to a project to allow us to get history of all builds of a project
 - We should not assume that prod or dev build are chronologically most recent

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
```diff
@@ -0,0 +1,44 @@
...
+      );
+
+      if (project === undefined) {
+        continue;
+      }
+
+      builds.push({
+        pages: JSON.stringify({
+          homePage: {
+            id: uuidv4(),
```
rpominov, line 27: For `Page.id` I think it's better to use UUID instead of `bson-objectid` that we use for `Instance.id`. Main reason is that UUID can be easily created in an SQL query if we ever need to create a Page using just SQL.

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
```diff
@@ -0,0 +1,44 @@
+import { PrismaClient, Prisma } from "./client";
+import { v4 as uuidv4 } from "uuid";
```
kof, line 2: could just map it to uuid for simplicity

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
```diff
@@ -0,0 +1,44 @@
...
+    const projects = await prisma.project.findMany();
+
+    const builds: Prisma.BuildCreateManyInput[] = [];
+
+    for (const tree of trees) {
+      const project = projects.find(
+        (project) =>
+          project.devTreeId === tree.id ||
+          project.prodTreeId === tree.id ||
+          JSON.parse(project.prodTreeIdHistory).includes(tree.id)
```
kof, line 17: I would also run it over zod schema to be sure

kof: I mean in all cases where we JSON.parse or stringify

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
```diff
@@ -0,0 +1,44 @@
...
+          },
+          pages: [],
+        }),
+        isDev: project.devTreeId === tree.id,
+        isProd: project.prodTreeId === tree.id,
+        projectId: project.id,
+      });
+    }
+
+    await prisma.build.createMany({ data: builds });
```
kof, line 42: did you mean to run this synchronously? or could all of these run in parallel?

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
rpominov: This is a bit tricky. All code that is used in a migration has to be immutable to changes that we make to the rest of the project. We can't just import something. We would have to copy everything we want to use into the migration folder.

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
rpominov: Not sure I understand the question. If you're asking why I use `createMany` instead of `create`, that's because with `createMany` Prisma should make just one SQL query (it may split in to several batches, but still less that a query for each row)

rpominov: Doesn't matter for the amount of data we a're working with. Just want to set a good example.

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
kof: oh that's an interesting point, definitely need to document that! maybe even add to each migration file as a comment and in a template, because this is SUPER non-obvious

kof: I guess we could copy the schema we need to the local modules, just like the client

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
kof: what I mean is  that this current implementation will run one query after another in that for of loop, but it could run in parallel, I see 2 options:

1. using map, return promises from queries without awaiting, then Promise.all - this will send many queries all in parallel
2. use transaction API https://www.prisma.io/docs/concepts/components/prisma-client/transactions#the-transaction-api, not sure if prisma does there anything different than in 1 (I recall they can optimize that, not sure though)

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
rpominov: > maybe even add to each migration file as a comment and in a template

Good idea! Will add to the template

> I guess we could copy the schema we need to the local modules, just like the client

Something to keep in mind, but for now will just copy Zod definitions to the migration file manually

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
rpominov: > one query after another in that "for of" loop

There're no queries in the loop. Entire migration is 3 queries.

> using map, return promises from queries without awaiting, then Promise.all - this will send many queries all in parallel

This will work only if Prisma uses several connections to the database AND if we won't wrap the entire migration into a transaction (a transaction can't be run over multiple connections afaik)

> use transaction API https://www.prisma.io/docs/concepts/components/prisma-client/transactions#the-transaction-api, not sure if prisma does there anything different than in 1 (I recall they can optimize that, not sure though)

Same. Prisma can't run queries in a transaction in parallel, that's just a limitation on Postgres level

kof: Either I am missing it or you are: we are running those queries from some server. That server has a latency to the database server.

When we send the query and await for a response, we are waiting for execution + latency to receive the results.
When we parallelize like I said - queries are send all at once, we are waiting for all responses, no matter which one comes back first.

So I am optimizing away the latency between the database server and the node server.

kof: If it's just 3 queries it may not matter, but in a complex migration that could be many queries

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
kof: that's what I meant, zod schema definitions, copied, could be separate files next to the migration

packages/prisma-client/prisma/migrations/20220909124542_move_to_builds/migration.ts
kof: you are right, I misread when this query happens because of indentations, I thought it's inside the for-of loop but its after that and you are already using transaction, so I guess prisma will send all queries together

apps/designer/app/shared/db/build.server.ts
rpominov: Keeping `homePage: Page` for now, may reconsider later

kof:
@rpominov "Project may refer to a prod build"

What do you mean by this? Just to clarify what I thought initially is that project is kinda always in dev mode, so refers a dev build. When we publish, we clone that build and reference it as a prod build. So prod build is never modified by accident, it's always secure and we always operate on the dev build. 

I guess you already know that, just making sure

rpominov:
@kof Yeah, we're on the same page I think. Sorry if I described it in a confusing way. Was trying to describe it from the perspective of how it's laid out in the database.

rpominov:
Please ignore the current solution for now, btw. It's still in a flux. I need to finish updating the code that works with data to see which way is better. Let's discuss after PR is ready for review.

rpominov:
@kof This is ready for review (can't request as you're the author 😅 )

apps/designer/app/canvas/shared/sync/use-sync.ts
```diff
@@ -17,7 +17,7 @@ export const useSync = ({ project }: { project: Project }) => {
         method: "post",
         body: JSON.stringify({
           transactions: entries,
-          treeId: project.devTreeId,
+          treeId: project.devBuild?.pages.homePage.treeId,
```
kof, line 20: this will depend on the currently selected page, right? What's the plan, is this temporary and needs a todo?

andrasbacsai:
Merged main and resolved all conflicts, requested by @kof  (there was a lot 😅). 

@rpominov Please review when you are back.

apps/designer/app/routes/rest/breakpoints.$projectId.ts
```diff
@@ -10,19 +10,21 @@ export const loader: LoaderFunction = async ({
...
+
+    const prodBuild = await db.build.loadProdByProjectId(params.projectId);
+
+    if (prodBuild === undefined) {
       throw new Error(
         `Project ${params.projectId} needs to be published first`
       );
     }
-    const data = await db.breakpoints.load(project.prodTreeId);
+    const data = await db.breakpoints.load(prodBuild.pages.homePage.treeId);
```
kof, line 24: breakpoints is per build, not per page, we need to transform it to use build id

apps/designer/app/routes/rest/props.$projectId.ts
```diff
@@ -9,8 +9,18 @@ export const loader: LoaderFunction = async ({
   params,
 }): Promise<Array<InstanceProps> | ErrorData> => {
   try {
-    const project = await db.project.loadById(params.projectId);
-    return await db.props.loadByProject(project, "production");
+    if (params.projectId === undefined) {
+      throw new Error(`Project ID required`);
+    }
+
+    const prodBuild = await db.build.loadProdByProjectId(params.projectId);
```
kof, line 16: wouldn't it read more nicely if we had build.loadByProjectId(projectId, 'prod') ?

apps/designer/app/shared/db/misc.server.ts
```diff
@@ -23,24 +23,16 @@ export const publish = async ({
...
-  });
-  const prodTreeIdHistory = project.prodTreeIdHistory;
-  if (project.prodTreeId) {
-    prodTreeIdHistory.push(project.prodTreeId);
+  // should not happen
+  if (project.devBuild == undefined) {
+    throw new Error(`project.devBuild is not loaded`);
   }
+
+  await db.build.createProd(project.devBuild, project.id);
```
kof, line 31: same suggestion here `build.create(project.id, project.devBuild, "prod")`

packages/prisma-client/migrations-cli/commands.ts
```diff
@@ -129,6 +129,26 @@ const ensureLastPending = async (migrationsName: string) => {
   }
 };
 
+const isNoopSql = (sqlScript: string) => {
```
kof, line 132: you probably wanna split these into a separate pr and we can merge it quickly

packages/project/src/db/build.ts
```diff
@@ -0,0 +1,119 @@
...
+      treeId: tree.id,
+    },
+    pages: [],
+  });
+};
+
+export const createDev = async (baseBuild?: Build) => {
+  if (baseBuild === undefined) {
+    const pages = await createPages();
+    return prisma.build.create({
```
kof, line 85: not awaiting?

packages/project/src/db/build.ts
```diff
@@ -0,0 +1,119 @@
...
+      data: { pages: JSON.stringify(pages) },
+    });
+  } else {
+    const pages = await clonePages(baseBuild.pages);
+    return prisma.build.create({
```
kof, line 90: same

packages/project/src/db/project.ts
```diff
@@ -119,43 +121,20 @@ export const clone = async (clonableDomain: string, userId: string) => {
...
-  ]);
-
-  const parsedProject = parseProject(project);
-  if (parsedProject === null) {
-    throw new Error(`Not found project "${clonableDomain}"`);
-  }
-  return parsedProject;
+  const devBuild = await db.build.createDev(prodBuild);
+
+  return create({
```
kof, line 133: await?

packages/project/src/db/types.ts
```diff
@@ -1,9 +1,33 @@
...
+import {
+  Project as BaseProject,
+  Build as DbBuild,
+} from "@webstudio-is/prisma-client";
 import type { Asset } from "@webstudio-is/asset-uploader";
-const TreeHistorySchema = z.array(z.string());
 
-export type Project = Omit<BaseProject, "prodTreeIdHistory" | "assets"> & {
-  prodTreeIdHistory: z.infer<typeof TreeHistorySchema>;
+// Project
```
kof, line 8: useful comment :)

rpominov:
Thanks @andrasbacsai !

packages/project/src/db/build.ts
rpominov: Not sure I understand

```
const fn = async () => {
  return promise
}
```

Is equivalent to

```
const fn = async () => {
  const result = await promise
  return result
}
```

Or do we consider this a bad practice?

kof: oh is it? I didn't think this through

apps/designer/app/canvas/shared/sync/use-sync.ts
rpominov: Yeah, temporary. We wanted to only change schema in this PR, and make minimal changes to the reset of the code, just to make it work.

Will add todos.

kof: gotcha, as you wish, as long as we can merge it and potentially release it, all good

andrasbacsai:
Merged the main again, because there was some problem with `projects` package.

packages/prisma-client/migrations-cli/commands.ts
rpominov: Sorry for mixing this up. I'll do better next time, but I'd safe time on splitting and merge this as is

rpominov (approved):

rpominov:
@kof Ready for review

packages/prisma-client/prisma/migrations/template.txt
```diff
@@ -10,7 +10,10 @@ import { PrismaClient } from "./client";
 //   It's better to copy it to the migration directory.
 
 export default () => {
-  const client = new PrismaClient();
+  const client = new PrismaClient({
+    // Uncomment to see the queries in console as the migration runs
+    // log: ["query", "info", "warn", "error"],
```
kof, line 15: you can also do thi https://www.prisma.io/docs/concepts/components/prisma-client/debugging

kof (commented):
lgtm, pelase check the preview env too

packages/prisma-client/prisma/migrations/template.txt
rpominov: I tried this, but it doesn't output queries in SQL format, only in some prisma internal format

rpominov:
Preview won't work because migrations are not applied

PR #234, opened 2022-08-21
Lazy load text editor

kof:
We need to start optimizing bundle size because it has grown substantially. First thing is to not load things that we don't need on first load.

In this case we will lazy load the text editor only first time once user enters editing mode of the instance

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


rpominov (approved):
Looks good 👍

PR #240, opened 2022-08-22
Refactor `pubsub` to use strongly-typed events between publishers and subscribers

Andarist:
## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [x] detailed
  - [x] with testing

---

1. I've refactored publish+subscribe from the SDK to `createPubsub`
2. I've created a single "instance" of the pubsub in the app (mainly to decouple the SDK from app-specific types etc)
3. I now use this throughout the app 
4. thanks to interface augmentations I augment the `PubsubMap` with the types specific to particular published events - this allows you to ensure that there is no type mismatch between a publisher and a subscriber. This solution has been chosen to avoid maintaining a single, big, list of events - with this approach you can add types "locally", close to the feature (they will be seen globally though but that's also the whole point of this)

I've tried my best to colocate particular augmentations with features but I definitely didn't do this in an ideal way. This requires somebody with a deeper understanding of the codebase and the domain - changing this should be as easy as moving particular declarations to other files.


apps/designer/app/canvas/shared/use-shortcuts.ts
```diff
@@ -1,38 +1,59 @@
...
-  publish<"togglePreviewMode">({ type: "togglePreviewMode" });
+  publish({ type: "togglePreviewMode" });
 };
 
 const publishSelectBreakpoint = ({ key }: HandlerEvent) => {
+  if (!key) {
+    throw new Error(
+      "`publishSelectBreakpoint` doesn't account for being called without a `key`"
+    );
+  }
```
Andarist, line 44: This is not ideal but I've felt that it's better to add such a throw here, just in case. The problem is that [this file](https://github.com/webstudio-is/webstudio-designer/blob/b913ac73d726ad41ca33bda2d44f1bd6fdf09ea9/apps/designer/app/designer/features/topbar/menu/menu.tsx) piggybacks on the shortcut events - but in there we don't have access (or at least not always) to the `key`.

Perhaps this file shouldn't do it at all and it should just use its own, explicit, events - but that would be a separate refactor.

Not all shortcut events are actually dispatched from that `menu.tsx` and those to which I've added those throws arent dispatched from there. I can't easily filter them out though because it's a list that is being looped on etc. So I've just made `HandlerEvent["key"]` optional for the time-being

apps/designer/app/canvas/shared/use-shortcuts.ts
```diff
@@ -1,38 +1,59 @@
...
   publish({
     type: "selectBreakpointFromShortcut",
-    payload: key,
+    payload: parseInt(key, 10),
```
Andarist, line 47: As far as I can tell this is a fix because the subscriber expects to receive an `index`, and that should be a number:
https://github.com/webstudio-is/webstudio-designer/blob/b913ac73d726ad41ca33bda2d44f1bd6fdf09ea9/apps/designer/app/designer/features/breakpoints/use-subscribe-shortcuts.ts#L10

apps/designer/app/designer/features/style-panel/use-style-data.ts
```diff
@@ -3,13 +3,21 @@ import type {
...
 import { useSelectedBreakpoint } from "../../shared/nano-states";
 import { parseCssValue } from "./parse-css-value";
 import { getInheritedStyle } from "./get-inherited-style";
 import { getCssRuleForBreakpoint } from "./lib/utils/get-css-rule-for-breakpoint";
 import { useRootInstance } from "~/shared/nano-states";
 
+declare module "~/shared/pubsub" {
+  export interface PubsubMap {
+    updateStyle: StyleUpdates;
+    [key: `previewStyle:${Instance["id"]}`]: StyleUpdates;
```
Andarist, line 17: Isn't TypeScript just beautiful? 😉

apps/designer/app/designer/features/topbar/menu/menu.tsx
```diff
@@ -140,9 +140,9 @@ export const Menu = ({ config, publish }: MenuProps) => {
         <DropdownMenuItem
           css={menuItemCss}
           onSelect={() => {
-            publish<"shortcut", string>({
+            publish({
               type: "shortcut",
-              payload: "undo",
+              payload: { name: "undo" },
```
Andarist, line 145: This was somewhat surprising - but as far as I could tell those call sites here probably were buggy as the subscriber expects an object:
https://github.com/webstudio-is/webstudio-designer/blob/b913ac73d726ad41ca33bda2d44f1bd6fdf09ea9/apps/designer/app/canvas/shared/use-shortcuts.ts#L167

apps/designer/app/shared/pubsub/index.ts
```diff
@@ -0,0 +1,9 @@
+import { createPubsub } from "@webstudio-is/react-sdk";
```
Andarist, line 1: This file is the core of the magic (together with the `createPublish` declaration in the SDK)

package.json
```diff
@@ -31,5 +31,6 @@
   "packageManager": "yarn@1.22.19",
   "lint-staged": {
     "*.{ts,tsx,js,json,css,md}": "prettier --write"
-  }
+  },
+  "prettier": {}
```
Andarist, line 35: I have a setting in my IDE that requires a Prettier config to reformat files so I've added this. I'm using this setting because I don't want to reformat repos that don't use Prettier at all

apps/designer/app/shared/shortcuts/index.ts
```diff
@@ -5,6 +5,7 @@ export const shortcuts = {
   preview: "cmd+shift+p, ctrl+shift+p",
   copy: "cmd+c, ctrl+c",
   paste: "cmd+v, ctrl+v",
+  delete: "backspace, delete",
```
Andarist, line 8: This is a runtime change and perhaps this should be reverted. I've added this to satisfy this call:
https://github.com/webstudio-is/webstudio-designer/blob/b913ac73d726ad41ca33bda2d44f1bd6fdf09ea9/apps/designer/app/designer/features/topbar/menu/menu.tsx#L192-L195
and I've matched the keys declared here:
https://github.com/webstudio-is/webstudio-designer/blob/b913ac73d726ad41ca33bda2d44f1bd6fdf09ea9/apps/designer/app/canvas/shared/use-shortcuts.ts#L107

apps/designer/app/canvas/shared/use-shortcuts.ts
kof: I think its better to make sure the key is always provided instead.

kof: Also key could totally stay a string, the other number type should change, dom key is string

apps/designer/app/canvas/shared/use-shortcuts.ts
kof: nah, lets make the key string there

apps/designer/app/designer/features/sidebar-left/lib/navigator/navigator.tsx
```diff
@@ -1,11 +1,18 @@
...
-import type { Instance, Publish } from "@webstudio-is/react-sdk";
+import type { Instance } from "@webstudio-is/react-sdk";
+import { type Publish } from "~/shared/pubsub";
 import { useSelectedInstancePath } from "~/designer/shared/instance/use-selected-instance-path";
 import { useSelectedInstanceData } from "~/designer/shared/nano-states";
 import { Tree } from "~/designer/shared/tree";
 import { useRootInstance } from "~/shared/nano-states";
 import { Header } from "../header";
 
+declare module "~/shared/pubsub" {
```
kof, line 10: what worries me about these is that their location is sort of random, how do we know where to put those declarations? Should we centralize them for things that are "shared" and not actually local?

apps/designer/app/designer/features/workspace/workspace.tsx
```diff
@@ -1,7 +1,23 @@
 import { Box, Flex, Toaster } from "@webstudio-is/design-system";
+import { type Instance, type InstanceProps } from "@webstudio-is/react-sdk";
 import { useCanvasWidth, useZoom } from "~/designer/shared/nano-states";
 import { CanvasTools } from "./canvas-tools";
-import { type Publish } from "@webstudio-is/react-sdk";
+import { type Publish } from "~/shared/pubsub";
+
+declare module "~/shared/pubsub" {
```
kof, line 7: same here, location of this is kinda random, because those events are used all over the place

apps/designer/app/shared/shortcuts/index.ts
kof: why would it need a revert? seems like a legit fix, I will go through all shortcuts in the end and test them

kof:
We should probably move that pubsub into a package right away

apps/designer/app/canvas/shared/use-shortcuts.ts
Andarist: `selectBreakpointFromShortcut` doesn't have to rely on literal keys coming from the events - it's already a "translated" event (sort-of). IMHO when possible it's better to refine data asap, closer to the publisher than to the subscriber. As this is going to be used as an index, I think that it makes sense to convert this here and have that consuming logic slightly simpler.

apps/designer/app/designer/features/sidebar-left/lib/navigator/navigator.tsx
Andarist: Ye, I think that putting shared things in the central spot would be better. It just wasn't always obvious to me where things should be put and which ones are local and which ones are not. Those random places in which you can find those augmentations right now are usually in files that publish those events.

apps/designer/app/canvas/shared/use-shortcuts.ts
Andarist: > I think its better to make sure the key is always provided instead.

In such a case - do you want to provide fake keys to those dispatched `shortcut` events? Or do you want to refactor them to explicit events (instead of piggybacking on the existing `shortcut` event)?

kof:
@andrasbacsai any idea why the publish fails?

apps/designer/app/canvas/features/wrapper-component/use-css.tsx
```diff
@@ -23,12 +23,9 @@ const usePreviewCss = ({ instance, css }: UseCssProps) => {
     StyleUpdates["updates"] | UpdatesReset
   >([]);
 
-  useSubscribe<string, StyleUpdates>(
-    `previewStyle:${instance.id}`,
-    ({ updates }) => {
-      setPreviewCss(updates);
-    }
-  );
+  useSubscribe(`previewStyle:${instance.id}`, ({ updates }) => {
```
rpominov, line 26: So cool that even this works!

andrasbacsai:
The package publishing fails because the current setup requires manually increasing the version before committing. 
Unfortunately, I could not set up a process where the GitHub Action checks for the commit message and decides automatically if a new version is required from the specific package because we are in a monorepo setup. It is hard to determine which package needs to be updated.

So I leave it to the most straightforward setup -> If you need a new version of a package, you manually bump the version inside the proper `package.json`

rpominov (approved):
Love this PR! Was so annoying to define all these types and get no type safety in return.

apps/designer/app/canvas/features/wrapper-component/use-css.tsx
kof: yeah

Issue #228, opened 2022-08-16
Create a TS HTTP API Client


kof:
As a developer I need a TS client to fetch all the project data over HTTP without having to build my own fetch calls, because this way we can optimize the calls and also have more flexibility in changing the HTTP API for now, since it's not public yet.

- [x] build the client that can fetch all project data
- [x] use it in cli
- [x] use it in worker

kof:
I think this was already done

PR #242, opened 2022-08-23
HTTP API package

andrasbacsai:
Related issue #228.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/http-client/src/index.d.ts
```diff
@@ -0,0 +1,7 @@
+export type IncludeTypes = "tree" | "props" | "breakpoints";
```
kof, line 1: it looks like we need those types to be installable as a dependency

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,37 @@
+import isoMorphicFetch from "isomorphic-fetch";
```
kof, line 1: why do you need isophorphic fetch? can just use fetch and set the required node version

andrasbacsai: Because with isomorphic fetch, we could use the server side (uses `node-fetch`) and browser (uses `window.fetch`) without knowing the environment.

packages/http-client/src/index.d.ts
andrasbacsai: Can you please elaborate on this?

packages/http-client/src/index.ts
kof: what I am saying I would rather isntall a polyfill instead of this and then only for node https://www.npmjs.com/package/node-fetch#providing-global-access

in all other environments (browsers, workers) we already have fetch, we should treat it in code like it is always provided

packages/http-client/src/index.d.ts
kof: what I mean is that we need to make sure that we can update main repo without hardcoding things as much as possible, I just checked this isn't used yet, so I may not have a full context why this selection is here

andrasbacsai: With the exported `loadProjectFromRest` function, you can include only data you would require from the REST API. By default all data is included.

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+const loadProjectFromRest = async ({
+  projectId,
+  host = process.env.DESIGNER_HOST || "localhost:3000",
+  include = { tree: true, props: true, breakpoints: true },
+}: {
+  projectId: string;
+  host?: string;
+  include?: Includes<boolean>;
+}): Promise<Project> => {
+  const domain = `${protocol}://${host}`;
```
kof, line 29: should use new URL api

kof: it will give you error if malformed

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    include.breakpoints && fetch(`${domain}/rest/breakpoints/${projectId}`),
+  ]);
+  return {
+    tree,
+    props,
+    breakpoints,
+  };
+};
+
+export { loadProjectFromRest };
```
kof, line 42:
```suggestion
export { loadProject};
```
Rest is an implementation detail, we can switch to anything else but API can stay the same

kof (changes requested):
added some feedback

packages/http-client/src/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+export type Includes<T> = { [include in IncludeTypes]: T };
+export type Project = {
```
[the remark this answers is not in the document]
andrasbacsai: `Includes` and `IncludeTypes` (or any similar) are not defined in the main project.

For Projects, that is okay. I fix types there.

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,42 @@
+import isoMorphicFetch from "isomorphic-unfetch";
+import type { Includes, Project } from "./index.d";
+
+const protocol = process.env.NODE_ENV === "production" ? "https" : "http";
+
+const fetch = async (url: string, options?: RequestInit) => {
+  const response = await isoMorphicFetch(url, options);
```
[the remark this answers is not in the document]
andrasbacsai: Ah, missed your last comment on the previous conversation.

andrasbacsai:
Fixed the issues.

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,32 @@
+import "./fetch-polyfill";
+import type { Includes, Project } from "./index.d";
+
+const protocol = process.env.NODE_ENV === "production" ? "https" : "http";
+
+const loadProject = async ({
```
kof, line 6: nit
```suggestion
export const loadProject = async ({
```

kof (approved):
Assuming you want to merge this PR, then in the next step reuse Project type definition from a new package

packages/http-client/src/index.ts
andrasbacsai: Fixed.

Issue #245, opened 2022-08-29
Data migrations


rpominov:
We have a problem with database migrations: Prisma does schema migrations very well, but has very poor support of data migrations. A data migration is when you need to move data from one column to another or from one table to another and/or modify JSON data somehow.

Here are the solutions I found:

## 1. Prisma custom migrations

https://www.prisma.io/docs/guides/database/developing-with-prisma-migrate/customizing-migrations

**TLDR**: You modify data in an SQL file

|                                                                     |                                                                                                                                              |
| ------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| SQLite support                                                      | In theory, but it makes it even more difficult to write a migration as you'd need to write SQL that can run in two database engines |
| Ability to run JavaScript in a migration                            | No, migration is an SQL file. Could be possible in a limited way with https://plv8.github.io/                                                |
| Access to a Prisma Client in a migration                            | No, migration is an SQL file.                                                                                                                |
| Changes to data are applied before the next schema migration is run | Yes                                                                                                                                          |
| Amount of work to implement the solution                            | Zero                                                                                                                                         |

## 2. RedwoodJS Data Migrations

https://redwoodjs.com/docs/data-migrations

**TLDR**: You modify data in a JavaScript file using Prisma Client

|                                                                     |                                                                                                                                                                                                              |
| ------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| SQLite support                                                      | In theory                                                                                                                                                                                           |
| Ability to run JavaScript in a migration                            | Yes                                                                                                                                                                                                          |
| Access to a Prisma Client in a migration                            | Yes. But you're given a Prisma Client corresponding to the current — for the time when you apply the migration — content of `schema.prisma`. This is a problem because you might need to manipulate fields that may no longer exist in the current schema. |
| Changes to data are applied before the next schema migration is run | No. You first run **all** schema migrations, and then **all** data migrations.                                                                                                                           |
| Amount of work to implement the solution                            | Not much                                                                                                                                                                                                     |

Other caveats:

- We'll probably have to add entire `@redwoodjs/core` as a dependency. Which is about 1GB of `node_modules`

## 3. Knex.js

https://knexjs.org/guide/migrations.html

**TLDR**: You modify data in a JavaScript file. You should be able to import the Prisma Client

|                                                                     |     |
| ------------------------------------------------------------------- | --- |
[...]

PR #249, opened 2022-08-30
Data migrations

rpominov:
https://github.com/webstudio-is/webstudio-designer/issues/245#issuecomment-1230389726

TODO:

 - [x] Command for creating schema migration (based on `prisma migrate diff`)
 - [x] Command for creating data migration
   - [x] Use Umzug's `create` to create the file
   - [x] Use `prisma generate` to create a frozen Prisma Client 
 - [x] Command for applying migrations
   - [x] If there're failed migrations from previous runs: show error, stop
   - [x] Apply migrations using Umzug
  - [ ] Command similar to `prisma migrate reset`
    - [ ] Confirmation prompt 
    - [ ] Use `prisma migrate reset` to clear DB, but prevent it from applying migrations somehow
      - [ ] Alternatively https://stackoverflow.com/a/13823560/478603 (Prisma does the same: https://github.com/prisma/prisma-engines/blob/b92d74444c1c0298714e9b140a85dd714b8dc7c4/migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs#L267-L278 . See also the `?schema` parameter of connection URL: https://www.prisma.io/docs/concepts/database-connectors/postgresql#connection-url)
    - [ ] Apply migrations using Umzug
  - [x] Command similar to `prisma migrate status`
  - [x] Command similar to `prisma migrate resolve`
  - [ ] Documentation
  - [ ] Test and document use-cases:
    - [ ] Setting up a new development environment
    - [ ] Creating and applying a schema migration
    - [ ] Creating and applying a data migration
    - [ ] Applying migrations added by other developers
    - [ ] Applying migrations in production / preview deployments
  - [x] Testing / cleanup

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/package.json
+++ b/apps/designer/package.json
@@ -13,7 +13,8 @@
     "watch:sdk": "relative-deps && relative-deps watch",
     "prettier": "prettier \"*.{js,md,json,ts}\" \"{docs,app}/**/*.{js,md,json,ts}\"",
     "storybook:run": "start-storybook -p 6006",
-    "storybook:build": "build-storybook"
+    "storybook:build": "build-storybook",
+    "migrations": "migrations --dev"
   },
   "dependencies": {
     "@lexical/react": "^0.3.5",
--- a/packages/prisma-client/migrations-cli/bin.js
+++ b/packages/prisma-client/migrations-cli/bin.js
@@ -1,0 +1,6 @@
+#!/usr/bin/env node
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+require("esbuild-register/dist/node").register();
+require("./engine").runCLI();
--- a/packages/prisma-client/migrations-cli/bin.js
+++ b/packages/prisma-client/migrations-cli/bin.js
@@ -1,6 +1,0 @@
-#!/usr/bin/env node
-
-/* eslint-disable @typescript-eslint/no-var-requires */
-
-require("esbuild-register/dist/node").register();
-require("./engine").runCLI();
--- a/packages/prisma-client/migrations-cli/umzug.ts
+++ b/packages/prisma-client/migrations-cli/umzug.ts
@@ -1,0 +1,96 @@
+import { Umzug } from "umzug";
+import fs from "node:fs";
+import * as prismaMigrations from "./prisma-migrations";
+import * as logger from "./logger";
+
+export const umzug = new Umzug({
+  migrations: {
+    glob: [
+      `./${"[0-9]".repeat(14)}_*`,
+      { cwd: prismaMigrations.migrationsDir },
+    ],
+    resolve(params) {
+      const sqlFilePath = prismaMigrations.getMigrationFilePath(
+        params.name,
+        "sql"
+      );
+      const tsFilePath = prismaMigrations.getMigrationFilePath(
+        params.name,
+        "ts"
+      );
+
+      if (fs.existsSync(sqlFilePath)) {
+        return {
+          ...params,
+          up: async () => {
+            await prismaMigrations.setStarted(params.name, sqlFilePath);
+            prismaMigrations.cliExecute(sqlFilePath);
+          },
+        };
+      }
+
+      if (fs.existsSync(tsFilePath)) {
+        return {
+          ...params,
+          up: async () => {
+            prismaMigrations.generateMigrationClient(params.name);
+
+            // eslint-disable-next-line @typescript-eslint/no-var-requires
+            const migration = require(tsFilePath);
+
+            if (typeof migration.default !== "function") {
+              throw new Error(
+                `Migration file's ${tsFilePath} default export must be a function`
+              );
+            }
+
+            await prismaMigrations.setStarted(params.name, tsFilePath);
+            await migration.default();
+          },
+        };
+      }
+
+      throw new Error(
+        `Couldn't find a migration.ts or migration.sql file in migrations/${params.name}`
+      );
+    },
+  },
+  context: {},
+  storage: {
+    logMigration(params) {
+      return prismaMigrations.setApplied(params.name);
+    },
+    unlogMigration() {
+      // this is for rollbacks, which we don't currently support
+      throw new Error("unlogMigration is not implemented");
+    },
+    async executed() {
+      const migrations = await prismaMigrations.getMigrations();
+      return migrations
+        .filter(prismaMigrations.isAppliedMigration)
+        .map((row) => row.migration_name);
+    },
+  },
+  logger: {
+    ...logger,
+    info: (event) => {
+      if (event.event === "migrating") {
+        logger.info(`Starting ${event.name}`);
+        return;
+      }
+      if (event.event === "migrated") {
+        logger.info(`${event.name} done in ${event.durationSeconds}s`);
+        return;
+      }
+      if (event.event === "up") {
+        logger.info(event.message);
+        return;
+      }
+      if (event.event === "created") {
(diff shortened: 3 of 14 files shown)


packages/prisma-client/migrations-cli/umzug.ts
```diff
@@ -1,13 +1,39 @@
+/* eslint-disable no-console */
+
 import { Umzug, FileLocker, memoryStorage } from "umzug";
 import fs from "node:fs";
 import path from "node:path";
+import child_process from "node:child_process";
 import { prisma } from "../src";
 
-const executeFile = async (filepath: string, timeout: number = 1000 * 60) => {
+const runPrismaCLI = (args: string[]) => {
```
kof, line 9: did you check if prisma has an api we can import and do normal function calls instead of executing a script in a shell?

packages/prisma-client/migrations-cli/bin.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+require("ts-node").register({ transpileOnly: true });
```
kof, line 3: couldn't you use tsup for executing ts? much faster, can't generate typings though, but I am assuming you only need it for execution, check out how its used in sdk/bin for e.g.

rpominov: Ok will try. Took "ts-node" from the examples I'm basing this all on.

packages/prisma-client/migrations-cli/umzug.ts
```diff
@@ -0,0 +1,203 @@
...
+          date.getMinutes(),
+          date.getSeconds(),
+        ]
+          .map((item) => item.toString().padStart(2, "0"))
+          .join("");
+      const name = path.basename(filepath);
+      return [
+        [
+          path.join(
+            __dirname,
```
kof, line 114: avoid using __dirname, its deprecated in latest node versions

packages/prisma-client/migrations-cli/umzug.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  const bin = path.resolve("./node_modules/.bin/prisma");
+  if (fs.existsSync(bin) === false) {
+    console.error(
+      "Could not find Prisma CLI binary. Running `yarn install` might help.\n"
+    );
+    process.exit(1);
+  }
+  try {
+    child_process.execFileSync(bin, args, {
```
kof, line 20: have you seen shelljs? https://www.npmjs.com/package/shelljs 

Would it be worth it? this code is not hard, but there might be edge cases I am not aware off, just as an idea in case we can't use prisma API directly, because wrapping the shell script is definitely less optimal

packages/prisma-client/migrations-cli/umzug.ts
```diff
@@ -0,0 +1,203 @@
...
+        console.log(event.message);
+        return;
+      }
+      if (event.event === "created") {
+        console.log(`created: ${event.path}`);
+        return;
+      }
+      console.log(event);
+    },
+    error: (event) => console.error(event),
```
kof, line 144: probably could use process.stdout/stderr so we can keep console stuff for debugging, we have a linter that prevents us checking in console calls that should have warned about these

packages/prisma-client/migrations-cli/umzug.ts
```diff
@@ -0,0 +1,203 @@
...
+    },
+    executed(meta) {
+      // @todo: read from _prisma_migrations
+      return tmpMemoryStorage.executed(meta);
+    },
+  },
+  create: {
+    template: (filepath) => {
+      const date = new Date();
+      const prefix =
```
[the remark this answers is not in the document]
rpominov: I don't see how I can create the string in the format I need using `toLocalDateString`

packages/prisma-client/migrations-cli/umzug.ts
rpominov: Do you know what is the alternative? I googled briefly and couldn't fine a clear answer. If, you don't know, don't worry, I'll look into it more later.

packages/prisma-client/migrations-cli/umzug.ts
rpominov: I thought about using some package, but still on the fence about it

packages/prisma-client/migrations-cli/umzug.ts
kof: https://stackoverflow.com/questions/46745014/alternative-for-dirname-in-node-js-when-using-es6-modules

rpominov: __dirname isn't deprecated https://nodejs.org/api/modules.html#__dirname

You're probably thinking of it being unavailable in the es6-module context. But this code is not in es6-module context.

rpominov: Just not sure `import.meta` will work in the non es6-module, but will try

kof: maybe there is a wrapper package that will work everywhere

packages/prisma-client/migrations-cli/bin.js
rpominov: Do you mean `esbuild-node-loader` like we use it here? https://github.com/webstudio-is/webstudio-designer/blob/7a99e3dfd7d5179e741f8327ddee8c0bde7f86d3/packages/react-sdk/package.json#L10

packages/prisma-client/migrations-cli/umzug.ts
rpominov: I now use `__dirname` in only one place. Will be easy to replace if we'll need to. Not sure we should spend any more time on this, not really that important.

packages/prisma-client/migrations-cli/umzug.ts
rpominov: Using `execa` now

packages/prisma-client/migrations-cli/bin.js
kof: yeah, sorry, tsup is for bundling and the loader is for running, both use esbuild, so its 100x faster than tsc

rpominov: Using `tsx` now

packages/prisma-client/migrations-cli/umzug.ts
rpominov: I've now added a logger that uses console. The idea is that for debug you use console as normal, and for CLI output you use the logger.

apps/designer/package.json
```diff
@@ -13,7 +13,8 @@
     "watch:sdk": "relative-deps && relative-deps watch",
     "prettier": "prettier \"*.{js,md,json,ts}\" \"{docs,app}/**/*.{js,md,json,ts}\"",
     "storybook:run": "start-storybook -p 6006",
-    "storybook:build": "build-storybook"
+    "storybook:build": "build-storybook",
+    "migrations": "migrations --dev"
```
rpominov, line 17: @kof For working with migrations during development it's currently set up like this:

```
$ cd apps/designer
$ yarn migrations # will output usage information
$ yarn migrations command # will run a command 
```

rpominov:
@kof I think this can be merged in this form. Remaining work that I'll do in a separate PR soon:

- The command for reseting the database
- Documentation

PR #251, opened 2022-08-31
Assets - Fix design of upload button and cog

SaraVieira:
## Before requesting a review

- [ x] made a self-review
  - [ ] conceptual
  - [ x] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

This PR updates tge Cog Icon and the upload button to match designs

CleanShot 2022-08-31 at 17 38 56@2x
CleanShot 2022-08-31 at 17 39 12@2x

CleanShot 2022-08-31 at 17 39 26@2x


apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
```diff
@@ -46,8 +46,17 @@ export const AssetTooltip = ({
...
-          color: "$hiContrast",
           transition: "opacity 100ms ease",
+
+          "svg path": {
+            fill: "$loContrast",
+          },
+
+          "&:hover": {
+            "svg path, svg circle": {
+              stroke: "$hiContrast",
```
kof, line 57: you coulnd't just set the current color?

kof: That would allow you to not write those svg selectors

kof: same for fill

packages/icons/src/upload.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { IconProps } from "./types";
+
+export const UploadIcon = React.forwardRef<SVGSVGElement, IconProps>(
+  ({ color = "currentColor", ...props }, forwardedRef) => {
+    return (
+      <svg
+        width="16"
+        height="16"
+        viewBox="0 0 16 16"
+        fill="none"
```
kof, line 11: does this attribute need to be here?

kof (changes requested):
some nits

packages/icons/src/upload.tsx
SaraVieira: nop, fixed

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-tooltip.tsx
SaraVieira: I did current color but still had to use the `svg` because of CSS specificity

kof: why? who is also defining that color? specificity means there is a conflict

SaraVieira: https://github.com/webstudio-is/webstudio-designer/blob/ad4286782ccb726e11ab1e426f53975b2dd15db7/apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx#L71

kof: I would fix both and use props instead of cascading selectors. Prop that is semantically saying what it is changing.

SaraVieira: I used JS for hover and passed props down

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
```diff
@@ -67,15 +68,17 @@ export const AssetThumbnail = (asset: Asset | UploadingAsset) => {
         alignItems: "center",
         padding: "0 $2",
         position: "relative",
-
-        "&:hover button": {
-          opacity: 1,
-        },
       }}
+      onMouseEnter={() => setIsBeingHovered(true)}
```
kof, line 72: oh got why this needed to be javascrtip event handling ... Sara!

kof: I will check this out now

SaraVieira: You said css bad

kof: OMG NO!

kof: semantic props

kof: you talk to a component over props and tell it what to do in a semantic way, component decides what css it needs to render

kof: writing cascading selectors is a thing you do inside a component targeting elements you control in that component, not the child components that may have conflicting selectors, by doing so you are breaching the component boundary.

cascading like this is a thing before components age, think of components as of  custom elements shadow-root, you wouldn't be able to cascade into the component like this

kof: passing handlers is not breaching the  boundary, but this is generally less efficient, so not a great way either

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info-tigger.tsx
```diff
@@ -41,16 +40,19 @@ export const AssetTooltip = ({
...
           cursor: "pointer",
-          color: "$hiContrast",
           transition: "opacity 100ms ease",
+          "&:hover": {
+            color: "$hiContrast",
+          },
         }}
       >
-        <GearIcon />
+        <GearIcon fill={String(theme.colors.loContrast)} />
```
kof, line 55: I am taking the css variable from the theme, passing it as a prop to the component. Alternative would be passing it from the parent  as a new css variable like --gear-icon-fill .... not sure what is cleaner ... thoughts?

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-info-tigger.tsx
```diff
@@ -41,16 +40,19 @@ export const AssetTooltip = ({
         title="Options"
         onClick={() => setTooltipOpen(true)}
         css={{
-          opacity: 0,
+          visibility: "var(--info-trigger-visibility)",
```
kof, line 43: visibility value passed from parent using css variable to avoid nested selectors that may have specificity conflicts later in the future

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
```diff
@@ -67,15 +67,15 @@ export const AssetThumbnail = (asset: Asset | UploadingAsset) => {
         alignItems: "center",
         padding: "0 $2",
         position: "relative",
-
-        "&:hover button": {
-          opacity: 1,
+        "--info-trigger-visibility": "hidden",
+        "&:hover": {
```
kof, line 71: when button is being hovered, we define the visibility variable that will be used inside the trigger, it acts as a public API

kof (approved):

apps/designer/app/designer/features/sidebar-left/panels/asset-manager/asset-thumbnail.tsx
kof: one problem that is still here is that this is completely untyped by ts, so maybe we should find a way to have the child component fully describe what we are passing, either by exporting the variable with valid values or a function that we can spread

rpominov: I saw this, not sure sure how this works, but can we do the same here? https://github.com/webstudio-is/webstudio-designer/blob/4164defb0818b5da9cc6ecca12abd885b78210ba/packages/design-system/src/components/control-group.tsx#L13

kof: This approach has still the problem with breaching Button's encapsulation, because you would define any styles on the Button, without Button explicitly allowing or being aware of the overrides, right?

kof: also you could go and write styles that reach into childs internal html structure if you wanted: `[`& ${Child} span`]` and by doing so create a an implicit dependency Child knows nothing about

rpominov: I general yes, but for `visibility` it seem fine to me as it just hides entire element without making any assumptions about its implementation

rpominov: But the solution with CSS variables is good too

kof: Yes, I am taking a stance on this case to have a single consistent solution we use everywhere. In some cases it won't be a huge deal in others it will. How do you consistently make sure at scale that everyone knows what to do?

It's hard, so if there is a solution that works well for every case, we should use it - that's how I think about it.

kof: Will push the same solution, just nicer written in a bit, stay tuned

kof: ok now its ready

rpominov (approved):
Great!

PR #250, opened 2022-08-31
fix: Github Actions

andrasbacsai:
There are 2 Github Actions flows:
1. Push changes to `package` dirs in `any branch` -> Only run `npm publish` with `--dry-run`, so contributors could check if the packages build correctly.
2. Push changes to `main` branch -> Run `npm publish` without `--dry-run`, production publish.

## Before requesting a review

- [ ] if the PR is WIP - use draft mode
- [ ] made a self-review
- [ ] added inline comments where things may be not obvious (the "why", not "what")
- [ ] what kind of review is needed?
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

- [ ] step by step interaction description and what is expected to happen

## Before merging

- [ ] tested locally and on preview environment (preview dev login: 5de6)
- [ ] updated [test cases](https://github.com/webstudio-is/webstudio-designer/blob/main/apps/designer/docs/test-cases.md) document
- [ ] added tests
- [ ] if any new env variables are added, added them to `.env.example` and the `designer/env-check.js` if mandatory

PR #253, opened 2022-09-02
Fix esm loader

andrasbacsai:
[This](https://github.com/nodejs/node/pull/42623) added a nasty feature (bug?) to the latest node versions (16/18), causing us to have this error in react-sdk:

```
@webstudio-is/react-sdk:build: Error [ERR_LOADER_CHAIN_INCOMPLETE]: "esbuild-node-loader 'resolve'" did not call the next hook in its chain and did not explicitly signal a short circuit. If this is intentional, include `shortCircuit: true` in the hook's return.
```

Now it [works](https://github.com/webstudio-is/webstudio-designer/runs/8156430960?check_suite_focus=true).

Github Actions will be fixed with [this](https://github.com/webstudio-is/webstudio-designer/pull/250).


packages/react-sdk/bin/generate-arg-types.ts
```diff
@@ -1,5 +1,5 @@
 import path from "path";
-import docgen from "react-docgen-typescript";
```
andrasbacsai, line 2: This requires as there is no default export, and tsx/node fails otherwise.

packages/react-sdk/package.json
```diff
@@ -7,8 +7,8 @@
   "scripts": {
     "dev": "tsup src/index.ts --format esm,cjs --watch --out-dir=lib",
     "build": "yarn build:mdn-data && yarn build:args && yarn build:lib",
-    "build:mdn-data": "node --experimental-loader esbuild-node-loader ./bin/mdn-data.ts ./src/css",
-    "build:args": "node --experimental-loader esbuild-node-loader ./bin/generate-arg-types.ts './src/components/*.tsx !./src/**/*.stories.tsx !./src/**/*.ws.tsx'",
+    "build:mdn-data": "tsx ./bin/mdn-data.ts ./src/css",
+    "build:args": "tsx ./bin/generate-arg-types.ts './src/components/*.tsx !./src/**/*.stories.tsx !./src/**/*.ws.tsx'",
```
kof, line 11: I am fine moving from  esbuild-node-loader  to tsx if it solves the problem same fast but better, but do you understand the underlying problem it is fixing for us?

also cc @rpominov since you started using  esbuild-node-loader and we are now migrating to tsx

kof: @andrasbacsai in this case  esbuild-node-loader can be removed from deps, right? I would also check other usages of it and replace everywhere

andrasbacsai: In case of @rpominov does not need it, I can remove it.

rpominov: I currently use `esbuild-register` for the migrations CLI in https://github.com/webstudio-is/webstudio-designer/pull/249

If there's reasons to use `txs` there, I can try to switch. But `esbuild-register` works fine for me, and I think it's faster than `txs`.

Also, maybe you should try replacing `node --experimental-loader esbuild-node-loader` with  `node -r esbuild-register`. I think this might be a better fix than switching to `tsx` if esbuild-register is indeed faster than tsx.

kof: tsx uses esbuild too, so there should be no perf difference, but please try and let us know, if there is, then we need to reconsider switching

kof (approved):
need to remove the previous package we use everywhere, but yes, lets go for it

PR #258, opened 2022-09-05
Style panel with css vars

kof:
This is not intended to be merged, but to serve as an example for the usage of css-vars.

Full implementation should happen in https://github.com/webstudio-is/webstudio-designer/pull/183 after we merge https://github.com/webstudio-is/webstudio-designer/pull/251 where the new css-vars package was introduced.


apps/designer/app/designer/features/style-panel/controls/select/select-control.tsx
```diff
@@ -1,8 +1,19 @@
+import { cssVars } from "@webstudio-is/css-vars";
 import { Grid, Select } from "@webstudio-is/design-system";
 import { getFinalValue } from "../../shared/get-final-value";
 import { PropertyName } from "../../shared/property-name";
 import type { ControlProps } from "../../style-sections";
 
+const gridTemplateColumnsVar = cssVars.define("grid-template-columns");
+
+export const selectControlCssVars = ({
```
kof, line 9: This represents a higher level API for CSS vars for SelectControl. Think of it as of react's context-like way to provide styles to a child component that is nested in the tree and we have no direct access to it over props.

apps/designer/app/designer/features/style-panel/controls/select/select-control.tsx
```diff
@@ -21,7 +32,15 @@ export const SelectControl = ({
   const setValue = setProperty(styleConfig.property);
 
   return (
-    <Grid columns={2}>
+    <Grid
+      columns={2}
+      css={{
+        gridTemplateColumns: cssVars.use(
```
kof, line 38: use the variable we defined

apps/designer/app/designer/features/style-panel/controls/select/select-control.tsx
```diff
@@ -21,7 +32,15 @@ export const SelectControl = ({
...
+          gridTemplateColumnsVar,
+          "repeat(2, 1fr)"
```
kof, line 40: use the fallback value, this is te same value that we get in Grid component from columns=2. I don't like the copy-paste here, since if we change columns to 3 this will not be very obvious.

Another option would be to introduce the CSS var straight in the Grid component and then reexport it from select control via that high level function

kof: removed columns prop, now we don't have duplication of the default value

apps/designer/app/designer/features/style-panel/sections/layout/layout.tsx
```diff
@@ -49,6 +49,7 @@ const layoutSectionFlexStyle = {
       },
     },
   },
+  ...selectControlCssVars({ gridTemplateColumns: "auto 1fr" }),
```
kof, line 52: This is how we explicitely define the template columns for the select control, without ever running into specificity or namespace conflicts

kof: anybody who reads this can trace it back to the select component and know exactly who is using this

kof:
Closing this since it was an example for CSS var for style panel

Issue #259, opened 2022-09-05
Custom migrations engine


rpominov:
The following is a description of a migrations engine that is already mostly implemented in #249 . This issue acts as a TODO for the remaining work and as a temporary documentation.

----

This is a drop-in replacement for Prisma's migration engine. It adds:
 - an ability to write a migration in TypeScript,
 - an ability to write a rollback for a migration.

A regular Prisma's migrations directory looks like this:

```
migrations/
  20220601192603_start/migration.sql
  20220608130959_adduser/migration.sql
```

With the new engine it will look something like this:

```
migrations/
  20220601192603_start/
    migration.sql
  20220608130959_adduser/
    migration.sql
    rollback.sql
  20220608130959_movedata/
    client/
      ...
    migration.ts
    rollback.ts
```

In a TypeScript migration script, you can use a Prisma client generated specifically for this migration. This client is frozen in time at the moment when you create a migration. So if at a later point a model will be removed from the main Prisma schema, in the migration you'll still have access to the old model. In other words, the Client in the migration corresponds to a state of the database where the preceding migrations have been applied, but the succeeding haven't.

```
Usage: migrations <command> [--dev]

Commands:
  create-schema <name> — Create a migration based on the changes in schema.prisma
  create-data <name>   — Create a migration that will change data rather than schema
  migrate              — Apply all pending migrations
  reset                — Clear the database and apply all migrations
  status               — Information about the state of the migrations
  resolve <applied|rolled-back> <name> — Mark a failed migration as applied or rolled-back

Arguments
  --dev                — Lets the CLI know that it's running in a development environment
```

TODO:

  - [x] Command for creating schema migration
  - [x] Command for creating data migration
  - [x] Command for applying migrations
  - [x] Command similar to `prisma migrate reset`
  - [x] Command similar to `prisma migrate status`
  - [x] Command similar to `prisma migrate resolve`
  - [ ] Rollbacks support (https://github.com/webstudio-is/webstudio-designer/issues/271)
  - [x] Documentation
  - [x] Test and document use-cases:
    - [x] Setting up a new development environment
    - [x] Creating and applying a schema migration
    - [x] Creating and applying a data migration
    - [x] Applying migrations added by other developers
    - [x] Applying migrations in production / preview deployments

PR #264, opened 2022-09-07
Migrations improvements

rpominov:
- #259

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/prisma-client/migrations-cli/commands.ts
```diff
@@ -1,21 +1,50 @@
...
   prismaMigrations.migrationsDir,
   "template.txt"
 );
 const lockfilePath = path.join(prismaMigrations.migrationsDir, "lockfile");
 
-const failWith = (message: string) => {
-  logger.error(message);
-  logger.error("");
-  process.exit(1);
+const confirm = async (
```
kof, line 17: there are higher level libs that allow you to do prompts in 1 loc

packages/prisma-client/migrations-cli/commands.ts
```diff
@@ -256,6 +276,29 @@ If you're sure no other process is running, please delete the lockfile:
...
+
+  ensureNoFailed(status);
+
+  if (status.pending.length === 0) {
+    logger.info("No pending migrations\n");
+    process.exit(0);
+  }
+
+  if (args.dev) {
+    logger.info("You're about to apply the following migration(s):");
```
kof, line 290: Btw I think this debug package would be even better than custom logger https://www.npmjs.com/package/debug

packages/prisma-client/migrations-cli/commands.ts
```diff
@@ -318,3 +371,24 @@ export const resolve = async ({
   logger.info(`Resolved ${migrationName} as rolled back`);
   logger.info("");
 };
+
+export const reset = async () => {
+  // just to make it read the migrations folder
+  // and fail early if something is wrong with the migration files
+  await getStatus();
+
+  if (args.dev === false) {
```
kof, line 380: I think instead of coupling dev and reset, we could make reset coupled with either prompt for confirmation or something like --really-sure-want-this

kof: coupling with dev solves the problem of protecting prod database in theory, but still kinda implicit logic, also I could potentially connect my local environment to prod (I shouldn't) and then it would just kill the db still

packages/prisma-client/migrations-cli/README.md
```diff
@@ -0,0 +1,50 @@
+## CLI Reference
+
+### `--dev`
+
+This flag can be added to any command. It informs the CLI that it’s used in a development environment, which slightly changes its behaviour:
+
+- automatically loads variables from `.env` files,
```
kof, line 7: instead of this, cli should just assume the .env file was already read and all variables are in process.env

packages/prisma-client/migrations-cli/README.md
```diff
@@ -0,0 +1,50 @@
...
+- uses a lockfile,
```
kof, line 8: which lockfile? for what?

packages/prisma-client/migrations-cli/README.md
```diff
@@ -0,0 +1,50 @@
...
+- asks for confirmation before applying migrations,
```
kof, line 9: this should be an interactive mode, not dev mode

packages/prisma-client/migrations-cli/README.md
```diff
@@ -0,0 +1,50 @@
...
+- allows resetting the database using the `reset` command.
```
kof, line 10: commented on this below, reset should have a different protection mechanism than implicit dev check

packages/prisma-client/migrations-cli/commands.ts
rpominov: I'm hesitant about adding dependencies usually. I worry about node_modules bloat and security.

https://twitter.com/DennisCode/status/1474972713861091329

packages/prisma-client/migrations-cli/commands.ts
rpominov: I'm trying to this similarly to Prisma https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-reset

Screenshot 2022-09-07 at 17 23 11

packages/prisma-client/migrations-cli/README.md
rpominov: I tried that. But didn't find an easy way to set this up.

packages/prisma-client/migrations-cli/README.md
rpominov: done

packages/prisma-client/migrations-cli/README.md
rpominov: Ok. Will resolve this comment as a duplicate

packages/prisma-client/migrations-cli/README.md
rpominov: >  for what?

To make sure you don't run two migration processes in parallel. I probably should say that in the doc.

> which lockfile?

Do you mean where it's located?

packages/prisma-client/migrations-cli/commands.ts
rpominov: The only reason we need a custom logger is to be able to print CLI messages while still using `no-console` rule. We don't need any other features.

packages/prisma-client/migrations-cli/commands.ts
kof: I am sure there are really nice well-made packages, I can recommend one I used if you want

packages/prisma-client/migrations-cli/commands.ts
kof: debug is not just console.log replacement, its about ability to control what will be logged using DEBUG env variable that you can pass before the call, e.g. I could do something like this DEBUG=migrations yarn migrations and it would log all the info for debugging that would be otherwise silent

or the other way around, if your output is not silent, I can silence it for CLI for example by DEBUG=*,-not_this

but again, just an idea

kof: btw this debug package is used by almost every popular package, you can try just putting DEBUG=prisma:engine

packages/prisma-client/migrations-cli/commands.ts
kof: damn ... it makes sense to be compatible with them, --force flag is a good touch for this, because otherwise you still need to confirm and can't really do that by accident

packages/prisma-client/migrations-cli/README.md
kof: lets chat

packages/prisma-client/migrations-cli/commands.ts
rpominov: I don't feel a need to use a dependency here to be honest

packages/prisma-client/migrations-cli/README.md
kof: yeah, I had no idea, need to explain that this tool generates an own lockfile to do XYZ, I thought you mean yarnlock or something

packages/prisma-client/migrations-cli/commands.ts
kof: I say it not only for the sake of reducing this line, but so that we have our set of dependencies we trust to build clis, this is not the only one we have, there is already one in https://github.com/webstudio-is/webstudio-cli

and will have more of them in the future ... makes sense when you look at it from global perspective?

packages/prisma-client/migrations-cli/commands.ts
rpominov: You can just run with `--dev` if you really want to do this

packages/prisma-client/migrations-cli/README.md
rpominov: Decided to keep as is in chat

packages/prisma-client/migrations-cli/commands.ts
rpominov: I think `debug` package is specifically for the debug output, basically a replacement for `console.debug()` but not for `console.info()`.

Overall I think it's a bit wasteful to focus on these little things right now. I'd just keep as is and forget about it unless there's a strong reason to change something.

packages/prisma-client/migrations-cli/commands.ts
rpominov: I think you right actually! It feels kinda out of place to enforce this using `--dev`. And Prisma doesn't seem to do any enforcement.

Will allow to run without `--dev`, user will still have to either confirm or run with `--skip-confirmation`.

rpominov:
@kof I didn't implement rollbacks yet. I think it should be a separate issue with `prio:2`

rpominov:
@kof Let's merge?

apps/designer/package.json
```diff
@@ -15,7 +15,8 @@
     "storybook:run": "start-storybook -p 6006",
     "storybook:build": "build-storybook",
     "migrations": "migrations --dev",
-    "migrations:deploy": "migrations migrate"
+    "migrate": "migrations migrate --dev --force",
+    "migrations:deploy": "migrations migrate --force"
```
kof, line 19: these two don't read very intuitively, migrate vs migrations:deploy maps to the same migrate command but differentiates in. --dev, anything we can do about it? Or is it just me not seeing full picture?

packages/prisma-client/migrations-cli/README.md
```diff
@@ -0,0 +1,151 @@
+## Intro
+
+This is a drop-in replacement for Prisma's migration engine. It adds:
+
+- an ability to write a migration in TypeScript,
```
kof, line 5: data migration, because prisma calls schema migration "migration"

packages/prisma-client/migrations-cli/README.md
```diff
@@ -0,0 +1,151 @@
...
+- an ability to write a rollback for a migration (\* to be implemented).
```
kof, line 6:
```suggestion
- an ability to write a rollback for data migration (to be implemented).
```

packages/prisma-client/migrations-cli/README.md
kof:
```suggestion
- an ability to write a data migration in TypeScript,
```

packages/prisma-client/migrations-cli/cli.ts
```diff
@@ -23,87 +18,92 @@ Commands:
...
+  --force  — Skips the confirmation prompt when running a dangerous command
 `;
 
 const main = async () => {
-  if (args.dev) {
-    dotenv.config();
-  }
-
-  const command = args._[0];
+  try {
```
kof, line 25: nit: could put this try/catch around main() call so the entire main function has less indentation - makes it easier to read

kof (approved):
Looking great, added some comments

packages/prisma-client/migrations-cli/README.md
rpominov: We can support rollbacks for schema migrations too. Or do you think we shouldn't?

apps/designer/package.json
rpominov: Yeah. It's kinda awkward. One is to allow to apply migrations from root for developers. Another is to be called during deployment by Vercel.

Would this be better?

```
"migrate": "migrations migrate  --force --dev",
"migrate:deploy": "migrations migrate --force"
```

packages/prisma-client/migrations-cli/README.md
kof: just want to be explicit what's meant, in that case data and schema migration

apps/designer/package.json
kof: what's the difference when we do it at vercel?

rpominov: ### `--dev`

This flag can be added to any command. It informs the CLI that it’s used in a development environment, which slightly changes its behaviour:

- automatically loads variables from `.env` files,
- uses a lockfile to avoid running more than one migration process at the same time.

rpominov: I was thinking about splitting `--dev` into two flags `--load-env` and `--use-lockfile`. But we'll still need these two separate scrips in `pakcage.json`

kof: Oh I read what --dev means, what I don't quite understand is how migrate:deploy is in the end different than "migrate"

kof: Maybe we don't actually need 2 separate scripts that a semantically slightly confusing if we for example:
- do dev mode by default and prod mode has to be passed via either NODE_ENV which we already have or --prod
- or separate flags that would turn it into prod mode


In the end we wouldn't need 2 scripts, it would be same script, just the CLI on vercel would have the right env/flags

rpominov: Not sure it's a good idea to rely on `NODE_ENV`. It's not always set up correctly.

Switching from `--dev` to `--prod` and calling with `--prod` at Vercel sounds good to me.

kof: cool, lets do this

kof: still don't quite understand how the prod flag will have a different outcome on the migrations :)

rpominov: After thinking a bit more, I think it's a bad idea to call with `--prod` at Vercel. Two reasons:

1. That locks us with `--prod`, and would not allow us to switch to `--load-env / --use-lockfile` later.
2. This makes Vercel know about our internal details of migrations.

I'd stick with two scripts in `package.json`. It's most correct in my opinion.

PR #268, opened 2022-09-11
refactor text-control to use combobox

natlusyht:
Based on https://github.com/webstudio-is/webstudio-designer/issues/52#issuecomment-1242118347 what is addressed by this PR is marked/listed below. Ones not marked are earmarked for having their own singular focused PR's.

- [x] do not send an update on blur if value has not changed.
- [x] on keyword values typing or keyboard up/down arrow should open keyword dropdown.
- [x] escape key when dropdown is open should close dropdown.
- [x] unify with select control dropdown style + refinements to dropdown style.
- [x] use combobox component.
- [x] scrolling menu bug https://share.descript.com/view/sG6IPs5kxOU
- [ ] refactor all those nested selectors that reach into components to  CSS variables, like in this example implementation https://github.com/webstudio-is/webstudio-designer/pull/258

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

Screenshot 2022-09-11 at 9 46 14 PM


packages/design-system/src/stitches.config.ts
```diff
@@ -271,6 +271,9 @@ export const {
       4: "400",
       max: "999",
     },
+    columns: {
+      1: "40% 60%",
```
kof, line 275: I am not quite sold about the reusability of this token, it sounds like a specific case, am I wrong? If yes please tell me how it can be reused everywhere

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -1,261 +1,201 @@
...
-  const isCurrentBreakpoint = useIsFromCurrentBreakpoint(styleConfig.property);
-  const textFieldRef = useRef<HTMLInputElement>(null);
-
   const value = getFinalValue({
     currentStyle,
     inheritedStyle,
     property: styleConfig.property,
   });
 
+  const getInputNode = (node: HTMLElement): HTMLInputElement => {
```
kof, line 37: why do you want to run a query selector instead of getting a ref React way???

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -1,261 +1,201 @@
...
+    return node
+      .closest("[data-control]")
+      ?.querySelector("input") as HTMLInputElement;
+  };
+  const isExpanded = (node: HTMLElement) => {
```
kof, line 42: same

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -1,261 +1,201 @@
...
+      if (Object(node)[id]) return;
+      Object(node)[id] = numericScrubControl(node, {
+        initialValue: value.value,
+        onValueChange: (event) => {
+          getInputNode(node).value = String(event.value);
+          handleChange("unit", event.value, true);
+        },
+      });
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kof, line 63: when I see this, something is definitely fishy ....

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -1,261 +1,201 @@
...
+                }}
+                onFocus={(event: FocusEvent<HTMLInputElement>) => {
+                  event.target.select();
+                }}
+                onBlur={(event) => {
+                  if (event.target.value !== event.target.getAttribute("value"))
+                    setValue(event.target.value);
+                }}
+                onKeyDown={(event) => {
+                  const target = event.target as HTMLInputElement;
```
kof, line 131: probably just need to define the event type instead of `as`

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -1,261 +1,201 @@
...
+                  if (event.code === "ArrowDown")
+                    currentValue = currentValue - currentDelta;
+                  const currentValueAsString =
+                    currentValue % 1
+                      ? currentValue.toPrecision(
+                          Math.abs(currentValue).toString().indexOf(".") + 2
+                        )
+                      : String(currentValue);
+                  target.value = currentValueAsString;
+                  setValue(currentValueAsString, { isEphemeral: true });
```
kof, line 160: some of this stuff could be a separate function to make it all more readable

apps/designer/app/designer/features/style-panel/sections/layout/layout.tsx
```diff
@@ -49,6 +45,18 @@ const layoutSectionFlexStyle = {
       },
     },
   },
+  // @todo justifyItems icons are using justifycontent icons atm
+  "& [data-property=justifyItems]": {
```
kof, line 49: why are you doubling down on nested selectors if we already agreed to migrate to css-vars

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -1,39 +1,98 @@
...
+          ...(isCurrentBreakpoint && {
+            bc: "$colors$blue4",
+            color: "$colors$blue11",
+            outline: "none",
+            "&:hover,&:active,&:focus": {
+              bc: "$colors$blue4",
+              color: "$colors$blue11",
+            },
+          }),
+          [`& ~ ${TextField}`]: {
```
kof, line 88: this shouldn't be necessary, it basically introduces knowledge of a TextField next to a property name component, breaks encapsulation again

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -1,261 +1,201 @@
...
+                {...toggleProps}
+                css={{
+                  visibility: items.length ? "visible" : "hidden",
+                  position: "absolute",
+                  right: "1px",
+                  top: "1px",
+                  width: "auto",
+                  height: "calc(100% - 2px)",
+                  px: "calc($1 / 2)",
+                  borderRadius: "$1",
```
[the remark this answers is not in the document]
okonet: I can extend the Input to support this kind of things easily.

apps/designer/app/designer/features/style-panel/sections/layout/layout.tsx
```diff
@@ -49,6 +45,18 @@ const layoutSectionFlexStyle = {
...
+    display: "none",
+  },
+  "& [data-property=columnGap]": {
+    gridArea: "6 / 1 / -1 / 7",
+    paddingRight: "10px",
+  },
+  "& [data-property=rowGap]": {
+    gridArea: "6 / 7 / -1 / -1",
+    paddingLeft: "10px",
+  },
```
okonet, line 59: I have no idea what's going on here. Maybe some comments?

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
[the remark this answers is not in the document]
kof: Yeah, I think something like this has to happen, because we can't be doing this custom CSS all over

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
okonet: Why is this needed?

packages/design-system/src/components/combobox.tsx
```diff
@@ -52,18 +52,36 @@ export const ComboboxTextField = forwardRef(ComboboxTextFieldBase);
...
-  padding: 0,
+const Listbox = styled("ul", {
   margin: 0,
+  padding: 0,
   overflow: "auto",
-  // @todo need some non-hardcoded value
-  maxHeight: 400,
-  minWidth: 230,
+  minWidth: 124,
+  maxHeight: 190,
```
okonet, line 60: Todo is removed but it wasn't done since we still use hard coded values. Also new values will probably break other parts of UI.

packages/design-system/src/components/icon-button.tsx
```diff
@@ -23,7 +23,7 @@ export const IconButton = styled("button", {
   "&::before, &::after": {
     boxSizing: "border-box",
   },
-  "&:hover, &:active": {
+  "&:focus, &:hover, &:active": {
```
okonet, line 26: I'd rather duplicate the rule in the next selector. Now it is much harder to understand what are the resulting styles going to be.

apps/designer/app/designer/features/style-panel/sections/layout/layout.tsx
```diff
@@ -1,28 +1,21 @@
...
-    "& > :first-child": {
-      gridTemplateColumns: "auto 1fr",
-      gap: "$space$2",
-      fontWeight: "500",
-      "& > :last-child": {
-        width: "fit-content",
-      },
-    },
-  },
+  // "& > [data-type=iconbuttonwithmenu]": {
```
kof, line 13: can this be removed?

apps/designer/app/designer/features/style-panel/sections/layout/layout.tsx
```diff
@@ -37,43 +30,102 @@ const layoutSectionFlexStyle = {
...
-    "& > :first-child": {
-      gridTemplateColumns: "repeat(1, 1fr)",
-      "& > :first-child": {
-        display: "none",
-      },
-    },
-  },
+};
+
+// @todo shouldn't need this
```
kof, line 35: any hint on what's stopping us to remove it?

apps/designer/app/designer/features/style-panel/sections/layout/layout.tsx
```diff
@@ -37,43 +30,102 @@ const layoutSectionFlexStyle = {
...
-        }}
-        batchUpdate={batchUpdate}
-      />
-      <FlexGrid
-        name="grid"
-        currentStyle={currentStyle}
-        batchUpdate={batchUpdate}
-      />
+      <Box css={gridStyle}>
+        {" "}
```
kof, line 89: probably accidental

apps/designer/app/designer/features/style-panel/sections/layout/shared/lock.tsx
```diff
@@ -39,13 +37,23 @@ export const Lock = ({
...
+          width: "100%",
+          justifyContent: "center",
+        }}
       >
-        {isPaired ? <LockCloseIcon /> : <LockOpenIcon />}
-      </IconButton>
+        <IconButton
+          css={{
+            "&:focus": {
+              outline: "none",
```
kof, line 49: we should fix it in the design system

apps/designer/app/designer/features/style-panel/sections/spacing/spacing.tsx
```diff
@@ -62,6 +62,7 @@ const styles = {
     display: "block",
     background: "transparent",
     color: "$hiContrast",
+    zIndex: 99,
```
kof, line 65: I fixed this same weird problem on main with zIndex: 0, I haven't fully understood why it was a problem in the first place though, its weird, there was nothing that indicates the need for z-index

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -1,39 +1,98 @@
...
+            outline: "none",
+          },
+          ...(isCurrentBreakpoint && {
+            bc: "$colors$blue4",
+            color: "$colors$blue11",
+            outline: "none",
+            "&:hover,&:active,&:focus": {
+              bc: "$colors$blue4",
+              color: "$colors$blue11",
+            },
```
kof, line 86: things like outline should go into icon button, the nudges with 1px nudges are questionable, lets figure out why we need those, there must be another way

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: I think @okonet just created a PR for a text field that allows rendering anything on the left or the right side of the input, so that TextField is in charge of it's layout and we don't have to position it from the outside.

packages/design-system/src/components/combobox.tsx
kof: forgotten? what can we do to avoid hardcoding those width/height values? maybe those should be on the consumer side?

packages/design-system/src/components/combobox.tsx
```diff
@@ -203,12 +222,14 @@ export const Combobox = <Item extends BaseItem>({
         </PopperAnchor>
         {renderPopperContent({
           style: { zIndex: 1 },
+          align: "end",
+          sideOffset: 8,
```
kof, line 226: this seems like an adjustment for a specific use case, cc @okonet

packages/design-system/src/components/combobox.tsx
```diff
@@ -134,6 +153,7 @@ type ComboboxProps<Item> = {
 export const Combobox = <Item extends BaseItem>({
   items,
   value,
+  selected = value,
```
kof, line 156: so you want value and selected value to be 2 different things? how does this fit with the API redesign @okonet ?

okonet: It doesn't. Why is this needed?

kof: so representation of the value inside the input.value is different than the value inside the items. It seems like a case for the itemToString function that would return the value without the unit, so it can be the same

okonet: Generally we should avoid specific use cases to introduce breaking changes in reusable components. 

I'm still not sure what problem you're trying to solve by this so maybe a small example in a Storybook so I could help you?

okonet: I believe this is a misused component case here. Unit should be a Select with a customized trigger component (IconButton) inside a `suffix` of the TextField. This is pretty much how it works now.

kof: So it would be a <Combobox suffix={<Select />} />, where combobox always receives values without units and Select always receives units without values

apps/designer/app/auth/login.tsx
```diff
@@ -44,59 +46,73 @@ export const Login = ({ errorMessage }: { errorMessage: string }) => {
...
-                </Flex>
-              </LoginButton>
-            </Form>
-            {data.isDevLogin && (
-              <>
-                {isDevLoginOpen ? (
-                  <Form action="/auth/dev" method="post">
+          <Flex direction="column" gap="4" css={{ width: 300 }}>
+            {errorMessage.length ? (
+              <__DEPRECATED__Text css={{ textAlign: "center" }} variant="red">
```
kof, line 53: please use Text right away

apps/designer/app/auth/login.tsx
```diff
@@ -44,59 +46,73 @@ export const Login = ({ errorMessage }: { errorMessage: string }) => {
...
+                      <GithubIcon />
+                    </Flex>
+                    Login with GitHub
+                  </Flex>
+                </LoginButton>
+              </Form>
+              <Form action="/auth/google" method="post">
+                <LoginButton enabled={data.isGoogleEnabled}>
+                  <Flex gap="2" align="center">
+                    <Flex css={{ width: 20, height: 20 }}>
```
kof, line 71: additional div to hold the dimensions looks non-ideal, what are some alternatives?

kof: also where are those 20px coming from?

kof:
@natlusyht some lint errors are there

kof:
- [ ] property name rendered twice 
Screenshot 2022-09-24 at 22 32 08

kof:
- [ ] scrub isn't working any more? even scrub cursor is not showing

kof:
@natlusyht 
1. take a value where you have  some non numeric
2. put it 0

kof:
@natlusyht that was in flex control, others, lets say Size > width don't have scrub either, is that expected?

kof:
I will let @okonet approve this one

apps/designer/app/designer/features/breakpoints/breakpoints.tsx
```diff
@@ -44,17 +43,17 @@ const BreakpointSelectorItem = ({
 }: BreakpointSelectorItemProps) => {
   return (
     <Flex align="center" justify="between" gap="3" css={{ flexGrow: 1 }}>
-      <__DEPRECATED__Text size="1" css={{ flexGrow: 1, color: "inherit" }}>
+      <Text size="1" css={{ flexGrow: 1, color: "inherit" }}>
```
kof, line 46: how did it pass tsc? size prop doesn't exist on new Text component

kof: @okonet should Text have flexGrow: 1 and color inherit by default baked in?

kof: @okonet should we add a variant for color="inherit"? <Text color="inherit" /> in addition to those others

kof: Actually regarding color there might be something wrong, what is the color that is used there? Do existing color variants not work?

okonet: It's hard to get the default for Text right. Sometimes I tend to use `color: inherit` as a default value so it can be composed easily, so it could make sense. On the other hand, what's not working with the color defined in the variant? Why is it needs to be overridden in this case?

As for flex, what I usually prefer is to make components take all available space and handle the layout outside of them. Would it work here?

I also think that Text should not be `block` or `flex` but instead `inline-block` or `inline-flex` since you have `Text` nested with `Text`. Right now it's not possible to do afaik.

apps/designer/app/auth/login.tsx
okonet: I added this since it would otherwise collapse the icon as far as I remember. I think the way how icon is made now needs reworking. I'd like Icon to have `size` prop that would define the size internally + currently Icon is not accessible so we probably need add a `label` prop so it can be used by assistive tech. Let's do a follow up for this?

kof: sure, please create an issue, to have a todo and so we can do it before we add more workarounds

packages/design-system/src/components/combobox.tsx
```diff
@@ -158,23 +207,22 @@ export const Combobox = <Item extends BaseItem>({
     getItemProps,
     selectedItem,
   } = useCombobox({
+    items: foundItems as Item[],
+    selectedItem: value ?? null, // Avoid downshift warning about switching controlled mode
+    /* @todo breaks input
```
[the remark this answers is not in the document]
okonet: What exactly does it break? `stateReducer` is responsible for the behavior, so competing it out won't solve anything. Please revert this and report the issue you're having so I can take a look at it.

packages/design-system/src/components/combobox.tsx
okonet: Yeah I'm also wondering what's this for...

packages/design-system/src/components/dropdown-menu.tsx
```diff
@@ -9,7 +9,8 @@ import { panelStyles } from "./panel";
 
 export const DropdownMenu = DropdownMenuPrimitive.Root;
 export const DropdownMenuArrow = styled(DropdownMenuPrimitive.Arrow, {
-  fill: "$panel",
+  fill: "$colors$slate4",
+  stroke: "$colors$slate1",
```
okonet, line 12: We should not use color tokens directly but rather semantic alternatives. What are you trying to express here?

packages/design-system/src/components/menu.tsx
```diff
@@ -26,7 +25,7 @@ export const itemCss = css(baseItemCss, {
   color: "$hiContrast",
   "&:focus, &[data-found]": {
     outline: "none",
-    backgroundColor: "$blue9",
+    backgroundColor: "$blue10",
```
okonet, line 28: Same here, please let's stop using color tokens directly. Instead we should stick to semantic tokens like `background`, `text`, etc.

packages/design-system/src/components/panel.tsx
```diff
@@ -1,10 +1,11 @@
...
 
 export const panelStyles = css({
-  backgroundColor: "$panel",
-  borderRadius: "$3",
+  overflow: "auto",
+  backgroundColor: "$colors$slate4",
+  borderRadius: "$1",
   boxShadow:
-    "$colors$shadowLight 0px 10px 38px -10px, $colors$shadowDark 0px 10px 20px -15px",
+    "0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15), inset 0 0 1px 1px $colors$slate1, 0 0 0 1px $colors$slate8",
```
okonet, line 8: Can we make it a token too?

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -18,16 +18,19 @@ export type NumericScrubDirection = "horizontal" | "vertical";
 
 export type NumericScrubValue = number;
 
+export type NumericScrubCallback = (event: {
+  target: HTMLElement;
+  value: NumericScrubValue;
+  preventDefault: () => void;
+}) => void;
```
okonet, line 25: Is this a custom event? Signature looks like it's `ChangeEvent`.

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -45,6 +48,7 @@ export const numericScrubControl = (
     maxValue = Number.MAX_SAFE_INTEGER,
     initialValue = 0,
     direction = "horizontal",
+    onValueInput = () => null,
```
okonet, line 51: What's the difference between those two?

packages/design-system/src/components/select.tsx
```diff
@@ -49,6 +51,9 @@ const StyledTrigger = styled(SelectPrimitive.SelectTrigger, {
 const StyledIcon = styled(SelectPrimitive.Icon, {
   display: "inline-flex",
   alignItems: "center",
+  justifyContent: "center",
+  height: "100%",
+  padding: "calc($space$1 / 2)",
```
okonet, line 56: If our spacing scale doesn't work, let's add a `todo` and fix it instead of adding more of `calc`.

packages/design-system/src/components/text-field.tsx
```diff
@@ -61,8 +61,8 @@ const InputBase = styled("input", {
 const TextFieldBase = styled("div", {
   // Custom
   display: "flex",
-  gap: "$1",
   backgroundColor: "$loContrast",
+  py: 1, // value of insert boxShadow
```
okonet, line 65: I don't understand this comment. Can it be re-written?

apps/designer/app/auth/login.tsx
```diff
@@ -50,9 +49,9 @@ export const Login = ({ errorMessage }: { errorMessage: string }) => {
           <Heading size="2">Login</Heading>
           <Flex direction="column" gap="4" css={{ width: 300 }}>
             {errorMessage.length ? (
-              <__DEPRECATED__Text css={{ textAlign: "center" }} variant="red">
+              <Text css={{ textAlign: "center", color: "$red10" }}>
```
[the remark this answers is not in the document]
kof: it has a few colors built-in: contrast, loContrast, hint. What is missing, should we allow an arbitrary color as well?

kof:
This is merged in https://github.com/webstudio-is/webstudio-designer/pull/313 and will continue there

Issue #272, opened 2022-09-12
Reuse code in edge deployments


andrasbacsai:
Currently, shared things are copy-pasted from the designer to the edge deployments. We need to fix this.

A possible solution is to publish the shared parts to an npm package, so they could be reused in several places.

andrasbacsai:
@kof Wdyt?

kof:
Right, we need to figure out how to meaningfully create those packages, so that they make sense for all use cases we have soon, which are
1. apps: designer, dashboard, auth, canvas
2. workers
3. cli

kof:
@andrasbacsai this one can be closed?

PR #277, opened 2022-09-13
Project package

andrasbacsai:
Related issue #272 

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/routes/rest/tree.$projectId.ts
```diff
@@ -1,7 +1,6 @@
 import { type LoaderFunction } from "@remix-run/node";
 import type { Tree } from "@webstudio-is/react-sdk";
-import * as db from "~/shared/db";
-
+import * as db from "@webstudio-is/project";
```
kof, line 3: db namespace now makes less sense because its kinda only project part, not all of it

andrasbacsai: Yes, it is just a temporary name. It won't stay like this.

andrasbacsai: Do you have any suggestion on the name? 
`projectdomain`?

E9s6ooLXsAAKmUP

kof: maybe just destructure the variables we use directly, without using this pattern `* as db` pattern, which is bad for tree-shaking anyways (I guess)

kof:
@okonet @rpominov @thysultan what we are trying to do here is to move things to a package that are going to be reused in other places , like I mentioned here https://github.com/webstudio-is/webstudio-designer/issues/272#issuecomment-1243502139

I don't know if "project" as a package is the best way to split it up. Ideally we would split up by domain/entity, project is also an entity, though in this case we are moving along with it a lot of related functionality that is being required from those other linked areas.

I guess we can split the project later if we need to, but right now its not quite obvious. Just want to make sure you are all on-board with this process.

apps/designer/app/designer/features/breakpoints/breakpoints.tsx
```diff
@@ -26,6 +25,7 @@ import {
 } from "./use-subscribe-shortcuts";
 import { ConfirmationDialog } from "./confirmation-dialog";
 import { useBreakpoints } from "~/shared/nano-states";
+import { sort } from "@webstudio-is/project";
```
kof, line 28: this is kinda too high level name to be exported there like this, because the name is too generic, we would probably need something like `breakpoint.sort()` or `breakpointUtils.sort()`

apps/designer/app/routes/rest/tree.$projectId.ts
kof: alternative is to import {db} from 'package', so that db is a namespace, not an import *

packages/project/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export * as project from "./project";
+export * as breakpoints from "./breakpoints";
+export * as tree from "./tree";
+export * as props from "./props";
+
+export * from "./shared/breakpoints";
+export * from "./shared/tree-utils";
```
kof, line 7: ok, here is the suggestion, lets export it something like this:

```suggestion
import * as project from "./project";
import * as breakpoints from "./breakpoints";
import * as tree from "./tree";
import * as props from "./props";

export const db = {project, breakpoints, tree, props}

export breakpoints from "./shared/breakpoints";
export tree from "./shared/tree-utils";
```

packages/project/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export * as project from "./project";
```
kof, line 1: and in addition I would move these to src/db folder since these are all db specific functions

okonet:
I'm still confused with the name of the package. Can we have multiple packages that have appropriate functionality or maybe just "utils"?

kof:
> I'm still confused with the name of the package. Can we have multiple packages that have appropriate functionality or maybe just "utils"?

The thing I am afraid of is that we would have to create 4-5 packages right now and I am not sure its worth it as of today. We will probably endup doing that long-term, but ...

Project as a package name is kinda domain of the project specific logic/data, so it would still be a good name even if we split it up into multiple packages ....

okonet:
Okay I see. It's fine with me then.

andrasbacsai:
Some futher explanation: a few parts needs to be in an npm package to be shareable with the edge version of Webstudio and to be in-sync, for example `canvas`.

apps/designer/app/designer/features/breakpoints/breakpoints.tsx
andrasbacsai: I moved all utils under `utils` export, so `utils.breakpoints.sort()` for example.

rpominov:
Oh, not a good time to do this. There will be a ton of conflicts with https://github.com/webstudio-is/webstudio-designer/pull/231 😅 

In general I understand the motivation about splitting by features. But I still think it's a bad idea to use NPM packages for technical reasons. I think we'll have a lot of problems like "how do I build this" / "how do I import that" / etc. And this only going to get worse. At least this is what my intuition tells, could be wrong.

I would split using directories rather than NPM packages, especially when we have to publish them in order to use.

There are other ways to reuse public code in a private repo. Like [submodules](https://github.blog/2016-02-01-working-with-submodules/) . I know they're clunky, but less than NPM packages, for our use case.

kof:
> Oh, not a good time to do this. There will be a ton of conflicts with https://github.com/webstudio-is/webstudio-designer/pull/231 😅

yeah, its unfortunate, but good thing is this PR doesn't change any logic, it only moves things

> There are other ways to reuse public code in a private repo. Like [submodules](https://github.blog/2016-02-01-working-with-submodules/) . I know they're clunky, but less than NPM packages, for our use case.

Ergonomics of submodules are also quite bad, I think in the end it all depends on tooling, so the question is if your feeling is based on the pre-turborepo and pre-esbuild time 

Another criteria is always the frequency of change. Something that frequently changes together, should ideally stay together, which is why when splitting, its important to keep functionality that changes together in the same place, to reduce the situations where you need cross-package changes. Overall creating a cross-package change in one PR right now is easy, it gets harder when you have to publish to use it in a separate repository, but that's what we have to do with the non-free part of the product. I am open to specific ideas though

apps/designer/app/routes/rest/project/clone.$domain.ts
```diff
@@ -1,5 +1,7 @@
 import { type LoaderFunction, redirect } from "@remix-run/node";
 import type { User } from "@webstudio-is/prisma-client";
+import { db as database } from "@webstudio-is/project";
+import type { Project } from "@webstudio-is/project";
 import * as db from "~/shared/db";
```
kof, line 5: these 2 db semantics are problem, one could be projectDb, another could be the userDb

rpominov:
> so the question is if your feeling is based on the pre-turborepo and pre-esbuild time

Kinda in between.

Not a strong opinion, and I don't have arguments other than "I have a bad feeling about this". If no one shares this feeling, let's go on.

kof:
> Not a strong opinion, and I don't have arguments other than "I have a bad feeling about this". If no one shares this feeling, let's go on.

I have bad feelings about typescript, prisma, CSS, HTML and a lot of things that are considered the best on the market right now haha, the question is if there is an alternative path and what are the tradeoffs :)

kof (approved):
minor nits there but generally I would merge

andrasbacsai:
I had several troubles with `submodules` in the past. I would stick with NPM modules for now and later on we can migrate if things change.

If this is merged, I’m happy to help you to fix all conflicts. 🙂

PR #284, opened 2022-09-16
Wait for designerReady event on the canvas before publishing data

kof:
It is possible that the js of designer is loaded later than the js of canvas, because everything is server-rendered

we just ran into a case where disabling cache has hit the bug

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [x] added tests


rpominov (commented):
Looks good, but I don't see how the new event delays the `loadBreakpoints` event

kof:
@rpominov that event is published from useHandleBreakpoints inside DesignMode component which isn't rendered until designer ready event is received, so it will publish only after that event first time, before that canvas will render elements tree but no DesignMode

rpominov (approved):

PR #292, opened 2022-09-20
[Design System] TextField: Add prefix and suffix props

okonet:
This PR introduces `prefix` and `suffix` and refactors `Combobox` to use them. It also makes the behavior of the Combobox a bit more smoother.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

- [ ] Check out stories for `TextField`
- [ ] Open Combobox first story
- [ ] Select an item using mouse (click on the dropdown icon)
- [ ] Press <kbd>Esc</kbd> to reset the state of the combobox
- [ ] Open the dropdown using <kbd>↓</kdb>
- [ ] Select one of the items
- [ ] Press <kbd>Backspace</kbd> to remove the selection. It should clear the input field right away.
- [ ] Start typing `or` to get filtered items and select `Orange` from the list

## Before merging

- [ ] added tests


packages/design-system/src/components/text-field.tsx
```diff
@@ -129,19 +141,72 @@ export const TextField = styled("input", {
...
+            alignItems: "center",
+            flexShrink: 0,
+            order: 0,
+          }}
+        >
+          {prefix}
+        </Flex>
+      )}
+
+      <InputBase ref={inputRef} disabled={disabled} {...textFieldProps} />
```
okonet, line 195: @Andarist can you please take a look at why those two refs types are not compatible? Thanks!

packages/design-system/src/components/text-field.tsx
```diff
@@ -32,28 +35,43 @@ export const TextField = styled("input", {
     fontFamily: "$untitled",
     color: "$hiContrast",
   },
+  "&::placeholder": {
+    color: "$slate9",
+  },
+
+  "&:disabled": {
+    color: "$slate8",
+    // @todo: This is not working now bue to cursor variant. Is it needed?
```
kof, line 44: not sure what you mean

kof (approved):
looking good!

kof:
Didn't test yet, let me know what to test

packages/design-system/src/components/text-field.tsx
```diff
@@ -129,19 +141,72 @@ export const TextField = styled("input", {
...
-    },
   },
   defaultVariants: {
     size: "1",
   },
 });
+
+type TextFieldProps = Omit<
+  React.ComponentProps<typeof TextFieldBase> & React.ComponentProps<"input">,
+  "prefix"
```
Andarist, line 152: An alternative fix for [the one posted below](https://github.com/webstudio-is/webstudio-designer/pull/292/files#r975490001). Perhaps this one is even better

```suggestion
  "prefix" | "ref"
```

packages/design-system/src/components/text-field.tsx
okonet: I removed the `cursor` variant since it wasn't used so I think this can be ignored.

okonet:
@natlusyht I resolved most of the regressions and simplified quite a lot of code in terms of custom styles. But there is quite a lot of code which I wouldn't touch since I don't really understand the implications. For example, sizing of `IconButtonWithMenu` etc. Could you please take a look at those?

taylornowotny:
Looking good so far!

My feedback is just please to style it consistent with the design.

taylornowotny:
> @taylornowotny please be more specific.

Please consult the design docs. The size, icons, colors, and text styles are all different.

okonet:
I didn't update the look of anything. This PR scope is not about visuals. Either way it would be helpful to get a link to the figma illustrating the difference although I don't think it should be part of this PR.

kof:
Here is the biggest visual regression I see, neither of them are conform to figma's design system, but the current PR makes it look totally broken, so what can we do? Do we need an intermediate step?

Screenshot 2022-09-22 at 13 39 11
Screenshot 2022-09-22 at 13 40 20

apps/designer/app/auth/login.tsx
```diff
@@ -44,59 +46,73 @@ export const Login = ({ errorMessage }: { errorMessage: string }) => {
...
-                </Flex>
-              </LoginButton>
-            </Form>
-            {data.isDevLogin && (
-              <>
-                {isDevLoginOpen ? (
-                  <Form action="/auth/dev" method="post">
+          <Flex direction="column" gap="4" css={{ width: 300 }}>
+            {errorMessage.length ? (
+              <__DEPRECATED__Text css={{ textAlign: "center" }} variant="red">
```
kof, line 53: could switch rightaway to Text, its easy

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -51,11 +52,12 @@ const Property = ({
 
   return (
     <Grid
-      gap="1"
+      gap={1}
```
kof, line 55: its an object key in the end, so strings are actually more accurate, but I don't care ...

packages/design-system/src/components/combobox.tsx
```diff
@@ -117,11 +113,12 @@ export const ComboboxPopperContent = PopperContent;
...
   label?: string;
-  items: Array<Item>;
+  placeholder?: string;
+  items: ReadonlyArray<Item>;
   value?: Item;
-  onItemSelect?: (value: Item) => void;
+  onItemSelect?: (value?: Item) => void;
   onItemHighlight?: (value?: Item) => void;
-  itemToString?: (item: Item | null) => string;
+  itemToString?: (item?: Item | null) => string;
```
kof, line 121: do you really want to support both: null and undefined? Usually I am trying to support undefined only, because it works as an optional value, supporting both is usually not great from logic perspective

okonet: This is mostly done to align with downshift types. I'm not a fan of both either and tried to make it simpler but couldn't succeed. Give it a try if you are fancy.

kof (changes requested):
some minor code comments, mostly we need to decide what to do with visually broken stuff 
@taylornowotny this PR is not about making it conform with DS, @okonet will investigate what we need to change globally to make it conform,

okonet:
The visual regressions are expected since current implementation is trying to adjust component's implementation to match the figma design system using custom CSS. I removed as much of it as possible since the proper strategy should be matching components in code and figma and not in the app code. I'm not sure how we could proceed here. We can expand the scope of the PR and continue here or create another one with visual changes based on this one. Both are fine with me but I believe that matching work can take quite a while.

PR #293, opened 2022-09-20
Pages routes

rpominov:
Going to update Remix routes to be aware of pages in this PR. No visible UI changes for the user still. 

Related issue: #54

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/routes/canvas/$projectId.$pageId.tsx
```diff
@@ -0,0 +1,2 @@
+export * from "./$projectId";
+export { default } from "./$projectId";
```
kof, line 2: I have done something similar in the past, but I think this is wrong, because we are not supposed to be importing these manually. I wonder if pageId can be just optional?

apps/designer/app/routes/canvas/$projectId.tsx
```diff
@@ -1,40 +1,33 @@
...
+    if (params.projectId === undefined) {
+      throw new Error("Missing projectId");
     }
-  }
 
-  return { errors: "Unexpected error", env };
+    return loadCanvasData(params.projectId, params.pageId);
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    sentryException({ message });
```
kof, line 28: maybe we should make sentry take error instances directly too? Seems like a pattern we will need all the time

apps/designer/app/routes/designer/$projectId.$pageId.tsx
```diff
@@ -0,0 +1,2 @@
+export * from "./$projectId";
+export { default } from "./$projectId";
```
kof, line 2: same question as with the other one

apps/designer/app/routes/canvas/$projectId.$pageId.tsx
rpominov: Remix doesn't seem to support optional dynamic segments. This solution seems fine to me, but I can change it if there're other way to achieve this.

apps/designer/app/routes/canvas/$projectId.tsx
rpominov: Yes, and it will be more correct to pass the error object as is to the Sentry API. It's designed to take error object not sure why we pass just the message

apps/designer/app/routes/designer/$projectId.$pageId.tsx
rpominov: Same answer :)

apps/designer/app/routes/canvas/$projectId.$pageId.tsx
kof: Indeed, there is an issue and a discussion but no implementation yet, that's weird, one thing I found is "splats"
https://remix.run/docs/en/v1/guides/routing#splats

Dunno if that's usable in a good way or if its better, e.g. can you have $projectId.*.tsx ? idk

apps/designer/app/routes/canvas/$projectId.tsx
kof: I think someone misunderstood sentry api then, feel free to change

rpominov: Already did

apps/designer/app/routes/canvas/$projectId.$pageId.tsx
rpominov: We could use splats, but then something like `/canvas/123/123/not_a_parameter_we_expect` will also be directed into our route and we'll have to produce 404 manually. Why not keep as is?

kof (approved):

apps/designer/app/routes/canvas/$projectId.$pageId.tsx
kof: yeah, that would be not ideal either

PR #295, opened 2022-09-20
Omit conflicting props

Andarist:
cc @okonet

### DIFF
--- a/packages/design-system/src/components/text-field.tsx
+++ b/packages/design-system/src/components/text-field.tsx
@@ -147,7 +147,11 @@ const TextFieldBase = styled("div", {
 });
 
 type TextFieldProps = Omit<
-  React.ComponentProps<typeof TextFieldBase> & React.ComponentProps<"input">,
+  Omit<
+    React.ComponentProps<typeof TextFieldBase>,
+    "onChange" | "onFocus" | "onBlur"
+  > &
+    React.ComponentProps<"input">,
   "prefix"
 > & {
   inputRef?: React.Ref<HTMLInputElement>;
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/text-field.tsx
```diff
@@ -147,7 +147,11 @@ const TextFieldBase = styled("div", {
 });
 
 type TextFieldProps = Omit<
-  React.ComponentProps<typeof TextFieldBase> & React.ComponentProps<"input">,
+  Omit<
```
Andarist, line 150: In general, this can be definitely improved - it would be good if you'd have a common solution for this kind of type merge. Note though that `Omit` also doesn't come for free - especially for such huge types as the ones here (those DOM-oriented types have over 200 properties).

packages/design-system/src/components/text-field.tsx
```diff
@@ -147,13 +147,14 @@ const TextFieldBase = styled("div", {
...
-  "prefix"
-> & {
-  inputRef?: React.Ref<HTMLInputElement>;
-  prefix?: React.ReactNode;
-  suffix?: React.ReactNode;
-};
+  React.ComponentProps<typeof TextFieldBase>,
+  "onChange" | "onFocus" | "onBlur" | "prefix"
+> &
+  React.ComponentProps<"input"> & {
```
okonet, line 153: I wonder if something like this would be even better:

```suggestion
  React.ComponentProps<typeof InputBase>,
  "prefix"
> & {
```

BTW still not sure why `prefix` needs to be omitted but it looks like this is this `prefix`: https://yeahexp.com/what-is-the-prefix-attribute-for-in-the-html-tag/

okonet (approved):

PR #296, opened 2022-09-20
Omit `ref` prop from `TextFieldProps`

Andarist:
cc @okonet

### DIFF
--- a/packages/design-system/src/components/text-field.tsx
+++ b/packages/design-system/src/components/text-field.tsx
@@ -148,7 +148,7 @@ const TextFieldBase = styled("div", {
 
 type TextFieldProps = Omit<
   React.ComponentProps<typeof TextFieldBase> & React.ComponentProps<"input">,
-  "prefix"
+  "prefix" | "ref"
 > & {
   inputRef?: React.Ref<HTMLInputElement>;
   prefix?: React.ReactNode;
@@ -191,7 +191,7 @@ export const TextField = React.forwardRef<
         </Flex>
       )}
 
-      <InputBase disabled={disabled} {...textFieldProps} ref={inputRef} />
+      <InputBase ref={inputRef} disabled={disabled} {...textFieldProps} />
 
       {suffix && (
         <Flex


packages/design-system/src/components/text-field.tsx
```diff
@@ -148,7 +148,7 @@ const TextFieldBase = styled("div", {
 
 type TextFieldProps = Omit<
   React.ComponentProps<typeof TextFieldBase> & React.ComponentProps<"input">,
-  "prefix"
+  "prefix" | "ref"
```
Andarist, line 151: I still think that [the other fix](https://github.com/webstudio-is/webstudio-designer/pull/292/files/211e91fba2ee77b20b8d99dfdccf7d746f084cd5#r975493771) was a little bit better. With this fix here `ref` is no longer part of the internal `props` (and it shouldn't be)

Happy to discuss this further.

okonet: I'm not sure I understand it correctly but wouldn't this mean `ref` prop would also disappear from the external API of the component? `TextFieldProps` is the public API for it.

Andarist: `ref` is a little bit magic - in here its type for the **external** types flow from the `forwardRef` itself, not for the exact types of the `props`. Notice how in here we don't specify the type for the `ref` at all and it's still there: [TS playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AocgEyTQBsUi5MBXAOw2AjbgBUkAzjADCuSGyRsYACjA4wAgFxwA3swgRlAIw11UPAL4BKZWxZ06FcgB5+Q0eG6SsGuEUwBeFcnQwAdGhEKDBIyJjSRgZwAPQAfFZo3EJwAGLQAO6MNFQO4s5wHojE-pgZWWHS5HBwsvJKRb5+uU5SAAp11jAAnmBIEJh8giJiLTCxADRuSJhGBbGqVdXV7hRLMdFwAHoA-IvLSDAsUDxmFqtwBuRGVtZpUJlQ2c0SUuoQU57exQFBIRWRMXi5CAA)

okonet:
Yeah I've implemented this already. Let's close.

PR #298, opened 2022-09-21
Use TS project references to improve the IDE experience

Andarist:


apps/designer/tsconfig.json
```diff
@@ -2,9 +2,20 @@
...
+    "composite": true,
     "paths": {
       "~/*": ["app/*"]
     },
-    "baseUrl": "."
-  }
+    "baseUrl": ".",
+    "rootDir": "./app"
+  },
+  "references": [
```
kof, line 12: would have to add every new package though, can't it use some asterix packages/* or something?

packages/asset-uploader/tsconfig.json
```diff
@@ -3,10 +3,13 @@
   "extends": "../../node_modules/@webstudio-is/tsconfig/base.json",
   "include": ["./src/**/*.ts"],
   "compilerOptions": {
+    "composite": true,
```
kof, line 6: can't this be in the packages/tsconfig?

kof:
I think its overall terrible if we have to maintain all those dependencies, there must be a better way or a way to automate it

okonet (approved):
No idea what's going on but I really need it :)

apps/designer/tsconfig.json
Andarist: ye, this is the problematic part - and why I've initially wanted to use a different strategy, that different strategy would require some additional scripts and stuff and would take me more time to prepare. I've figured out that it's better to roll with this strategy, for now, to get improvements ASAP than to wait on that other strategy.

We can't use the asterisk here because the listed packages allow TS to build the dependency graph (it's somewhat weird but it can't figure this out on its own using just `package.json`) and the dep graph it needed here to "build" those packages in the correct order.

We could scriptify regenerating those references in a precommit hook or something. That would mitigate the maintainability concern.

okonet:
@kof I think Nx has a script for that but yeah it's 💩

packages/asset-uploader/tsconfig.json
Andarist: > the configs can probably be deduped better later - i didnt attempt to do this just yet

apps/designer/tsconfig.json
kof: we should def. have a script and we should check if someone else has solved that before, because it sounds like there must be something, maybe even specifically for turborepo?

.husky/pre-commit
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
+yarn update:ts-references
```
kof, line 4: is this something we need to do at pre-commit or rather at postinstall?

kof: ah no we are commiting those so yeah pre-commit

bin/update-ts-project-references.ts
```diff
@@ -0,0 +1,133 @@
...
+type DependencyGraph = Map<string, Set<string>>;
+type TSConfig = { references?: { path: string }[] };
+
+const stringArrayEquals = (arrA: string[], arrB: string[]): boolean => {
+  return (
+    arrA.length === arrB.length && arrA.every((itemA, i) => itemA === arrB[i])
+  );
+};
+
+let prettier: typeof import("prettier") | undefined;
```
kof, line 21: shouldn't we just make sure prettier is always installed as a requirement? its very unlikely that it isn't

I mean if I understand correctly this code is trying to make it optional

packages/asset-uploader/tsconfig.json
```diff
@@ -1,12 +1,14 @@
-// Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
```
kof, line 1: this comment is gone but we are still using the workaround for extends with a relative path instead of `extends: "@webstudio-is/tsconfig/base.json"`

bin/update-ts-project-references.ts
```diff
@@ -0,0 +1,133 @@
...
+      dependencies.add(dependency);
+    }
+  }
+
+  return graph;
+};
+
+const updateFile = async (filePath: string, content: string) => {
+  await fsP.writeFile(filePath, format(filePath, content));
+  cp.spawnSync("git", ["add", filePath]);
```
kof, line 57: nit: could use async spawn from fsP too

bin/update-ts-project-references.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+const updateTSConfig = async (
```
kof, line 60: nit: we are using this convention
```suggestion
const updateTsConfig = async (
```

packages/project/tsconfig.json
```diff
@@ -1,12 +1,18 @@
-// Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
 {
   "extends": "../../node_modules/@webstudio-is/tsconfig/base.json",
   "include": ["./src/**/*.ts"],
   "compilerOptions": {
+    "composite": true,
```
kof, line 5: still wanna put it into the reusable config?

packages/project/tsconfig.json
```diff
@@ -1,12 +1,18 @@
...
     "module": "commonjs",
     "target": "ES2020",
     "strict": true,
     "baseUrl": ".",
-    "outDir": "./lib"
-  }
+    "outDir": "./lib",
+    "rootDir": "./src"
+  },
+  "references": [
```
kof, line 13: we should probably generate a comment here that these are generated and from where ... just in case someone tries to update these manually

kof:
@Andarist looking great! added a few minor comments

packages/asset-uploader/tsconfig.json
Andarist: ye, so at first I've tried to preserve this but it's way easier to treat those files as JSON - where comments are not supported. I've briefly experimented with loading this as JSON5 and manipulating it with [`golden-fleece`](https://github.com/Rich-Harris/golden-fleece) but the results were not satisfactory. Maybe I've just made silly mistake, although its API is pretty straightforward, I've figured out that it's probably more time-efficient to abandon this and just rollback to simple JSONs

packages/project/tsconfig.json
Andarist: related: https://github.com/webstudio-is/webstudio-designer/pull/298#discussion_r1002433115

bin/update-ts-project-references.ts
Andarist: Ok, I've made this change and also adjusted the `TSConfig` type to `TsConfig`

bin/update-ts-project-references.ts
Andarist: This is not about making it optional - it will still throw when updating the file if Prettier is missing. It's just lazy-loading it to make this script faster when there are no updates to be made (which should be true for most of its runs). I will add a comment explaining this

kof: got it, maybe worth the comment, wasn't obvious

packages/asset-uploader/tsconfig.json
kof: Oh I didn't realize this. Yeah it seems like its a complication, but that package is supposed to handle it? I can give it a try if you want

bin/update-ts-project-references.ts
Andarist: `fsP` is just `fs/promises` - so it doesn't have any APIs related to the `child_process`. I could use [util.promisify()](https://nodejs.org/api/util.html#utilpromisifyoriginal) if you'd prefer to avoid this being synchronous. Note that this only gets executed when there is a tsconfig change which should be rare.

packages/project/tsconfig.json
Andarist: Done. I don't fully know how it will impact your tsconfig that targets the remix app (since it also extends the base) but I don't think it should affect it negatively. Worst case scenario - we gonna revert this change and put those back into individual configs.

bin/update-ts-project-references.ts
kof: or right, its node api being inconsistent, because there is no child_process/promises ... ignore it, its fine, I thought its a no-brainer

packages/asset-uploader/tsconfig.json
kof: you mean its not that easy to just put this property into the package? I am not sure what is this a problem to just move the flag to the central config

kof: ah already resolved

packages/asset-uploader/tsconfig.json
Andarist: Ok, so I've tried it again and it seems to work fine. It was producing weird formatting but since I've started to use Prettier here anyway it just fixes it after golden fleece is done with its magic.

You can take a look at my experiment here:
https://github.com/Andarist/webstudio-designer/compare/Andarist:68cd9cc...Andarist:11adbbb

Should I push those commits here?

kof: I think its quite important to be able to add comments there over the long run. One option is to use comments as keys if that makes everything easier, not sure this works though here, consumers would have to ignore those and types would need to pass

kof: looked at the changes - I see no problem, looks great!

kof: just add the comment for autogenerated references and what script generates them

Andarist: Added those comments, the PR is ready for a final review

kof (approved):
looks great!

rpominov:
This seem to have broke `yarn build`. It fails with:


```
src/db/create.ts(6,8): error TS6305: Output file '/Users/roman/projects/webstudio-designer/packages/prisma-client/lib/index.d.ts' has not been built from source file '/Users/roman/projects/webstudio-designer/packages/prisma-client/src/index.ts'.
```

kof:
I saw this error also locally, then it went away after something I did, I was hoping its not gonna happen on CI but it does. Feels like some caching issue

Issue #286, opened 2022-09-18
Allow assets upload over 4.5mb at once


kof:
It turned out vercel has a limit on the formdata size we send using one request https://vercel.com/guides/how-to-bypass-vercel-body-size-limit-serverless-functions

Vercel responds with error 413 

```
Request Entity Too Large
FUNCTION_PAYLOAD_TOO_LARGE
```
Now we need to find a solution for this asap, because 4.5mb is basically acceptable for one image, but even then not always. When user selects multiple images that limit is gonna be a problem right away.

What are our options?

kof:
Maybe as a workaround we could split the upload into a request per upload, so that 4.5 MB limit applies per file, which is acceptable for now. It's actually a problem only if a user selects a bunch of files at once so the total limit 4.5mb becomes too small.

kof:
Now we upload each file separately, so 4.5mb per file is reasonable for now, will be released with #299

PR #299, opened 2022-09-22
Render the fonts list and font-face

kof:
- rendering font-family style with fallbacks on canvas
- uploading multiple formats for fonts (e.g. same font but woff, ttf, etc), renders them in a single font face using multi-src
- deleting a font using the menu (deletes all assets for the corresponding family)
- fixes https://github.com/webstudio-is/webstudio-designer/issues/286
- partially implements image picker for background image

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/design-system/src/components/combobox.tsx
```diff
@@ -4,6 +4,7 @@ import {
   useEffect,
```
kof, line 4: all the combobox changes are temporarily for the POC

kof:
@okonet this is still WIP, just wanted to have your pre-review, no need to test yet and ignore the combobox changes, I will rebase once we merge the new one

apps/designer/app/canvas/canvas.tsx
```diff
@@ -117,7 +117,7 @@ export const Canvas = ({ data }: CanvasProps): JSX.Element | null => {
   }
   const isDesignerReady = useSubscribeDesignerReady();
   useInitializeBreakpoints(data.breakpoints);
-  globalStyles();
+  useGlobalStyles({ assets: data.assets })();
```
rpominov, line 120: Why this strange API where hook returns a function that we call on every render?
Also, calling functions in render is against React conventions unless it's a hook or a pure function.

packages/asset-uploader/src/utils/filter-by-type.ts
```diff
@@ -1,12 +1,12 @@
...
 import type { AssetType } from "../types";
-import { FONT_FORMATS } from "../constants";
 
 export const filterByType = <PartialAsset extends { format: string }>(
   assets: Array<PartialAsset>,
   type: AssetType
 ) => {
   return assets.filter(({ format }: PartialAsset) => {
-    const isFont = FONT_FORMATS.includes(format as typeof FONT_FORMATS[number]);
+    const isFont = format in FONT_FORMATS;
```
rpominov, line 9: Looks like a bug because:

```
'ttf' in ['ttf']
> false
```

Same in some other places if you search by `in FONT_FORMATS`

apps/designer/app/canvas/canvas.tsx
kof: Yeah, there are basically 2 functions

globalCss() that returns a function, where globalCss is supposed to be called in the module scope only once and the function it returns would be called inside the component.

Since I need to call globalCss when component renders first time, because font-faces are based on data, I needed a way to memorize that call, so I wrapped it into useCallback. useEffect doesn't work here well, since we need that for ssr.

I could use useMemo (with side effect, which would be wrong too) or I could call the useCallback directly inside useGlobalStyles hook so that that second invocation is hidden in the hook.

Any other options?

packages/asset-uploader/src/utils/filter-by-type.ts
kof: FONT_FORMATS is an object, I think you looked at the previous state where it was an array

apps/designer/app/canvas/canvas.tsx
kof: Actually I just noticed its being called many times even with the same assets object, so its not memorizing properly, I am looking for a better solution now

kof: Found a better way using useRef, still hiding the problem with the underlying  globalCss API, but at least this is not leaking outside of the hook

packages/asset-uploader/src/utils/filter-by-type.ts
kof: actually changed it now to use Map, isn't much of importance but is nicer

rpominov (approved):

apps/designer/app/canvas/canvas.tsx
kof: another option is something like this useEffect + useInsertionEffect

https://github.com/cssinjs/jss/blob/master/packages/react-jss/src/createUseStyles.js#L15

apps/designer/app/designer/shared/assets/use-serial-submit.ts
```diff
@@ -0,0 +1,41 @@
...
+  options: {},
+};
+
+/**
+ * Wrapps useSubmit hook to submit form data serially.
+ * This is useful when you want to submit multiple files, but send them one after another.
+ * Specifically Vercel has a single upload limit of 4.5MB, so we need to send them one by one so that this limit is per file, not per total upload
+ */
+export const useSerialSubmit = () => {
+  const submit = useSubmit();
```
rpominov, line 18: Not sure if this should be part of this PR, but in general I think `useFetcher` is much better for actions like uploading assets, or deleting a font — all "small" forms

packages/asset-uploader/src/delete.ts
```diff
@@ -3,24 +3,25 @@ import { deleteFromDb } from "./db";
...
-  id: string;
-  name: string;
-}): Promise<Asset> => {
-  const currentAsset = await prisma.asset.findUnique({
-    where: { id },
+export const deleteAssets = async (
+  ids: Array<string>
+): Promise<Array<Asset>> => {
+  const assets = await prisma.asset.findMany({
+    where: { OR: ids.map((id) => ({ id })) },
```
rpominov, line 12: I'd try `in` https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#in

apps/designer/app/designer/shared/assets/asset-upload.tsx
```diff
@@ -1,63 +1,89 @@
...
+      }
+      const options = {
+        method: form.method as FormMethod,
+        action: form.action,
+        encType: form.enctype as FormEncType,
+      };
+      submit(formData, options);
+      toPreviewAssets(formData)
+        .then(onSubmit)
+        .catch((error: unknown) => {
```
okonet, line 81: Not sure why this type cast is needed

apps/designer/app/designer/shared/fonts-manager/fonts-manager.tsx
```diff
@@ -1,18 +1,146 @@
...
+      <Flex
+        {...getComboboxProps()}
+        css={{ flexDirection: "column", gap: "$3" }}
+      >
+        <TextField
+          type="search"
+          prefix={<MagnifyingGlassIcon />}
+          {...getInputProps()}
+        />
+        <fieldset>
```
okonet, line 123: This was supposed to be an example only. In our code we should use something from DS or if it's missing, Flex with as prop?

apps/designer/app/designer/shared/fonts-manager/fonts-manager.tsx
```diff
@@ -1,18 +1,146 @@
...
+          <legend>Choose an item</legend>
```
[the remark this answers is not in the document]
kof: wip, behind a feature flag

apps/designer/app/designer/shared/assets/use-serial-submit.ts
kof: rewriting to useFetcher right now

apps/designer/app/designer/shared/assets/asset-upload.tsx
kof: will check

kof: removed, not sure why it was used

apps/designer/app/designer/shared/assets/use-serial-submit.ts
kof: done

packages/asset-uploader/src/delete.ts
kof: good hint, done

apps/designer/app/designer/shared/assets/use-assets.ts
```diff
@@ -1,53 +1,99 @@
...
+    });
+    toPreviewAssets(formData)
+      .then((previewAssets) => {
+        setAssets([...previewAssets, ...assets]);
+      })
+      .catch((error) => {
+        if (error instanceof Error) {
+          toast.error(error.message);
+        }
+      });
```
rpominov, line 90: Is it ok that we submit anyway even if this error happen?

PR #302, opened 2022-09-23
Integrate lost-pixel

okonet:
This integrates [lost-pixel](https://www.lost-pixel.com) with design sysytem.


turbo.json
```diff
@@ -12,6 +12,10 @@
     },
     "storybook:run": {
       "cache": false
+    },
+    "storybook:build": {
+      "dependsOn": ["@webstudio-is/icons#build"],
```
kof, line 17: what is special about the icons package? is it because it's the monorepo package? if yes wouldn't it be easier to maintain if we used asterisk without specific package name in it?

kof (approved):
Looking good, my biggest concern is maintaining those dependency names.

okonet:
What dependency names?

kof:
Our build is using esbuild, its building all of those very quickly, feel free to measure. Except the things we have in react-sdk which shouldn't be part of the main build command anyways, let me measure how fast the build is actually without those slow parts we can avoid.

kof:
Well idk, its up to you, if you feel like building all is going to be a problem

okonet:
I don't see a problem ATM. If we encounter it in the future we can revisit. I'm okay with the this proposed approach for now.

okonet:
Frankly I don't know why we need to build icons package from source before we can consume it. Maybe we should look into that before anything else.

kof:
> Frankly I don't know why we need to build icons package from source before we can consume it. Maybe we should look into that before anything else.

Also my question here, icons are built at publish time, so they should be installable from npm, unless the build is not using it from npm and uses local package and has to build

okonet:
No if we agree on icons being like this for now.

turbo.json
okonet: I'm resolving this one to proceed for now. We agreed to look into builds in a separate PR

PR #303, opened 2022-09-23
Speedup sdk build, avoid building things that haven't changed

kof:
Since we are committing those build artefacts, there is no point in building them every time main build command is executed

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/package.json
+++ b/packages/react-sdk/package.json
@@ -6,7 +6,7 @@
   "homepage": "https://webstudio.is",
   "scripts": {
     "dev": "tsup src/index.ts --format esm,cjs --watch --out-dir=lib",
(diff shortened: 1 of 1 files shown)


rpominov (approved):
Great!

I think there is also an issue with these artefacts that formatting is not consistent. I sometimes get these files in my diff after a build even thought I didn't change anything.

This PR likely solves this issue as well.

kof:
yeah, there is the issue that formatter and the thing we generate are not equal, but if we update those manually, it will autoformat when updated, so we are good for now

PR #304, opened 2022-09-25
Feature style panel flex child section

ghost:
Branching from https://github.com/webstudio-is/webstudio-designer/pull/268
Depends on(Don't merge before): https://github.com/webstudio-is/webstudio-designer/pull/313

- add/update toggle-control.
- add floating panel.

Figma of what's to be added/update for reference.

Screenshot 2022-09-25 at 6 26 10 AM

Screenshot 2022-09-25 at 6 26 02 AM

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/popover.tsx
+++ b/packages/design-system/src/components/popover.tsx
@@ -25,6 +25,8 @@ const StyledContent = styled(PopoverPrimitive.Content, panelStyles, {
   "&:focus": {
     outline: "none",
   },
+  display: "flex",
+  flexDirection: "column",
 });
 
 type PopoverContentPrimitiveProps = React.ComponentProps<
@@ -67,7 +69,7 @@ type PopoverHeaderProps = {
 
 export const PopoverHeader = ({ title }: PopoverHeaderProps) => {
   return (
-    <>
+    <Box css={{ order: -1 }}>
       <Flex
         css={{ height: 40, paddingLeft: "$spacing$9" }}
         align="center"
@@ -87,7 +89,7 @@ export const PopoverHeader = ({ title }: PopoverHeaderProps) => {
         </PopoverClose>
       </Flex>
       <Separator />
-    </>
+    </Box>
   );
 };
 
--- a/apps/designer/app/designer/features/style-panel/shared/value-picker-popover.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/value-picker-popover.tsx
@@ -56,8 +56,8 @@ export const ValuePickerPopover = ({
           align="start"
           css={{ width: PANEL_WIDTH }}
         >
-          <PopoverHeader title={title} />
           {content}
+          <PopoverHeader title={title} />
         </PopoverContent>
       </PopoverPortal>
     </Popover>


packages/design-system/src/components/toggle-group.tsx
```diff
@@ -30,11 +31,12 @@ export const Item = styled(ToggleGroupPrimitive.Item, {
...
   "&:hover": { backgroundColor: "$slateA3" },
+  "&:focus": { boxShadow: "0 0 0 $spacing$2 $colors$blue10", zIndex: 1 },
   "&[data-state=on]": {
-    backgroundColor: "$slateA5",
-  },
-  "&:focus": {
-    position: "relative",
-    boxShadow: `0 0 0 2px black`,
+    backgroundColor: "$slate5",
+    "&[data-breakpoint=true]": {
```
kof, line 37: I feel like we talked about it several times by now.  A design system component doesn't know the designer specific stuff like breakpoint. It's business logic related

apps/designer/app/designer/features/style-panel/controls/toggle/toggle-control.tsx
```diff
@@ -0,0 +1,41 @@
+import { Flex, ToggleGroup, Tooltip } from "@webstudio-is/design-system";
+import { StyleProperty } from "@webstudio-is/react-sdk";
+import { useIsFromCurrentBreakpoint } from "../../shared/use-is-from-current-breakpoint";
+
+export type ToggleGroupControlProps = {
+  property: StyleProperty | StyleProperty[];
```
kof, line 6: is there a concrete use case of passing an array of properties? If not we should simplify and use single property.

apps/designer/app/designer/features/style-panel/controls/toggle/toggle-control.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const ToggleGroupControl = ({
+  property,
+  value = "",
+  items = [],
+  onValueChange,
+}: ToggleGroupControlProps) => {
+  const isCurrentBreakpoint = useIsFromCurrentBreakpoint(property);
+  return (
+    <ToggleGroup.Root
```
kof, line 20: Lets export the Toggle group in the ds as `export const ToggleGroupRoot` ... we have it currently inconsistent, but I think this is better ... alternatively the way you did it here and then renaming the rest

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -0,0 +1,267 @@
...
+  currentStyle,
+  sectionStyle,
+}: RenderCategoryProps) => {
+  const setAlignSelf = setProperty("alignSelf");
+  const setOrder = setProperty("order");
+  const setSizing = createBatchUpdate();
+  const getSizingValue = () => {
+    switch (
+      String(currentStyle.flexGrow?.value) +
+      String(currentStyle.flexShrink?.value)
```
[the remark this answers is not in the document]
kof: also this function can be outside of cmponent

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -0,0 +1,267 @@
...
+              child: <FlexSizingShrinkIcon />,
+              label: "Shrink if needed",
+              value: "shrink",
+            },
+            {
+              child: (
+                <ValuePickerPopover
+                  title="Sizing"
+                  content={
+                    <Grid
```
kof, line 152: a separate component would be much easier to read

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -0,0 +1,267 @@
...
+                  </Flex>
+                </ValuePickerPopover>
+              ),
+              label: "More sizing options",
+              value: "",
+            },
+          ]}
+        />
+      </Grid>
+      <Grid css={{ gridTemplateColumns: "4fr auto" }}>
```
kof, line 204: same thing here, would it make sense for it to be separate component? (same file)

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -0,0 +1,267 @@
...
+              case "0":
+              case "1":
+              case "-1": {
+                setOrder(value);
+                break;
+              }
+            }
+          }}
+          value={String(currentStyle.order?.value)}
+          items={[
```
kof, line 222: same thing here, would it make sense for it to be separate component? (same file)

kof: rather a function

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -0,0 +1,267 @@
...
+              label: "Make first",
+              value: "1",
+            },
+            {
+              child: <OrderLastIcon />,
+              label: "Make last",
+              value: "-1",
+            },
+            {
+              child: (
```
kof, line 239: this one could be a component

packages/icons/src/align-self-baseline.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
```
kof, line 1: we started generating icon components out of SVG, check out the package scripts

packages/icons/src/order-first.tsx
```diff
@@ -0,0 +1,47 @@
...
+          width="4"
+          height="8"
+          x="10.9"
+          y="4"
+          stroke={color}
+          strokeWidth="2.2"
+          mask="url(#a)"
+          rx="1"
+        />
+        <mask id="b" fill="#fff">
```
kof, line 29: is svg becoming part of the global document when rendered directly as a component? I wonder if that id is only visible inside the SVG itself or it just became a global id

kof (changes requested):
first round of review done

packages/icons/src/order-first.tsx
kof: I understand its from figma, my question is if that's ok or is it a document-global id

packages/icons/src/align-self-baseline.tsx
kof: cc @TrySound

TrySound: Props are spread and ref is added. See generated icons. Fill on the root should be enough. I dont see the issue.

packages/icons/src/order-first.tsx
TrySound: We can add prefixes for ids in svg with svgo plugin

packages/icons/src/order-last.tsx
```diff
@@ -0,0 +1,48 @@
...
+          width="4"
+          height="8"
+          x="10.9"
+          y="4"
+          stroke={color}
+          strokeWidth="2.2"
+          mask="url(#a)"
+          rx="1"
+        />
+        <mask id="b" fill="#fff">
```
kof, line 30: this one seems like it's doing something, so we can't just remove it, we need the plugin

packages/design-system/src/components/toggle-group.tsx
```diff
@@ -30,11 +31,6 @@ export const Item = styled(ToggleGroupPrimitive.Item, {
...
   "&:last-child": { borderTopRightRadius: 4, borderBottomRightRadius: 4 },
   "&:hover": { backgroundColor: "$slateA3" },
-  "&[data-state=on]": {
-    backgroundColor: "$slateA5",
-  },
-  "&:focus": {
-    position: "relative",
-    boxShadow: `0 0 0 2px black`,
-  },
+  "&:focus": { boxShadow: "0 0 0 $spacing$2 $colors$blue10", zIndex: 1 },
```
kof, line 34: was there again an issue with stacking context or are you fixing it preemptively? I wonder what we should standardized on, `1` feels random.

Maybe the issue was again somewhere else?

apps/designer/app/designer/features/style-panel/controls/toggle/toggle-control.tsx
```diff
@@ -0,0 +1,66 @@
...
+      value={value}
+      onValueChange={onValueChange}
+      css={{ width: "fit-content" }}
+    >
+      {items.map(({ child, label, value }, index) => {
+        return (
+          <ToggleGroupControlItem
+            key={index}
+            value={value}
+            state={isCurrentBreakpoint ? "set" : undefined}
```
kof, line 37: "set" is also application level thing, sorry, you either need to add that state to a new styled component like a last time, or it's something that we need everywhere and its not "on" state it already has, then you have to explain what use case it has somewhere else

kof: ohhh its local component now, ignore  that comment above

apps/designer/app/designer/features/style-panel/controls/toggle/toggle-control.tsx
```diff
@@ -0,0 +1,66 @@
...
+      })}
+    </ToggleGroupRoot>
+  );
+};
+
+const ToggleGroupControlItem = styled(ToggleGroupItem, {
+  variants: {
+    state: {
+      set: {
+        "&[data-state=on]": {
```
kof, line 53: why do you need to override the on state this way? is that fixing a specificity issue?

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -0,0 +1,277 @@
...
+    </ValuePickerPopover>
+  );
+};
+
+const getSizingValue = (flexGrow: string, flexShrink: string) => {
+  if (flexGrow === "0" && flexShrink === "0") return "none";
+  if (flexGrow === "1" && flexShrink === "0") return "grow";
+  if (flexGrow === "0" && flexShrink === "1") return "shrink";
+  return "";
+};
```
kof, line 277: this file looks now 1000x more readable now

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -28,7 +28,6 @@ export const PropertyName = ({ property, label, css }: PropertyProps) => {
                   color: "$hiContrast",
                 }),
           }}
-          htmlFor={property}
```
kof, line 31: that was for clicking on property to result in focusing the field, why removing?

kof:
Please link the new issues that for SVG stuff

1. id auto generation
2. generating

packages/design-system/src/components/toggle-group.tsx
kof: please add that as a comment, super non-obvious

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: it didn't work? then we need to fix it?

kof (changes requested):
almost there

apps/designer/app/designer/features/style-panel/controls/toggle/toggle-control.tsx
kof: damn, this isn't great from stitches side ... what options do we have?

packages/design-system/src/components/popover.tsx
```diff
@@ -77,10 +77,14 @@ export const PopoverHeader = ({ title }: PopoverHeaderProps) => {
         <PopoverClose asChild>
           <IconButton
             size="2"
-            css={{ marginRight: "$spacing$5" }}
+            css={{
+              marginRight: "$spacing$5",
+              "&:not(:hover):focus": { background: "none" },
```
kof, line 82: seems like you were trying to achieve this on focus without hover

Screenshot 2022-11-15 at 16 45 51


But in figma it is specified like this:


Screenshot 2022-11-15 at 16 46 45

Either we need to change it in figma by @taylornowotny and change it on IconButton directly or leave it as-is.

Adhoc changing it in one particular popup is wrong.

kof (approved):
🎉

PR #310, opened 2022-09-28
fix: http-client should not have process.env

andrasbacsai:
`process.env` is not defined in the worker state, so URL is defined in a different way.

### DIFF
--- a/packages/http-client/src/index.test.ts
+++ b/packages/http-client/src/index.test.ts
@@ -2,21 +2,27 @@ import { loadProject } from "./index";
 
 const existingProject = "ab3619d3-831f-46f8-abb6-609558cffd99";
 const notPublishedProject = "7ec397c6-b3d0-4967-9073-9d83623fcf8e";
+const webstudioAPIUrl = "http://localhost:3000";
 const nonExistingProject = "1";
 
 describe("getProjectDetails", () => {
   test("loads existing project", async () => {
-    const response = await loadProject({ projectId: existingProject });
+    const response = await loadProject({
+      webstudioAPIUrl,
+      projectId: existingProject,
+    });
     expect(response.tree.id).toBeTruthy();
   });
   test("loads non-existing project", async () => {
     const response = await loadProject({
+      webstudioAPIUrl,
       projectId: nonExistingProject,
     });
     expect(response.tree.errors).toBe("Project required");
   });
   test("loads not published project", async () => {
     const response = await loadProject({
+      webstudioAPIUrl,
       projectId: notPublishedProject,
     });
     expect(response.tree.errors).toBe(
--- a/packages/http-client/src/index.test.ts
+++ b/packages/http-client/src/index.test.ts
@@ -1,32 +1,24 @@
 import { loadProject } from "./index";
 
-const existingProject = "ab3619d3-831f-46f8-abb6-609558cffd99";
-const notPublishedProject = "7ec397c6-b3d0-4967-9073-9d83623fcf8e";
-const webstudioAPIUrl = "http://localhost:3000";
-const nonExistingProject = "1";
+const existingProjectId = "ab3619d3-831f-46f8-abb6-609558cffd99";
+const notPublishedProjectId = "7ec397c6-b3d0-4967-9073-9d83623fcf8e";
+const apiUrl = "http://localhost:3000";
 
 describe("getProjectDetails", () => {
   test("loads existing project", async () => {
     const response = await loadProject({
-      webstudioAPIUrl,
-      projectId: existingProject,
+      apiUrl,
+      projectId: existingProjectId,
     });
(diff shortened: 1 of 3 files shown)


packages/http-client/src/index.test.ts
```diff
@@ -2,21 +2,27 @@ import { loadProject } from "./index";
 
 const existingProject = "ab3619d3-831f-46f8-abb6-609558cffd99";
 const notPublishedProject = "7ec397c6-b3d0-4967-9073-9d83623fcf8e";
+const webstudioAPIUrl = "http://localhost:3000";
```
kof, line 5:
```suggestion
const apiUrl = "http://localhost:3000";
```

packages/http-client/src/index.test.ts
```diff
@@ -2,21 +2,27 @@ import { loadProject } from "./index";
 
 const existingProject = "ab3619d3-831f-46f8-abb6-609558cffd99";
```
kof, line 3:
```suggestion
const existingProjectId = "ab3619d3-831f-46f8-abb6-609558cffd99";
```

packages/http-client/src/index.ts
```diff
@@ -1,18 +1,19 @@
...
 
-const protocol = process.env.NODE_ENV === "production" ? "https" : "http";
-
 export const loadProject = async ({
+  webstudioAPIUrl = null,
   projectId,
-  host = process.env.DESIGNER_HOST || "localhost:3000",
   include = { tree: true, props: true, breakpoints: true },
 }: {
+  webstudioAPIUrl: string | null;
```
kof, line 9: 1. there is no non-webstudio api url right?
2. if its optional, use optional type, almost never null, we are using null type basically when we have to deal with dom apis which use null, mostly I am trying to hide null in the abstraction and never expose it

packages/http-client/src/index.ts
```diff
@@ -1,18 +1,19 @@
...
   projectId: string;
-  host?: string;
   include?: Includes<boolean>;
 }): Promise<Project> => {
-  const baseUrl = new URL("/", `${protocol}://${host}`);
+  if (!webstudioAPIUrl) {
+    throw new Error("Webstudio API URL is required.");
+  }
+  const baseUrl = new URL("/", `${webstudioAPIUrl}`);
```
kof, line 16: what is the first slash for if that's an actual URL?

packages/http-client/src/index.ts
```diff
@@ -1,18 +1,19 @@
...
-  host = process.env.DESIGNER_HOST || "localhost:3000",
   include = { tree: true, props: true, breakpoints: true },
 }: {
+  webstudioAPIUrl: string | null;
   projectId: string;
-  host?: string;
   include?: Includes<boolean>;
 }): Promise<Project> => {
-  const baseUrl = new URL("/", `${protocol}://${host}`);
+  if (!webstudioAPIUrl) {
```
kof, line 13: once you only have undefined
```suggestion
  if (webstudioAPIUrl === undefined) {
```

kof (approved):
A number of cosmetic change requests there, approving so you can merge once fixed

packages/http-client/src/index.ts
kof: wait this doesn't make any sense, this URL is not optional and so shouldn't accept a null at type level, but then validate in code to not accept null, it should only accept string strictly and so you don't even have to do this runtime check

kof: Unless you think the url could come from env and be undefined or empty string, in this case you could do new URL(apiUrl) and then check for hostname for example, so you have an actual validation

packages/http-client/src/index.ts
```diff
@@ -1,18 +1,19 @@
...
+  apiUrl: string;
   projectId: string;
-  host?: string;
   include?: Includes<boolean>;
 }): Promise<Project> => {
-  const baseUrl = new URL("/", `${protocol}://${host}`);
+  if (apiUrl === undefined) {
+    throw new Error("Webstudio API URL is required.");
+  }
+  const baseUrl = new URL(`${apiUrl}`);
```
kof, line 16: no need for template

packages/http-client/src/index.ts
```diff
@@ -1,18 +1,19 @@
...
-  host = process.env.DESIGNER_HOST || "localhost:3000",
   include = { tree: true, props: true, breakpoints: true },
 }: {
+  apiUrl: string;
   projectId: string;
-  host?: string;
   include?: Includes<boolean>;
 }): Promise<Project> => {
-  const baseUrl = new URL("/", `${protocol}://${host}`);
+  if (apiUrl === undefined) {
```
kof, line 13: Again, I mentioned in the CR that if type is not optional, this logic doesn't make sense, what makes sense is validating after new URL the set hostname

PR #313, opened 2022-09-29
Feature:  CSSValueControl

okonet:
### Common:
- Free text editing
- Enter or blur calls onChangeComplete
- Value prop can be of type "invalid" and render invalid mode of the input (red outline)
### Unit mode:
- When entire text is a number we automatically switch to unit mode on keydown
- Unit selection on unit button click or focus+enter
- When selecting unit arrow keys are used to navigate unit items
- When selecting unit Enter key or click is used to select item
- When selecting unit Escape key is used to close list
- Key up and down on focused input increment/decrement the value
  - shift key modifier increases/decreases value by 10
  - option/alt key modifier increases/decreases value by 0.1
  - no modifier increases/decreases value by 1
- Scrub interaction
### Keywords mode:
- When any character in the input is not a number we automatically switch to keywords mode on keydown
- Filterable keywords list (click on chevron or arrow down to show the list)
- Arrow keys are used to navigate keyword items
- Enter key or click is used to select item when list is open
- Escape key is used to close list
- When hovering over keywords list, onItemHighlight is called
Features outside of this input (non standard):
- Typing number + unit (e.g. "12px") in unit mode will change the selected unit on blur/enter
- Evaluated math expression: "2px + 3em" (like CSS calc())

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


kof:
Question: why not CssValueControl? We do try to define a value, right?

okonet:
Can rename if that makes it clear. To me it's a control for a property but that's true, it's it's value we are setting.

apps/designer/app/designer/features/style-panel/controls/css-property/css-property.tsx
```diff
@@ -0,0 +1,246 @@
...
+      event.currentTarget.value = String(roundedValue);
+      onChange?.({
+        type: "unit",
+        value: roundedValue,
+        unit: value?.unit ?? "px",
+      });
+    }
+  };
+
+  const unitStateReducer = useCallback((state, actionAndChanges) => {
```
kof, line 78: can't this be a separate function outside of component? I personally really really don't like huge components

apps/designer/app/designer/features/style-panel/controls/css-property/css-property.tsx
```diff
@@ -0,0 +1,246 @@
...
+      if (event.altKey) currentDelta = 0.1;
+      numberValue = numberValue + currentDelta * direction;
+      const roundedValue =
+        numberValue % 1
+          ? Number(
+              numberValue.toPrecision(
+                Math.abs(numberValue).toString().indexOf(".") + 2
+              )
+            )
+          : numberValue;
```
kof, line 68: this entire block could be a separate pure function

kof:
1. Theming in storybook looks like this for me, was it the same for you?

Screenshot 2022-09-30 at 15 23 36

okonet:
3 and 4 are related to TextField implementation and I’m aware of them. I can make this work next week. Please disregard it for now. 

As for the rest, some of those issues are related to how the state is preserved / handled outside of the component. We can introduce an additional state for units to remember what unit was previously selected but this shows how complex the whole thing is. 

As for the general UX of the combo box: it’s debatable since making combo box input display the value has its drawbacks:

1. We can’t display complex items like icons etc there
2. you can’t see the whole list of items before you clear the input which can be very annoying for longer items. What’s more annoying that it’s not clear why the drop-down isn’t appearing at all when an item is selected. In my latest implementation of the combo box pressing backspace or starting to type anything would remove the selection. 
3. it’s impossible to implement multiple selection in the future without this. 

All of that being said I’m not saying that my solution is perfect and I admit the problems mentioned above. Let’s discuss?

okonet:
I think the easiest way to continue would be for you to implement it the way you'd like it to be and create a PR for it. One thing I'd like us to avoid though is changing behavior of DS components to accommodate for a specific app use case without considering other use cases. Try to cover all the use cases in the storybook when you're implementing it. So if you're questioning the combo box behavior let's change it in design system first and then use in this component so we can have consistent behavior everywhere. Does that make sense?

okonet:
BTW none of the videos can't be viewed for some reason neither on iPhone nor my Mac... Any idea why is that? Any way to re-share them?

kof:
@okonet did you use safari? It works perfectly in chrome, but I just tried safari and lol its just spinning

okonet:
1. Yeah I'm using Safari as my main browser. But also links look weird to me since if you simply drop a video here it will convert it to a video tag AFAIK. Maybe it's some codec issue. I'll take a look in Chrome then.

okonet:
So after watching videos I can now give a bit of comment:

1. Interestingly it works in the storybook. Might be something I didn't finish in the text-control integration. I bet it's missing `itemToString`...
2. The CSSProperty control is designed in a way it's not holding application's state but gives two callback. This way it's decoupled from the application's logic and the way the app handles its state. So in order to fix 2 and 5 we need to introduce some local state in the TextControl but I think the better solution is to make the Combobox not loose currently selected value until the final selection has happened. It should be an easy change that would solve some current issues.
3. p 6 is solvable with a local state inside the component. It's rather an easy fix.

kof:
@okonet I have just tested it again: having the selected item look like input value text and then not able to select it or clear it when I start typing feels unlogical/wrong.

Either we need to change the design to display that selected item as a chip or we should render it as an input value. But the way it is right now is no good.

I don't believe we actually ever wanted the chip-like design in there ... to me it should feel like an input all the way

kof:
@okonet, me and @natlusyht  were trying to remember what you said why you want to prevent user from editing the value including the selected value. Can you help us understand?

Right now it feels like the wrong approach to the UI we were trying to accomplish.

okonet:
I already commented that I'm open for changing this. Please feel free to change that.

kof:
@okonet perfect, @natlusyht will be on it today

kof:
@okonet please review

taylornowotny:
My comments:

Clicking the unit selector and then clicking away should not leave the unit selector focused. Either the input should be focused or nothing should be focused. In Webflow focus goes away entirely if you click away from the input. 

Clicking once anywhere in an input that is not already selected should highlight the text. The current single-click behavior should be what happens on double-click.

Limit units to the units in the design. The unit list is currently unecessarily long. For example we don't need centimeters in a digital-native app and we don't need Hz or kHz for any reason at all. Better to start simple and add more options later if needed.

For most inputs with units the list looks like this, but this list may be different for different properties. The design will always specify what units to include in the unit list.
Screen Shot 2022-10-14 at 11 49 39 AM

kof:
- [ ] entering a value in keyboard mode and hitting enter should close keywords list

kof:
- [ ] This interaction doesn't preserve the unit (Unit mode > select all > backspace > enter number)

https://user-images.githubusercontent.com/52824/195978905-bc14d810-3638-4711-a0fa-0f15bc61677a.mp4

kof:
- [ ] On blur should not activate the property name label showing that it has changed when the value hasn't changed, input can enter a state where two things are focused at the same time: the input box and the unit select. probably Tab should tab to the next input instead.

https://user-images.githubusercontent.com/52824/195978926-e0b58eb5-ea52-4fb9-a945-cea983b0bbb2.mp4

kof:
- [ ] side by side the hover states of the unit/keyword suffix icon don't match, keyword suffix being closer to the figma doc design.

image

taylornowotny:
Agree with Sultan, remove caret while scrubbing.

- [ ] On first click, the input should become focused with the value highlighted and no caret visible. On second click, the caret should be placed at the end of the value and the value should not be highlighted.
    - What currently happens is: On first click, caret is placed at the cursor position even if that’s in between characters, then when the click is released the caret goes away and the whole value is selected. It would be better if the whole value was selected immediately on click with no caret. Second click behaves just like first click.

- [ ] After_ scrubbing, number is de-selected and caret is placed at the end. I think the value should stay selected with no caret present after scrubbing.

- [ ] The scrub cursor activates immediately on click but it should be on click+hold with a ~500ms delay.

- [ ] After scrubbing, cursor jumps to the input field. This feels unnatural - can we stop that from happening?

- [ ] The scrub is working great! Nice & intuitive.

- [ ] When focus moves to the unit select button, focus should be removed from the input. Can’t focus two things at the same time.

- [ ] When a unit is selected from the unit menu, the focus should return to the input. Currently focus goes away after selecting a unit so pressing tab after selecting a unit does nothing. If the user can type in the input field then that input field should have a focused state.

- [ ] Selecting an input adds the ‘set’ state even when no changes are made to the value. The set state should only apply after the value is edited.

kof:
- [ ] check the font

kof:
> number should not be a unit, should be — dash if anything.

yeah this one is also a question, because technically its a number unit in css

kof:
I am thinking to merge this before resolving all the things, to unblock other things and avoid conflicts. Ideally at least a code review would be great @okonet

kof:
> When a unit is selected from the unit menu, the focus should return to the input. Currently focus goes away after selecting a unit so pressing tab after selecting a unit does nothing. If the user can type in the input field then that input field should have a focused state.

@taylornowotny solved it by not having the input half-focused after unit selection, does that work? Or do you think after unit selection focus should actually land on the input field?

kof:
> When focus moves to the unit select button, focus should be removed from the input. Can’t focus two things at the same time.

@taylornowotny I just noticed that the focused state of the unit button in figma is not outline, its background, so once this is fixed, I think it will look natural, having focus outline over the input and the blue background on the unit when unit is focused

kof:
@taylornowotny @natlusyht I created a follow-up issue, lets continue discussing there, because this PR needs to be merged, it has too many changes that impact other work

https://github.com/webstudio-is/webstudio-designer/issues/333

PR #315, opened 2022-10-03
Feature css value control branch

natlusyht:
This is a branch of [feature](https://github.com/webstudio-is/webstudio-designer/pull/313) with only text-control reverted to it's previous state.

The issue I mentioned in https://github.com/webstudio-is/webstudio-designer/pull/268/files#r979314152 is still present though.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
+++ b/apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
@@ -1,7 +1,31 @@
-import { CssPropertyControl } from "~/designer/features/style-panel/controls/css-property/css-property";
+import { FocusEvent, KeyboardEvent, useRef, PointerEvent } from "react";
+import {
+  Box,
+  Text,
+  IconButton,
+  DropdownMenu,
+  DropdownMenuTrigger,
+  DropdownMenuPortal,
+  DropdownMenuContent,
+  DropdownMenuRadioItem,
+  DropdownMenuRadioGroup,
+  TextField,
+  Combobox,
+  numericScrubControl,
+} from "@webstudio-is/design-system";
 import { getFinalValue } from "../../shared/get-final-value";
 import { ControlProps } from "../../style-sections";
-import { StyleValue } from "@webstudio-is/react-sdk";
+import { units } from "@webstudio-is/react-sdk";
+import { ChevronDownIcon } from "@webstudio-is/icons";
+import { PropertyIcon } from "../../shared/property-name";
+
+const sortedUnits = units
+  .slice(0)
+  .sort((v) =>
+    ["%", "px", "rem", "em", "ch", "vh", "vw", "hv", "vmin", "vmax"].includes(v)
+      ? -1
+      : 1
+  );
 
 export const TextControl = ({
   currentStyle,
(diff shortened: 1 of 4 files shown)


apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -58,6 +58,11 @@ export const PropertyIcon = forwardRef<
         ref={forwardRef}
         size="1"
         css={{
+          // @note DO NOT REMOVE WITHOUT REPLACING WITH TOKENS
```
kof, line 61: The idea Andrey and I are pushing is to actually not do the nudges at all and instead fix it later properly with changes in the scales etc, so the preference is towards have small suboptimal visual things instead of heaving these

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -15,37 +39,195 @@ export const TextControl = ({
...
-        type: "keyword",
-        value: item.name,
-      }))}
-      onChange={(value) => {
-        console.log("change", value);
-        handleChange(value, true);
+    <Combobox
+      name={styleConfig.property}
+      items={items}
+      value={String(value.value)}
```
kof, line 101: we discsussed that combobox should receive the value object, not the value.value string, why is it back?

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -15,37 +39,195 @@ export const TextControl = ({
...
+              suffix={
+                <IconButton
+                  {...(value.type !== "unit" && toggleProps)}
+                  css={{
+                    visibility: items.length ? "visible" : "hidden",
+                    position: "absolute",
+                    right: "1px",
+                    top: "1px",
+                    width: "auto",
+                    height: "calc(100% - 2px)",
```
kof, line 173: again, we said no nudges

kof: positioning should be done via layout
sizes should be done using token values, if there is none that fits the exact pixel value we need - take the next closest, we have to fix tokens, not workaround with nudges

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
kof: of course it doesn't, you have to pass to the input the value.value

okonet: I think you must define `itemToString` to make it work. This will also fix filtering.

okonet: I think I'll make it a required prop so it removes this confusion.

kof: \when rendering the value in the input, he can just pass value.value, its in his control where TextField is getting rendered

okonet: I'm confused. What is "that"? It was working in my branch. Maybe take a look there? Just a heads up: I'm working on some combo box changes in the original branch to 1. Make custom usages easier and 2. To revert the input and prefix behavior I introduced earlier.

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -90,6 +90,8 @@ export const numericScrubControl = (
         break;
       }
       case "pointerdown": {
+        // light touches don't register corresponding pointerup
+        if (event.pressure === 0) break;
```
kof, line 94: @natlusyht will send this in a separate pr

kof:
Closed in favor of https://github.com/webstudio-is/webstudio-designer/pull/313

PR #319, opened 2022-10-05
Minimal pages panel 

rpominov:
This PR is another step to implement pages support: #54

Goals of this PR:

 - [x] Add a pages panel behind a feature flag
   - [x] Show pages list
   - [ ] Add page button should add a new page
   - [x] Clicking on a page in the list should open it in the editor
   - [ ] Clicking on the menu icon should open settings panel
     - [ ] Settings panel allows to change page name
     - [ ] Settings panel allows to delete the page
 

-------------------------

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


rpominov:
@kof I didn't do everything what I planed initially for this PR. Will do in separate PRs as you suggested.

apps/designer/app/routes/canvas/$projectId.$pageId.tsx
```diff
@@ -1,2 +0,0 @@
-export * from "./$projectId";
-export { default } from "./$projectId";
```
rpominov, line 2: This was a bad idea indeed. This confuses Remix somehow so it doesn't load data when you navigate in browser

kof:
Weird it passed before

kof:
seems like btoa exists in node 16 and doesn't in 14 and I have no idea why we support 14 anyways, so I would just drop 14 everywhere

rpominov:
Yea, good idea! Should I do it?

kof:
yes, please

apps/designer/app/designer/designer.tsx
```diff
@@ -296,7 +309,13 @@ export const Designer = ({ config, project, page }: DesignerProps) => {
         </Workspace>
       </Main>
       <SidePanel gridArea="sidebar" isPreviewMode={isPreviewMode}>
-        <SidebarLeft publish={publish} />
+        <SidebarLeft
+          publish={publish}
+          project={project}
+          pages={pages}
```
kof, line 315: We can't pass all of this to sidebar component, because this pattern won't let us scale the panels. Each feature panel should get the data on its own, e.g. using nano state

apps/designer/app/designer/features/sidebar-left/panels/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
+// @todo: consider moving code from index.tsx to separate files similar to other panels
```
kof, line 1: so far we have ben using index files to export public api only, they are like a definition of public API, so we don't write anything else but exports in there

apps/designer/app/designer/features/sidebar-left/panels/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+export const TabContent = ({
+  onSetActiveTab,
+  project,
+  pages,
+  currentPageId,
+  config,
+}: TabContentProps) => {
+  const pagesTree = useMemo(() => pages && toTreeData(pages), [pages]);
```
kof, line 78: if pages can be undefined, then the type is wrong

rpominov: Forgot to update. I used nano state at first with `undefined` as initial value. But this code will stay as I'm moving back to nano states.

apps/designer/app/designer/features/sidebar-left/panels/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const navigate = useNavigate();
+  const handleSelect = (pageId: string) => {
+    navigate(`${config.designerPath}/${project.id}?pageId=${pageId}`);
```
kof, line 82: There is this future use case that would be great to make possible: remotely connecting the designer to any site/app that uses SDK. Think of it as of chrome devtools connecting to v8 over protocol.

The question with the url's is that they are kindof owned by the application on canvas. Lets brainstorm how can we allow the application on canvas to fully own itself, including the urls.

apps/designer/app/designer/features/sidebar-left/panels/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  return (
+    <>
+      <Header title="Pages" onClose={() => onSetActiveTab("none")} />
+      <TreeNode
```
kof, line 88: we have also Instances tree, lets call them distinctively PagesTreeNode

rpominov: I plan to refactor this part in future PRs. This will be more complicated after more features are added, then it will be more clear how to structure the code better.

apps/designer/app/designer/features/sidebar-left/panels/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        onSelect={handleSelect}
+        itemData={pagesTree}
+        {...staticTreeProps}
+      />
+    </>
+  );
+};
+
+// @todo: proper icon
+export const icon = <StackIcon />;
```
kof, line 100: can use the "page" icon from figma

apps/designer/app/designer/features/sidebar-left/panels/pages/index.tsx
kof: I just realized that the TreeNode is the same for both instances and pages

rpominov: Yes, but there's also `InstanceTreeNode` that uses `TreeNode` inside. Probably will be the same for pages.

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -149,12 +149,15 @@ export type TreeNodeProps<Data extends { id: string }> = {
   level?: number;
   animate?: boolean;
   forceHoverStateAtItem?: string;
+
+  hideRoot?: boolean;
```
kof, line 153: just a thought: instead of a flag, it could be based on data, if data is array - there is no root, because it's basically the children

rpominov:
TODO in following PRs:

 - [ ] Use `page.path` as canvas URL.
 - [ ] Refactor code that uses `TreeNode`

packages/design-system/src/components/tree/tree-node.tsx
rpominov: May do it this way, but this complicates things

rpominov: Will think more about it when I'll be adding more functionality to the list. Will leave as is in this PR.

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -0,0 +1,126 @@
...
+    return (
+      <TreeNodeLabel
+        isSelected={props.isSelected}
+        text={props.data.data.name}
+      />
+    );
+  },
+};
+
+export const TabContentWithData = ({
```
kof, line 71:
```suggestion
export const Pages = ({
```

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -0,0 +1,126 @@
...
+  onSetActiveTab,
```
kof, line 72: pass onClose, move onSetActiveTab to TabContent, so that Pages has nothing to do with Tabs

kof: I did a similar thing for image manager recently, because we started reusing image manager and fonts manager UI that we use in tabs also in a floating window, we will need the same for pages because we will want to be able to link a page from props panel, so that page manager acts like a page picker

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
kof: And would put it into a separate module right away, can be local for now, we will move it to shared later, once we create page picker

kof: for context https://github.com/webstudio-is/webstudio-designer/pull/299

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
kof: Oh btw you willl want to use the hooks inside the Pages component, because you will want that pages component to get the data automatically when used as a picker in floating window

kof (changes requested):
lets make Pages content reusable outside of tabs and we are good to merge

apps/designer/app/designer/designer.tsx
```diff
@@ -268,6 +300,14 @@ export const Designer = ({ config, project, page }: DesignerProps) => {
     [publishRef, onRefReadCanvasWidth, onRefReadCanvas]
   );
 
+  const page = useMemo(() => {
```
kof, line 303: I feel like you will need this in many places, maybe instead of placing pageId into nano state it should be the page

kof (approved):

apps/designer/app/designer/designer.tsx
rpominov: Yeah, not sure. This is an intermediary PR so a lot of small things like this are still in flux, will be easier to tell which way is better later.

PR #322, opened 2022-10-06
CSS Value Input that combines the TextField and Combobox updates all in one

kof:
- [x] use new combobox and textfield in props
- [ ] fix visual regressions in style panel

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -10,19 +10,99 @@ import type { SelectedInstanceData } from "~/shared/canvas-components";
...
+    isOpen,
+  } = useCombobox({
+    items: itemsProp,
+    value,
+    itemToString: (item) => item ?? "",
+    onItemSelect,
+  });
+
+  return (
+    <ComboboxPrimitive>
```
okonet, line 65: I don’t like this API. They hide the real thing with no benefit. What is primitive and what it’s doing? What is anchor? Keep it as is by using Popper?

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -10,19 +10,99 @@ import type { SelectedInstanceData } from "~/shared/canvas-components";
...
+            name="prop"
+            placeholder="Property"
+            readOnly={isReadonly}
+            state={isInvalid ? "invalid" : undefined}
+            suffix={
+              <IconButton {...getToggleButtonProps()}>
+                <ChevronDownIcon />
+              </IconButton>
+            }
+            onKeyPress={(event) => {
```
okonet, line 79: I’d move this to props getter to be safe you don’t override what’s inside.

apps/designer/app/designer/features/props-panel/props-panel.tsx
kof: I could call it ComboboxPopper etc to make it more clear, I think it makes sense to reexport just to make sure we can do something in combobox later if needed

kof: changed

apps/designer/app/designer/features/props-panel/props-panel.tsx
kof: done

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -10,19 +10,100 @@ import type { SelectedInstanceData } from "~/shared/canvas-components";
...
+    isOpen,
+  } = useCombobox({
+    items: itemsProp,
+    value,
+    itemToString: (item) => item ?? "",
+    onItemSelect,
+  });
+
+  return (
+    <ComboboxPopper>
```
okonet, line 65: I still don't see any value in re-exporting at least not at this point but I don't care that much about it so it's okay.

packages/design-system/src/components/combobox.tsx
```diff
@@ -77,32 +118,26 @@ type useComboboxProps<Item> = {
   ) => Partial<UseComboboxStateChangeOptions<Item>>;
 };
 
-// eslint-disable-next-line func-style
-export function useCombobox<Item>({
+export const useCombobox = <Item,>({
```
okonet, line 121: I think it's a stupid rule to have. What's wrong with function keyword? How is the const better considering the ugly TS syntax?

packages/design-system/src/components/combobox.tsx
```diff
@@ -118,14 +153,10 @@ export function useCombobox<Item>({
...
 
   useEffect(() => {
-    cachedItems.current = items;
-  }, [items]);
-
-  useEffect(() => {
-    if (!isOpen) {
-      setFilteredItems(cachedItems.current);
+    if (isOpen === false) {
+      reset();
```
okonet, line 157: `resetFilter` would read better IMO.

okonet (approved):
Feel free to merge as all my comments are stylistic.

packages/design-system/src/components/combobox.tsx
kof: Another option is to enable Item extends unknown in the linter, so that it looks it least less like a hack. If you really want to discuss function keyword, we can do that separately

kof:
@okonet I am going to need to fix the styling still, before I merge, it still looks broken

packages/design-system/src/components/combobox.tsx
okonet: I actually don't want to discuss code style since it has 0 or even negative value. Spending time on fixing such issues just because we have this rule is a waste of everyone's time IMO. But it looks like you have a strong option about it so let's keep doing what we doing 🤷🏻‍♂️

kof: I agree, it doesn't really matter, I feel strongly about it because I am trying to enforce a more consistent structure of code, meaning the order of definition - first define, then use. With named functions people can create very different structure and make it harder to read overall, you can't scan the code consistently.
When a few people work on the project - it doesn't matter. At scale this flexibility makes it harder to read.
In this particular case it wouldn't matter, because its a small file that will stay small etc. But someone will look it up and do the same somewhere else, where its not a good thing any more. So I am trying to set an example.

kof:
@okonet going to merge this after last check from you, combobox in style is still visually wrong, but will get rid of it by porting the work from other PRs

PR #323, opened 2022-10-07
New unified route for canvas / preview / local publish

rpominov:
- Routes changed:

| Name | Old | New |
| - | - | - |
| Canvas | `/canvas/{projectId}?pageId={pageId}` | `/{pagePath}?projectId={projectId}&mode=edit`<br/>or<br/>`{projectDomain}.localhost/{pagePath}?mode=edit`
| Preview | `/preview/{projectId}?pageId={pageId}` | `/{pagePath}?projectId={projectId}&mode=preview`<br/>or<br/>`{projectDomain}.localhost/{pagePath}?mode=preview`
| Published | `{projectDomain}.localhost`<br/> (only home page was available) | `/{pagePath}?projectId={projectId}`<br/>or<br/>`{projectDomain}.localhost/{pagePath}`

- Canvas was unresponsive after you change the page, because it reloads and waits for `designerReady` forever. This is fixed.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/shared/use-designer-ready.ts
```diff
@@ -1,18 +1,30 @@
...
 declare module "~/shared/pubsub" {
   export interface PubsubMap {
-    designerReady: undefined;
-    designerReadyAck: undefined;
+    canvasReady: undefined;
+    canvasReadyAck: undefined;
   }
 }
 
 export const useSubscribeDesignerReady = () => {
```
kof, line 12: would make sense to also rename the hook and the file, no?

rpominov: I did that at first, but then realised it kinda still checks that designer is ready, but now by receiving `canvasReadyAck` instead of `designerReady`

kof:
@andrasbacsai since we are introducing pages and routing for multi-page sites, we need to check what needs to be adopted  from the publishing perspective

kof:
@andrasbacsai the only route we care on a published site is `{projectDomain}.localhost/{pagePath}`, where `pagePath` is user-defined and comes from the database

apps/designer/app/routes/$.tsx
```diff
@@ -0,0 +1,128 @@
...
+      env,
+    };
+  }
+};
+
+const Content = () => {
+  const data = useLoaderData<Data>();
+
+  if ("errors" in data) {
+    return <p>{data.errors}</p>;
```
kof, line 115: we added ErrorMessage component, but haven't started using it consistently, we should use it

apps/designer/app/routes/$.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  const Outlet =
+    data.mode === "edit"
+      ? () => <Canvas data={data} />
+      : () => <InstanceRoot data={data} />;
+
+  // @todo This is non-standard for Remix, is there a better way?
+  // Maybe there is a way to tell remix to use the right outlet somehow and avoid passing it?
+  return <Root Outlet={Outlet} />;
```
kof, line 125: The outlet is basically picked based on the route, so if we do custom routing, remix has no way to know

apps/designer/app/routes/$.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const loader: LoaderFunction = async ({
+  request,
+}): Promise<Data | Response> => {
+  try {
+    const canvasRequest = isCanvasRequest(request);
+
+    if (canvasRequest === false) {
+      return redirect(config.dashboardPath);
+    }
```
kof, line 82: if its not a request for canvas, shouldn't we yield to the router to handle it? What happens if its some other route e.g. /designer/something ?

apps/designer/app/routes/$.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+type Mode = "edit" | "preview" | "published";
+
+type Data =
+  | (CanvasData & { env: Env; mode: Mode })
+  | { errors: string; env: Env };
+
+type ProjectIdentifier = { type: "id" | "domain"; value: string };
+
+export const isCanvasRequest = (
```
kof, line 19: We should add tests for this at this stage, if pass request.url and request.headers, it will be easy to test

kof: Tests would be more important right now just to see all the cases it handles

apps/designer/app/routes/$.tsx
```diff
@@ -0,0 +1,128 @@
...
+    if (projectIdParam !== null) {
+      projectId = { type: "id", value: projectIdParam };
+    }
+  }
+
+  if (projectId === undefined) {
+    return false;
+  }
+
+  const modeParam = url.searchParams.get("mode");
```
kof, line 53: Maybe a bit more organized

if (modes.includes(mode) === false) { 
   throw new Error
}
return {projectId, mode, pathname}

kof: this would handle the case if mode param is wrong, so we don't fallback to published

apps/designer/app/routes/$.tsx
```diff
@@ -0,0 +1,128 @@
...
+    "";
+  const [userDomain, wstdDomain] = host.split(".");
+  if (wstdDomain === "wstd" || wstdDomain?.includes("localhost")) {
+    projectId = { type: "domain", value: userDomain };
+  }
+
+  if (projectId === undefined) {
+    const projectIdParam = url.searchParams.get("projectId");
+    if (projectIdParam !== null) {
+      projectId = { type: "id", value: projectIdParam };
```
kof, line 45: projectId as an object semantically seems a bit weird, because I assumed projectId is the actual id.

kof: I would just make the type part of the response object so that there is. {type: 'id', projected, mode, pathname}

apps/designer/app/routes/$.tsx
kof: isCanvasRequest sounds like a function that returns booleans only, but since it can return object, it needs a better name

apps/designer/app/routes/$.tsx
rpominov: Working on it. `publish({ type: "canvasWidth", payload: 600 });` that it does, doesn't work. Working on a fix.

apps/designer/app/routes/$.tsx
rpominov: I'm a bit confusing. Do you suggest a change or just clarifying the code comment?

apps/designer/app/routes/$.tsx
rpominov: This is for the case when you open `localhost:3000`, it just redirects you to dashboard. This logic is copied from `index.tsx`

apps/designer/app/routes/$.tsx
rpominov: Done

apps/designer/app/routes/$.tsx
rpominov: Renamed `projectId` to `projectIdObject`, hope that's less confusing.

I think `{type: 'id', projected, mode, pathname}` would be also incorrect semantically because `id / domain` is a type of data that is used to identify the project, it's not the type of the entire object that the functions returns, if that makes sense.

apps/designer/app/routes/$.tsx
kof: Oh it hought you commented lol, I was probably me, so this is me talking to past me

apps/designer/app/routes/$.tsx
kof: will it allow us having a route like /designer/something ? with the normal route handling

apps/designer/app/routes/$.tsx
kof: maybe call it parseUrl or something and return undefined when can't parse?

apps/designer/app/routes/$.tsx
kof: you could make it 2 separate object depending on if the data is identified by id vs domain ... or maybe there could be some other idea, projectIdObject is clear but not nice

kof: I think semantics of this still look weird:
```
    const project =
      projectId.type === "id"
        ? await db.project.loadById(projectId.value)
        : await db.project.loadByDomain(projectId.value);
        ```
        
        Because in one case projectId.value is a domain and in another case it's an id and reading this, you won't clearly see the difference unless you go digging the type

Comment (kof) [reply apps/designer/app/routes/$.tsx]: This would read much more straight forward
```
    const project =
      'projectId' in parsedUrl
        ? await db.project.loadById(parsedUrl.projectId)
        : await db.project.loadByDomain(parsedUrl.domain);
```

File: apps/designer/app/routes/$.tsx (hunk 4 above)
Comment (rpominov) [reply apps/designer/app/routes/$.tsx]: Sure. Remix fallbacks to `$.tsx` only if it didn't find any other matching route.

File: apps/designer/app/routes/$.tsx (hunk 3 above)
Comment (rpominov) [reply apps/designer/app/routes/$.tsx:127]: Yeah 😅

File: apps/designer/app/routes/$.tsx (hunk 5 above)
Comment (rpominov) [reply apps/designer/app/routes/$.tsx]: - added tests
- renamed to `getCanvasRequestParams`
- returns `undefined` if doesn't match

File: apps/designer/app/shared/router-utils/canvas-request-params.test.ts [hunk 8]
```diff
@@ -0,0 +1,91 @@
+import { getCanvasRequestParams } from "./canvas-request-params";
+
+const makeRequest = (url: string, headers: { [key: string]: string }) => ({
+  url,
+  headers: {
```
kof, line 5: isn't headers just a Map instance?

apps/designer/app/shared/router-utils/canvas-request-params.ts
```diff
@@ -0,0 +1,58 @@
...
+      projectIdObject: ProjectIdentifier;
+      mode: CanvasRouteMode;
+      pathname: string;
+    }
+  | undefined => {
+  const url = new URL(request.url);
+
+  const pathname = url.pathname;
+
+  if (Object.values(config).some((path) => pathname.startsWith(path))) {
```
kof, line 23: this would return true for non-canvas paths too, e.g. dashboardPath, is that what you wanted?

apps/designer/app/routes/$.tsx
rpominov: Did it as you suggested in the last comment. Good idea!

apps/designer/app/shared/router-utils/canvas-request-params.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  if (projectDomain !== undefined) {
+    return {
+      projectDomain,
+      mode,
+      pathname,
+    };
+  }
+
+  return undefined;
```
kof, line 73: did typescript force you to write that? its normally the default behavior to return undefined, just wondering

apps/designer/app/shared/router-utils/canvas-request-params.ts
rpominov: All paths that are remaining in `config` are not canvas paths. This line is correct. It checks for example that you're trying to access `/dashboard` and `getCanvasRequestParams` returns `undefined` in this case

apps/designer/app/shared/router-utils/canvas-request-params.ts
rpominov: No, I simply wrote it without thinking twice about it :)

apps/designer/app/shared/router-utils/canvas-request-params.ts
```diff
@@ -0,0 +1,74 @@
...
+    return undefined;
+  }
+
+  let projectDomain = undefined;
+  let projectId = undefined;
+
+  const projectIdParam = url.searchParams.get("projectId");
+  if (projectIdParam !== null) {
+    projectId = projectIdParam;
+  }
```
kof, line 36: nit: here is a one-liner to do the same thing you did in 5 lines
`const projectId =  url.searchParams.get("projectId") ?? undefined`
but also why do you care about null if you never return a null from your function, so you literally won't expose null from the interface

apps/designer/app/shared/router-utils/canvas-request-params.test.ts
rpominov: I don't think so. In typescript it's defined separately from Map.

apps/designer/app/shared/router-utils/canvas-request-params.ts
rpominov: Fixed

apps/designer/app/shared/router-utils/canvas-request-params.ts
kof: but there is no guarantee that config will only have non-canvas paths, right? also that config is not just for paths, it could get something else added to it ... later and the person who adds to config won't know this logic depends on it in such a way

apps/designer/app/designer/features/breakpoints/use-update-canvas-width.ts
```diff
@@ -32,7 +31,16 @@ export const useUpdateCanvasWidth = () => {
...
   }, [selectedBreakpoint, setCanvasWidth]);
 
-  useSubscribe("canvasWidth", setCanvasWidth);
+  // Handle the case when canvas loads with a error
+  // And doesn't send any messages that it's supposed to send
+  useEffect(() => {
+    if (canvasWidth === 0) {
+      const timeout = setTimeout(() => {
+        setCanvasWidth(600);
+      }, 1000);
```
kof, line 40: this sounds fishy, because it's an arbitrary number, what if it loads with an error and it takes longer?

kof (approved):

apps/designer/app/designer/features/breakpoints/use-update-canvas-width.ts
rpominov: The delay is to avoid setting it to 600 before it's set to a breakpoint value, which causes a glitch. If it takes longer to load the error that's not an issue.

apps/designer/app/shared/router-utils/canvas-request-params.ts
rpominov: I've removed this check from `getCanvasRequestParams` now, and added an alternative check in `apps/designer/app/entry.server.tsx`. The new check doesn't use `config`.

rpominov:
Not sure these new routes is a good idea overall. We're basically allowing anyone to create arbitrary pages on ours site, the only limitation is that the page will have to have a `?projectId` parameter. For example someone can create something like `https://alpha.webstudio.is/about?projectId=...` or `/donations` and send people emails asking to donate to webstudio but in reality put their credentials at the page. Probably there're a lot more security concerns, I'm just not a security expert.

kof:
A valid concern, it could be all on a subdomain?

kof:
ah it's already on the subdomain alpha.webstudio.is, you think this is not enough?

kof:
We also need to change the URL schema for the launch from "alpha" subdomain, so its a good time to rethink it. An option could be to run the entire designer on wstd.io

kof:
Another idea is to use design.webstudio.is as a designer domain, this way we could also drop the /designer path and on canvas use build.webstudio.is, which would be for the build

rpominov:
> ah it's already on the subdomain alpha.webstudio.is, you think this is not enough?

In any case users will be able to put anything at the same domain where we run the designer. To me this seems very dangerous.

kof:
> In any case users will be able to put anything at the same domain where we run the designer. To me this seems very dangerous.

well its not the same domain if we use something like build.webstudio.is, which is a different subdomain

PR #326, opened 2022-10-10
Isolated domains for user-editable content

### LINKED ISSUE #324: Isolated domains for user-editable content
Currently users are able to put arbitrary content at the same domain where we run the designer app. For example at `https://alpha.webstudio.is/preview/{projectId}` and [with the new routing it will be even worse](https://github.com/webstudio-is/webstudio-designer/pull/323#issuecomment-1272344796).

This is a security concern. This may allow bad actors to:
 - steal cookies,
 - create fake login pages,
 - create other fake pages like `/donations`,
 - etc.
 
Proposed solution:

Any user generated content should be available only from an isolated domain like `{projectId or projectDomain}.alpha.webstudio.is`. Ideally something like this:

- https://test1-61bhz.wstd.io/ — published website
- https://test1-61bhz.wstd.live/ — canvas / preview / etc.

This can be achieved using wildcard domains at [Vercel](https://vercel.com/blog/wildcard-domains) or [Cloudflare](https://developers.cloudflare.com/dns/manage-dns-records/reference/wildcard-dns-records/).

For the preview deploys at Vercel it's ok to keep everything at the same domain as they are used only for development, and setting up separate domain for preview deploys can be extra tricky.

rpominov:
Fixes #324

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/.env.example
```diff
@@ -26,3 +26,12 @@ ASSET_CDN_URL=
 PUBLISHER_ENDPOINT=
 PUBLISHER_HOST=
 PUBLISHER_TOKEN=
+
+# Base URL for accessing user-generated content (canvas / preview / etc)
+# Should contatin protocol and host (e.g. https://example.com)
+# Will be detected automatically in development environments like localhost or preview deployment
+USER_CONTENT_BASE_URL=
```
kof, line 33: How about BUILD_ORIGIN

Screenshot 2022-10-11 at 18 08 04

apps/designer/.env.example
```diff
@@ -26,3 +26,12 @@ ASSET_CDN_URL=
...
+
+# If set to `true`, user-generated content will be available 
+# only at {project.domain} subdomain of {USER_CONTENT_BASE_URL}
+USER_CONTENT_REQUIRE_SUBDOMAIN=
```
kof, line 37: lets add as comment when do we need it?

apps/designer/app/routes/$.tsx
```diff
@@ -52,8 +57,18 @@ export const loader: LoaderFunction = async ({
       pathname
     );
 
+    if (canvasData === undefined) {
+      throw json("Page not found", { status: 404 });
```
kof, line 61: Never saw  json() being thrown that works?

kof: ah you are catching it below

apps/designer/app/shared/router-utils/index.ts
```diff
@@ -1 +1 @@
-export * from "./canvas-request-params";
+export * from "./user-content-params";
```
kof, line 1: user content is what we call "build", right?

apps/designer/app/shared/router-utils/user-content-params.ts
```diff
@@ -0,0 +1,90 @@
...
+  const { USER_CONTENT_BASE_URL } = process.env;
+  if (USER_CONTENT_BASE_URL !== undefined && USER_CONTENT_BASE_URL !== "") {
+    return USER_CONTENT_BASE_URL;
+  }
+
+  // Local development special case
+  const host = getRequestHost(request);
+  if (
+    process.env.NODE_ENV === "development" &&
+    /^(.*\.)?localhost(:\d+)?$/.test(host)
```
kof, line 32: why not new URL(host).hostname?

apps/designer/app/shared/router-utils/user-content-params.ts
```diff
@@ -0,0 +1,90 @@
...
+// A subtype of Request. To make testing easier.
+type MinimalRequest = {
+  url: string;
+  headers: { get: (name: string) => string | null };
+};
+
+const getRequestHost = (request: MinimalRequest): string =>
+  request.headers.get("x-forwarded-host") || request.headers.get("host") || "";
+
+export const getUserContentBaseUrl = (request: MinimalRequest): string => {
```
kof, line 22: seems like it would be nicer to pass the env as an argument, given how the tests look like

apps/designer/app/shared/router-utils/user-content-params.ts
```diff
@@ -0,0 +1,90 @@
...
+      mode: CanvasRouteMode;
+      pathname: string;
+    }
+  | undefined => {
+  const url = new URL(request.url);
+
+  const requestHost = getRequestHost(request);
+  const userContentHost = new URL(getUserContentBaseUrl(request)).host;
+
+  if (process.env.USER_CONTENT_REQUIRE_SUBDOMAIN !== "true") {
```
kof, line 68: please check that all new env variables are validated if they are required in env-check.ts

apps/designer/app/shared/router-utils/user-content-params.ts
```diff
@@ -0,0 +1,90 @@
...
+  const baseHost = rest.join(".");
+  if (baseHost === userContentHost) {
+    return {
+      projectDomain,
+      mode: getMode(url),
+      pathname: url.pathname,
+    };
+  }
+
+  return;
```
kof, line 89: no need

kof (approved):
some minor things

apps/designer/.env.example
rpominov: No strong opinion, but USER_CONTENT describes the intent better in my view. 

As for ORIGIN instead of BASE_URL, need to check definition of ORIGIN, but probably a good idea.

apps/designer/.env.example
rpominov: Do you mean at which deployment type?

apps/designer/app/shared/router-utils/index.ts
rpominov: Not sure. Do you mean something related to Build type in database?

In this PR it means content at canvas / preview. We can use any term, but "build" is not intuitive for me at all, at least right now.

apps/designer/app/shared/router-utils/user-content-params.ts
rpominov: I don't understand. This code checks that `host` looks like `"localhost"` or `"foo.localhost"` or `"foo.localhost:3000"`.

apps/designer/app/shared/router-utils/user-content-params.ts
rpominov: Done

apps/designer/app/shared/router-utils/user-content-params.ts
rpominov: Feels weird to pass env since it's globally available. Maybe just to simplify tests?

apps/designer/.env.example
kof: the problem with user content is that this is the second term that describes user's build, we need to agree on one, unless you see a difference between user content and build

apps/designer/.env.example
kof: yeah, practically when do we use it

apps/designer/app/shared/router-utils/index.ts
kof: no. the entire thing on canvas is user's "build", tat's how we call it so far in a lot of places, it's what user builds

apps/designer/app/shared/router-utils/user-content-params.ts
kof: oh I see, I just wish we would avoid regexes if we can use some well-tested parser, at least add the examples please to make it clear what its for

apps/designer/app/shared/router-utils/user-content-params.ts
kof: I get it, though its in the end better then all the mocking strategies I have seen so far, it can be optional and default to process.env, so normal consumers don't have to pass it

apps/designer/app/shared/router-utils/index.ts
rpominov: Ok. Still not intuitive for me to be honest 😅  
But now I know, and will rename here.

kof:
Looking nice!

PR #328, opened 2022-10-13
Ability to add a page

rpominov:
Another part of #54 

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -1,30 +1,84 @@
...
-  }
+const nonEmptyString = (fieldName: string) => {
+  const message = `${fieldName} can't be empty`;
+  return z
+    .string({ invalid_type_error: message, required_error: message })
+    .transform((v) => v.trim())
+    .refine((val) => val !== "", message);
+};
+
+const createPageInputSchema = zfd.formData({
```
kof, line 17: 1. Schema name is always capitalized because it corresponds to the type. 
2. I started using Schema suffix and then regretted it, because when not having it we cal literally have the same name for both schema and the type without conflicts, its actually what they use as a convention in zod

```
const Dog = z.object({
  name: z.string(),
  age: z.number(),
});

// extract the inferred type like this
type Dog = z.infer<typeof Dog>;
```


```suggestion
const CreatePageInput = zfd.formData({
```

rpominov: I think keeping `Schema` suffix can be useful in cases when you want to export both the type and the schema

kof: that's a thing I didn't know when I used the suffix, you can apparently export both and then when importing you can say if you want the schema or the type using the type 

import {type Test} from '...' // imports type
import {Test} from '...' // imports schema

kof: oh wait, you are saing `import { type Test, Test } from "...";` is impossible???

rpominov: Yeah, TypeScript doesn't allow you to import both under the same name.

kof: oh you don't need to import both, if you import just the schema namespace you can use it as a type

kof: So basically type declaration on the same namespace creates an alias on that namepsace and so it becomes both type and code and can be used for both

rpominov: Yeah, it works. Cool!

```
import { Test } from "./pages.$projectId";
const _test: Test = Test.parse("test");
```

apps/designer/app/designer/features/sidebar-left/lib/header.tsx
```diff
@@ -2,33 +2,44 @@ import { Flex, IconButton, Text } from "@webstudio-is/design-system";
...
-type HeaderProps = {
+type BaseHeaderProps = {
   title: string;
-  isClosable?: boolean;
-  onClose?: () => void;
+  actions: React.ReactNode;
 };
 
-export const Header = ({ title, isClosable = true, onClose }: HeaderProps) => {
+export const BaseHeader = ({ title, actions }: BaseHeaderProps) => {
```
kof, line 10: Here are suggestions:
1. lets rename actions to "suffix". We started using this pattern in design system, its more generic, so you can pass there anything you want, not only actions
2. lets rename BaseHeader to Header and delete the current Header
3. Lets export frequently used buttons to be composed with the header, e.g. the CloseButton

so the typical header with close button would look like this:

```
<Header suffix={<CloseButton onClick={...}/>}>
```

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -110,16 +137,44 @@ export const TabContent = (props: TabContentProps) => {
...
-      selectedPageId={currentPageId}
-    />
+    <>
+      <PagesPanel
+        onClose={() => props.onSetActiveTab("none")}
+        onNewPage={() => {
+          if (newPageOpen === false) {
+            setNewPageOpen(true);
+          }
+          if (newPageOpen && newPageState === "idle") {
```
kof, line 155: is that preventing from closing the new page content when it's saving or loading? Should we be preventing it?

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const isSubmitting = fetcher.state !== "idle";
+
+  const prevFetcher = useRef(fetcher);
+  useEffect(() => {
+    if (prevFetcher.current.state !== fetcher.state) {
+      onFetcherStateChange?.(fetcher.state);
+      if (fetcher.state === "idle" && fetcher.data !== undefined) {
+        if ("ok" in fetcher.data) {
+          onSuccess?.(fetcher.data.page);
+        } else {
```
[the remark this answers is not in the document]
kof: generally prefer early returns and short conditions assigned over variables, makes it easier to read

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
rpominov: > is that preventing from closing the new page content when it's saving or loading?

Yeah

> Should we be preventing it?

Do you think it's a bad idea? Form submition may fail, and if the panel is closed by that time user will not see the errors.

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -1,6 +1,89 @@
...
+import { z } from "zod";
+import { CreatePageData } from "~/designer/features/sidebar-left/panels/pages/settings";
+import { sentryException } from "~/shared/sentry";
+
+const nonEmptyString = z
+  .string({
+    invalid_type_error: "Can't be empty",
+    required_error: "Can't be empty",
+  })
+  .transform((v) => v.trim())
```
kof, line 14: please don't use single char vars

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -1,6 +1,89 @@
...
+  .refine((val) => val !== "", "Can't be empty");
```
kof, line 15: please use full words, e.g.value

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -1,6 +1,89 @@
...
+      (path) => path === "" || path.startsWith("/"),
+      "Must start with a /"
+    )
+    .refine(
+      (path) => /^[-_a-z0-9\\/]*$/.test(path),
+      "Only a-z, 0-9, -, _ and / are allowed"
+    ),
+});
+
+const handlePUT = async (
```
kof, line 33: Lets use strict camel case without accounting for abbreviations or anything. imageUrl etc. ... makes it much easier to keep consistent and convert automatically when needed
```suggestion
const handlePut = async (
```

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -1,6 +1,89 @@
...
+  return { ok: true, page: newPage };
+};
+
+export const action: ActionFunction = async ({ request, params }) => {
+  try {
+    if (params.projectId === undefined) {
+      throw new Error(`Project ID required`);
+    }
+
+    if (request.method.toLowerCase() === "put") {
```
kof, line 74: isn't it consistently upper case from remix? So it doesn't depend what you used to send it, would be simpler to use it as-is instead of converting the case every time, well unless it's actually not consistent

packages/design-system/src/components/tooltip.tsx
```diff
@@ -82,3 +82,31 @@ export const Tooltip = React.forwardRef(function TooltipWrapper(
...
 });
+
+export const InputErrorsTooltip = ({
+  errors,
+  children,
+}: {
+  errors: string[] | undefined;
+  children: React.ComponentProps<typeof Tooltip>["children"];
+}) => {
+  const content = errors?.map((error, i) => (
```
kof, line 93: single char variable

packages/design-system/src/components/tooltip.tsx
```diff
@@ -82,3 +82,31 @@ export const Tooltip = React.forwardRef(function TooltipWrapper(
     </TooltipPrimitive.Root>
   );
 });
+
+export const InputErrorsTooltip = ({
+  errors,
+  children,
+}: {
+  errors: string[] | undefined;
```
kof, line 90: nit
```suggestion
  errors?: string[];
```

packages/design-system/src/components/tooltip.tsx
```diff
@@ -82,3 +82,31 @@ export const Tooltip = React.forwardRef(function TooltipWrapper(
...
+  children: React.ComponentProps<typeof Tooltip>["children"];
```
kof, line 91: Nit: import ComponentProps, Fragment etc. Reads nicer but definitely a subjective preference

apps/designer/app/routes/rest/pages.$projectId.ts
rpominov: I don't know whether it's consistent. This is not documented as far as I know

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
kof: Generally preventing user from opening/closing things and making them to wait is always bad.
In this case what we can do is:
1. indicate we are sending, so they know they may miss something if they close (it's an edge case anyways, it should be fast enough not to notice)
2. if its easy enough we can resort to a toast if user has already closed, but if it makes logic complicated - don't worry, its not a normal case

Important is that the user knows what's happening when sending the state.

apps/designer/app/routes/rest/pages.$projectId.ts
kof: Actually I just remembered that I was using a lower case POST/DELETE on the client and received upper case on the server, so it is consistent

kof: in the sense that on the request.method its always upper case

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
rpominov: Yeah, `#2` will be too much complication. I've made it so the button says `Creating...` for now, maybe not ideal solution but this is not a final design anyway

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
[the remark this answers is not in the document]
rpominov: We cannot return early, because we want `prevFetcher.current = fetcher;` at the end to always run.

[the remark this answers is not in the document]
rpominov: Not sure I understand the idea with `hasFetched`. The way you wrote it will not work for two reasons:

1. TypeScript will not understand that if  `hasFetched` is true, then `fetcher.data` is not undefined
2. If we move `hasFetched` together with `"ok" in fetcher.data`, the else condition will not work as needed

Let's keep as is? Don't see any reason to spend time discussing this really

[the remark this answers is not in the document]
kof: I haven't spent much time in the example, the point I am making is to avoid nested if/else style, because indentation makes it harder to read. Its probably subbjective

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -86,7 +96,19 @@ const PagesPanel = ({
 
   return (
     <>
-      <Header title="Pages" onClose={onClose} />
+      <Header
+        title="Pages"
+        suffix={
+          <>
+            {onNewPage && (
+              <IconButton size="2" onClick={onNewPage} aria-label="New Page">
```
kof, line 104: I would inline a callback to avoid leaking the click event to a handler that shouldn't know about implementation detail and to make semantics of onNewPage make more sense because that callback could be called without a click event from some other logic, just theoretically.

apps/designer/app/designer/features/sidebar-left/panels/pages/settings-panel.tsx
```diff
@@ -0,0 +1,61 @@
...
+const CollapsibleContent = styled(Collapsible.Content, {
+  overflow: "hidden",
+  flexGrow: "1",
+  display: "flex",
+  flexDirection: "column",
+  '&[data-state="open"]': { animation: `${openKeyframes} 200ms ease-out` },
+  '&[data-state="closed"]': { animation: `${closeKeyframes} 200ms ease-out` },
+});
+
+export const SettingsPanel = ({
```
kof, line 37: In the end this panel will be needed for more than pages, but we can keep it here for now if you think its risky to make it reusable right away

kof:
can you please enable pages on preview over env? So we can test it

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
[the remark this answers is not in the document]
rpominov: I get it, but there're reasons to write it this way in this case

apps/designer/app/designer/features/sidebar-left/panels/pages/settings-panel.tsx
rpominov: Will do this later, it's kinda WIP yet

rpominov:
> can you please enable pages on preview over env? So we can test it

Enabled

rpominov:
Just noticed header buttons have broke at some point. Fixing in a new PR.

Screenshot 2022-10-19 at 17 27 14

rpominov:
The `width: "min($6, 100%)"` introduced in this commit broke the buttons https://github.com/webstudio-is/webstudio-designer/commit/9d5d219806a1ce435fd1ef961d114af9d909e245

Not sure how to best to fix this. The `min` does not allow container of the buttons to grow by content it seems

Issue #333, opened 2022-10-17
Remaining CssValueInput todos


kof:
- [x] This interaction doesn't preserve the unit (Unit mode > select all > backspace > enter number)

https://user-images.githubusercontent.com/52824/195978905-bc14d810-3638-4711-a0fa-0f15bc61677a.mp4

- [x]  On blur should not activate the property name label showing that it has changed when the value hasn't changed,

- [x] default cursor over input should be "default", not scrub

- [x] don't bind pointermove event if there was no pointerdown event

- [x] allow vertical movement when scrubbing horizontally, so that user can avoid covering the number by moving up/down

- [x] match the styles of the combobox items with figma

- [x] Questions around scrubbing behavior:
  - should we select text when scrub starts?
  - should scrub cursor be delayed using threshold distance?
  - when input is focused, should scrub be still enabled?
  - when input focused and scrub starts, should text get selected?

taylornowotny:
There's one UX pattern I'm not sure about that I would like to discuss. When scrubbing, if the cursor is dragged away from the input field, the cursor will jump back to the input when the click is released. 

This doesn't feel right to me, but this is how figma does it. I'm curious to know what others think. In my view, it's never a good idea to take control of the cursor position from the user. It feels wrong, like if your own hand suddenly moved position in space without you moving it. 

That's all subjective though - what do others think?


https://user-images.githubusercontent.com/86495896/196234290-206dc1c1-9787-4c6a-aec4-2473f6dfad74.mov

kof:
@taylornowotny that's basically web platform limitation, when you do mouseup, we show you your actual cursor, its where it has always been. While you are dragging we hide the real cursor and show you a fake one. This is how we can go over the window edge and come out on the other side. 

This is not fixable.

taylornowotny:
> @taylornowotny that's basically web platform limitation, when you do mouseup, we show you your actual cursor, its where it has always been. While you are dragging we hide the real cursor and show you a fake one. This is how we can go over the window edge and come out on the other side.
> 
> This is not fixable.

Understood! All good then. Users who are accustomed to the figma behavior may expect this behavior anyways

taylornowotny:
Just a few remaining issues before it's perfect!

- [ ] The scrub cursor shows immediately on hover. The primary action that an input will be used for is changing the value, while scrubbing is secondary. For that reason I think the cursor should be the default cursor when the input is hovered.
[...]

PR #334, opened 2022-10-18
Fix stale unit change in scrub logic

kof:
https://github.com/webstudio-is/webstudio-designer/issues/333

Reproduction: 

1. Click on any unit text control
2. Select a different unit
3. Start scrubbing 
4. See unit is not reverted back to previous value

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -83,15 +83,11 @@ const useScrub = (options: {
 }) => {
   const inputRef = useRef<HTMLInputElement | null>(null);
   const optionsRef = useRef(options);
-
-  useEffect(() => {
-    optionsRef.current = options;
-  }, [options]);
+  optionsRef.current = options;
 
   useEffect(() => {
     const { value, onChange, onChangeComplete } = optionsRef.current;
     if (value.type !== "unit" || inputRef.current === null) return;
-
     const scrub = numericScrubControl(inputRef.current, {
       initialValue: value.value,
       onValueInput(event) {
@@ -109,7 +105,7 @@ const useScrub = (options: {
     });
 
     return scrub.disconnectedCallback;
-  }, [options.value.type]);
+  }, [options.value.type, options.value?.unit]);
 
   return inputRef;
 };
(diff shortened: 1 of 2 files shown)


apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -109,7 +105,7 @@ const useScrub = (options: {
     });
 
     return scrub.disconnectedCallback;
-  }, [options.value.type]);
+  }, [options.value.type, options.value?.unit]);
```
rpominov, line 108: I would use `options.value` inside the `useEffect` body instead of using `optionsRef.current.value`, so that eslint would detect it as a dependency and catch bugs like this.

Maybe would also put only `onChange` and `onChangeComplete` into a ref to make more clear the intention of ignoring changes of the callbacks identity.

Doesn't really affect behaviour, so not super important if hard to do for some reason, just would be a bit more correct.

rpominov (approved):

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: The problem is that we are calling onChange here and that onChange is going to update the value, so we end up in an infinite loop, I agree though what I did doesn't look nice and clear at all

kof: trying something out rn

kof: ok I tried to make it better, not sure I succeeded

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -76,36 +76,52 @@ const useHandleOnChange = (
...
-  optionsRef.current = options;
+  const onChangeRef = useRef(onChange);
+  onChangeRef.current = onChange;
+  const onChangeCompleteRef = useRef(onChangeComplete);
+  onChangeCompleteRef.current = onChangeComplete;
+  const valueRef = useRef(value);
+  valueRef.current = value;
+
+  const type = valueRef.current.type;
+  const unit = type === "unit" ? valueRef.current.unit : undefined;
```
rpominov, line 97: This code should be identical to:

```
const type = value.type;
const unit = type === "unit" ? value.unit : undefined;
```

Because the line above does `valueRef.current = value`.

Otherwise this is more or less what I meant, yeah :)

PR #335, opened 2022-10-19
Style fixes for buttons and unit icon-button

kof:
## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -25,13 +25,12 @@ import {
 } from "react";
 import { useUnitSelect } from "./unit-select";
 
-const defaultKeywords: [] = [];
-const defaultValue: UnsetValue = { type: "unset", value: "" };
+const unsetValue: UnsetValue = { type: "unset", value: "" };
 
 const isNumericString = (input: string) =>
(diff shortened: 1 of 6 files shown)


apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -200,8 +200,8 @@ type CssValueInputProps = {
 
 export const CssValueInput = ({
   property,
-  value = defaultValue,
-  keywords = defaultKeywords,
+  value = unsetValue,
+  keywords = [],
```
rpominov, line 204: The previous version when array was created outside was a bit better because now a new value will be created on every render, which may cause extra updates down the line if `keywords` are used as a dependency somewhere.

Not sure whether this is a concern in this case, but I'm guessing that's why it was outside before

rpominov (approved):

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: nah, don't think its a concern, was outside because it was used inside as a fallback and would actually prevent an update in case keywords were default, so that its the same ref, but its not used any more twice, so I moved it

PR #351, opened 2022-10-22
Fonts manager & selector

kof:
## Scope

- fonts manager (click on font family value in style panel)
- text-field component fixes
- search-field component

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


kof:
Todo

- [ ] update list after upload
- [ ] make TextField type search use search icon and close button, reuse it in style panel and fonts manager
- [ ] when no uploaded fonts, keys navigation doesn't match items, selection of new font has weird effects
- [ ] style font input-button (should text-field handle type button?)

apps/designer/app/designer/shared/fonts-manager/fonts-manager.tsx
```diff
@@ -80,62 +92,116 @@ export const FontsManager = ({ value, onChange }: FontsManagerProps) => {
...
+    onItemSelect(value) {
+      if (value !== null) {
+        onChange(value.label);
+      }
+    },
+  });
+
+  const items = groupItems(filteredItems);
+
+  useEffect(resetFilter, [fontItems.length, resetFilter]);
```
rpominov, line 123: Is the intention to run `resetFilter` when `fontItems.length` changes? Should `resetFilter` be a dependency?

Also, React does not guarantee that it will run effect only when dependencies change, they are just a hint for performance. Although, I couldn't find this in their current docs, so maybe they've gave up and guarantee it now 😄

apps/designer/app/designer/shared/fonts-manager/fonts-manager.tsx
```diff
@@ -80,62 +92,116 @@ export const FontsManager = ({ value, onChange }: FontsManagerProps) => {
...
+                {item.type === "category" && index !== 0 && (
+                  <Separator css={{ my: "$1" }} />
+                )}
+                <ListboxItem
+                  {...getItemProps({
+                    item,
+                    index,
+                    highlighted: item === openMenuItem ? true : undefined,
+                  })}
+                  disabled={item.type === "category"}
```
rpominov, line 185: Instead of using a special "category" item, I would probably split array into two here, and render:

 - Category header
 - First array
 - Category header
 - Second array

kof: yep, I hit a bunch of bugs because of how I did it, will redo it now like you said

apps/designer/app/designer/shared/fonts-manager/fonts-manager.tsx
```diff
@@ -1,70 +1,56 @@
...
+import { ItemMenu } from "./item-menu";
+import { Listbox, ListboxItem } from "./listbox";
 
-const getItems = (
-  assets: Array<Asset | PreviewAsset>
-): Array<{ label: string }> => {
-  const system = Array.from(SYSTEM_FONTS.keys()).map((label) => ({ label }));
+type Item = {
+  label: string;
+  type: "uploaded" | "system" | "category";
```
rpominov, line 16: Do we still need category here?

kof: nope, removing

taylornowotny:
- [ ] Tabbing from the search field to a font item and pressing ENTER/SPACE does not change the active font.
- [ ] Arrow keys can be used to move the selection between list items when the search field is focused, but arrow keys cannot be used to move the focus between list items. I think it would be preferable for arrow keys to work in both cases. As a user I expect to be able to navigate any list using arrow keys (after focus has moved to the list with TAB).
- [ ] Do not show the "subsection line" above the topmost category. Looks good when there's an uploaded font category but when there's only system fonts, the line should be hidden.
Screen Shot 2022-10-27 at 1 56 13 PM
Screen Shot 2022-10-27 at 1 57 20 PM
- [ ] Deleting a custom font while the custom font is current adds Arial to the search field. It should select Arial automatically, but Arial shouldn't be in the search field because that hides the other fonts until the user deliberately clears the search field. Also, I'm not sure that Arial is actually being selected. It seems like it's selected in the menu but not actually applied to the project.
- [ ] When moving the focus with TAB, the last list item will remain focused while the focus moves to the close (X) button. Pressing ENTER closes the window as expected. The highlight should be removed from the last item when focus moves off of that item.
Screen Shot 2022-10-27 at 2 17 25 PM

Additional note: There is a need for a progress indicator when uploading fonts. I will design this so we can talk about it and add it later.

taylornowotny:
Here's a video of what happens when you delete a custom font that is the current font.
https://share.descript.com/view/ugPrj53F4tk

kof:
@taylornowotny gread finds, I am going to fix them all in a separate PR

PR #352, opened 2022-10-24
Fix a crash in canvas when "Block third-party cookies" is enabled in Chrome

### LINKED ISSUE #347: Different domain in iframe does not work with "Block third-party cookies" enabled in Chrome
To reproduce:

 - Go to chrome://settings/cookies
 - Set "Block third-party cookies" radio-button to selected
 - Open designer at prod (https://alpha.webstudio.is/designer/{projectID})
 - The canvas is unable to connect to the designer (for example, clicking on canvas doesn't make an item selected)
 
Need to find a way to make it work with the setting enabled. If that's impossible, implement some UI that tells the user to disable the setting.

---

https://bugs.chromium.org/p/chromium/issues/detail?id=357625
https://github.com/remix-run/remix/issues/3659

rpominov:
Fixes #347

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/canvas.tsx
+++ b/apps/designer/app/canvas/canvas.tsx
@@ -65,6 +65,7 @@ const useElementsTree = () => {
     if (rootInstance === undefined) return;
 
     return createElementsTree({
+      embeded: true,
       instance: rootInstance,
       breakpoints,
       Component: WrapperComponentDev,
--- a/apps/designer/app/canvas/canvas.tsx
+++ b/apps/designer/app/canvas/canvas.tsx
@@ -65,7 +65,7 @@ const useElementsTree = () => {
     if (rootInstance === undefined) return;
 
     return createElementsTree({
-      embeded: true,
+      sandbox: true,
       instance: rootInstance,
       breakpoints,
       Component: WrapperComponentDev,
--- a/packages/react-sdk/src/tree/session-storage-polyfill.tsx
+++ b/packages/react-sdk/src/tree/session-storage-polyfill.tsx
@@ -43,5 +43,8 @@ const polyfill = function () {
 }.toString();
 
 export const SessionStoragePolyfill = () => (
-  <script dangerouslySetInnerHTML={{ __html: `(${polyfill})()` }} />
+  <script
+    dangerouslySetInnerHTML={{ __html: `(${polyfill})()` }}
+    suppressHydrationWarning
+  />
 );
--- a/packages/react-sdk/src/tree/session-storage-polyfill.tsx
+++ b/packages/react-sdk/src/tree/session-storage-polyfill.tsx
@@ -2,11 +2,6 @@
 
 // The code is based on this discussion https://bugs.chromium.org/p/chromium/issues/detail?id=357625
 const polyfill = function () {
-  // run only if we're in an iframe
-  if (window.location === window.parent.location) {
-    return;
-  }
-
   try {
     const key = "__session_storage_availability_test__";
     sessionStorage.setItem(key, "test"); // test
--- a/packages/react-sdk/src/tree/session-storage-polyfill.tsx
+++ b/packages/react-sdk/src/tree/session-storage-polyfill.tsx
@@ -2,6 +2,11 @@
 
 // The code is based on this discussion https://bugs.chromium.org/p/chromium/issues/detail?id=357625
 const polyfill = function () {
+  // run only if we're in an iframe
+  if (window.location === window.parent.location) {
+    return;
+  }
+
   try {
     const key = "__session_storage_availability_test__";
     sessionStorage.setItem(key, "test"); // test
--- a/packages/react-sdk/src/tree/session-storage-polyfill.tsx
+++ b/packages/react-sdk/src/tree/session-storage-polyfill.tsx
@@ -1,0 +1,35 @@
+// This is a temporary work around for https://github.com/remix-run/remix/issues/3659
+
+// The code is based on this discussion https://bugs.chromium.org/p/chromium/issues/detail?id=357625
+const polyfill = `(function () {
+  try {
+    const key = "__session_storage_availability_test__";
+    sessionStorage.setItem(key, "test");
+    sessionStorage.removeItem(key);
+  } catch (e) {
+    const data = new Map();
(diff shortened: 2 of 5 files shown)


rpominov:
Build fails due to https://github.com/webstudio-is/webstudio-designer/pull/298#issuecomment-1288769465

packages/react-sdk/src/tree/session-storage-polyfill.tsx
```diff
@@ -0,0 +1,35 @@
+// This is a temporary work around for https://github.com/remix-run/remix/issues/3659
```
kof, line 1: uhh, this looks fishy: consumers will think sessionStorage works, but it basically just shifts the problem to warnings instead of error. Shouldn't consumers be testing if sessionStorage is available before using it?

kof: what is currently using sessionStorage?

rpominov: Remix uses it to restore scroll position. They need to fix it on their side. Hopefully they'll do it as the issue is already open, then we can remove this workaround.

packages/react-sdk/src/tree/session-storage-polyfill.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+// The code is based on this discussion https://bugs.chromium.org/p/chromium/issues/detail?id=357625
+const polyfill = `(function () {
```
kof, line 4: its such a weird way to do the polyfill ... and then render it specifically before scrollrestoration over react ...
can't it just be imported as a normal script, before anything else runs?

kof: we will have the same issue if we try access sessionStorage in some script if it runs before this

packages/react-sdk/src/tree/session-storage-polyfill.tsx
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    const key = "__session_storage_availability_test__";
+    sessionStorage.setItem(key, "test");
+    sessionStorage.removeItem(key);
```
kof, line 8: I am confused by the test and the implementation of this, is setItem forbidden or removeItem?
If both why do we need both for the test?
If both why do we not warn in all polyfilled methods?

rpominov: The test is the first line `setItem`. The second line is just a cleanup.

packages/react-sdk/src/tree/session-storage-polyfill.tsx
rpominov: > render it specifically before scrollrestoration

The fact that it goes before `<ScrollRestoration>` is accidental, I just needed to put it before all scripts.

> can't it just be imported as a normal script

Not sure putting it in something like `entry.client.tsx` will work, as it'll probably go after `<ScrollRestoration>` in that case, but I can try.

apps/designer/app/canvas/canvas.tsx
```diff
@@ -65,6 +65,7 @@ const useElementsTree = () => {
     if (rootInstance === undefined) return;
 
     return createElementsTree({
+      embeded: true,
```
kof, line 68: typo, but also maybe there is a better name, e.g. "sandbox"
```suggestion
      embedded: true,
```

rpominov: Renamed to sandbox

kof: if using "embed", correct word would be probably "embed" as a noun

packages/react-sdk/src/tree/session-storage-polyfill.tsx
```diff
@@ -0,0 +1,41 @@
...
+        setItem: (key: string, val: unknown) => {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `Session storage is unavailable due to Error "${
+              (error as Error).message
+            }". A polyfill is used to set value of "${key}". The value will be lost when the page is reloaded.`
+          );
+          data.set(key, String(val));
+        },
+        getItem: (key: string) => {
```
kof, line 25: so getItem won't throw?

kof (approved):

packages/react-sdk/src/tree/session-storage-polyfill.tsx
```diff
@@ -0,0 +1,41 @@
...
+        },
+        clear: () => {
+          data.clear();
+        },
+      },
+    });
+  }
+}.toString();
+
+export const SessionStoragePolyfill = () => (
```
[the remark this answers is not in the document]
rpominov: No, because there's no code that tries to use LocalStorage without checking for availability.
The polyfill is a workaround for the bug in Remix where it uses SessionStorage without checking that it's available.

packages/react-sdk/src/tree/session-storage-polyfill.tsx
rpominov: Yes, it won't. Not sure I understood the question :)

kof: its weird that setItem throws an error but getItem doesn't ... so reading from sessionStorage is allowed?

kof: I am talking about native sessionStorage api, because you are warning only in setItem as if the getItem isn't the same problem

packages/react-sdk/src/tree/session-storage-polyfill.tsx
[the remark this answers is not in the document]
kof: I am talking about potential future 3rd party code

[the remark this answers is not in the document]
rpominov: I think we should assume that other code will not have bugs like Remix. This polyfill is not an ideal solution, I don't think it's a good idea to add one for LocalStorage just in case.

PR #354, opened 2022-10-24
Remaining CssValueInput todos

ghost:
Solves the issues presented in: https://github.com/webstudio-is/webstudio-designer/issues/333

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/remix.config.js
+++ b/apps/designer/remix.config.js
@@ -18,6 +18,6 @@ module.exports = {
   serverBuildPath: "api/index.js",
   // publicPath: "/build/",
   watchPaths: () => {
-    return ["../../packages/**/*"];
+    return ["../../packages/**/lib/**.js"];
   },
 };
--- a/packages/prisma-client/package.json
+++ b/packages/prisma-client/package.json
@@ -8,6 +8,7 @@
     "typecheck": "tsc --noEmit",
     "lint": "eslint ./src --ext .ts,.tsx --max-warnings 0",
     "checks": "yarn typecheck && yarn lint",
+    "dev": "prisma generate --watch && tsup src/index.ts --format esm,cjs --watch --out-dir=lib",
     "build": "prisma format && prisma generate && rm -fr lib tsconfig.tsbuildinfo && tsc"
   },
   "bin": {
@@ -15,10 +16,12 @@
   },
   "devDependencies": {
     "@types/uuid": "^8.3.4",
-    "prisma": "^4.3.1"
+    "prisma": "^4.3.1",
+    "tsup": "^6.1.3",
+    "typescript": "4.7.4"
   },
-  "main": "src/index.ts",
-  "types": "src/index.ts",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
   "license": "MIT",
   "private": false,
   "dependencies": {


apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -263,7 +270,7 @@ export const CssValueInput = ({
 
   const suffix =
     value.type === "keyword" ? (
-      <IconButton {...getToggleButtonProps()}>
+      <IconButton css={{ width: "auto" }} {...getToggleButtonProps()}>
```
kof, line 273: what is auto solving here?

apps/designer/app/designer/features/style-panel/shared/css-value-input/unit-select.tsx
```diff
@@ -30,21 +31,31 @@ for (const [key, value] of unitRenderMap.entries()) {
...
 }: UseUnitSelectType) => {
   const [isOpen, setIsOpen] = useState(false);
   const renderUnits = useMemo(
-    () => units.map((unit) => unitRenderMap.get(unit) ?? unit),
+    () =>
+      units
+        .filter((unit) => {
+          return isValid(
+            property,
+            value?.value + String(unit).replace("number", "")
```
kof, line 55: this can use toValue, right?

packages/design-system/src/components/combobox.tsx
```diff
@@ -24,12 +24,12 @@ import { Box } from "./box";
 import { Grid } from "./grid";
 
 const Listbox = styled("ul", panelStyles, {
-  padding: 0,
+  padding: "$1 0",
```
kof, line 27: nit: `py: '$1' `reads better

packages/react-sdk/src/css/units.ts
```diff
@@ -29,4 +29,5 @@ export const units = [
   "vw",
   "x",
   "%",
+  "number",
```
kof, line 32: file was generated, its even in the comment above

packages/react-sdk/src/stitches/to-css.ts
```diff
@@ -2,9 +2,10 @@ import type { CSS } from "./stitches";
 import type { StyleProperty, StyleValue, CssRule, Breakpoint } from "../css";
 import { DEFAULT_FONT_FALLBACK, SYSTEM_FONTS } from "@webstudio-is/fonts";
 
-export const toValue = (value: StyleValue): string => {
+export const toValue = (value: StyleValue | undefined): string => {
```
kof, line 5:
```suggestion
export const toValue = (value?: StyleValue): string => {
```

packages/react-sdk/src/stitches/to-css.ts
```diff
@@ -2,9 +2,10 @@ import type { CSS } from "./stitches";
...
+  if (value === undefined) return "";
   if (value.type === "unit") {
-    return value.value + (value.unit === "number" ? "" : value.unit);
+    return value.value + value.unit;
```
kof, line 8: are you saying that we now have a unit type that has an empty string value.unit instead of number?

packages/react-sdk/src/css/units.ts
kof: I added export const Unit = z.union([z.enum(units), z.literal("number")]); in css/schema number as a valid one, but that's kinda wrong, these two are now out of sync

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -110,13 +110,18 @@ const useScrub = ({
           unit,
           value: event.value,
         });
+        inputRef.current?.blur();
+        inputRef.current?.parentElement?.setAttribute("data-state", "focus");
```
kof, line 114: there are 2 problems here:
1. you are accessing parent element in a way that is fragile - if we change the structure - this will suddenly start adding attribute on a wrong element
3. focus has actual meaning in the sense it is focused, but in this case you are actually unfocusing, so you are kinda trying to do something like visually-focused or something. 

I just needed the same thing [here](https://github.com/webstudio-is/webstudio-designer/pull/351/files#diff-e2bb0fb4fd1b7d2d36fb8c2c21769648fc2154d3226fae44d2e559357d648639R176)

 and called it. "active" like the active CSS pseudo selector ... I am not sure if my semantics better, wdyt?
 also I did it over prop, so to set it you would have to set it over local state

kof: I should probably turn it into variant="active", because its just a visual thing and `variant="ghost" active` shouldn't be possible at the same time I guess

kof: changed it to be variant over there

apps/designer/app/designer/features/style-panel/shared/css-value-input/unit-select.tsx
kof: turns out same thing has been already used in `parseCssValue` function which should also use toValue ... we need a single place that turns internal CSS value object to CSS string

kof:
- [ ] Some values can't have a numeric value, e.g. overflow: 3px would be illegal

We already have that information from the CSS database so we could prevent even going into that mode or alternatively fix validation of the input. Right now it allows to enter unit value for overflow for example for some reason. It should be in an invalid state with that value

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: will merge totay-tomorrow that other pr so you can use the variant

taylornowotny:
This is looking great! The click-and-hold to scrub interaction is really smooth. I think users are going to like that

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: I found a solution for all known icon inside textfield problems: check this out, lets wait until I merge and we can rebase this one ad4e57a83e28bea3538edd057e9fef9c724dfb7e

kof: should be able to remove width:auto with the updates in main

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: btw this has been merged, you can set that state as a prop state="active"

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -83,8 +84,9 @@ const useScrub = ({
   value: StyleValue;
   onChange: (value: StyleValue) => void;
   onChangeComplete: (value: StyleValue) => void;
-}) => {
+}): [React.MutableRefObject<HTMLInputElement | null>, boolean] => {
   const inputRef = useRef<HTMLInputElement | null>(null);
+  const [inputActiveState, setInputActiveState] = useState(false);
```
kof, line 89: its a boolean, so it can be named isInputActive

apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
```diff
@@ -17,5 +24,15 @@ export const getFinalValue = ({
...
   }
+  if (
+    currentValue &&
+    currentValue.type === "unit" &&
+    isValid(
+      property,
+      toValue({ ...currentValue, value: Math.abs(currentValue.value) })
+    ) === false
+  ) {
+    return { ...(currentValue as unknown as InvalidValue), type: "invalid" };
```
kof, line 35: same here with assertion

apps/designer/app/designer/features/style-panel/shared/parse-css-value.ts
```diff
@@ -2,9 +2,11 @@ import hyphenate from "hyphenate-style-name";
 import { units } from "@webstudio-is/react-sdk";
 import type { StyleProperty, StyleValue, Unit } from "@webstudio-is/react-sdk";
 
-const unitRegex = new RegExp(`${units.join("|")}`);
+const unitRegex = new RegExp(
+  `${(units as unknown as string[]).concat("number").join("|")}`
```
kof, line 6: same here with assertions

packages/design-system/src/components/panel.tsx
```diff
@@ -4,7 +4,8 @@ export const panelStyles = css({
   overflow: "auto",
   backgroundColor: "$slate4",
   borderRadius: "$1",
-  boxShadow: "0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15)",
```
kof, line 7: I remember having updated this shadow from figma, just a few days ago, lets 
1. discuss how this is wrong
2. normalize asking questions if you see we are back and forth on something

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -39,6 +39,7 @@ type NumericScrubState = {
   offset: number;
   velocity: number;
   direction: string;
+  timeout: unknown;
```
kof, line 42: why would that be unknown

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -39,7 +39,7 @@ type NumericScrubState = {
   offset: number;
   velocity: number;
   direction: string;
-  timeout: unknown;
+  timeout?: ReturnType<typeof window.setTimeout>;
```
kof, line 42: thats timerId

kof: timeout is the argument you pass to setTimeout, what you receive is an id for that timer

apps/designer/app/designer/features/style-panel/shared/css-value-input/unit-select.tsx
```diff
@@ -31,22 +37,29 @@ for (const [key, value] of unitRenderMap.entries()) {
...
   ...props
 }: UseUnitSelectType) => {
   const [isOpen, setIsOpen] = useState(false);
   const renderUnits = useMemo(
-    () => units.map((unit) => unitRenderMap.get(unit) ?? unit),
-    [units]
+    () =>
+      units
+        .filter((unit) => {
+          return isValid(property, toValue({ ...value, unit } as typeof value));
```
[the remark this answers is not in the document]
Andarist: the object passed to `toValue` has a property `value` that is of type: `number | undefined` but the expected type of that property is `string | number | string[]`. Here, `undefined` is not compatible with the expected type. The `undefined` appear here because the `value` identifier here has an optional value (its `value` property is `number` though, so it's compatible).

So for example this works OK:
```ts
          if (value) {
            toValue({ ...value, unit });
          }
```
But we need to also handle the other case somehow (when `value === undefined`). 

This one type checks OK:
```ts
  const renderUnits = useMemo(
    () =>
      value &&
      units
        .filter((unit) => {
          return isValid(property, toValue({ ...value, unit }));
        })
        .map((unit) => unitRenderMap.get(unit) ?? unit),
    [units, property, value]
  );
```

but it changes the type of `renderUnits` from `string[]` to `string[] | undefined` and this causes a problem later on.

In practice, this won't be a problem because we have an early return based on `value == undefined` but we can't express a relationship between `value` and `renderUnits` in a way that would allow TS to understand that checking `value` also checks~ `renderUnits` (unless we combine both into an object/array somehow).

The easiest fix is to change the latter check. So the next version could look smth like this:
```ts
  const renderUnits = useMemo(
    () =>
      value &&
      units
        .filter((unit) => {
          return isValid(property, toValue({ ...value, unit }));
        })
        .map((unit) => unitRenderMap.get(unit) ?? unit),
    [units, property, value]
  );

  const renderValue = value && unitRenderMap.get(value.unit);

  if (renderUnits == undefined || renderValue === undefined) {
    return [isOpen, null];
  }
```
That, in turn, will result in some further problems with the argument passed to `onChange`. So it seems that the easiest fix is to just check all of those things in the `if` responsible for the early return. The final version could look smth like this:
```ts
  const renderUnits = useMemo(
    () =>
      value &&
      units
        .filter((unit) => {
          return isValid(property, toValue({ ...value, unit }));
        })
        .map((unit) => unitRenderMap.get(unit) ?? unit),
    [units, property, value]
  );

  const renderValue = value && unitRenderMap.get(value.unit);

  if (value === undefined || renderUnits == undefined || renderValue === undefined) {
    return [isOpen, null];
  }
```

Is this extra added check somewhat redundant? Yes. Is this extra added check purely to satisfy the TS? Yes. 😉

apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
Andarist: The reported error here is just terrible:
> Type '"invalid"' is not assignable to type '"unit"'.ts(2322)

The presence of the error is legit though and by casting here we risk some runtime failures elsewhere. You want to end up with `InvalidValue` here. That type has its `value` property typed as `string`:
https://github.com/webstudio-is/webstudio-designer/blob/cfd40d748b936d23caa3b6c60a8385c5df7f77c8/packages/react-sdk/src/css/schema.ts#L34
but the `currentValue`'s type here is `UnitValue` (we are within `if (currentValue.type === "unit")` here) and its `value` property is typed as `number`:
https://github.com/webstudio-is/webstudio-designer/blob/cfd40d748b936d23caa3b6c60a8385c5df7f77c8/packages/react-sdk/src/css/schema.ts#L12

apps/designer/app/designer/features/style-panel/shared/parse-css-value.ts
Andarist: `.concat` is a generic function and its type definitions only allow you to pass items of the type contained in the source array. Basically, for a given `T[]` concat can only produce `T[]`. Your `T` here is a list of some string literals (representing units) and `"number"` is not one of them thus it's not accepted as a valid argument because it's not assignable to `T`.

As a workaround for this I recommend using spread - which allows you to create new array types out of the input "arguments":
```ts
const unitRegex = new RegExp(`${[...units, "number"].join("|")}`);
```

packages/design-system/src/components/panel.tsx
kof: We seem to be looking at different figma, lets figure this out:

## case 1: 

https://www.figma.com/file/EfzZ2w94KpmUviZJSBgxCd/Designer-UI?node-id=1602%3A41906

```
box-shadow: 0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15);
```

Screenshot 2022-11-01 at 10 37 39


## case 2

https://www.figma.com/file/EfzZ2w94KpmUviZJSBgxCd/Designer-UI?node-id=2327%3A47295

```
box-shadow: 0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15);
```

Screenshot 2022-11-01 at 10 40 11

kof: @natlusyht is @taylornowotny helpful? I think he confirmed what I was saying

kof (changes requested):
lst thing is with the shadow

packages/design-system/src/components/panel.tsx
kof: ohhh now I understand

kof: btw background color is gray4 instead of slate4, it was already wrong, not a regression, but since we looked please take it along

kof (approved):

PR #362, opened 2022-10-27
Ability to edit basic page settings and delete a page

rpominov:
Another part of #54

TODO:

 - [x] Add a menu button that opens the settings panel
 - [x] Via the settings panel user should be able to:
   - [x] Edit path and title
   - [x] Delete the page

----------------

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -60,36 +67,63 @@ const staticTreeProps = {
...
   onNewPage?: () => void;
   onSelect: (pageId: string) => void;
   selectedPageId: string;
+  onEdit?: (pageId: string) => void;
+  editingPageId?: string;
 }) => {
   const [pages] = usePages();
   const pagesTree = useMemo(() => pages && toTreeData(pages), [pages]);
 
+  const renderItem = useCallback(
```
kof, line 90: maybe this should be a component Item?

rpominov: Sorry, don't understand what you mean :)

rpominov: Ah, you mean instead of passing `renderItem` function to the tree, pass a component?

I think that we'll be slightly worse as we would have to create components dynamically during render, etc. But overall I don't have a strong opinion.

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -129,14 +164,19 @@ export const TabContent = (props: TabContentProps) => {
...
-    navigate(`${props.config.designerPath}/${project.id}?pageId=${pageId}`);
+    if (pageId === "home") {
+      navigate(`${props.config.designerPath}/${project.id}`);
+    } else {
+      navigate(`${props.config.designerPath}/${project.id}?pageId=${pageId}`);
+    }
   };
 
-  const [newPageOpen, setNewPageOpen] = useState(false);
+  const NEW_PAGE = "new-page";
```
kof, line 178: we usually use this naming convention only when we export those constants to be reused across project, so that they are immediately visible as contacts, inline - just normal camel case

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -146,19 +186,41 @@ export const TabContent = (props: TabContentProps) => {
     <>
       <PagesPanel
         onClose={() => props.onSetActiveTab("none")}
-        onNewPage={() => setNewPageOpen((current) => !current)}
+        onNewPage={() =>
```
kof, line 189: maybe better onCreateNewPage

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -7,26 +7,119 @@ import {
...
+    }
+  });
+
+  const resetErrors = (fieldName: keyof typeof fieldIds) =>
+    setFieldErrors(({ [fieldName]: _, ...rest }) => rest);
+
+  return (
+    <>
+      <Group>
+        <Label htmlFor={fieldIds.name} size={2}>
```
kof, line 82: Label is based on __DEPRECATED__Text component, we need to update it to use Text component which doesn't have size prop any more, because we switched to semantic variants

rpominov: I can update it, but should this be a part of this PR?

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -7,26 +7,119 @@ import {
...
+        <InputErrorsTooltip errors={fieldErrors.name}>
+          <TextField
+            state={fieldErrors.name && "invalid"}
+            id={fieldIds.name}
+            name="name"
+            disabled={disabled}
+            value={values?.name}
+            onChange={(event) => {
+              resetErrors("name");
+              onChange?.("name", event.target.value);
```
kof, line 94: its better to have a single argument in in those handlers, like with the native event handlers, in this case:
`onChange({field: 'name', value: event.target.value})`

On the related note: is it really necessary to handle onChange for every field separately instead of handling onChange on the form component and then, new FormData(event.currentTarget) and then validate the entire form data every time, so you have a single place for entire form

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -127,3 +164,149 @@ export const NewPageSettings = ({
     </>
   );
 };
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type EditPageData = FetcherData<{}>;
```
kof, line 169: I am sure there is a better way to express this type afaik linter suggests different ways when you try to use {}

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
rpominov: There're couple reasons handle changes separately:

1. We want to send to the server only changed data because it reduces the risk of overriding more recent data on the server with potentially outdated data on the client, also a bit smaller payloads.
2. It's easier to track whether we have data to submit when we have a separate object with only changes.

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
rpominov: None of the alternative do what I need, and this works perfectly. I think this is an edge case where linter is wrong

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
kof: I think you are doing it here so you can know which field has changed and submit only changed fields. Another way of doing it which is potentially simpler: just iterate over the data and compare the fields before submission

rpominov: I think the current way is simpler. I'd keep it as is in this PR, and consider again when I will work on adding "meta" support (nested repeating forms), because it may require refactoring anyway.

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -7,26 +7,119 @@ import {
...
+const normalizeErrors = (
+  errors: string | ZodError["formErrors"]
+): ZodError["formErrors"] =>
+  typeof errors === "string"
+    ? { formErrors: [errors], fieldErrors: {} }
+    : errors;
+
+type EditablePage = Omit<Page, "treeId" | "id">;
+
+const FormFields = ({
```
kof, line 43: would be probably easier to maintain if FormFields would be a controlled component, without the fetcher and then you would add the logic outside. This way you could for example render the form in storybook without fetching logic and also it wouldn't have much to debug

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -127,3 +164,149 @@ export const NewPageSettings = ({
...
+    onDeleted?.();
+  };
+
+  if (page === undefined) {
+    return null;
+  }
+
+  return (
+    <>
+      <Header
```
kof, line 286: same here, try to decouple the render component from all that logic above

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
```diff
@@ -0,0 +1,82 @@
...
+  createContext,
+  useCallback,
+  useContext,
+  useRef,
+  type ReactNode,
+} from "react";
+import { useOnFetchEnd } from "./on-fetch-end";
+
+/* This utility is designed for the case when you want to send a request using Remix's fetcher,
+ * but you don't want the request to be cancelled when the component where useFetcher() is called unmounts.
```
kof, line 12: I am actually surprised that remix cancels the fetch when component unmounts

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type AddTaskFn = <Data>(
+  target: Task<Data>["target"],
+  options: Task<Data>["options"],
+  callback?: Task<Data>["callback"]
+) => void;
+
+const Context = createContext<AddTaskFn | undefined>(undefined);
+
+export const PersistentFetcherProvider = ({
```
kof, line 29: why do you need a provider for this vs a simple globally handled queue like we have in canvas/shared/sync/queue ?

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
rpominov: I don't know 🤷‍♂️

packages/design-system/src/components/icon-button.tsx
```diff
@@ -66,8 +55,44 @@ export const IconButton = styled("button", {
         width: "$8",
       },
     },
+    variant: {
```
kof, line 58: this is a big change to the design system, we need to think through what are all the variants we have and make it all consistent ... lets have a discussion on this ... seems like a major pain point with all the design system stuff right now

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
rpominov: We need a global fetcher created using `useFetcher` which can be done only in a React component.

packages/design-system/src/components/toast.tsx
```diff
@@ -130,3 +131,22 @@ export const Toaster = () => {
 };
 
 export { toast };
+
+export const toastNonFieldErrors = (
```
kof, line 135: 1. not sure if this should be here on the level of design system
2. design system doesn't know zod yet and I am not sure it should
3. I don't understand the semantics of "toastNonFieldErrors"

rpominov: > not sure if this should be here on the level of design system

Yeah, probably better to put it in `shared` in the app.

> I don't understand the semantics of "toastNonFieldErrors"

"Given form errors from Zod, show a toast for each error that does not correspond to a known field"

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -111,30 +126,195 @@ const hoverStyle = {
...
+        },
+      },
+    },
     enableHoverState: {
-      true: { "&:hover:after": hoverStyle },
+      true: {
+        "&:hover:after": hoverStyle,
+
+        // Suffix is also visible on hover, which we don't track using JavaScript
+        [`&:hover ${ItemButton}`]: {
```
kof, line 161: instead of the ItemButton selector we could use the CSS variables all the way, similar to how I did it with `imageInfoTriggerCssVars()`
1. component exports a high-level function that returns all variables
2. consumer uses it to set the variables

in this case it would look like this

`'&:hover': itemButtonCssVars('hover')`

below it would be:

`&:hover: suffixContainerCssVars('hover')`

rpominov: But why not use `ItemButton` selector? This is a local component, so there's no breaking of encapsulation.

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
kof: Not necessarily if you do it right next

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
kof: current way spreads the logic for how you handle this data all over the form, it looks simpler now but its more custom and more code

finding changed fields is actually quite easy, its a simple loop you would do once before validation and submission, it will allow you to get rid of all the logic in the form and make it headless, then handle the validation/submission in a separate hook from outside

rpominov: > finding changed fields is actually quite easy, its a simple loop you would do once before validation and submission,

We need both representations anyway complete and partial. But it's easier to go from partial to complete (`latest = {...onServer, ...changes}`) than to go from complete to partial (`changes = diff(onServer, latest)`)

> it will allow you to get rid of all the logic in the form and make it headless, then handle the validation/submission in a separate hook from outside

This is separate issue, you've mentioned below. Will extract to a hook, yes.

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -3,7 +3,11 @@ import { db } from "@webstudio-is/project/index.server";
 import { type Pages, utils } from "@webstudio-is/project";
 import { zfd } from "zod-form-data";
 import { z } from "zod";
-import { CreatePageData } from "~/designer/features/sidebar-left/panels/pages/settings";
+import {
+  CreatePageData,
+  DeletePageData,
+  EditPageData,
+} from "~/designer/features/sidebar-left/panels/pages/settings";
```
kof, line 10: 1. we never import things from features except of inside the component that puts features together, like Designer
2. when there is an index.ts, we never reach into that folder to get random files, index.ts acts as a public interface, everything it didn't export is private. There is a linter for this we haven't installed yet, can you please create a ticket to add it?

You probably want to move those types to some shared location and make sure that we don't import the js from that module if we only need types. I was wondering sometimes if our compiler is smart enough to do that or do we need to be careful and put `type` keyword after import

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
kof: lets dig into this, we need to know

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
kof: right, fetcher can't be created outside

packages/design-system/src/components/toast.tsx
kof: ah so the semantics is this: toastUnknownFieldErrors

kof: Looks like we could use shared/form-utils or something

packages/design-system/src/components/tree/tree-node.tsx
kof: 1. Because this way you can't create a different Button component that would be passed to the ItemContainer. You are coupling these.
2. We might need to get rid of stitches long term, so the less features from stitches we use the better. It will be much easier to get rid of it.
3. I think we should leverage CSS variables to the maximum, they give us both: decoupling, all native CSS things like :hover, customization and even dynamic updates. I think long term CSS props should be heavily used

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
rpominov: I can create an issue and ask. Couldn't find anything in existing issues or docs. I the source code it looks like this:

https://github.com/remix-run/remix/blob/471ab259fba5adbdd70de71ee7bc7e874c1f4db9/packages/remix-react/components.tsx#L1469-L1474

https://github.com/remix-run/remix/blob/471ab259fba5adbdd70de71ee7bc7e874c1f4db9/packages/remix-react/transition.ts#L496-L498

packages/design-system/src/components/toast.tsx
rpominov: `toastUnknownFieldErrors` wouldn't be right because it shows 2 kinds of errors:

1. Errors of entire form
2. Errors of fields that we didn't expect to receive (don't have fields for in UI)

So the name would fit for `#2` but not `#1`

packages/icons/src/menu.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from "react";
+import { IconProps } from "./types";
+
+export const MenuIcon = React.forwardRef<SVGSVGElement, IconProps>(
```
kof, line 4: isnt this the same as DotsHorizontalIcon?

rpominov: Not sure, will check.

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
kof: Maybe look in remixes discussions/issues or start your own. Basically we need to figure out what's remixes plan for fetching data like you did in persistent fetcher

rpominov: Ah, didn't check discussions. It was asked there before, but there're only answers by community members:

https://github.com/remix-run/remix/discussions/3585 — suggested to not unmount until done
https://github.com/remix-run/remix/discussions/2623#discussioncomment-2508116 — suggested to use context like I did
https://github.com/remix-run/remix/discussions/2462#discussioncomment-2423735 — suggested to not unmount until done

Not sure if there's a point in trying to ask again.

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
rpominov: Keeping as is unless there're reasons to go with component

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
rpominov: Will do a separate PR: https://github.com/webstudio-is/webstudio-designer/issues/381

packages/icons/src/menu.tsx
rpominov: `DotsHorizontalIcon` is from `radix-ui` and it's slightly different.

packages/design-system/src/components/icon-button.tsx
rpominov: Removed

kof:
got conflicts

rpominov:
Resolved

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -60,36 +69,80 @@ const staticTreeProps = {
...
+          {...props}
+          suffix={
+            onEdit && (
+              <Flex css={{ mr: "$2" }} align="center">
+                {isEditing ? (
+                  <ChevronRightIcon />
+                ) : (
+                  <MenuButton
+                    aria-label="Page settings"
+                    parentSelected={isSelected}
```
kof, line 125: isParentSelected

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -129,14 +183,19 @@ export const TabContent = (props: TabContentProps) => {
...
-  const handleSelect = (pageId: string) => {
+  const handleSelect = (pageId: "home" | Page["id"]) => {
     if (project === undefined) {
       return;
     }
-    navigate(`${props.config.designerPath}/${project.id}?pageId=${pageId}`);
+    if (pageId === "home") {
+      navigate(`${props.config.designerPath}/${project.id}`);
+    } else {
+      navigate(`${props.config.designerPath}/${project.id}?pageId=${pageId}`);
```
kof, line 193: do this path building logic anywhere else? if yes we should have a reusable function

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -146,19 +205,41 @@ export const TabContent = (props: TabContentProps) => {
...
+            onSuccess={(page) => {
+              setEditingPageId(undefined);
+              handleSelect(page.id);
+            }}
+          />
+        )}
+        {editingPageId !== newPageId && editingPageId !== undefined && (
+          <PageSettings
+            onClose={() => setEditingPageId(undefined)}
+            onDeleted={() => {
```
kof, line 232: would probably still call it onDelete

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -84,45 +168,186 @@ export const NewPageSettings = ({
...
+      return;
+    }
+    // We use submitPersistently instead of fetcher.submit
+    // because we don't want the request to be canceled when the component unmounts
+    submitPersistently<EditPageData>(
+      toFormData({ id: pageId, ...submittedValues, ...unsavedValues }),
+      { method: "post", action: `/rest/pages/${projectId}` },
+      (data) => {
+        if ("errors" in data) {
+          toastNonFieldErrors(normalizeErrors(data.errors), []);
```
kof, line 272: I remember I was confused about that name "toastNonFieldErrors" and had a suggestion

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -65,15 +64,75 @@ const handlePut = async (
...
+
+  const devBuild = await db.build.loadByProjectId(projectId, "dev");
+
+  if (devBuild.pages.homePage.id === id) {
+    return fieldError("id", `Can't delete the home page`);
+  }
+
+  await db.build.deletePage(devBuild.id, id);
+
+  return { ok: true };
```
kof, line 117: we should normalize this to something we already started using: { status: "ok" } and maybe even wrap it into a function like you did with fieldError and use it everywhere, so that we have this consistently throughout entire application

kof: using {status: 'ok'}, gives us a way to also have some other status e.g. status: "terrible" :)

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -113,3 +172,12 @@ export const loader: LoaderFunction = async ({
     };
   }
 };
+
+const fieldError = (fieldName: string, error: string) => {
```
kof, line 176: we should normalize this kind of utilities in a shared thing, because its not a singular use case, this should be useful in a lot of places, we need form-utils

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
kof: well that's on itself is a good answer, so we are on the right pass, since we don't want to force user to wait, then we have to use context

packages/design-system/src/components/icon-button.tsx
```diff
@@ -47,7 +47,7 @@ export const IconButton = styled("button", {
         width: "$3",
         "&:hover, &:focus": {
           background: "none",
-          borderColor: "transparent",
+          border: "none",
```
kof, line 50: there was a problem with this: border is holding space, so if you remove border on hover, size of the icon will change and things may shift

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -111,30 +170,185 @@ const hoverStyle = {
...
+  itemData,
+  level,
+  shouldRenderExpandButton,
+  isExpanded,
+  children,
+  suffix,
+  suffixWidth = suffix ? "$sizes$5" : "0",
+  alwaysShowSuffix = false,
+  forceFocus = false,
+  selectionTrigger = "click",
```
kof, line 241: better name: selectionEvent trigger is often used as an element, jsx instance

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -111,30 +170,185 @@ const hoverStyle = {
...
+  parentIsSelected?: boolean;
+  selectedItemId?: string;
+  onSelect?: (itemId: string) => void;
+  level: number;
+  isAlwaysExpanded: boolean;
+  shouldRenderExpandButton: boolean;
+  isExpanded: boolean;
+};
+
+export const TreeItemBody = <Data extends { id: string }>({
```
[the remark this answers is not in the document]
kof: its usually a sign that the composition model is bad, e.g. component knows too much and should be composed out of multiple components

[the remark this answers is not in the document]
kof: without passing all the props to this parent component

packages/icons/src/menu.tsx
kof: can we use one for both? Is design in figma using different icons here and in fonts manager? I think we only need one icon for this but I may be wrong, lets clarify it

kof (changes requested):
found a few more things to clarify

apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
rpominov: Yeah. I answered there that `toastUnknownFieldErrors` wouldn't be ideal either, but thinking again it's not that important, will rename.

packages/design-system/src/components/tree/tree-node.tsx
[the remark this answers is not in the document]
rpominov: The idea is that as a user you just pass through all props without thinking about them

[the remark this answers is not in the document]
rpominov: Check how it's used

packages/design-system/src/components/icon-button.tsx
kof: hmm, I can't find what was wrong ... maybe its something else

packages/design-system/src/components/tree/tree-node.tsx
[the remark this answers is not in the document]
kof: still means this component knows a lot of stuff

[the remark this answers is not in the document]
kof: if you don't have a better idea, lets leave it for now, but def. looks wrong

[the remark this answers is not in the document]
rpominov: Seems fine to me because it knows a lot about other component in the same file. They are coupled and colocated, so it's ok imo

packages/icons/src/menu.tsx
rpominov: Yeah, I've replaced with `MenuIcon` in fonts

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
rpominov: What if instead of `config` that we pass around, we'll have something like `shared/routes-paths.ts` with:

```
export const designerPath({projectId, pageId}: {projectId: string, pageId?: string}) => {
  if (pageId) {
    return `/designer/${projectId}?pageId=${pageId}`
  }
  return `/designer/${projectId}`
}

// same for every path that we need in the app...
```

1. Wouldn't have to pass around config
2. Builds the entire path
3. If we need paths to differ in different deployments we can have conditions like `process.env.NODE_ENV === ...` in this file

rpominov: Also found this https://github.com/yesmeck/remix-routes . But not sure about it:

1. Adds another build step (https://github.com/yesmeck/remix-routes#setup)
2. Works only for Remix routes, but we won't necessarily use Remix for e.g. canvas (or it can be Remix, but a separate app)
3. Not sure it will support all our use cases even for Remix routes as well as custom written functions

kof: 1. yeah path-utils
2. we don't actually need to pass around config, its static, we can import it directly

kof: remix-routes package - yeah not sure about it

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
```diff
@@ -0,0 +1,82 @@
...
+      const task: Task<unknown> = { target, options, callback };
+      if (fetcher.state === "idle" && queue.current.length === 0) {
+        currentRequestCallback.current = task.callback;
+        fetcher.submit(task.target, task.options);
+      } else {
+        queue.current.push(task);
+      }
+    },
+    [fetcher]
+  ) as AddTaskFn;
```
kof, line 69: we should be able to do better, need TS help?

apps/designer/app/shared/form-utils/use-ids.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const useIds = <FieldName extends string>(
+  fieldNames: readonly FieldName[]
+) => {
+  if (process.env.NODE_ENV !== "production") {
+    // It's OK to disable because process.env.NODE_ENV is not going to change
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const fieldNamesLength = useRef<number>(fieldNames.length);
+    if (fieldNamesLength.current !== fieldNames.length) {
+      throw new Error("changing fieldNames length is not supported");
```
kof, line 12: also in case of thrown error, rules of hooks will be broken

apps/designer/app/shared/form-utils/validation.ts
```diff
@@ -0,0 +1,79 @@
...
+  knownFields?: readonly string[]
+) => {
+  for (const message of errors.formErrors) {
+    toast.error(message);
+  }
+
+  if (knownFields !== undefined) {
+    for (const fieldName in errors.fieldErrors) {
+      if (knownFields.includes(fieldName) === false) {
+        for (const message of errors.fieldErrors[fieldName] as string[]) {
```
kof, line 60: also should be possible to not assert, need help?

apps/designer/app/shared/form-utils/use-ids.ts
rpominov: Sure, but then the entire app will crash so it doesn't matter

apps/designer/app/shared/form-utils/validation.ts
kof: seems like you should be checking if fieldName is in fieldErrors instead of asserting it

rpominov: These `as` statements are fine IMO. I don't see how they can lead to bugs. But sure, if you or @Andarist knows how to improve this, let improve.

kof (approved):
Approving in advance, please 
- please check how we are building those routes overall and maybe create utils or create a separate issue for this if you like
- please try to avoid type assertions and do the actual checks instead, its safer because we actually know its correct then rather than relying on the logic that happens somewhere else which may change while assertion stays in place and renders typechecks inaccurate

apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
rpominov: Will do a separate PR: https://github.com/webstudio-is/webstudio-designer/issues/389

apps/designer/app/shared/form-utils/validation.ts
kof: The way I see it is you are asserting that those field errors for a certain field exist, but the code doenstn't prove that, its just telling compiler this is true.

You are probably checking it by checking knownFields and known fields automatically have keys in fieldErrors, but this logic is implicit and has no explicit verification. Sometimes its too hard to make it more obvious, so we do this type of stuff, but we should def. try to avoid it

rpominov: Yeah, this one was actually a potential bug, I misunderstood originally what TS complains about. Fixed.

apps/designer/app/shared/fetcher/persistent-fetcher.tsx
rpominov: Asked in chat for help. But I think this is reasonable in this case. Will merge, and do another PR if we figure out a solution.

PR #363, opened 2022-10-27
Watch less files with Remix

Andarist:
cc @kof


apps/designer/remix.config.js
```diff
@@ -18,6 +18,9 @@ module.exports = {
   serverBuildPath: "api/index.js",
   // publicPath: "/build/",
   watchPaths: () => {
-    return ["../../packages/**/*"];
+    return [
+      "../../packages/**/lib/**.js",
+      "../../packages/prisma-client/src/**.ts",
```
Andarist, line 23: `prisma-client` is the only package consumed directly from `src`:
https://github.com/webstudio-is/webstudio-designer/blob/bf5b8a6f0bdaaac5d23f084e98feea014f9a29ab/packages/prisma-client/package.json#L20

I'm not exactly sure why that is - this was already set up like that.

kof: oh maybe we just didn't do the same thing we do with everything else?

kof: lets check it out, maybe we don't need this exception

kof: I just remembered the difference with prisma client is that its being generated using build command and prisma generates a client into its own folder inside node_modules, so maybe it should be left as-is

cc @rpominov

kof: so there are basically 2 types code here, one that comes from src and that comes from prisma client inside nodemodules that is generated from prisma/schema/prisma

I don't think we need to watch the schema generation, because it is changed rarely, but when it changes, it doesn't update src or lib because it literally doesn't change those, but then the client is updated and it would be a problem if user changes the schema and then the entire system ignores the change in the new client

@rpominov, @Andarist please check if this situation works

I think we can also consume this package from the lib instead of src for consistency, its probably just an oversight

rpominov: > I think we can also consume this package from the lib instead of src for consistency, its probably just an oversight

+1. I tried that seem to work fine.

kof: what about changin prisma schema and getting the updated client everywhere?

rpominov: Yeah, I've tested that. After I've run `yarn build`, it was updated.

Andarist: Regenerating prisma files will always require a build for that (or a watch mode for that). Both `packages/prisma-client/src` and `packages/prisma-client/lib` should be able to consume this stuff in the very same way though - so the fact that Prisma requires an extra step shouldn't matter much for this thing here.

The only difference, that I think of, is that you will need to generate `lib` for that package. But you already do that for all other packages - so it shouldn't be a problem at all.

A patch with those changes could look like this:
https://github.com/webstudio-is/webstudio-designer/commit/c6412ad9ea6727c0c1b23393c87d3ac086351d61

Note that as a bonus, I've included `prisma generate --watch` in the `dev` script. So you shouldn't even have to rerun `build` manually to update the prisma files while developing.

kof: if I understand correctly we can remove that line and point the prisma-client to the lib folder for consistency, right?

Andarist: I think so, yes - I've pushed out that extra commit to this PR just now.

kof:
For some reason Lost Pixel Bot is hanging, not sure why

kof:
@Andarist can you push an empty commit please? turns out I can't because pr is from your branch

kof:
Separate PR has triggered actions to rerun and now they are passing, I assume strongly that Lost Pixel Bot is hanging when PR is from a different repo seems like a permissions issue

PR #364, opened 2022-10-30
Fonts remaining todos

kof:
- [ ] list of weights font has available in font-weight property
- [ ] Additional note: There is a need for a progress indicator when uploading fonts. I will design this so we can talk about it and add it later. 
- [x] check if all new states are added to storybook
- [x] maybe add some more tests in storybook for the new things?
- [x] delete font-face from canvas when deleting the font asset
- [x] A font with fallback shows fallback in the stylepanel and then that value can't be matched with values from fonts list to show current font
- [x] create a reusable list component in design system
- [x] Tabbing from the search field to a font item and pressing ENTER does not change the active font.
- [x] Arrow keys can be used to move the selection between list items when the search field is focused, but arrow keys cannot be used to move the focus between list items. I think it would be preferable for arrow keys to work in both cases. As a user I expect to be able to navigate any list using arrow keys (after focus has moved to the list with TAB).
- [x] Do not show the "subsection line" above the topmost category. Looks good when there's an uploaded font category but when there's only system fonts, the line should be hidden.
Screen Shot 2022-10-27 at 1 56 13 PM
Screen Shot 2022-10-27 at 1 57 20 PM

- [x] When moving the focus with TAB, the last list item will remain focused while the focus moves to the close (X) button. Pressing ENTER closes the window as expected. The highlight should be removed from the last item when focus moves off of that item.
Screen Shot 2022-10-27 at 2 17 25 PM

- [x] Here's a video of what happens when you delete a custom font that is the current font.
https://share.descript.com/view/ugPrj53F4tk

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


taylornowotny:
Here's a sneaky bug. When there are no results for a search, the panel gets stuck on "File not found". Clicking X will refresh the results, but deleting the search terms with the keyboard does not.

https://share.descript.com/view/wGfRP4lK5MR

rpominov (approved):

PR #370, opened 2022-11-01
Enforce the version of Yarn

istarkov:
According to this https://yarnpkg.com/cli/set/version this is +- recommended way to use the same package yarn version.
Having that my system yarn version is berry this solves issues with yarn for me and for anyone else with non 1.xx yarn.


kof (approved):

TrySound (approved):

istarkov:
#377

PR #373, opened 2022-11-02
Tweak the Stitches module

Andarist:
Just some minor changes to the structure of this module. No functional changes or anything but this way it will stay a little bit closer to the intention.

### DIFF
--- a/packages/react-sdk/src/stitches/stitches.ts
+++ b/packages/react-sdk/src/stitches/stitches.ts
@@ -1,31 +1,30 @@
-import {
-  createStitches,
-  globalCss as globalCssImport,
-  type CSS,
-  css as createCss,
-} from "@stitches/core";
+import { createStitches, type CSS } from "@stitches/core";
 import type { Breakpoint } from "../css";
 
 let media = {};
 
-// @todo needs fixing
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-let stitches: any;
+const createStitchesInstance = () => createStitches({ media });
+
+type StitchesInstance = ReturnType<typeof createStitchesInstance>;
+
+let stitches: StitchesInstance | undefined;
 
 export { type CSS };
 
 const getCachedConfig = () => {
   if (stitches === undefined) {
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/stitches/stitches.ts
```diff
@@ -1,31 +1,30 @@
...
 import type { Breakpoint } from "../css";
 
 let media = {};
 
-// @todo needs fixing
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-let stitches: any;
+const createStitchesInstance = () => createStitches({ media });
+
+type StitchesInstance = ReturnType<typeof createStitchesInstance>;
```
Andarist, line 8: This way I "capture" the exact inferred type without supplying any generics manually etc.

packages/react-sdk/src/stitches/stitches.ts
```diff
@@ -1,31 +1,30 @@
...
   if (stitches === undefined) {
-    stitches = createStitches({ media });
+    stitches = createStitchesInstance();
   }
   return stitches;
 };
 
-export const css: typeof createCss = (...args) => {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const css: StitchesInstance["css"] = (...args: any[]) => {
```
Andarist, line 22: This is closer to the intention because we reexport a proxy to your own instance and not to some default instance embedded in the Stitches package.

`any[]` is needed here though because TS doesn't understand that this is merely a proxy function and while instantiating the generic call embedded in this function it fails to prove that this is OK.

kof: are we still geting any ts validation for css()?

kof (approved):

packages/react-sdk/src/stitches/stitches.ts
Andarist: Yes, that flows from the declared type (`StitchesInstance['css']`). We only suppress the typechecking "internally" within this function - but externally this doesn't change anything.

PR #377, opened 2022-11-02
Use corepack to setup yarn

TrySound:
See https://nodejs.org/api/corepack.html

Corepack is a way to setup specific yarn/pnpm version on the fly. Builtin with Node 16.9+

Requires to run `corepack enable`

Replaces https://github.com/webstudio-is/webstudio-designer/pull/370

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "packageManager": "yarn@1.22.19",
   "name": "turbo",
   "version": "0.0.0",
   "private": true,
@@ -32,7 +33,7 @@
   },
   "engines": {
     "npm": ">=7.0.0",
-    "node": "16"
+    "node": ">=16.9"
   },
   "dependencies": {
     "golden-fleece": "^1.0.9"
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "typescript": "4.7.4"
   },
   "engines": {
-    "npm": ">=7.0.0",
     "node": ">=16.9 <17"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "typescript": "4.7.4"
   },
   "engines": {
-    "node": ">=16.9 <17"
+    "node": "16"
   },
   "dependencies": {
     "golden-fleece": "^1.0.9"
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "engines": {
     "npm": ">=7.0.0",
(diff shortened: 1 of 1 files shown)


istarkov (approved):

kof:
@rpominov because otherwise you can get yarn 2 or 3

istarkov:
corepack enable? do I need to run it by myself?

TrySound:
Yes, it's not enabled by default. So need to add it into setup process.

TrySound:
> in that case we can delete the yarn artifact we committed before?

You can even remove globally install yarn

package.json
```diff
@@ -32,7 +33,7 @@
   },
   "engines": {
     "npm": ">=7.0.0",
-    "node": "16"
+    "node": ">=16.9"
```
kof, line 36: btw we set a fixed version here to just avoid any surprises, because its a complex product, not a library, I would rather keep it

kof:
Turns out corepack is experimental and only available in node 19. I don't understand how it helps us right now

TrySound:
See, it's experimental only from node.js point of view and lack of npm support. So mostly political. Though actual implementation pretty stable and ported into all versions with active support. It works locally and help CI a lot.

image

TrySound:
It helps to get the right yarn version in any project packageVersion is specified. Latest yarn 1, yarn 2+ adds it along with `yarn init`.

package.json
kof: to be clear: I am against this change, can create additional work right now

kof:
so corepack exists in node 16? great

kof:
any update on this?

TrySound:
Oh, there is rebase button.

All is ready. I fixed node range but still require higher than 16.9 which supports corepack.

kof:
if we need to upgrade node version, we need to do it everywhere, including vercel, ci etc.

kof:
Please just use a fixed node version in the engines, I am fine upgrading.

TrySound:
Alright, though it's not fixed version. It's range `16.*.*`

kof:
So how do we make sure people use the version we specified? We need to enforce the corepack setting somehow. Also this pr should then remove the yarn artifact from the repo, right?

kof:
What about this one?

TrySound:
People can not use corepack though we still can enforce with engines field.

kof:
ifthey can't, then we still have to keep the local yarn version, then I would say lets not add it

Issue #196, opened 2022-07-27
Image component and src prop


kof:
As a user I want to be able to drop an image component onto canvas then select src prop from asset manager

## Basics

- [x] create basic Image component for sdk
- [x] don't load image if its off screen (lazy prop by default, that's it?)
- [x] add srcset and sizes to image component
- [x] handle File type in props panel that lets you select image
- [x] define what we can do for high res screens and low res screens (srcset?)
- [x] we should be able to know what image sizes are needed in each depending on the breakpoint and/or layout to pass with the src to optimizer the right size

## Questions

- [ ] should user be able to put arbitrary URL instead of selecting an image from asset manager?

kof:
Check out image component from astra https://docs.astro.build/en/guides/integrations-guide/image/

I don't know if we can even use it, if not - it can be a big inspiration

istarkov:
Images:


# img vs picture tag on responsive resolutions.


# IMG


## 1. _img_ tag is well suited if on all breakpoints you have same width, heights for the image.

In that case all you need is single `srcset` with dpr set.

```html
<img 
  srcset="https://via.placeholder.com/128 1x, https://via.placeholder.com/256 2x" 
  src="https://via.placeholder.com/256" 
  width="120" height="120"
>
```

For devices with 

- `window.devicePixelRatio = 1` will be loaded `https://via.placeholder.com/128` 
- `>=2` will be loaded `https://via.placeholder.com/256`

There is no need to support dpi >= 2 because of physical limitations of oled screens

see explanation: https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/capping-image-fidelity-on-ultra-high-resolution-devices.html



## 2. img tag also is ok to use if aspect-ratio is +-fixed, i.e. like


```html
<img
  alt=""
  sizes="
    (max-width: 600px) 100vw, 
    (max-width: 1200px) 50vw, 
    30vw
  "
  srcset="
    https://via.placeholder.com/64     64w,
    https://via.placeholder.com/96     96w,
    https://via.placeholder.com/128   128w,
    https://via.placeholder.com/256   256w,
    https://via.placeholder.com/384   384w,
    https://via.placeholder.com/640   640w,
    https://via.placeholder.com/750   750w,
    https://via.placeholder.com/828   828w,
    https://via.placeholder.com/1080 1080w,
    https://via.placeholder.com/1200 1200w,
    https://via.placeholder.com/1920 1920w,
[...]

Issue #192, opened 2022-07-27
Decide how we can express properties that are shown by default in props panel


kof:
We need a way to manually express properties that should be visible by default without searching for them, for each component

kof:
We are currently mixing very different concerns:
- required
- default prop
- shortlisted prop (to be always visible)

We need to 
-  express that a prop can not be removed from the panel props list, but can have an empty value (not defined)
- express a required prop that has a default

PR #378, opened 2022-11-02
[sdk] Image

istarkov:
Responsive Image component for sdk, ref #196. 
Inspired by nextjs Image.

JSON changes were made to make it possible to see images without broken attributes in storybook.

**Per commit review would give best results.**

Additionally:
- Removed default `auto` on width/height img tag attributes - should be numbers based on spec.
- `max-width: 100%` added to the Image control as a default, solves initial placement (force Image to fit parent width)
- `defaultValue.value` changed to `defaultValue`

TODO:
- [x] Jest test for gen attributes
- [x] Return width/height into defaultProps as empty strings [comment](https://github.com/webstudio-is/webstudio-designer/pull/378#pullrequestreview-1166284537)
- [x] Storybook (examples for fixed width/height, object-cover, aspect-ratio, fill container, custom sizes)
- [x] Check it works in the designer-ui.
- [x] Check if it works with caching.
   see [link](https://webstudio.is/cdn-cgi/image/width=640,quality=80,format=auto/https://assets-dev.webstudio.is/0beWEKDTrKqaZkh5bdN0peW0oOrcXN_1663019526493.jpg) contains `vary: Accept, Accept-Encoding`  

PS: `img` tag left as having that [format=auto](https://developers.cloudflare.com/images/image-resizing/url-format/#format) is working seems like no good use case for `picture` tag exists.

Need to be done next PRs:

- css aspectRatio support
- local transformations support?

Probable future todos:
- Knowledge of Image aspect-ratio would allow cropping images serverside.
- Knowledge of image screen size on various breakpoints will allow creating near-perfect `sizes` instead of DEFAULT_SIZES.
- 404 etc processing with CSS - https://bitsofco.de/styling-broken-images/ (has some opinionated issues) or js solution with custom user fallback.
- Slow networks.
- Early hints for high priority images https://blog.cloudflare.com/early-hints/


packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
+        : { src: imagePlaceholderSvg };
+
+    return <img {...props} {...imageAttributes} decoding="async" ref={ref} />;
+  }
 );
 
 Image.defaultProps = {
   src: "",
   loading: "lazy",
-  width: "auto",
```
istarkov, line 14: Removed width/height from UI. Need to be returned with "" and transformed through getInt() inside `image`

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
+  let resizeOrigin = "";
+
+  if (process.env.NODE_ENV !== "production") {
+    if (!allSizes.includes(width)) {
+      warnOnce("Width must be only from allowed values");
+    }
+
+    // No image transformation in development
+    resizeOrigin = "https://webstudio.is";
+  }
```
rpominov, line 40: Is this for localhost? It's strange then that `https://webstudio.is` is used, because AFAIK locally we "upload" assets to local filesystem. But not sure about that. /cc @kof

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
+
+  const params = [
+    `width=${width}`,
+    quality != null && `quality=${quality}`,
+    // TODO: Must be checked how it works with caching (like avif image is cached and then read by the browser without avif support)
+    "format=auto",
+  ].filter(Boolean);
+
+  return `${resizeOrigin}/cdn-cgi/image/${params.join(",")}/${src}`;
```
rpominov, line 49: So in production `resizeOrigin` will be `""`?

packages/react-sdk/src/utils/warn-once.ts
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+let warnOnce = (_: string) => {};
```
rpominov, line 2: `eslint-disable` can be avoided if you replace `{}` with `undefined`

packages/react-sdk/src/components/image.tsx
istarkov: Locally we don't have cloudflare image resize, how to test/check without?

packages/react-sdk/src/components/image.tsx
istarkov: Yes, as for every host we have image resize enabled for now.

istarkov: This forces no dns lookup, + ability to push high-priority images using http2

istarkov: See https://blog.cloudflare.com/early-hints/

packages/react-sdk/src/utils/warn-once.ts
istarkov: Copypasted from nextjs, I think (not sure) that empty body is explicitly set here, so smart build can fully remove all calls to this function in production.

istarkov: For example esbuild
```
function a(_: string) {}
const c = a('hello');
```
`yarn esbuild --minify --tree-shaking=true --loader=ts < file.js`

minify this to empty code


BUT 

```
const a = (_: string) => {};
a('hello')
```

is not minified.

I don't know does above minified or not.

packages/react-sdk/src/components/image.tsx
istarkov: Locally we can emulate Cloudflare resize with imgproxy inside docker compose. And use imgproxy [loader transform]( https://github.com/webstudio-is/webstudio-designer/pull/378/files#diff-2b9aebd3d7549d8e39f1c7f9cb934b5909220081cef2f6b8cfc2fc93b712140eR21) for local development. In that case docker compose will be a requirement.

istarkov: Or stick to no transform at all just adding fake `cdn-cgi/image` route locally

packages/react-sdk/src/components/image.tsx
kof: Default props have been a major pain in that ass also because we need a way to say what props we want to __always__ show for a component, because they are frequently used and the rest can be added by the user.

So we def. need to decouple defaultProps.

kof: So I am open to proposals on how we can do both, what you need and expressing props shown by default on props panel

packages/react-sdk/src/components/image.tsx
kof: this is a good question, because if you are building locally and actually need to see the page with same sizes you would get in prod, we need to build our own resizing service that runs locally.

We will probably build it at some point in any case, but can we avoid building it now? Would loading original images be a huge problem for now? Technically we don't expect people to use webstudio locally at this stage, this is something for later, after beta rollout

packages/react-sdk/src/utils/warn-once.ts
kof: 2 thoughts:
1. this would be great to have everywhere
2. maybe we could even overwrite console.warn to log only once?

kof: is there an npm package that does exactly this? so we can use it in designer and everywhere else?

packages/react-sdk/src/components/image.tsx
kof: As discussed in a call:
1. lets use asset id in props/styles
2. lets build a single function that returns a path to the image depending on the environment, so we don't need a proxy locally for now
3. lets use that single function in image manager in the designer as well as in SDK components

packages/react-sdk/src/utils/warn-once.ts
istarkov: cc @TrySound, @rpominov  
the idea of all such helpers that they and all calls into must be eliminated at production, even if it's not wrapped like in my code with `if (process.env.NODE_ENV) { warnOnce`

In the past all such helpers came with babel etc transforms. But nowadays we have rollup, esbuild, webpack and this feature is tool-dependent. Probably no npm with support for everything exists, BTW my knowledge is not enough.

kof: in jss and many other places we use https://github.com/alexreardon/tiny-warning/blob/master/src/index.js which warns in non-production env only and works everywhere well, but it doesn't do the warnOnce logic, so we kinda need the same but for warnOnce

rpominov: > maybe we could even overwrite console.warn to log only once?

Please don't. I can easily imagine spending a day questioning my sanity because logs don't work :)

kof: Here is the exact package you need https://github.com/satya164/warn-once/blob/main/index.js

kof: And I know the guy, he is great

kof: > Please don't. I can easily imagine spending a day questioning my sanity because logs don't work :)

yep, dumb idea

istarkov: btw needs to be checked

kof: it shouldn't, that's not how it works in jss with tiny-warning at least, @TrySound was helping there to implement it, I wouldn't worry about it right now, please go ahead use that warnOnce package, if something is off we can still fix rollup for this later

packages/react-sdk/src/components/image.stories.tsx
```diff
@@ -1,18 +1,145 @@
...
-const Template: ComponentStory<typeof ImagePrimitive> = (args) => (
-  <ImagePrimitive {...args} />
-);
+type ImageProps = ComponentProps<typeof ImagePrimitive>;
 
-export const Image = Template.bind({});
+const ImageTemplate: ComponentStory<
+  React.ForwardRefExoticComponent<ImageProps & { css: CSS }>
+> = (args) => {
+  const className = css({ maxWidth: "100%" })({
```
kof, line 18: We discussed that we will be moving away from stitches on canvas long term in favor of something statically extractable,  so it would be good if we didn't have to rewrite stories, its currently only used in one place.

Alternative would be to use style prop, would that work for you?

packages/react-sdk/src/components/image.stories.tsx
```diff
@@ -1,18 +1,145 @@
...
+HeroImage.parameters = {
+  docs: {
+    description: {
+      story: `"sizes" attribute explicitly equal to 100vw allowing to skip the default behavior. See DEFAULT_SIZES in the Image component. Load images depending on the viewport width.`,
+    },
+  },
+};
+
+HeroImage.args = {
+  src: "https://assets-dev.webstudio.is/0beWEKDTrKqaZkh5bdN0peW0oOrcXN_1663019526493.jpg",
```
kof, line 142: lets use some local dummy image instead

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
-type ImageProps = ComponentProps<typeof defaultTag>;
+type ImageProps = ComponentProps<typeof defaultTag> & { quality?: number };
+
+/**
+ * max(...imageSizes) must be less then min(...deviceSizes)
+ * idea is to support images with vw size less than 100vw
+ * https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/image-config.ts
+ **/
+const imageSizes = [16, 32, 48, 64, 96, 128, 256, 384];
+const deviceSizes = [640, 750, 828, 1080, 1200, 1920, 2048, 3840];
```
kof, line 14: What about using breakpoints user has in the project defined? We have default values and sdk/src/css/breakpoints.ts and user can change them and add new once too

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
+    // No image transformation in development
+    resizeOrigin = "https://webstudio.is";
+  }
+
+  const params = [
+    `width=${width}`,
+    quality != null && `quality=${quality}`,
+    // TODO: Must be checked how it works with caching (like avif image is cached and then read by the browser without avif support)
+    "format=auto",
+  ].filter(Boolean);
```
kof, line 47: there is a API for this: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
+
+  return `${resizeOrigin}/cdn-cgi/image/${params.join(",")}/${src}`;
+};
+
+/**
+ * https://github.com/vercel/next.js/blob/canary/packages/next/client/image.tsx
+ **/
+const getWidths = (
```
kof, line 55: this function needs unit tests

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
...
+  width: number | undefined,
+  sizes: string | undefined
+): { widths: number[]; kind: "w" | "x" } => {
+  if (sizes) {
+    // Find all the "vw" percent sizes used in the sizes prop
+    const viewportWidthRe = /(^|\s)(1?\d?\d)vw/g;
+    const percentSizes = [];
+    for (let match; (match = viewportWidthRe.exec(sizes)); match) {
+      percentSizes.push(parseInt(match[2]));
```
kof, line 64: wow, if they made simple mistakes like that, maybe we shouldn't copy their code at all, also we need to check the license, when we copy even if its MIT, we need to do attributions in proper places
I would rather rewrite it

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,187 @@
 import React, { forwardRef, type ElementRef, type ComponentProps } from "react";
```
kof, line 1: we need to think where to put the entire logic that is not specific to our image component implementation (if we think this is actually possible)

reason is that someone could write an image component for a different framework, e.g. vue and reuse this logic.

If you agree, then maybe something like src/component-utils ?

packages/react-sdk/src/components/image.tsx
istarkov: Cloudflare especially wrote in docs:
The path is not URL-encoded, so the resizing URL can be safely constructed by concatenating /cdn-cgi/image/options and the original image URL. So its just path without need of any url encoding. URLSearchParams has nothing with path creating, URL is also not needed here

packages/react-sdk/src/components/image.tsx
kof: I understand that these are popular device sizes and image sizes. Vercel has no info about used media queries on the site I would assume. Another question is if their image resize service is even designed like ours, which allows to pass any width/height dynamically.

In our case we COULD use our known breakpoints to generate optimal sizes specifically designed for the right media query?

Wdyt?

packages/react-sdk/src/arg-types/utils.ts
```diff
@@ -31,6 +31,7 @@ export const propsToArgTypes = (
     }
 
     const control = mapControlForType(prop);
+
```
kof, line 34: would be better to send such changes as separate PR, because this is easy to approve and merge and generates a huge diff, other things are harder to get fully done

packages/react-sdk/src/components/image.tsx
kof: I see, in this case, its not a correctness thing, more like a style thing. Reading it I had no idea if its gonna be encoded later, so I would much prefer to use a safe API. Its just better to standardize on using new URL and new URLSearchParams IMO

kof: Next person in the code base will read your code and build a URL like you did, but in a place where its unsafe. Just don't give them a reason to copy this.

packages/react-sdk/src/components/image.tsx
istarkov: Having that assetId should be used at the upper level component, (not in react-sdk)
I added for now additional Image prop `optimize` defaulting to `false`

With it, image behavior is the same as it was before.

Then at the upper level, we can read the image by assetId and decide whether we need any optimizations. In case of optimization is enabled we have a default loader with a fixed image path. Or we can pass any other image `loader`.

packages/react-sdk/src/components/image.stories.tsx
istarkov: Done

packages/react-sdk/src/components/image.stories.tsx
istarkov: Done, as discussed `USE_CLOUDFLARE_IMAGE_TRANSFORM` added, and organisation logo used.

packages/react-sdk/src/components/image.tsx
istarkov: Added algo description https://github.com/webstudio-is/webstudio-designer/blob/8663e1cd5a6c8519156a3eacb6784f45b9e3b800/packages/react-sdk/src/component-utils/image.ts#L2
With whats now and possible optimisations

packages/react-sdk/src/components/image.tsx
istarkov: Now no url.

packages/react-sdk/src/components/image.tsx
istarkov: radix added

packages/react-sdk/src/components/image.tsx
istarkov: Moved there with Vercel Licence details

packages/react-sdk/src/arg-types/utils.ts
istarkov: https://github.com/webstudio-is/webstudio-designer/pull/386
After merge/rebase that changes would gone

packages/react-sdk/src/components/image-dev.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const pathname = `/cdn-cgi/image/${pathParams.join(",")}/${src}`;
+
+  const url = new URL(pathname, resizeOrigin);
+  return url.href;
+};
+
+const localImageLoader: ImageLoader = ({ width, src, quality }) => {
+  // Just emulate like we really resize the image
+  const params = new URLSearchParams();
+  params.set("width", `${width}`);
```
kof, line 55:
```suggestion
  params.set("width", width);
```

packages/react-sdk/src/components/image-dev.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+    <ImagePrimitive
+      {...args}
+      optimize={true}
+      loader={imageLoader}
+      style={style}
+    />
+  );
+};
+
+export const FixedWidthImage = ImageTemplate.bind({});
```
kof, line 81: btw we probably don't need to use template and bind thing, see for example packages/design-system/src/components/select.stories.tsx

packages/react-sdk/src/components/image-dev.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+      story: `"sizes" attribute explicitly equal to 100vw allowing to skip the default behavior. See DEFAULT_SIZES in the Image component. Load images depending on the viewport width.`,
+    },
+  },
+};
+
+HeroImage.args = {
+  src: imageSrc,
+  sizes: "100vw",
+  style: { aspectRatio: "3/4", objectFit: "cover", width: "100%" },
+};
```
kof, line 200: btw the good thing about stories is that we also have visual regression tests and this is gonna run on ci and ensure the images look the same

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,129 @@
...
+  generateImgAttrs,
+  allSizes,
+  type ImageLoader,
+} from "../component-utils/image";
+
+/**
+ * Default image loader in case of no loader provided
+ * https://developers.cloudflare.com/images/image-resizing/url-format/
+ **/
+const cloudflareImageLoader: ImageLoader = ({ width, src, quality }) => {
```
kof, line 13: why not moving all of this logic to component-utils?

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,129 @@
...
+        ? optimize
+          ? generateImgAttrs({
+              src: imageProps.src,
+              width,
+              quality,
+              sizes,
+              loader,
+            })
+          : { src: imageProps.src }
+        : { src: imagePlaceholderSvg };
```
kof, line 100: I would try to move all of it into component utils, add a fewtests and this way allow someone to create a component in a different library than react while keeping the entire logic as-is

packages/react-sdk/src/components/image.tsx
kof: Btw. this is the issue I was talking about, wanna take it next?

kof: https://github.com/webstudio-is/webstudio-designer/issues/192

packages/react-sdk/src/component-utils/image.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      src: "https://webstudio.is/logo.webp",
+      srcSet: undefined,
+      sizes: undefined,
+      quality: 100,
+    });
+
+    expect(imgAttr).toMatchInlineSnapshot(`
+      Object {
+        "sizes": undefined,
+        "src": "/cdn-cgi/image/width=256,quality=100,format=auto/https://webstudio.is/logo.webp",
```
kof, line 18: is this correct? `format=auto/https://webstudio.is/logo.webp` between the options and the URL I would have expected a comma: `format=auto,https://webstudio.is/logo.webp` but you know better

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,85 @@
...
+import { getImageAttributes, type ImageLoader } from "../component-utils/image";
+
+const getInt = (x: unknown): number | undefined => {
+  if (typeof x === "number") {
+    return Math.round(x);
+  }
+
+  if (typeof x === "string") {
+    const v = Number.parseInt(x, 10);
+    if (`${v}` === x) {
```
kof, line 11: why do you have template here lol
```suggestion
    if (parsedValue === value) {
```

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,85 @@
 import React, { forwardRef, type ElementRef, type ComponentProps } from "react";
+import { getImageAttributes, type ImageLoader } from "../component-utils/image";
+
+const getInt = (x: unknown): number | undefined => {
```
kof, line 4: we have a convention to never use single-char variables like this
```suggestion
const getInt = (value: unknown): number | undefined => {
```

packages/react-sdk/src/components/image.tsx
kof: So following this logic Number.parseInt("10.5", 10) === 10
and it will be false because "10" !== ''10.5"

so you will end up returning undefined, which is probably wrong, because you wouldn't be supporting floating comma string values

while above you do Math.round() which tells me that you actually do want to support floating comma.

packages/react-sdk/src/component-utils/image.test.ts
istarkov: Its right `/`

packages/react-sdk/src/components/image.tsx
kof: so to be correct, you probably need to use Number.parseFloat()

istarkov: parseInt('101.1', 10) gives 101 I'm just checking that it was right Int in x
`${101}` !== `101.1` and should work for intergers only

istarkov: I don't like js `==`comparison because its some kind of undefined behaviour (too much to remember), so convert into right type

packages/react-sdk/src/components/image.tsx
istarkov: IMO in math-like utils short x, y, z, a, b are ok. Sometimes x is just x and has no name or value ;-)

packages/react-sdk/src/components/image.tsx
kof: yes totally, react-sdk sould be on the level of a generic component library from knowledge perspective + meta data specific to webstudio

packages/react-sdk/src/components/image.tsx
kof: so you want to return undefined when user passes "10.5"?

packages/react-sdk/src/components/image-dev.stories.tsx
istarkov: type of set `set(name: string, value: string): void;`

packages/react-sdk/src/components/image.tsx
kof: in this simple function it doesn't mattter, but we need to follow convention, because the next person will write Xs all over the place and point at your function without context as prior art.

always set an example, always assume people won't understand your special context

packages/react-sdk/src/components/image.tsx
kof: > I don't like js `==`comparison because its some kind of undefined behaviour (too much to remember), so convert into right type

me neither, in fact if you look into the code base, you should only find == when comparing to null, to cover both null and undefined, we never use == for anything else

kof: Please add an eslint rule for both

1. == (only allowed to compare to null)
2. some min length for variables min length

istarkov: Yes.  By spec it's an integer value. Having the right inputs someday we can remove all that checks.

kof: Then I don't understand why we don't return undefined if we pass `10.5` as a number, but instead we round it to `10`

istarkov: Fixed, really strange, thank you!

packages/react-sdk/src/components/image-dev.stories.tsx
istarkov: Found serious issue with inline styles during the rewrite, changed default image display type on `block`.

packages/react-sdk/src/components/image-dev.stories.tsx
```diff
@@ -61,12 +60,29 @@ const imageLoader = USE_CLOUDFLARE_IMAGE_TRANSFORM
...
+    ImageProps & { style?: HTMLAttributes<"img">["style"] }
   >
 > = (args) => {
-  const style = { maxWidth: "100%", ...args.style };
+  warnOnce(
+    !(
+      imageMeta.defaultStyle?.maxWidth?.value === 100 &&
+      imageMeta.defaultStyle?.maxWidth?.unit === "%"
+    ),
+    "We expect that the image default style has maxWidth: 100%"
```
kof, line 73: I wonder where we could do this validation so that it can check also user-defined meta if user provides own components, probably in the utility function that is used by the component

kof (approved):
Looking good!

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,86 @@
 import React, { forwardRef, type ElementRef, type ComponentProps } from "react";
+import { getImageAttributes, type ImageLoader } from "../component-utils/image";
+
+const getInt = (value: unknown): number | undefined => {
```
kof, line 4: I would still move more stuff out of the component into component-utils

packages/react-sdk/src/components/image.tsx
```diff
@@ -1,24 +1,86 @@
...
+    const imageAttributes = getImageAttributes({
+      src: imageProps.src,
+      srcSet: imageProps.srcSet,
+      sizes: imageProps.sizes,
+      width,
+      height,
+      quality: Math.max(Math.min(getInt(quality) ?? DEFAULT_QUALITY, 100), 0),
+      loader,
+      optimize: optimize ?? DEFAULT_OPTIMIZE,
+    }) ?? { src: imagePlaceholderSvg };
```
kof, line 57: I would creat one function in component-utils that takes the props and does all of this

istarkov: `ComponentProps<typeof defaultTag>` is react specific i.e event handlers like onLoad etc.
So we can't use it as function arg type. 
Also I would prefer to always have the most narrow type for most functions.

In ideal world `width`, `height` would be already validated `int | undefined` from data, 
so I leaved them here.

As no reason to have them as strings in a system if spec says that it's integer numbers.

kof: you can type the props you need in that function, don't understand the prob

packages/react-sdk/src/components/image-dev.stories.tsx
```diff
@@ -0,0 +1,169 @@
+// Story for image development, see https://github.com/webstudio-is/webstudio-designer/issues/387
+
+import React, { type ComponentProps, type HTMLAttributes } from "react";
```
TrySound, line 3: No default import please. Use only named.

packages/react-sdk/src/components/image-dev.stories.tsx
```diff
@@ -0,0 +1,169 @@
...
+      }}
+    />
+  </div>
+);
+
+/**
+ * "sizes" attribute explicitly equal to 100vw allowing to skip the default behavior.
+ * See DEFAULT_SIZES in the Image component. Load images depending on the viewport width.
+ **/
+export const HeroImage: ComponentStory<React.FunctionComponent> = () => (
```
TrySound, line 159: Perhaps no need to add types at all?

packages/react-sdk/src/components/image-dev.stories.tsx
kof: lets add an eslint plugin for this if it exists?

packages/react-sdk/src/components/image-dev.stories.tsx
istarkov: Could be usefull later to add args or params

istarkov: `i.e. HeroImage.params = {docs: {...}}`

packages/react-sdk/src/components/image-dev.stories.tsx
istarkov: Mb I'm wrong but since React 17 we don't need to import * as React in every place jsx is used?
I haven't written on React for 1-2 years soon, and mb its tooling dependent but even ts shows me that React is unused

istarkov: I would leave it for next PRs as now it's in every place I see in sdk. Easier to fix all in a single PR.

Added here https://github.com/webstudio-is/webstudio-designer/issues/397#issuecomment-1304891614

packages/react-sdk/src/components/image.props.json
```diff
@@ -2538,5 +2538,41 @@
...
+      }
+    ],
+    "required": false,
+    "type": {
+      "name": "ImageLoader"
+    },
+    "control": {
+      "type": "text"
+    }
+  },
```
rpominov, line 2559: Just something to keep in mind, not necessarily an issue depending on how we going to use this prop: 

We keep props in DB as a JSON object, and it's impossible to store a function there.

rpominov (approved):

PR #379, opened 2022-11-02
Remaining style fixes for fonts manager

kof:
## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


taylornowotny:
All looks great now!

taylornowotny (approved):

packages/icons/src/cross-circled-filled.tsx
```diff
@@ -16,8 +16,6 @@ export const CrossCircledFilledIcon = React.forwardRef<
       ref={forwardedRef}
     >
       <path
-        fill-rule="evenodd"
-        clip-rule="evenodd"
```
rpominov: Strange that removed here but renamed in `magnifying-glass.tsx`? Is this a mistake?

rpominov (approved):

packages/icons/src/cross-circled-filled.tsx
kof: I just checked they were not needed it seems (didn't change the appearance) ... but idk maybe I should have left them in?

rpominov: It's either because these values are defaults or because the shape is such that it doesn't matter. But I'd copy SVG as is from Figma just to be safe.

kof: yeah, they didn't matter in this case, will readd them nevertheless

PR #383, opened 2022-11-03
Generate actual `/server` entries in packages

Andarist:
I think that this should solve your watching-related problems while also fixing some other stuff.

The watch-related problem was that you didn't trigger any rebuilds during dev when editing files in packages that were only impacting your server files. The reason behind this was that those files were not part of the `tsup` config - so it didn't rebuild when you were editing those.

I've fixed that by including the server entries in the `tsup` configs. To make it work I also had to introduce `server.js` and `server.d.ts` files in the affected packages. Without those tools (TS, jest, Remix) wouldn't know how to find your dist files. In the future, it might be possible to use `package.json#exports` to achieve this goal. However, I don't recommend trying to do this right now because you'd have to use `moduleResolution: 'node16'` in TypeScript, and that would require you to add `.js` extension to all of your relative imports etc. This is a bigger change that should be done separately if you'd like to pursue this.

By creating actual entry points for those server files we also cut down on the final bundlesize a little bit because without this Remix was pulling files from `/src` when importing server files but for client/common files it was pulling from `/lib`. This means that the code shared by those files wasn't actually shared at runtime.


packages/asset-uploader/package.json
```diff
@@ -37,5 +37,13 @@
     "!*.test.*"
   ],
   "license": "MIT",
-  "private": false
+  "private": false,
+  "tsup": {
```
Andarist, line 41: I had to move the tsup-related config somewhere because we want to preserve `.server.js` convention. It's not possible to define entries that contain dots using CLI args because `tsup` thinks that we want to create a nested object (it splits the given value by a dot and it's not possible to "escape" it anyhow)

rpominov (approved):
> This means that the code shared by those files wasn't actually shared at runtime.

Great that this is fixed, could also cause very confusing bugs because two instances of some library would be initialised.

package.json
```diff
@@ -41,5 +41,8 @@
   "lint-staged": {
     "*.{ts,tsx,js,json,css,md}": "prettier --write"
   },
-  "prettier": {}
+  "prettier": {},
+  "resolutions": {
+    "esbuild": "0.15.13"
+  }
```
Andarist, line 47: `esbuild@0.15.8` is needed as it comes with bundling fixes: https://github.com/evanw/esbuild/releases/tag/v0.15.8

apps/designer/remix.config.js
```diff
@@ -9,10 +9,7 @@ module.exports = {
   // server. This does not understand the vercel lambda module format,
   // so we default back to the standard build output.
   server: process.env.NODE_ENV === "development" ? undefined : "./server.js",
-  // This makes sure remix bundles monorepo packages which are always
-  // sym-linked and can't be hoisted. You could also manually name packages
-  // e.g. ["@webstudio-is/ui", ...]
-  serverDependenciesToBundle: [/@webstudio-is\/.*/, "pretty-bytes"],
```
Andarist, line 15: I didn't see

apps/designer/remix.config.js
```diff
@@ -9,10 +9,7 @@ module.exports = {
   // server. This does not understand the vercel lambda module format,
   // so we default back to the standard build output.
   server: process.env.NODE_ENV === "development" ? undefined : "./server.js",
-  // This makes sure remix bundles monorepo packages which are always
-  // sym-linked and can't be hoisted. You could also manually name packages
-  // e.g. ["@webstudio-is/ui", ...]
```
Andarist, line 14: I didn't see a strong reason to bundle them. Monorepo packages should behave like regular packages - for the most part. 

Removing them from there fixes bundling issues around the `sharp` module. This module **can't** be bundled.

packages/project/package.json
```diff
@@ -18,12 +18,12 @@
     "@webstudio-is/react-sdk": "*",
     "bson-objectid": "^2.0.2",
     "immer": "^9.0.12",
-    "nanoid": "^3.2.0",
     "slugify": "^1.6.5",
     "uuid": "^9.0.0",
     "zod": "^3.19.1"
   },
   "devDependencies": {
+    "nanoid": "^3.2.0",
```
Andarist, line 26: This is weird-ish and shouldn't be required. However, `nanoid` is an ESM-only package and as such, it can't be required. So we need to bundle it into `@webstudio-is/project` with `tsup` (and that's why it has been moved from deps to devDeps). This way Remix won't try to `require` it.

apps/designer/app/routes/$.tsx
```diff
@@ -8,7 +8,7 @@ import { useLoaderData } from "@remix-run/react";
 import { InstanceRoot, Root } from "@webstudio-is/react-sdk";
 import { loadCanvasData, type CanvasData } from "~/shared/db";
 import config from "~/config";
-import { db } from "@webstudio-is/project/index.server";
+import { db } from "@webstudio-is/project/server";
```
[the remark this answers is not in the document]
Andarist: This actually still references `something.server` because this `server` file just reexports `index.server.js`. However, I've retested this - and it seems that this part doesn't work now.

I think the fix here is more important than this Remix safeguard. I could investigate separately how to make this work later on though. I think though that maybe it would be better to introduce a throwing expression there that would work regardless of the Remix convention, what do you think? It would still prevent the modules from being imported by the client code.

Andarist:
@kof @rpominov I think that this is good to go now but as always - please test it out or let me know asap if anything unexpected happens after this lands.

apps/designer/remix.config.js
kof: sharp should be only used serverside ever

packages/project/package.json
kof: omg .... there must be some generic solution to that cc @TrySound

kof: problem is its going to confuse people for being in devDependencies and they will potentially move it back to dependencies

apps/designer/remix.config.js
Andarist: yes, and it was - but after my changes, it started to be bundled by Remix (whereas on the main branch `require('sharp')` is left in the bundle). I'm not exactly sure why that has changed - it's probably something related to how Remix/esbuild interpret externals or to how Remix uses this `serverDependenciesToBundle`. By bundling sharp we screwed up a relative require contained in that package:
https://github.com/lovell/sharp/blob/9fa516e8496d96fc12a4cba25216e467e36d62b7/lib/sharp.js#L7

packages/project/package.json
Andarist: ESM-only packages tend to be problematic in the ecosystem. This is even called out by the Remix docs here:
https://remix.run/docs/en/v1/pages/gotchas#importing-esm-packages

This thing here was somewhat required by the other change that I've made to `serverDependenciesToBundle`:
https://github.com/webstudio-is/webstudio-designer/pull/383#discussion_r1014602193

Because now I don't bundle the `@webstudio/*` packages I can't just bundle something that they require - it's impossible. 

Perhaps I should try to figure out why `sharp` has started to be bundled now instead but I think it's better to land this to fix your immediate issues and try to investigate this further separately. Once I would get to the bottom of the sharp thing we could revert this change.

Note though that I stand by the comment made about the `serverDependenciesToBundle`. Usually, there shouldn't be any reason to bundle your monorepo packages like this. I suspect this was already a bandaid for the problems like this one - consuming ESM-only packages deeper in the dep graph etc. So while conceptually it shouldn't have to be done, there might be some technical limitation like this that might make it a lesser evil.

It's also worth noting that by relying on ESM-only packages your packages are currently a little bit broken anyway (if they would be consumed from npm). Your packages are at the moment CJS-only and that is incompatible with ESM-only packages.

Andarist: > problem is its going to confuse people for being in devDependencies and they will potentially move it back to dependencies

Sure, I agree with that. That's why I'm proposing to land this as a stopgap solution - if it doesn't break anything for your right now - and improve that soon.

kof: > @webstudio/*

There was no way using a different matcher than this? one that understands .server?

Andarist: Hmm, those are somewhat unrelated. Your server can still import regular non-.server files and you could end up with the same problem.

To work around the esm-only issue and to continue bundling your monorepo packages for your server bundles we'd like to:
- use `serverDependenciesToBundle: ['@webstudio/*']`
- control external and put `sharp` there. However, Remix doesn't allow you to do that. **Maybe**, you can declare `sharp` as a dependency of your designer app and that would force Remix to infer it as being external. This could be confusing as well though because your Remix app doesn't actually directly depend on `sharp` so it would look redundant. 

In Remix it seems that the bundle/not-bundle decision for the server entry is taken [here](https://github.com/remix-run/remix/blob/3832d457b34d7a62633bf86a2606971263e710cb/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L15-L19). So I could investigate this. On the other hand... once you start bundling some stuff I really think that you should just go all in. I'd have to think about it more to assess the exact risks but my hot take is that "partial bundling" can almost always manifest in some obscure bugs. They might be edge cases and you might never encounter them but that doesn't mean that they are impossible.

kof: Oh I don't care if we bundle something or not, esbuild is fast enough. I only care that we 
1. don't add serverside stuff to the client
2. have normal way of defining dependencies

Andarist: Do you want any of those fixed as part of this PR? Or can I investigate those in followups?

kof: up to you, I would try and bundle everything with esbuild so we never ever have to worry about esm vs cjs

kof: Lets create a follow-up issue if that's better

kof (approved):

apps/designer/app/routes/$.tsx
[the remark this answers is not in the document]
rpominov: If after we merge this PR what was previously server only code will appear in client side bundles we probably should solve this before merging, or maybe I misunderstood something.

> I think though that maybe it would be better to introduce a throwing expression there that would work regardless of the Remix convention, what do you think? It would still prevent the modules from being imported by the client code.

Not sure I understand. Do you mean throwing an error in browser context so the code wouldn't run in the browser?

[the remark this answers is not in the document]
rpominov: Maybe we should actually just do `import { db } from "@webstudio-is/project";` everywhere, and remove these entry points. 

It should just work because for the client bundles remix does not import loaders and actions functions so they will be tree-shaken and all imports that they use along with them.

Remix docs say that `.server` shouldn't be necessary usually.

And for the rare cases when it's necessary, I think a better approach is to create a `*.server` file in the app rather than in a package.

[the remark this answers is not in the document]
rpominov: @kof do you know why we started to import from `.server` rather than normally? Is it to be extra safe or have we actually had a problem of backend code appearing in the browser?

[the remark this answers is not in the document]
kof: its an actual problem once you end up importing in client side server stuff, remix will crash at you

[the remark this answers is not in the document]
kof: creating .server files in the app instead of a package is indeed an option, as long as contributors to the package always get the error if that change can't be used in the app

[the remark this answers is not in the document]
rpominov: I've pulled this PR locally, run `yarn dev`, and got this error in the browser: Screenshot 2022-11-05 at 18 09 37

So it seems `import ... from ".../server"` is not the same as `import ... from ".../something.server"` for Remix.

I guess we have two options:

 1. Remove `/server` entries and put all exports in `index.ts` normally. In the app import via intermediary `*.server.ts` files
 2. Just rename these `server.js` files to `index.server.js` so the imports would look like `import ... from ".../index.server"`
 
I like `#1` a bit better but no strong opinion, and not sure `#2` will work.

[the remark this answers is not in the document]
Andarist: > If after we merge this PR what was previously server only code will appear in client side bundles we probably should solve this before merging, or maybe I misunderstood something.

Nothing should change - but the risk of introducing a server-related import to the client bundle increases here. I will take a second look into how Remix actually implements their `.server` convention and maybe I will be able to do smth about it.

> Not sure I understand. Do you mean throwing an error in browser context so the code wouldn't run in the browser?

Yes, smth like that. With that, it wouldn't rely on the Remix convention and it would work in the same way "everywhere".

> It should just work because for the client bundles Remix does not import loader and action functions so they will be tree-shaken and all imports that they use along with them.

Unfortunately, it's not always that simple and thus relying on this introduces some new risks.

> And for the rare cases when it's necessary, I think a better approach is to create a *.server file in the app rather than in a package.

That I agree with - `*.server` convention feels mostly relevant for the app code. Packages should be more reusable and should work outside of Remix etc. That being said - sometimes it's useful to have different entries per environment. For example, React comes with `react-dom` and `react-dom/server` (and even `react-dom/client` that has been added in React 18).

My first instinct would be to say that:
- you either have a lib that works in any env and you only provide a single entrypoint 
- or you have a lib that needs different implementation per environment but it still works in all~ of them and then u can have a single entrypoint but you need to alias that entrypoint per env (nowadays this gets easier with `package.json#exports`)
- or you have a library that only works in a subset of the available envs. You can have only a single entrypoint there as importing it in an incompatible env would crash one way or another so you probably would notice it
- or you have a library where a subset of the API only works in some specific env. In this case, I recommend having multiple entrypoints (like in this PR).

What you probably don't want to have is a single entrypoint in which some APIs are isomorphic but some APIs are env specific. I would classify this as a code smell. I mean that this probably ain't a good idea:
```js
import { worksEverywhere, onlyWorksInNode } from 'pkg'
```

Of course, there are no hard rules here and you might sometimes break out of the listed scenarios and still be fine... but the line is blurry and it gets harder to control this. The mentioned scenarios are somewhat straightforward and I would stick to them. It's especially important not only because of the imported APIs but because of what they might depend on.

The mentioned tree-shaking scenario and isomorphic files like in the case of Remix blur the line BUT the "mixed" entrypoint can still be a problem. Tree-shaking (with `sideEffects: false` in the imported packages) might perform much better if a file has a mixed client/server code while not having mixed imports - if imports are either isomorphic, client-only, or server-only, you can more safely rely on bundling and tree-shaking.

> cannot initialize routeModules

Hmm, I've seen this before some of my fixes but I've thought that I already got rid of this. Thanks for checking and I gonna take another pass at this.

Note that in this case, it might be unrelated to the `.server` convention. I got this exact error before simply because one client-only chunk had a `ReferenceError` in it and it failed to initialize some global variables used by Remix for rehydration. But if this still happens I need to recheck why this happens now.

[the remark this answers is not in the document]
kof: The problem is that when writing code in the package, you don't know if its being used serverside. If we can show the user rightaway that the change is now going to bring the server-side code to the client - then we may be able to get away with it.

It has to crash the app when user is bringing server code to the client. If we can do that, we can keep .server to the app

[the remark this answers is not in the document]
kof: Another big worry of mine is that client code can get stuff in that its not supposed to get just because we write the code in packages in a way that can't be treeshaken and in addition there is no .server hint.

I am super worried about growing bundle size for the client and it has as of right now already big problems we need to investigate. It is way bigger than it should be.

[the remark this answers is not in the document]
kof: It may have to do with the fact that we have canvas and parent document and we load twice the things that are actually the same and could be taken from browser cache. I am not sure.

I bet our bundler can't see the difference between iframe stuff and main window: e.g. see https://webstudio-designer-7humw6tf0-webstudio-is.vercel.app/designer/5e1de84c-25c6-4cfb-913f-280c295a4199

half of the files are the same, visible by the size right away. So we probably 1/2 from 1.5mb is unnecessary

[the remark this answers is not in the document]
kof: Just mentioning this to make you aware, maybe there is a way to handle all of this better, but maybe its completely unrelated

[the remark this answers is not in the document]
Andarist: > It has to crash the app when user is bringing server code to the client. If we can do that, we can keep .server to the app

Ye, so with the `throw`-based idea this should very much be possible. 

> Another big worry of mine is that client code can get stuff in that its not supposed to get just because we write the code in packages in a way that can't be treeshaken and in addition there is no .server hint.

Isn't this always the risk if you rely on tree-shaking? I mean, you don't truly control what gets included and what is not. Using a single package can bring megabytes to your bundles. There is no granular control over that and the best mitigation is to monitor your bundles, your dependencies, and sometimes improve your own module graph if needed.

I might be confusing what do you have in mind though. Could you define "client code can get stuff in that its not supposed to get"? Are you purely concerned here from the bundlesize perspective? Or do you mean potential runtime crashes etc?

> It may have to do with the fact that we have canvas and parent document and we load twice the things that are actually the same and could be taken from browser cache. I am not sure.

This could be mitigated using multiple entrypoints. The bundler should be able to share common chunks between multiple entries. I'm not sure if this is supported by Remix and esbuild in the way that I know and understand this. It would very much be possible (and the best solution imho) in the webpack/rollup world. I can only assume that the same should be supported here too. To assess this I would have to take a closer look at your setup related to this stuff.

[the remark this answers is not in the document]
Andarist: I think that the fact that your `pkg/index.server` worked was a little bit accidental. The comment in the Remix's source code suggests that they only want to allow this convention for the files in the `/app` directory and they even mention there that it should be allowed for node modules to use `.server.js` with different semantics:
https://github.com/remix-run/remix/blob/cb7416beb87f2f008f05bafb0c25d263bf59851f/packages/remix-dev/compiler/plugins/emptyModulesPlugin.ts#L18-L26

The problem~ with their implementation is that they simply call `path.resolve(args.resolveDir, args.path)` and that doesn't resolve module specifiers. So the "resolved" path looks something like:
```
~/Code/webstudio-designer/apps/designer/app/routes/designer/@webstudio-is/project/index.server
```
and that satisfies their check. Should it though? I don't quite think so - I think this could even be reported as a bug to the Remix team.

I **think** that I've managed to fix most of the issues by reverting some of the changes introduced here previously and by introducing `sideEffects: false` to your packages. Thanks to this flag unused imports in your isomorphic files might be dropped entirely by the bundler (`import { worksEverywhere, onlyWorksInNode } from 'pkg'` might still be problematic but conceptually you try to stay away from exporting things like that so you should be fine).

I still think that you:
- either shouldn't bundle your monorepo packages on the server
- or bundle everything for the server.

This might be a separate discussion though so I've removed those changes from this PR and I hope that this will now work as expected.

Note that you are still using esm-only `nanoid` and this might create problems (just like any esm-only package consumed from your CJS bundle). But let's discuss and solve this separately later.

rpominov (approved):
Seems to work great. Tested locally and in preview. Total size transfered is the same as in `main` in preview (at least in `/designer`)

TrySound (approved):

kof:
merged 😅

PR #385, opened 2022-11-03
Font weights

kof:
- [x] render only valid font weights as select box
- [x] hide weights that are not available with the currently selected font
- [x] correctly handle inherited font-family/weight

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


taylornowotny (approved):

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const fontFamily = getFinalValue({
+    currentStyle,
+    inheritedStyle,
+    property: "fontFamily",
+  });
+
+  const { labels, selectedLabel, availableFontWeights } = useFontWeights(
+    toValue(fontFamily, { withFallback: false }),
+    toValue(fontWeight)
+  );
```
rpominov, line 72: It seems that `fontWeight` here might be not one of `fontFamily`'s available weights. Is this case handled well?

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  if (fontWeight === undefined) return null;
```
rpominov, line 74: Is this just for TS, or is it possible for current `fontWeight` to be `undefined`? If possible, is it the right behaviour to not render the control in such case?

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
```diff
@@ -0,0 +1,103 @@
...
+      options={labels}
+      value={selectedLabel}
+      onChange={(label) => {
+        const selected = availableFontWeights.find(
+          (option) => option.label == label
+        );
+        if (selected) {
+          setValue(selected.weight);
+        }
+      }}
```
rpominov, line 89: Not ideal that labels are used as ids. Shouldn't be an issue, but maybe we need a comment in the code above where labels are built saying that they must be unique.

packages/fonts/src/font-data.ts
```diff
@@ -41,32 +22,47 @@ export const parseSubfamily = (subfamily: string) => {
...
+};
+
+export const normalizeFamily = (family: string, subfamily: string) => {
+  let simplifiedFamily = family;
+  subfamily
+    .trim()
+    .split(" ")
+    .forEach((word: string) => {
+      simplifiedFamily = simplifiedFamily.replace(
+        new RegExp(word.trim(), "i"),
```
rpominov, line 44: Special regexp characters must be escaped in `word`

packages/fonts/src/font-data.ts
```diff
@@ -41,32 +22,47 @@ export const parseSubfamily = (subfamily: string) => {
...
+    const { alt } = fontWeights[weight];
+    if (subfamilyLow.includes(name) || subfamilyLow.includes(alt)) {
       break;
     }
   }
-  return { style, weight };
+  return { style, weight: Number(weight) };
+};
+
+export const normalizeFamily = (family: string, subfamily: string) => {
```
rpominov, line 37: I couldn't understand what this function does 😅

packages/fonts/src/font-data.ts
rpominov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll — might be helpful

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
kof: yes, that's exactly the point of that function to get the available font weights, refactored it into 2 functions and renamed to make it more clear

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
kof: I believe its possible, depending on what getFinalValue() does now or will do in the future, right now it uses both, style from our data and from the computed style browser, so if non of them have that value, it can be undefined

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
kof: those are a fixed set of font weight, 9 to be precise, they are standardized, no way of clashing, will add a comment since this wasn't obvious

packages/fonts/src/font-data.ts
kof: you are right, I am using regex just because of case sensitivity option, I can rewrite it and not use regex

packages/fonts/src/font-data.ts
kof: explained, helped?

rpominov: Yes, but the logic inside the function seem very unpredictable. For example: `normalizeFamily("Robolder Bold", "Bold")` will produce `"Roer Bold"`

kof: you are right, I was operating on a subset of values

rpominov: Still feels kinda unpredictable. Is it required that `normalizeFamily("Roboto Bold", "Bold") === normalizeFamily("Roboto Black", "Black")`? If yes, how do we know user won't upload some weird font for which this will not be the case?

rpominov: Also, why do we assume they will be split by " "? What if it `normalizeFamily("RobotoBold", "Bold")`.

Maybe I'm overthinking it, and the output of this function is not important. I don't understand well where the result passed down the line and what the requirements for it.

kof: > Also, why do we assume they will be split by " "? What if it normalizeFamily("RobotoBold", "Bold").

In that case we don't care, the font name is RobotoBold then, its weird but its the name. What we are trying to separate is the weight keyword from the font name, which happens to be the case most of the time in font naming.

kof: > Is it required that normalizeFamily("Roboto Bold", "Bold") === normalizeFamily("Roboto Black", "Black")
Yes we kind need to let user use "Roboto" and bold is the weight we allow to select separately. Same font name, different weights

kof: > If yes, how do we know user won't upload some weird font for which this will not be the case?

We normalize what we can normalize, if we can't - it will be displayed as a standalone separate font, user can still use it, but its a separate font name.

rpominov: Ok, got it 👍

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
kof: oh yeah, we need to think later about this check in all controls, atm it never happened, maybe because there is always a value, so in this case its just typescript satisfaction:
e.g. getComputedStyle(document.body.parentElement) returns more properties than we support in style panel, things that aren't set have initial value setup, so I think it actually returns the entire CSS, even though the spec says it only returns things that apply. Maybe it doesn't return properties that can't be applied to a certain type of element at all ..
its an edge case in any case, so maybe we should just silence typescript and remove this check everywhere

kof: should be separate pr though, because this check is in all controls

PR #391, opened 2022-11-04
Text control prefix icon

ghost:
Returns the prefix icon that was removed.
- API for prefix icon
- Add tooltip to css-input-value so that it's accessibly when property-name is not present.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing
Screenshot 2022-11-05 at 2 13 24 AM

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -268,7 +268,7 @@ export const CssValueInput = ({
   });
 
   const shouldHandleEvent = useCallback((node) => {
-    return !suffixRef.current?.contains?.(node);
+    return suffixRef.current?.contains?.(node) === false;
   }, []);
   const [scrubRef, inputRef, isInputActive] = useScrub({
     value,
@@ -296,7 +296,7 @@ export const CssValueInput = ({
   const prefix = icon && (
     <CssValueInputIconButton
       state={isCurrentBreakpoint ? "set" : undefined}
-      css={value.type == "unit" ? { cursor: "ew-resize" } : {}}
+      css={value.type == "unit" ? { cursor: "ew-resize" } : undefined}
     >
       {icon}
     </CssValueInputIconButton>


apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -19,34 +22,51 @@ export const TextControl = ({
...
-        const prevValue = toValue(value);
-        const nextValue = toValue(styleValue);
-        if (prevValue === nextValue) return;
-        setCurrentValue(undefined);
-        setValue(toValue(styleValue));
-      }}
-    />
+    <Tooltip
+      content={styleConfig.label}
+      delayDuration={600}
```
kof, line 28: should we have a single source of all delays in style panel?

apps/designer/app/designer/features/style-panel/shared/property-icon.tsx
```diff
@@ -0,0 +1,17 @@
...
+import type { PropertyProps } from "../style-sections";
+import { forwardRef } from "react";
+
+export const PropertyIcon = forwardRef<
+  HTMLButtonElement,
+  PropertyProps & { children: JSX.Element }
+>(({ property, children }) => {
+  const isCurrentBreakpoint = useIsFromCurrentBreakpoint(property);
+  return (
+    <TextFieldIconButton state={isCurrentBreakpoint ? "breakpoint" : undefined}>
```
kof, line 12: 2 problems:
1. We also need to show the resize icon for the scrub
2. TextFieldIconButton isn't really aware of feature specific concepts like a breakpoint, so you are leaking abstraction now.

We need to solve this  by making CssValueInput control both of those things, since its the only one aware of both.

Do you know if we are going to need PropertyIcon anywhere else than CssValueInput? If no - this can be moved into css-value-input folder.

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -249,10 +265,15 @@ export const CssValueInput = ({
     },
   });
 
-  const [inputRef, isInputActive] = useScrub({
+  const shouldHandleEvent = useCallback((node) => {
+    if (suffixRef.current?.contains?.(node)) return false;
+    return true;
```
kof, line 270:
```suggestion
    return suffixRef.current?.contains?.(node)
```

packages/design-system/src/components/text-field.tsx
```diff
@@ -314,8 +328,8 @@ export const TextField = React.forwardRef<HTMLDivElement, TextFieldProps>(
           ref={mergeRefs(internalInputRef, inputRef ?? null)}
         />
 
-        {prefix && <PrefixSlot>{prefix}</PrefixSlot>}
-        {suffix && <SuffixSlot>{suffix}</SuffixSlot>}
+        {prefix && <PrefixSlot ref={prefixRef}>{prefix}</PrefixSlot>}
```
kof, line 331: you control the prefix element outside, so you should be getting the ref to it instead of using the slot

packages/design-system/src/components/text-field.tsx
```diff
@@ -55,6 +55,14 @@ export const TextFieldIconButton = styled(
             color: "white",
           },
         },
+        highlight: {
```
kof, line 58: sure ,but then later you also need the orange one, for inherited values

kof: We should rather use that icon button and have its own states in CSS input

styled(TextFieldIconButton, {...specific stuff for style panel})

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
kof: ping

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -249,10 +266,14 @@ export const CssValueInput = ({
     },
   });
 
-  const [inputRef, isInputActive] = useScrub({
+  const shouldHandleEvent = useCallback((node) => {
+    return !suffixRef.current?.contains?.(node);
```
kof, line 270: hard to read, don't do the !thing please

```suggestion
    return suffixRef.current?.contains?.(node) === false;
```

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -270,17 +291,31 @@ export const CssValueInput = ({
...
-        <ChevronDownIcon />
-      </TextFieldIconButton>
-    ) : value.type === "unit" ? (
-      unitSelectElement
-    ) : null;
+  const isCurrentBreakpoint = useIsFromCurrentBreakpoint(property);
+  const prefix = icon && (
+    <TextFieldIconButton
+      state={isCurrentBreakpoint ? "set" : undefined}
+      css={value.type == "unit" ? { cursor: "ew-resize" } : {}}
```
kof, line 298: I believe it can be undefined 
```suggestion
      css={value.type == "unit" ? { cursor: "ew-resize" } : undefined}
```

packages/design-system/src/components/combobox.tsx
```diff
@@ -82,7 +82,9 @@ export const ComboboxListboxItem = forwardRef(ListboxItemBase);
 
 export const ComboboxPopper = Popper;
 
-export const ComboboxPopperContent = PopperContent;
+export const ComboboxPopperContent = styled(PopperContent, {
+  zIndex: 10,
```
kof, line 86: we haven't needed it so far, why starting now? on design system level we shouldn't need zIndex

packages/design-system/src/components/text-field.tsx
```diff
@@ -55,6 +55,22 @@ export const TextFieldIconButton = styled(
             color: "white",
           },
         },
+        set: {
```
kof, line 58: like I said in a PR before, text field has no idea of set/inherited semantics, its a component that is used outside of style panel.

I don't know why its not clear automatically.

kof: create your own component in css value input:

```ts
const CssValueInputIconButton = styled(TextFieldIconButton, {
  variants: {
    set: {},
    inherited: {}
  }
})
```

kof: or something like this

packages/design-system/src/components/combobox.tsx
kof: got it, lets us $1 instead of 10
```suggestion
  zIndex: '$1',
```

packages/design-system/src/components/text-field.tsx
kof: active is a generic state for this icon button, its the same across all usages, not just style panel

kof (approved):

PR #392, opened 2022-11-05
Add a file transform to jest to handle assets

### LINKED ISSUE #387: Importing static assets into react-sdk stories ruins test for use-props-logic.test.ts
Steps to reproduce.

Add `import localLogoImage from "../../storybook-assets/logo.webp";` to [react-sdk image story](https://github.com/webstudio-is/webstudio-designer/blob/main/packages/react-sdk/src/components/image.stories.tsx#L5)

Run `yarn checks` at the project root

output:

```log
@webstudio-is/designer:checks: FAIL app/designer/features/props-panel/use-props-logic.test.ts
@webstudio-is/designer:checks:   ● Test suite failed to run
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     Jest encountered an unexpected token
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     By default "node_modules" folder is ignored by transformers.
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     Here's what you can do:
@webstudio-is/designer:checks:      • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
@webstudio-is/designer:checks:      • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
@webstudio-is/designer:checks:      • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
@webstudio-is/designer:checks:      • If you need a custom transformation specify a "transform" option in your config.
@webstudio-is/designer:checks:      • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     You'll find more details and examples of these config options in the docs:
@webstudio-is/designer:checks:     https://jestjs.io/docs/configuration
@webstudio-is/designer:checks:     For information about custom transformations, see:
@webstudio-is/designer:checks:     https://jestjs.io/docs/code-transformation
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     Details:
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     /root/webstudio/webstudio-designer/packages/react-sdk/storybook-assets/logo.webp:1
@webstudio-is/designer:checks:     ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){RIFF�"
@webstudio-is/designer:checks:                                                                                           
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:     SyntaxError: Invalid or unexpected token
@webstudio-is/designer:checks: 
@webstudio-is/designer:checks:       11 | // eslint-disable-next-line @typescript-eslint/ban-ts-comment
@webstudio-is/designer:checks:       12 | // @ts-ignore
@webstudio-is/designer:checks:     > 13 | const logo_webp_1 = __importDefault(require("../../storybook-assets/logo.webp"));
@webstudio-is/designer:checks:          |                                     ^
@webstudio-is/designer:checks:       14 | exports.default = {
@webstudio-is/designer:checks:       15 |     title: "Components/Image",
@webstudio-is/designer:checks:       16 |     component: image_1.Image,
```

Andarist:
closes #387

Note that the issue mentions adding an asset import to a story in `react-sdk`. Right now you also get type errors there on those. In the designer app you don't get those because types for assets are declared within the `@remix-run/dev` typedefs that you reference:
https://github.com/webstudio-is/webstudio-designer/blob/a9c2b0e6b9a82ea7f1ff626c47519b8404a67b04/apps/designer/remix.env.d.ts#L1

This is a separate problem though and should be handled separately.


apps/designer/jest.config.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @type {import('@jest/types').Config.InitialOptions}
+ */
 module.exports = {
```
Andarist, line 4: q: any strong reason why this jest config is not reusing the shared one?

packages/jest-config/index.js
```diff
@@ -1,8 +1,12 @@
+/**
+ * @type {import('@jest/types').Config.InitialOptions}
+ */
 module.exports = {
   testEnvironment: "node",
   testMatch: ["<rootDir>/src/**/*.test.ts"],
   transform: {
     "^.+\\.tsx?$": "esbuild-jest",
+    "^.+\\.webp$": require.resolve("./file-transform.js"),
```
Andarist, line 9: alternatively, you could use a "reverse" regex like `"^(?!.*\\.(js|jsx|mjs|cjs|ts|tsx|css|json)$)"`. This would handle every non-js/css/json asset with this `file-transform`. I'm personally more comfortable with the explicit allowlist though.

packages/jest-config/file-transform.js
```diff
@@ -0,0 +1,7 @@
+const last = (arr) => arr[arr.length - 1];
+
+module.exports = {
+  process(src, filename) {
+    return `module.exports = ${JSON.stringify(last(filename.split("/")))};`;
```
istarkov, line 5: mb native api?
```
const path = require("path");
path.basename("hello/world.js")
```

Andarist: Ye, this is also an option - I mostly did this to avoid ESlint complaints:
```
Require statement not part of import statement.eslint[@typescript-eslint/no-var-requires](https://typescript-eslint.io/rules/no-var-requires)
```

Happy to use `require` and suppress the error though if you prefer that

istarkov: I just have no idea how it will work for windows users, or do we support them ;-)

istarkov (approved):

packages/jest-config/file-transform.js
Andarist: I think that this might come here already normalized - but ye, I'm not 100% sure about this.

kof: to avoid eslint complaints we can disable it for the line)))

apps/designer/jest.config.js
kof: we probably just forgot to move it once we created shared one

kof: seems like we even improved it in designer but not for other packages, I remember trying to make inline snapshots without success in a package and gave up, lets merge the 2 together and move it

packages/jest-config/file-transform.js
Andarist: ok, i went that way

apps/designer/jest.config.js
Andarist: let's try to unify it in some followup PR then

kof: https://github.com/webstudio-is/webstudio-designer/issues/393

kof (approved):

kof:
Is this waiting to get merged? @Andarist you can merge yourself actually, right? I think you are in the core team permissions group

kof:
we should unify jest config (remove the copy paste from apps/designer and just use it from packages everywhere), it has already resulted in a merge conflict

PR #394, opened 2022-11-05
Hydrate text editor instances

TrySound:
isNew and createInstance flags makes state management more complex. Instead tree of instances updates should be idempotent. This will let us do more straightforward mapping from lexical nodes to designer tree.


kof:
seems to be working fine, I wonder what I was thinking as I created this mess lol

kof (approved):

PR #399, opened 2022-11-06
[jest] Allow inline snapshots

istarkov:
Cherry picked from #378

`expect(obj).toMatchInlineSnapshot()`
cause `Jest: Couldn't locate all inline snapshots.` error without source maps support enabled.


rpominov (approved):

kof (approved):
we should remove that config in the app and make it use this one everywhere

Issue #289, opened 2022-09-20
Trying to make a part of a Bold text Italic results in duplication of text


rpominov:
1. Start editing a Paragraph. For example, it has text "hello world"
2. Make entire text bold "hello world" -> "**hello world**"
3. Try to make the word "world" italic
4. The result is not  "**hello _world_**" but "**hello world**_world_"

TrySound:
How to represent overlap in navigator? If we remove style their whole range is removed. Should we add something like "Italic Bold Text"?

rpominov:
Webflow does something like this:

```
Paragraph
 |- Bold
 |- Bold
     |- Italic
```

We should probably do the same, but "Italic Bold Text" might be also a good option.

/cc @kof

kof:
Aren't we technically nesting bold in italic or italic in bod depending on what was first? If we are nesting them in HTML then we should represent the same structure in navigator I guess?

PR #412, opened 2022-11-07
Reimplement text editor

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/289

Previous implementation relied on custom instance nodes which broke a lot of builtin behaviour.

With new editor I use only builtin nodes. To preserve styling when entering editing mode className is computed and assigned to lexical element. Instances are stored in map and reused by updates converted and className renderer.

Toolbar now shows "checked" state and allows to toggle off styling.

### DIFF
--- a/apps/designer/app/canvas/features/text-editor/index.ts
+++ b/apps/designer/app/canvas/features/text-editor/index.ts
@@ -1,0 +1,3 @@
+export { TextEditor } from "./text-editor";
+// hack for React.lazy
+export { TextEditor as default } from "./text-editor";
--- a/apps/designer/app/canvas/features/text-editor/index.ts
+++ b/apps/designer/app/canvas/features/text-editor/index.ts
@@ -1,3 +1,3 @@
 export { TextEditor } from "./text-editor";
-// hack for React.lazy
+// React.lazy requires default export
 export { TextEditor as default } from "./text-editor";
--- a/apps/designer/app/canvas/features/text-editor/interop.ts
+++ b/apps/designer/app/canvas/features/text-editor/interop.ts
@@ -1,0 +1,100 @@
+import {
+  type ElementNode,
+  $getRoot,
+  $createTextNode,
+  $createParagraphNode,
+  TextNode,
+  LineBreakNode,
+  ParagraphNode,
+} from "lexical";
+import { $createLinkNode, LinkNode } from "@lexical/link";
+import type { ChildrenUpdates, Instance } from "@webstudio-is/react-sdk";
+import { createInstanceId } from "~/shared/tree-utils";
+
+// Map<nodeKey, Instance>
+export type Refs = Map<string, Instance>;
+
+const $writeUpdates = (
+  node: ElementNode,
+  updates: ChildrenUpdates,
+  refs: Refs
+) => {
+  const children = node.getChildren();
+  for (const child of children) {
+    if (child instanceof ParagraphNode) {
+      if (updates.length !== 0) {
+        updates.push("\n");
+      }
+      $writeUpdates(child, updates, refs);
+    }
+    if (child instanceof LineBreakNode) {
+      if (updates.length !== 0) {
+        // @todo should we visually distinct line breaks and paragraphs?
+        updates.push("\n");
+      }
+    }
+    if (child instanceof LinkNode) {
+      // @todo support link children styling
+      const text = child.getTextContent();
+      const id = refs.get(child.getKey())?.id ?? createInstanceId();
+      updates.push({ id, component: "Link", text });
+    }
+    if (child instanceof TextNode) {
+      // @todo support nesting bold into italic and vice versa
+      const text = child.getTextContent();
+      if (child.hasFormat("bold")) {
+        const id = refs.get(child.getKey())?.id ?? createInstanceId();
+        updates.push({ id, component: "Bold", text });
+      } else if (child.hasFormat("italic")) {
+        const id = refs.get(child.getKey())?.id ?? createInstanceId();
+        updates.push({ id, component: "Italic", text });
+      } else {
+        updates.push(text);
+      }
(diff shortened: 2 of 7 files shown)


rpominov (approved):

apps/designer/app/canvas/features/text-editor/index.ts
```diff
@@ -0,0 +1,3 @@
+export { TextEditor } from "./text-editor";
+// hack for React.lazy
```
istarkov, line 2: What the hack?

apps/designer/app/canvas/features/text-editor/toolbar-connector.tsx
```diff
@@ -0,0 +1,113 @@
...
+const isSelectedLink = (selection: RangeSelection) => {
+  const node = getSelectedNode(selection);
+  const parent = node.getParent();
+  return $isLinkNode(parent) || $isLinkNode(node);
+};
+
+export const ToolbarConnectorPlugin = () => {
+  const [editor] = useLexicalComposerContext();
+
+  // control toolbr state on data or selection updates
```
istarkov, line 40: typo toolbr

istarkov (approved):
Magic ;-)

apps/designer/app/canvas/features/text-editor/index.ts
kof: Please add a more specific comment, I have no idea :)

apps/designer/app/canvas/features/text-editor/interop.ts
```diff
@@ -0,0 +1,100 @@
...
+  refs: Refs
+) => {
+  const children = node.getChildren();
+  for (const child of children) {
+    if (child instanceof ParagraphNode) {
+      if (updates.length !== 0) {
+        updates.push("\n");
+      }
+      $writeUpdates(child, updates, refs);
+    }
```
kof, line 29: this would make it clear that we aren't going to do anything later, so we can stop reading
```suggestion
      continue;
    }
```

apps/designer/app/canvas/features/text-editor/interop.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (child instanceof LineBreakNode) {
+      if (updates.length !== 0) {
+        // @todo should we visually distinct line breaks and paragraphs?
```
kof, line 32: can you explain what the question is? I it the difference between creating new paragraphs when hitting enter? vs using the same paragraph with newlines inside?

apps/designer/app/canvas/features/text-editor/interop.ts
```diff
@@ -0,0 +1,100 @@
...
+        updates.push("\n");
+      }
+    }
+    if (child instanceof LinkNode) {
+      // @todo support link children styling
```
kof, line 37: Basically we can't make children of a link text bold right now, right?

apps/designer/app/canvas/features/text-editor/interop.ts
```diff
@@ -0,0 +1,100 @@
...
+      const text = child.getTextContent();
+      const id = refs.get(child.getKey())?.id ?? createInstanceId();
+      updates.push({ id, component: "Link", text });
+    }
+    if (child instanceof TextNode) {
+      // @todo support nesting bold into italic and vice versa
```
kof, line 43: oh I thought part of the rewrite was to support these

apps/designer/app/canvas/features/text-editor/interop.ts
```diff
@@ -0,0 +1,100 @@
...
+  ParagraphNode,
+} from "lexical";
+import { $createLinkNode, LinkNode } from "@lexical/link";
+import type { ChildrenUpdates, Instance } from "@webstudio-is/react-sdk";
+import { createInstanceId } from "~/shared/tree-utils";
+
+// Map<nodeKey, Instance>
+export type Refs = Map<string, Instance>;
+
+const $writeUpdates = (
```
kof, line 17: I guess no that its a simple function, we could add a few tests (in case its simple enough to do, otherwise we have to postpone this for later)

apps/designer/app/shared/tree-utils/create-instance.ts
```diff
@@ -5,6 +5,10 @@ import {
   components,
 } from "@webstudio-is/react-sdk";
 
+export const createInstanceId = () => {
```
kof, line 8: Why? to make sure instance id's are created the same way everywhere?

Then we need to update all places that do that as well: there is a couple of them if you search for ObjectId()

apps/designer/app/canvas/features/text-editor/interop.ts
TrySound: If use default browser styling paragraphs have not editable space (margin) between them, while line break is just a newline. Lexical allows to enter line break with shift+enter.

apps/designer/app/canvas/features/text-editor/interop.ts
TrySound: Not yet, same in current implementation. Need to change updaters structure. Will do in separate PR. Now just basic editing support.

apps/designer/app/canvas/features/text-editor/interop.ts
TrySound: It will be next PR. For now at least editing and deleting of these fragments works.

apps/designer/app/canvas/features/text-editor/interop.ts
TrySound: We will change this a lot when support nesting and other features.

kof (approved):
approving, because these changes are great, there is a bunch of things to add/improve, especially tests, but its up to you if you want to split PRs

PR #423, opened 2022-11-08
Allow text editor styles nesting

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/289

Added support for styles inside links and styles inside each other. Updates can have children updates and ids in updates become optional.

For now code is repetitive in some cases. Will be more general eventually.

Though still we getting some complexity because lexical represents italic and bold as single node when intersected.

image


TrySound:
Btw interesting point
https://github.com/facebook/lexical/issues/2418#issuecomment-1171020000

packages/react-sdk/src/tree/create-elements-tree.tsx
```diff
@@ -8,7 +8,11 @@ import { SessionStoragePolyfill } from "./session-storage-polyfill";
 
 export type ChildrenUpdates = Array<
   | string
-  | { id: Instance["id"]; component: Instance["component"]; text: string }
+  | {
+      id: null | Instance["id"];
```
kof, line 12: if its optional, lets make it optional/undefined, not null. I am trying to fight null whenever I can, only using it where I can't avoid it

kof (approved):
At this int I don't understand half of that code but I bet you do haha

packages/react-sdk/src/tree/create-elements-tree.tsx
istarkov: null is not that bad, instead of `undefined`, it shows that the developer explicitly set it as null, or the developer must explicitly set it as null.

`undefined` or optional don't give any information - is it mistake or what

kof: Usually if its not set, its not set. You shouldn't care why. Unless you can explain how this information is relevant here, lets use undefined. Generally undefined works with default arguments etc. and nicer to type

kof: I meant if its undefined. Unset in objects is a different thing, property still exists here.

kof: This is btw mainly how I differentiate if it was set to undefined vs it wasn't set at all. `property in object`

istarkov: https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes must be set if not already

kof: I probably do explicit delete/unset semantics when I need something to be gone, on the level of an action/event, not on a level of a value that has been set to null

kof: in this case I still don't understand if we need to know that something was "deleted"

kof: as opposed to being simply optional

Issue #400, opened 2022-11-06
Support all remaining toolbar options in text editing WF has


kof:
<img width="337" alt="Screenshot 2022-11-06 at 16 50 00" src="https://user-images.githubusercontent.com/52824/200180713-72fc52ac-4c36-46ca-b3e7-c3ca12251a45.png">

PR #426, opened 2022-11-08
Allow span fragment in text editor

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/400

- added span component to sdk
- added span support to text editor toolbar

Lexical has a great utility $patchStyleText which does exaactly what we need, split selection into stylable element. Though there is a catch. It requires to apply some styles. We do styling lazily and with clases so it doesn't fit to trigger split. To workaroud it I instead patch with css custom property which does not affect styling and considered a trigger for lexical.

image

## Before requesting a review

- [x] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/components/span.ws.tsx
+++ b/packages/react-sdk/src/components/span.ws.tsx
@@ -1,0 +1,15 @@
+import { BrushIcon } from "@webstudio-is/icons";
+import type { WsComponentMeta } from "./component-type";
+import { Span } from "./span";
+
+const meta: WsComponentMeta<typeof Span> = {
+  Icon: BrushIcon,
+  Component: Span,
+  canAcceptChildren: false,
+  isContentEditable: false,
+  label: "Span Text",
+  isInlineOnly: true,
+  isListed: false,
+};
+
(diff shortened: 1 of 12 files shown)


packages/react-sdk/src/components/span.ws.tsx
```diff
@@ -0,0 +1,15 @@
+import { BrushIcon } from "@webstudio-is/icons";
+import type { WsComponentMeta } from "./component-type";
+import { Span } from "./span";
+
+const meta: WsComponentMeta<typeof Span> = {
+  Icon: BrushIcon,
+  Component: Span,
+  canAcceptChildren: false,
+  isContentEditable: false,
+  label: "Span Text",
```
kof, line 10: Label is for semantic meaning of that component, Bold Text and Italic Text make sense, this case I would call it "Styled Text" I guess

kof (approved):
one comment about label

rpominov (approved):

PR #432, opened 2022-11-09
Switch on initialVisibleProps

istarkov:
Draft Preview to show the idea.

ref #192

The idea is that instead of multiple sources of properties must be one.
That simplifies all things and removes questions like 
- props ordering
- bugs in `react-docgen-typescript` generation (it doesn't allow to make non-optional properties required etc)
- moves all props we need to show initially into the single place
- no clear way to define default props for functional components exists, `defaultProps` should gone some day for functional components, ({a='default', b, v}) 'default' is not exists in type information.

It's defined at `${component}.ws.ts` file and called  `initialVisibleProps`

Type on `initialVisibleProps` forces you to put into ALL required properties of the component with default values.
Also, it forces default values to be the right type too.
image

In the case of inexisted properties:
image

I think we can make properties like `tag` as required, no reason to make it optional IMO.
Also in case of structure like `key => value` is not enough its possible to force with types to move `required` etc props into `initialVisibleProps` to have that parameters explicit at least for internal components.

### DIFF
--- a/packages/react-sdk/src/components/box.ws.ts
+++ b/packages/react-sdk/src/components/box.ws.ts
@@ -18,6 +18,7 @@ const meta: WsComponentMeta<typeof Box> = {
   isInlineOnly: false,
   isListed: true,
   label: "Box",
+  initialVisibleProps: { tag: "div" },
 };
 
 export default meta;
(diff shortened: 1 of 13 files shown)


packages/react-sdk/src/components/box.ws.ts
```diff
@@ -18,6 +18,7 @@ const meta: WsComponentMeta<typeof Box> = {
   isInlineOnly: false,
   isListed: true,
   label: "Box",
+  initialVisibleProps: { tag: "div" },
```
kof, line 21: still thinking about a better name, maybe `showProps` or `initialProps`?

kof: The initial value should be part of the component defaults. Here we are just setting the flag to show the prop control

kof: could be: `initialProps: ['tag']`

PR #435, opened 2022-11-09
Save db state after docker compose down

TrySound:
With previous docker setup I have a problem when switched between projects had to setup migrations again and loose current session.

Docker compose volumes solve this problem and preserve current db state. Can be destroyed with `docker compose down -v`.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/package.json
+++ b/apps/designer/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "description": "Webstudio Designer UI as a service",
   "scripts": {
+    "bootstrap": "docker compose up -d && migrations migrate --dev --force",
     "build": "remix build",
     "dev": "remix dev",
     "start": "remix-serve build",
--- a/apps/designer/package.json
+++ b/apps/designer/package.json
@@ -3,7 +3,6 @@
   "version": "0.1.0",
   "description": "Webstudio Designer UI as a service",
   "scripts": {
-    "bootstrap": "docker compose up -d && migrations migrate --dev --force",
     "build": "remix build",
     "dev": "remix dev",
     "start": "remix-serve build",
--- a/apps/designer/docker-compose.yaml
+++ b/apps/designer/docker-compose.yaml
@@ -1,4 +1,7 @@
-version: '3.1'
+version: "3.1"
+
+volumes:
+  db:
 
 services:
   db:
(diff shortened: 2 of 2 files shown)


apps/designer/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "description": "Webstudio Designer UI as a service",
   "scripts": {
+    "bootstrap": "docker compose up -d && migrations migrate --dev --force",
```
istarkov, line 6: Document here? https://github.com/webstudio-is/webstudio/blob/main/docs/contributing.md

istarkov (approved):

apps/designer/package.json
kof: now that's getting interesting. I was against adding docker to the core, but was ok if its a simple separate file, now its leaking into package.json.

How can we make the whole docker thing fully separated?

kof: Can you use a bin/docker.sh or something?

kof: or move the entire thing compose and scripts for docker and whatever else is going to be needed into a separate repository

kof: or move everything into a docker folder both script shortcut and the compose file to avoid docker spreading

kof: maybe even docker root folder and there docker/designer/...files - that would be even better

TrySound: Are you not a fan of docker?

kof: yep not a fan at all, had to fight it a lot with playwright and jenkins ... its always slow on mac ... too much indirection with file system and what not, lets not start a discussion here, it will be a long one, short story is docker is not the future of development, its just a shorter way to start a external applications like a database ... I will never develop everything inside it and since its just one postgress database here, I don't need it at all personally

I see the benefit of having it for postgress for people who don't want to install it manually

kof (changes requested):
please lets figure out how to contain docker stuff in one folder or something

apps/designer/package.json
kof: Also maybe we will need some other software that we want to make easy to setup

TrySound:
Alright, let's remove script but at least add volumes

kof (approved):

kof:
> Alright, let's remove script but at least add volumes

I think long term you will need scripts and what not, so probably better to move the docker stuff to a folder so it can live there happily  in isolation

rpominov (approved):

PR #439, opened 2022-11-10
Migrate fontSize & lineHeight and borderRadius tokens

kof:
In figma we currently only use 3 fontSizes (I added the rest just to avoid breaking current UI). 

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/design-system/src/components/icon-button-with-menu.tsx
```diff
@@ -17,7 +17,7 @@ const StyledContent = styled(DropdownMenuPrimitive.Content, {
 
 const itemStyles = {
   all: "unset",
-  fontSize: "$2",
```
rpominov, line 20: Strange. Here 2 replaced with 3 ...

packages/design-system/src/components/tabs.tsx
```diff
@@ -29,7 +29,7 @@ export const TabsTrigger = styled(TabsPrimitive.Trigger, {
   },
   [`& > ${Text}`]: {
     color: "$gray9",
-    fontSize: "$fontSizes$1",
+    fontSize: "$fontSizes$fontSize$3",
```
rpominov, line 32: Is this ok?

packages/design-system/src/components/icon-button-with-menu.tsx
kof: that's correct, $2 in previous scale is 13px, in new one we don't have it so it's 12px which is $3

rpominov (approved):

PR #443, opened 2022-11-10
Setup svgo-jsx for icons package

TrySound:
Ref https://github.com/svg/svgo-components/tree/main/packages/svgo-jsx

Icons can be generated from .svg sources to react components. svgo-jsx allows to do it easily.

Template reproduce existing components. Took 3 icons from material icons though they do not have viewBox so added removeDimensions svgo plugin to add it before predefined size (16) is added.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/icons/package.json
```diff
@@ -6,7 +6,7 @@
   "homepage": "https://webstudio.is",
   "scripts": {
     "dev": "tsup --watch",
-    "build": "rm -fr lib tsconfig.tsbuildinfo && tsc",
+    "build": "rm -fr lib tsconfig.tsbuildinfo src/generated && svgo-jsx && node ./svgo-jsx-index.js && tsc",
```
kof, line 9: this should be a separate build step, one we call only manually for now when we add a new svg icon

TrySound: So you want to commit generated icons?

packages/icons/svgo-jsx-index.js
```diff
@@ -0,0 +1,19 @@
+const fs = require("fs/promises");
```
kof, line 1: lets use ts file and run it like we run all bin scripts e.g. in SDK from /bin folder

packages/icons/svgo-jsx.config.js
```diff
@@ -0,0 +1,52 @@
+const path = require("path");
```
kof, line 1: same, can it use ts?

packages/icons/package.json
```diff
@@ -6,7 +6,7 @@
   "homepage": "https://webstudio.is",
```
kof, line 6: for file names convention is always dash separated (format_clear.svg)

packages/icons/svgo-jsx.config.js
TrySound: Yep, worked with tsx

packages/icons/bin/svgo-jsx-index.ts
```diff
@@ -0,0 +1,16 @@
+import * as fs from "fs/promises";
+import * as path from "path";
+
+(async () => {
+  try {
+    const list = await fs.readdir(path.join(process.cwd(), "./src/gen"));
```
kof, line 6: we don't really need a special folder I think :)

TrySound: it's useful when we delete svg and want to synchronize list of generated icons

kof: got it

kof: wait, then you can delete whatever svg was actually deleted, you know a ll svgs

kof: ah no, if its deleted you don't know if the other one exists and is not manual, yeah we need a generated folder

kof (approved):

packages/icons/svgo-jsx.config.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const config = {
+  inputDir: "./icons",
+  outputDir: "./src/gen",
+  template,
+  transformFilename,
+  svgProps: {
+    width: "16",
+    height: "16",
+    fill: "{color}",
```
rpominov, line 36: Something to keep in mind: in the current icons, color isn't always in `fill`, sometimes it's `stroke={color}`. In case we'll move existing icons to this or something...

rpominov (approved):

packages/icons/svgo-jsx.config.ts
TrySound: We'll figure it out

PR #444, opened 2022-11-11
[designer] Assets picker

istarkov:
ref #196

- Image reverted to ordinary Image
- Added supports of "ordinary" Image `src` selection with the Asset manager. (The Idea that we save asset + prop value simultaneously, so "ordinary" controls will work without issues, i.e. in case of the image we save `prop="src" value="/imagpath" asset={...tonof info here...}`)
- Image optimizations removed (will be created in standalone package in next PR)
- Now UserProps has type https://github.com/webstudio-is/webstudio-designer/blob/c36519540c9d3493acfe0dcdd1c118ac4b6196be/packages/react-sdk/src/user-props/schema.ts#L33-L39
- Db type is: https://github.com/webstudio-is/webstudio-designer/blob/6d5cc571cb4e3bf00ec95f166a0daf9420ef50ac/packages/react-sdk/src/user-props/schema.ts#L4-L10

Next PRs:
1. Restore the Image component with optimizations at standalone package
2. Pass the Image component to SDK and use it instead of the original in case of optimizations are needed.


apps/designer/app/canvas/features/wrapper-component/wrapper-component.tsx
```diff
@@ -64,7 +66,26 @@ export const WrapperComponentDev = ({
     [setSelectedElement]
   );
 
-  const userProps = useUserProps(instance.id);
+  const defaultTransformAsset = React.useCallback(
```
kof, line 69: lets import {useCallback} style

apps/designer/app/designer/features/props-panel/control.stories.tsx
```diff
@@ -7,22 +7,14 @@ export default {
 
 export const Primitive: ComponentStoryObj<typeof Control> = {
   args: {
-    type: "text",
+    component: "Image",
+    userProp: { prop: "src", id: "some_id", value: "" },
```
kof, line 11: we never use snake case, only in CONSTANTS_LIKE stuff, lets not even make an example lol to prevent this idea

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -142,84 +220,150 @@ const RangeControl = ({
     </Flex>
   </Flex>
 );
+*/
 
 const NotImplemented = () => <div />;
 
-type PrimitiveControlProps = BaseControlProps & {
-  type: "array" | "boolean" | "date" | "number" | "object" | "text";
+const includes = <T extends string>(arr: readonly T[], v: string): v is T => {
```
kof, line 227: I think we need this as a reusable utility somewhere ... is there a good package that already has it or can we put it somewhere else?

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -142,84 +220,150 @@ const RangeControl = ({
...
+  return arr.includes(v as never);
 };
 
-type ColorControlProps = BaseControlProps & {
-  type: "color";
-  presetColors?: Array<string>;
+const isStringArray = (arr: unknown): arr is Array<string> => {
```
kof, line 231: this could be also a good utility in the same belt

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -142,84 +220,150 @@ const RangeControl = ({
...
-    | "check"
-    | "inline-check"
-    | "select"
-    | "multi-select";
-  options: Array<string>;
-};
+  const defaultValue = argType?.control.defaultValue;
+  const type = argType?.control.type;
+
+  if (type == null) {
```
kof, line 255: can type be null and undefined? if yes then its correct

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -142,84 +220,150 @@ const RangeControl = ({
...
+    warnOnce(
+      type == null,
```
kof, line 257: that's just true
```suggestion
      true,
```

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -142,84 +220,150 @@ const RangeControl = ({
...
 
-export type ControlProps =
-  | PrimitiveControlProps
-  | OptionsControlProps
-  | FileControlProps
-  | ColorControlProps
-  | RangeControlProps;
+  if (typeof type !== "string") {
+    warnOnce(
+      typeof type !== "string",
```
kof, line 265:
```suggestion
      true,
```

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -142,84 +220,150 @@ const RangeControl = ({
...
+        value={value}
+        defaultValue={defaultValue}
+        onChange={onChangePropValue}
+      />
+    );
+  }
+
+  if (
+    includes(
+      [...radioControlTypes, ...checkboxControlTypes, ...selectControlTypes],
```
kof, line 309: this merge can be done once in the module scope

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -200,8 +210,14 @@ export const PropsPanel = ({
       <PlusIcon />
     </Button>
   );
+
   return (
-    <>
+    <Box
+      css={{
+        position:
+          "relative" /* Hack - value picker popover positioning depends on it */,
```
kof, line 218: now we gotta fix it once for both cases, I think we should have that container in a location for both, seems like the closest shared folder is designer/shared/inspector/pane.tsx

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -19,11 +19,16 @@ declare module "~/shared/pubsub" {
...
-  id: UserProp["id"],
-  field: "prop" | "value",
-  value: UserProp["prop"] | UserProp["value"]
-) => void;
+// First infer is used to iterate over Union type
+type UserPropValue = UserProp extends infer IterUserProp
+  ? IterUserProp extends { value: infer Value }
+    ? Value
+    : never
+  : never;
```
kof, line 27: almost too complex type for my test, please check if it can be simplified some how

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -19,11 +19,16 @@ declare module "~/shared/pubsub" {
...
+
+type handleChangePropName = (id: UserProp["id"], name: string) => void;
```
kof, line 29: types are allways capitalized

packages/asset-uploader/src/env.ts
```diff
@@ -0,0 +1,3 @@
+export const env = {
+  VERCEL_ENV: process.env.VERCEL_ENV,
```
kof, line 2: can we please abstract away VERCEL_ENV and use something that will not change no matter if we use vercel or something else?

packages/asset-uploader/src/utils/format-asset.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  projectId: "id",
+  size: 2135,
+  format: "png",
+  createdAt: new Date(),
+  description: "",
+  meta: JSON.stringify({ width: 200, height: 200 }),
+};
+
+describe("getAssetPath", () => {
+  const OLD_ENV = process.env;
```
kof, line 18: `OLD_ENV` where is this coming from?

kof, line 18: I got confused for a second, we use SCREAMING_CASE only when we need to export constants like a config, not when its a local constant

packages/asset-uploader/src/utils/format-asset.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    };
+  });
+
+  afterAll(() => {
+    process.env = OLD_ENV;
+  });
+
+  describe("local", () => {
+    test("returns none on Location.FS if VERCEL_ENV not set", () => {
+      process.env.VERCEL_ENV = undefined;
```
kof, line 33: do you like the mocking more than passing env as a second argument that defaults to process.env? It's clean and doesn't depend on jest magic

packages/asset-uploader/src/utils/format-asset.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      const { formatAsset } = require("./format-asset");
```
kof, line 35: the above suggestion would also prevent the need for require

packages/asset-uploader/src/utils/format-asset.ts
```diff
@@ -24,8 +25,19 @@ export const formatAsset = (asset: DbAsset): Asset => {
     };
   }
 
+  let transform: ImageAsset["transform"] = "none";
+
+  if (asset.location === "REMOTE") {
+    transform = "relative";
+
+    if (env.VERCEL_ENV != null) {
```
kof, line 33: same here, how can we have a flag that is independent of the specific environment name like "vercel", "cloudflare" or whatever

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,75 @@
...
 import { prisma } from "@webstudio-is/prisma-client";
+import { formatAsset } from "@webstudio-is/asset-uploader/server";
 
 export const loadByTreeId = async (treeId: Tree["id"]) => {
   const instancePropsEntries = await prisma.instanceProps.findMany({
     where: { treeId },
   });
 
+  // Fin all asset_id in all props
+  const assets_id: string[] = [];
```
kof, line 19: oh nooo, now snake landed in code and I had to create an issue for another eslint rule https://github.com/webstudio-is/webstudio-designer/issues/445 which is now yours 🙃

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,75 @@
...
+
+  for (const instanceProps of instancePropsEntries) {
+    const props = JSON.parse(instanceProps.props);
+    const dpProps = zUserDbProps.parse(props);
+
+    for (const dbProp of dpProps) {
+      if ("asset_id" in dbProp) {
```
kof, line 26: same

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,75 @@
...
   const instancePropsEntries = await prisma.instanceProps.findMany({
     where: { treeId },
   });
 
+  // Fin all asset_id in all props
+  const assets_id: string[] = [];
+
+  for (const instanceProps of instancePropsEntries) {
+    const props = JSON.parse(instanceProps.props);
+    const dpProps = zUserDbProps.parse(props);
```
kof, line 23: zod schema always starts capital letter, just like its type

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,75 @@
...
+          resProps.push({
+            id: dbProp.id,
+            prop: dbProp.prop,
+            asset,
+            required: dbProp.required,
+          });
+        } else {
+          // Can be ok if asset was deleted having that no db constraint checks exists
+          // TODO: remove throw on something else
+          throw new Error(`asset with ${dbProp.asset_id} not found`);
```
kof, line 63: just fui for consistent error style
```suggestion
          throw new Error(`Asset with assetId "${dbProp.asset_id}" not found`);
```

packages/react-sdk/src/component-utils/image-loaders.ts
```diff
@@ -0,0 +1,48 @@
...
+ * https://developers.cloudflare.com/images/image-resizing/url-format/
+ **/
+export const cloudflareRelativeImageLoader: ImageLoader = ({
+  width,
+  src,
+  quality,
+}) => {
+  if (process.env.NODE_ENV !== "production") {
+    warnOnce(
+      !allSizes.includes(width),
```
kof, line 15: easier to scan, please trust me on this

```suggestion
      allSizes.includes(width) === false,
```

packages/react-sdk/src/component-utils/image-loaders.ts
```diff
@@ -0,0 +1,48 @@
+import warnOnce from "warn-once";
+import { allSizes, type ImageLoader } from "./image";
+
+/**
+ * Default image loader in case of no loader provided
+ * https://developers.cloudflare.com/images/image-resizing/url-format/
+ **/
+export const cloudflareRelativeImageLoader: ImageLoader = ({
```
kof, line 8: just realized that normal open source package user will not have cloudflare installed, so they don't need this cloudflare logic to be there at all, what would be a better place for CF specific stuff?

packages/react-sdk/src/component-utils/image-loaders.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/**
+ * Same as cloudflareRelativeImageLoader just uses full URL for tranform path
+ */
+export const cloudflareRemoteImageLoader: ImageLoader = (props) => {
+  // Cloudflare docs say that we don't need to urlencode the path params
+  const pathname = cloudflareRelativeImageLoader(props);
+
+  const resizeOrigin = "https://webstudio.is/";
```
kof, line 33: can't be hardcoded

packages/react-sdk/src/components/component-type.ts
```diff
@@ -2,6 +2,7 @@ import type { Style } from "../css";
 import { z } from "zod";
 import React from "react";
 import { IconProps } from "@webstudio-is/icons";
+import { TransformAsset } from "~/user-props";
```
kof, line 5: use type
```suggestion
import type { TransformAsset } from "~/user-props";
```

packages/react-sdk/src/components/component-type.ts
```diff
@@ -31,6 +33,7 @@ export const WsComponentMeta = z.lazy(() =>
       isListed: z.boolean(),
       label: z.string(),
       children: z.optional(z.array(z.string())),
+      transformAsset: z.optional(z.any()),
```
kof, line 36: any???

packages/react-sdk/src/components/image.ws.tsx
```diff
@@ -17,6 +23,26 @@ const defaultStyle = {
...
+
+const ensureBoolean = (value: unknown): boolean | undefined => {
+  if (typeof value === "boolean") return value;
+  return undefined;
+};
+
+const assertUnreachable = (_arg: never) => {
+  throw new Error("Didn't expect to get here");
+};
+
```
kof, line 45: looks like a set of useful generic utils like in some other case before

packages/react-sdk/src/components/image.ws.tsx
```diff
@@ -26,6 +52,44 @@ const meta: WsComponentMeta<typeof Image> = {
   isInlineOnly: false,
   isListed: true,
   label: "Image",
+  transformAsset: (assetProp, props) => {
```
kof, line 55: now that makes meta non-serializable, not sure this is not a problem

packages/react-sdk/src/components/image.ws.tsx
```diff
@@ -26,6 +52,44 @@ const meta: WsComponentMeta<typeof Image> = {
...
+    }
+
+    const { asset } = assetProp;
+
+    let loader = localImageLoader;
+
+    if ("transform" in asset) {
+      switch (asset.transform) {
+        case "relative":
+          loader = cloudflareRelativeImageLoader;
```
kof, line 67: again, image has no idea of cloudflare

kof: I think ultimately this entire transform belongs somewhere else

kof (changes requested):
good initial draft, lets clean it up

packages/react-sdk/src/components/image.ws.tsx
kof: discussed: we want a universal transformers interface that can be sticked together like middleware. This one is not universal right now

apps/designer/app/designer/features/props-panel/control.tsx
istarkov: Disagree, people even don't reuse npm packages already exists in big repos, and reusing a one-liner like this is nearly impossible. I think I wrote string array type guard for the first time in my life, and the same for includes. They are rare, the main reason is to be type guards, and not every piece of code should has own module or be in some utils. Even so - reusability is near impossible.

apps/designer/app/designer/features/props-panel/control.tsx
istarkov: it can, its even `any` as control is not defined on argtype

apps/designer/app/designer/features/props-panel/control.tsx
kof: yeah maybe this one is a bit too special

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -117,109 +160,186 @@ const BooleanControl = ({
...
+      `Control type for prop: ${
+        userProp.prop
+      } component: ${component} is not a string but ${typeof type}`
+    );
+
+    return <NotImplemented />;
+  }
+
+  if (isFeatureEnabled("assets")) {
+    if (component === "Image" && userProp.prop === "src") {
```
rpominov, line 254: Why we check `component` and `userProp.prop` directly rather than going via `const type = argType?.control.type` like for all other control types? Is this temporary?

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -117,109 +160,186 @@ const BooleanControl = ({
...
+  return Array.isArray(arr) && arr.every((v) => typeof v === "string");
 };
 
-type RangeControlProps = BaseControlProps<number> & {
-  type: "range";
-  min: number;
-  max: number;
-  step: number;
+const assertUnreachable = (_arg: never) => {
+  throw new Error("Didn't expect to get here");
```
rpominov, line 205: Would be good to include `_arg` in the message. Just in case :)

packages/asset-uploader/src/utils/get-asset-path.ts
```diff
@@ -8,18 +8,28 @@ const fsEnv = FsEnv.parse(process.env);
...
+    );
+
+    return locUrl.pathname;
   }
 
   if (asset.location === Location.REMOTE && s3Envs.success) {
     if (s3Envs.data.ASSET_CDN_URL) {
       const cndUrl = new URL(s3Envs.data.ASSET_CDN_URL);
-      cndUrl.pathname = encodeURIComponent(asset.name);
+      cndUrl.pathname = encodeURIComponent(encodeURIComponent(asset.name));
```
rpominov, line 27: encodeURIComponent twice? Same below.

packages/react-sdk/src/prop-transforms/image.ts
```diff
@@ -0,0 +1,87 @@
...
+        quality: ensureStringOrNumber(imageValueProps.quality),
+        loader,
+        // User can overwrite optimize
+        optimize: ensureBoolean(imageValueProps.optimize) ?? true,
+      });
+
+      if (imageAttributes != null) {
+        for (const [key, value] of Object.entries(imageAttributes)) {
+          res.push({
+            id: `imgt_${gId++}`,
```
rpominov, line 77: What `imgt` means and what `g` means in `gId`? 😅

apps/designer/app/canvas/canvas.tsx
```diff
@@ -129,6 +130,13 @@ export const Canvas = ({ data }: CanvasProps): JSX.Element | null => {
   const isPreviewMode = useSubscribePreviewMode();
   const elements = useElementsTree();
 
+  propTransforms.initComponentPropsTransformer(
+    "Image",
+    propTransforms.Image.createPropsTransformer(
+      data.params?.propTransforms?.Image ?? null
+    )
+  );
```
kof, line 138: should be more generic, for any component

apps/designer/app/routes/$.tsx
```diff
@@ -76,7 +76,16 @@ export const loader: LoaderFunction = async ({
       throw json("Page not found", { status: 404 });
     }
 
-    return { ...canvasData, env, mode };
+    const params = { ...canvasData.params };
```
kof, line 79: why do we need `params` and not canvasData.propTransform directly?

apps/designer/app/routes/$.tsx
```diff
@@ -76,7 +76,16 @@ export const loader: LoaderFunction = async ({
...
+
+    if (env.RESIZE_ORIGIN != null) {
+      params.propTransforms = params.propTransforms ?? {};
+      params.propTransforms.Image = {
+        resizeOrigin: env.RESIZE_ORIGIN,
+      };
+    }
```
kof, line 86: these are technically not props, these are transformer configs, we configure a middleware, then the middleware will receive component props and know what to do with it.

kof: I think this datastructure makes more sense to me, wdyt?

```
canvasData.propTransforms = [
  {
    component: 'Image',
    config:  [resizeOrigin: env.RESIZE_ORIGIN,}
  }
]
```

kof: but looking at his all makes me wonder if we could do what you suggested initial - flat config for all transformers, transformer takes what it needs. In the worst case we can make this here work later, right?

packages/asset-uploader/src/utils/get-asset-path.ts
istarkov: Yep otherwise spaces, cyrillic etc is not working. 
On upload we do following https://github.com/webstudio-is/webstudio-designer/blob/6488305087e661923415971a36d491727e336072/packages/asset-uploader/src/targets/s3/upload.ts#L85 (otherwise s3 will respond with error I think)

Then we need to encode filename into a path, but filename is already `filename = encodeURIComponent(asset.name)` see link above
Having that here we need `encodeUriComponent(filename)` we got 2nd encode ;-)

packages/react-sdk/src/prop-transforms/image.ts
istarkov: Just need some id non intersected with others. 
g means global ;-) If we have some convention about mod level state vars tell me. 
Usualy I use underscore after var, i.e. `id_`

apps/designer/app/designer/features/props-panel/control.tsx
istarkov: Its typescript hack, see `_arg: never` so any call to this function with any non `never` argument would cause ts error. `throw new Error` just instead of comment. IMO no need to check typescript in runtime adding any logic on never arg or output it

apps/designer/app/designer/features/props-panel/control.tsx
istarkov: Because we want this only for selected component `Image`, for known prop name 'src' only.

`type` there is coming from storybook and is `string` or like, there are no type `asset` or like on our components.

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -117,109 +160,186 @@ const BooleanControl = ({
...
+    );
+    return <NotImplemented />;
+  }
+
+  if (typeof type !== "string") {
+    warnOnce(
+      true,
+      `Control type for prop: ${
+        userProp.prop
+      } component: ${component} is not a string but ${typeof type}`
```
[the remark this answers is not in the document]
istarkov: `Control type for prop:` blabla `is not a string` will fix

apps/designer/app/designer/features/props-panel/control.tsx
rpominov: I mean what if a user reports a bug saying they're getting a error "Didn't expect to get here"? It would be better for us if user comes with something like "Didn't expect to get here. Got: something".

packages/asset-uploader/src/utils/get-asset-path.ts
rpominov: Ok, just wanted to make sure this is not a mistake

packages/react-sdk/src/prop-transforms/image.ts
rpominov: Can we do something like `image_transform_${globalId}` so it would be less confusing? (if I guessed correctly what `imgt` means)

packages/asset-uploader/src/utils/get-asset-path.ts
istarkov: Ill add my comments into a code as its really confusing and probably need some investigations, like storing fileName on asset

packages/react-sdk/src/prop-transforms/image.ts
istarkov: Done

apps/designer/app/designer/features/props-panel/control.tsx
istarkov: Moved msg out of scope. Yep found a case that it can be executed in runtime - for example, if a developer will forget to set `return` in the code below, it would be no ts error.

apps/designer/app/designer/features/props-panel/control.tsx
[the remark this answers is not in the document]
istarkov: fixed

apps/designer/app/routes/$.tsx
istarkov: Ok transforms now are in a standalone folder, and just transforms https://github.com/webstudio-is/webstudio-designer/blob/8539fbf2f41d4763ff308da662106ed843164c88/packages/react-sdk/src/prop-transforms/index.ts#L1

In the future, it can be an argument to sdk components.

All logic to create transforms etc are  moved here 

https://github.com/webstudio-is/webstudio-designer/blob/32b33b2d176f5687d8c2a28b42663dc5b5cdcee9/packages/react-sdk/src/user-props/prop-transforms.ts#L27

And now more generic.

https://github.com/webstudio-is/webstudio-designer/blob/32b33b2d176f5687d8c2a28b42663dc5b5cdcee9/packages/react-sdk/src/tree/root.ts#L36

packages/react-sdk/src/user-props/prop-transforms.ts
```diff
@@ -0,0 +1,64 @@
+import warnOnce from "warn-once";
+import type * as components from "../components";
+import type { TransformProps } from "~/user-props";
+import type * as propTransforms from "~/prop-transforms";
+
+// Bad rule, in modern ts {} is good type see https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#improved-intersection-reduction-union-compatibility-and-narrowing
+// eslint-disable-next-line @typescript-eslint/ban-types
```
rpominov, line 7: I had a case where this rule was in the way too, and `{}` worked perfectly

rpominov (approved):

packages/react-sdk/src/components/meta.ts
```diff
@@ -32,14 +32,32 @@ const meta = {
...
-type MetaProps = Record<
-  string,
-  {
-    type: string;
-    required: boolean;
-    defaultValue: null | string;
-    options?: string[];
-  }
->;
+type MetaProp =
```
istarkov, line 35: In case of we switch to `ts` instead of `json` generation.
I.e. `as const`, I would be able to generate this types from ts.

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -117,107 +145,170 @@ const BooleanControl = ({
...
-  presetColors?: Array<string>;
+const ImageControl = ({ asset, onChange }: ImageControlProps) => {
+  return (
+    <ValuePickerPopover
+      title="Images"
+      content={
+        <ImageManager
+          onChange={(asset) => {
+            onChange(asset);
+          }}
```
kof, line 161:
```suggestion
          onChange={onChange}
```

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -117,107 +145,170 @@ const BooleanControl = ({
...
+          onChange={onChangePropValue}
+          options={options ?? DEFAULT_OPTIONS}
+          type={type}
+        />
+      );
+    }
+
+    assertUnreachable(
+      type,
+      `Did't expect to be here, unknown control type ${type}`
```
kof, line 304: funny, but
```suggestion
      `Unknown control type ${type}`
```

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -117,107 +145,170 @@ const BooleanControl = ({
...
+    );
+  }
+
+  warnOnce(
+    true,
+    `Control type ${type} not implemented for prop: ${userProp.prop} component: ${component}`
```
kof, line 310:
```suggestion
    `Control type "${type}" is not implemented for prop: "${userProp.prop}" in component "${component}"`
```

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,73 @@
 import {
   type Tree,
   type AllUserProps,
-  UserProps,
+  type UserProp,
+  // TODO: move zod schema into project package
```
kof, line 5: forgotten? or wanted to do with a separate PR? 
inn any case we use the `@todo` style comment

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,73 @@
...
   const instancePropsEntries = await prisma.instanceProps.findMany({
     where: { treeId },
   });
 
+  // Fin all assetId in all props
+  const assetIds: string[] = [];
+
+  for (const instanceProps of instancePropsEntries) {
+    const props = JSON.parse(instanceProps.props);
+    const dpProps = UserDbProps.parse(props);
```
kof, line 23: just call it parsedProps
```suggestion
    const parsedProps = UserDbProps.parse(props);
```

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,73 @@
...
+
   return instancePropsEntries.map((instanceProps) => {
     const props = JSON.parse(instanceProps.props);
-    UserProps.parse(props);
+    const resProps: UserProp[] = [];
+    // Can be changed onto `props as ` having that above is already checked, no optmizations now
+    const dpProps = UserDbProps.parse(props);
+
+    for (const dbProp of dpProps) {
+      if (dbProp.assetId != null) {
```
kof, line 51: can we please keep the nesting level low?
```suggestion
      if (dbProp.assetId == null) {
         resultProps.push(parsedProp)
         continue
      }
```

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,73 @@
...
+  const assetsMap = new Map(
+    assets.map((asset) => [asset.id, formatAsset(asset)])
+  );
+
   return instancePropsEntries.map((instanceProps) => {
     const props = JSON.parse(instanceProps.props);
-    UserProps.parse(props);
+    const resProps: UserProp[] = [];
+    // Can be changed onto `props as ` having that above is already checked, no optmizations now
+    const dpProps = UserDbProps.parse(props);
```
kof, line 48:
```suggestion
    const parsedProps = UserDbProps.parse(props);
```

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,73 @@
...
+
+    for (const dbProp of dpProps) {
+      if (dbProp.assetId != null) {
+        const asset = assetsMap.get(dbProp.assetId);
+        if (asset) {
+          resProps.push({
+            ...dbProp,
+            asset,
+          });
+        } else {
```
kof, line 58:
```suggestion
          continue;
        } 
```

packages/project/src/db/props.ts
kof: cleaned it up for you

```ts
return instancePropsEntries.map((instanceProps) => {
  const props = JSON.parse(instanceProps.props);
  // Can be changed onto `props as ` having that above is already checked, no optmizations now
  const parsedProps = UserDbProps.parse(props);
  const resultProps: UserProp[] = [];

  for (const parsedProp of parsedProps) {
    if (parsedProp.assetId == null) {
      resultProps.push(parsedProp);
      continue;
    }
    
    const asset = assetsMap.get(dbProp.assetId);
    if (asset) {
      resProps.push({
        ...dbProp,
        asset,
      });
      continue
    } 
    // Can be ok if asset was deleted having that no db constraint checks exists
    // @todo: remove throw on something else
    throw new Error(`Asset with assetId "${dbProp.assetId}" was not found`);
  }

  return {
    ...instanceProps,
    props: resultProps,
  };
});
```

packages/project/src/db/props.ts
```diff
@@ -1,22 +1,73 @@
...
+          // Can be ok if asset was deleted having that no db constraint checks exists
+          // TODO: remove throw on something else
+          throw new Error(`Asset with assetId "${dbProp.assetId}" not found`);
```
kof, line 61: Deleted image that is still being referenced from the props is a normal thing, we need to handle it gratefully without breaking

kof (changes requested):
almost there, hang on

packages/project/src/db/props.ts
istarkov: Separate

packages/project/src/db/props.ts
istarkov: IMO dbProps is perfectly fine as gives information that this props are somehow different from userProps.
Changed resultProps onto userProps to make it more visible

packages/project/src/db/props.ts
istarkov: db is better name, we have different UserProps representation on UI and DB side

kof (approved):

Issue #429, opened 2022-11-09
Optimistically update local state when editing or deleting pages


rpominov:
- [ ] When user edits page name it should update immediately in the page list
- [ ] When user deletes a page it should disappear immediately in the page list

kof:
@rpominov this was fixed, right?

rpominov:
No 😄

PR #447, opened 2022-11-14
Pages UI fixes

### LINKED ISSUE #431: Pages UX fixes
- [x] Add placeholders to fields "/home" for Path, "My awesome site" for Title etc.
- [x] When opening page settings the Name field should gain focus
- [x] When creating a new page, Name should be pre-filled with some value, and text in Name should be selected so it's easy to delete the pre-filled value
- [x] When creating a new page, Path and Title should be generated automatically from Name, unless user have edited them
- [x] The close button tooltip should be aligned normally (centred like other tooltips)
- [x] Better design for the "Create" button
- [x] Better settings panel animation ( #376 )
- [x] Home page should not have a path field

rpominov:
Fixes #431

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/designer/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -145,6 +156,36 @@ export const NewPageSettings = ({
   onSuccess?: (page: Page) => void;
   projectId: string;
 }) => {
+  const [pages] = usePages();
+
+  const nameToPath = (name: string) => {
```
kof, line 161: seems like a good case to move the function out of  component

rpominov:
Not in this PR:
 - "Create" button doesn't look like in Figma (especially hover state) — #450
 - Optimistic updates — #429
 
 Will do these in other PRs

kof:
Looking good, animation is too slow, let me tweek it a bit, will commit directly here if that's ok

kof:
1. we should not animate width/height, it can turn out jerky because it will have to recalculate layout all the time, mostly can use use transforms, opacity.
2. adding opacity makes it look smoother
3. When animating something on open, it has to go really fast right away, then slow down or whatever. Same timing, but if you start slow - it feels slow, it has to show immediate reaction

kof (approved):
Send you a PR for animation tweaks, feel free to merge in https://github.com/webstudio-is/webstudio-designer/pull/459

PR #456, opened 2022-11-17
Removing stitches from sdk and canvas entirely

kof:
Removing stitches from sdk and canvas entirely

- [x] stop using stitches for helper css
- [x] migrate font faces rendering on canvas to use the engine 
- [x] remove stitches from canvas logic
- [ ] move our CSS data types to the engine or to standalone package

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/shared/styles.ts
```diff
@@ -4,59 +4,95 @@ import { setInstanceStyleMutable } from "~/shared/tree-utils";
...
+
+  addGlobalRules(globalStylesCssEngine, { assets });
+
+  for (const style of helperStyles) {
+    globalStylesCssEngine.addPlaintextRule(style);
+  }
+
+  if (typeof document !== "undefined") {
+    globalStylesCssEngine.render();
+  }
```
rpominov, line 94: Seems odd that we do all these side effects during render rather than in useEffect. If there's a reason for this, I think we need a code comment.

rpominov (approved):

apps/designer/app/canvas/shared/styles.ts
kof: true I took a bad shortcut, because I need it to be rendered first, before all other components styles get rendered

kof: Another option is to render a component first and then call this hook inside it and use layout effect right away. So that it renders as early as possible

rpominov: Sounds like use-case for useLayoutEffect, yeah

rpominov (approved):

PR #459, opened 2022-11-17
Page settings animation

kof:
I tweaked the animation

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/designer/features/sidebar-left/panels/pages/pages.tsx
```diff
@@ -186,7 +187,16 @@ const PagesPanel = ({
   }
 
   return (
-    <>
+    <Box
+      css={{
```
kof, line 191: I don't quite like this, maybe you have a better idea: we need to slide out the panel, from behind the tree panel, so we need tree panel to take full hight, have background etc, so it can cover it up

PR #461, opened 2022-11-17
Move css-data to a package

kof:
Moving CSS data out of sdk package.

This also moves all tree utils back to project package

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/css-data/bin/mdn-data.ts
```diff
@@ -2,9 +2,9 @@ import { definitionSyntax, type DSNode } from "css-tree";
 import properties from "mdn-data/css/properties.json";
 import units from "mdn-data/css/units.json";
 import syntaxes from "mdn-data/css/syntaxes.json";
-import type { StyleValue } from "../src/css";
-import { popularityIndex } from "./popularity-index";
-
+//import type { StyleValue } from "../src/css";
+import { popularityIndex } from "../src/popularity-index";
+type StyleValue = any;
```
kof, line 7: will fix in the next PR, was moving types and discovered a rabbit whole of code duplication we got

rpominov (approved):
Looks good

apps/designer/app/canvas/canvas.tsx
```diff
@@ -55,7 +55,7 @@ const useElementsTree = () => {
       if (rootInstance === undefined) return;
 
       const { instanceId, updates } = change;
-      setInstanceChildrenMutable(instanceId, updates, rootInstance);
+      utils.tree.setInstanceChildrenMutable(instanceId, updates, rootInstance);
```
TrySound, line 58: This is not optimizable by bundler. We will get `a.tree.setInstanceChildrenMutable` across whole bundle

kof: yeah, that's expected, we moved it there long time ago, alternative would be to export it as treeUtils so we can import treeUtils and not utils, right?

kof: this PR is touching it because somebody brought back to the app files we moved to project quite some time ago

kof: we don't expect these project stuff to be tree shaken since it's all in the designer, but maybe we are wrong by ignoring it in general and setting a bad example

PR #462, opened 2022-11-18
Restrict some prefixes for page paths

### LINKED ISSUE #449: Don't allow user to use some paths for their pages (e.g. the path we load css from)
We're going to use a prefix `/s` for our system paths inside the build. For example `/s/css`. So we must not allow users to use `/s` or `/s/*` for their pages.

This is a separate issue from #448 (things like `/s/css` will be in the same namespace as user's pages in any case)

rpominov:
Fixes #449

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project/src/db/types.ts
+++ b/packages/project/src/db/types.ts
@@ -28,7 +28,7 @@ export const pathValidators = (
   baseValidator: ZodType<string>
 ): ZodType<string> =>
   baseValidator
-    .refine((path) => path !== "", "Can't be just empty")
+    .refine((path) => path !== "", "Can't be empty")
     .refine((path) => path !== "/", "Can't be just a /")
     .refine(
       (path) => path === "" || path.startsWith("/"),
--- a/packages/project/src/db/types.ts
+++ b/packages/project/src/db/types.ts
@@ -50,6 +50,7 @@ export const pathValidators = (
     )
     .refine(
       // Remix serves build artefacts like JS bundles from /build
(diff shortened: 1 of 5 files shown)


packages/project/src/db/types.ts
```diff
@@ -38,7 +42,22 @@ const Page = z.object({
     .refine(
       (path) => /^[-_a-z0-9\\/]*$/.test(path),
       "Only a-z, 0-9, -, _ and / are allowed"
-    ),
+    )
+    .refine(
+      // We use /s for our system stuff like /s/css or /s/uploads
+      (path) => path !== "/s" && path.startsWith("/s/") === false,
+      "/s prefix is reserved"
```
kof, line 49:
```suggestion
      "/s prefix is reserved for the system"
```

packages/project/src/db/types.ts
```diff
@@ -38,7 +42,22 @@ const Page = z.object({
...
+    )
+    .refine(
+      // Remix serves build artefacts like JS bundles from /build
+      // And we cannot customize it due to bug in Remix: https://github.com/remix-run/remix/issues/2933
+      (path) => path !== "/build" && path.startsWith("/build/") === false,
+      "/build prefix is reserved"
```
kof, line 55:
```suggestion
      "/build prefix is reserved for the system"
```

kof (approved):

PR #463, opened 2022-11-18
Cross page / project / browser window copy-paste

### LINKED ISSUE #453: Cross-page copy-paste
Turns out copy/paste on the same page is working, but not cross-page.

rpominov:
Fixes #453

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/designer/features/topbar/menu/menu.tsx
```diff
@@ -183,10 +185,12 @@ export const Menu = ({ publish }: MenuProps) => {
...
-                payload: { name: "paste" },
-              });
+              // @todo: need to read from clipboard here. is it possible?
+              //
+              // publish({
+              //   type: "shortcut",
+              //   payload: { name: "paste" },
+              // });
             }}
           >
```
rpominov, line 195: Might not be possible to keep copy / paste menu items. 

Webflow seem to support copy, but for paste it says:

Screenshot 2022-11-22 at 15 34 10

Didn't research it well yet.

kof: yeah, probably because it has to be a native copy function, so you can't trigger that copy event, you can only listen to it and add the data to it

rpominov: I've created a separate issue for this #499 . I think we can do this later. Maybe there will be a better support in browsers by then too.

rpominov:
So cool being able to copy from a project in preview to my local project 🙃

apps/designer/app/shared/copy-paste/serialize.ts
```diff
@@ -0,0 +1,25 @@
+import { Instance, UserProp } from "@webstudio-is/react-sdk";
+import { z } from "zod";
+
+const TYPE = "@webstudio/instance/v1" as const;
```
kof, line 4: lets call it v0.1

kof: v1 is usually referenced as a stable, as in 1.0.0

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      const props =
+        data.props && data.props.length > 0
+          ? utils.props.cloneUserProps(data.props)
+          : undefined;
+
+      onPaste(instance, props);
+    };
+
+    document.addEventListener("copy", handleCopy);
```
kof, line 121: would it make sense to add capture: true option? so that underlying components can't prevent it

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
```diff
@@ -0,0 +1,130 @@
...
+import { type InstanceCopyData, serialize, deserialize } from "./serialize";
+
+type Props = {
+  selectedInstanceData?: InstanceCopyData | undefined;
+  allowAnyTarget?: boolean;
+  onPaste?: (instance: Instance, props?: UserProp[]) => void;
+  onCut?: (instance: Instance) => void;
+};
+
+export const useInstanceCopyPaste = (props: Props): void => {
```
kof, line 13: btw if we want to some day render any framework on canvas, we need to start writing down the logic as much as possible outside of react stuff and then only wrap what's needed with a react hook, I think majority of this logic is universal, only useEffect/cleanup is react specific

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
rpominov: Do you mean custom components? Not sure, but I think if they want to prevent it for some reason we should probably allow. Maybe they want to implement their own logic

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
rpominov: Not sure it's a good idea to make code more complex by adding indirection just in case we need it in some far future

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
kof: any code that would listen to that event down the line, alright lets wait and see if we hit a problem

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
kof: its not that far I think, within the next 6month, but also, I feel like it can be nicely written, using good patterns and this would show us a way forward with this, otherwise everyone will think this way

rpominov: Extracted

rpominov:
@kof Is this good to merge now?

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  return true;
+};
+
+type Props = {
+  selectedInstanceData?: InstanceCopyData | undefined;
+  allowAnyTarget?: boolean;
+  onPaste?: (instance: Instance, props?: UserProp[]) => void;
+  onCut?: (instance: Instance) => void;
```
kof, line 42: these are not really optional, right?

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
```diff
@@ -0,0 +1,163 @@
...
+    setProps(props: Props) {
+      currentProps = props;
+    },
+  };
+};
+
+// Everything is extrcated from hook,
+// to make it easier to remove React from canvas if we need to
+export const useInstanceCopyPaste = (props: Props): void => {
+  const eventsHandler = useMemo(() => createEventsHandler(), []);
```
kof, line 153: seems you could even do
```suggestion
  const eventsHandler = useMemo(createEventsHandler, []);
```

kof:
Tested, works great, nice!

kof (approved):

apps/designer/app/shared/copy-paste/use-instance-copy-paste.ts
rpominov: They are, but no sure why I did that. Will fix.

packages/react-sdk/src/user-props/schema.ts
```diff
@@ -11,7 +11,7 @@ const UserPropDb = z.object({
 
 export const UserDbProps = z.array(UserPropDb);
 
-const UserUserProp = z
+export const UserUserProp = z
```
rpominov, line 14: This type has changed in `main` in two important ways since I made this PR. Can't use it as is now. Need to figure out how to update

rpominov:
@TrySound Could you help me with Jest here, please? It's above my head. 

It fails with: `Must use import to load ES Module`

I tried:

 - adding `extensionsToTreatAsEsm` to jest.config.js
 - re-exporting `@webstudio-is/jest-config` from the designer config (with designers' `testMatch`)
 
 But couldn't make either work

packages/react-sdk/src/user-props/schema.ts
rpominov: @istarkov I've defined Asset in `zod` and moved transform to where we write to DB. Please check.

rpominov:
Almost got Jest working. But it fails with following now:

```
SyntaxError: The requested module 'react-use' does not provide an export named 'useDebounce'
```

PR #466, opened 2022-11-18
Generate icons index.ts with after hook

TrySound:
Ref https://github.com/svg/svgo-components/tree/main/packages/svgo-jsx#how-to-generate-indexjs-if-you-really-need-it

Use svgo/jsx "after" hook instead of custom script for index.ts generation. This makes setup slightly simpler and more portable.

Use tsx as loader to run svgo-jsx instead of using .bin as entry point. This is safer when use in various package managers.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


kof (approved):
ah index is just the list of exports

PR #472, opened 2022-11-19
[release-fix] Flex gap icon is not button

istarkov:
Flex Section - the left icon isn't supposed to have greyish background

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -8,6 +8,7 @@ import {
   ComboboxListbox,
   ComboboxListboxItem,
   numericScrubControl,
+  TextFieldIcon,
   TextFieldIconButton,
   styled,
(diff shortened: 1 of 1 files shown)


apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -367,7 +368,7 @@ export const CssValueInput = ({
   );
 };
 
-const CssValueInputIconButton = styled(TextFieldIconButton, {
+const CssValueInputIconButton = styled(TextFieldIcon, {
```
kof, line 371: @natlusyht do you remember why you used icon button instead of icon? Was it supposed to be interactive

TrySound (approved):

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: Seems like it should be an icon indeed

kof (approved):

PR #476, opened 2022-11-19
Check all .ts and .tsx in packages src folder

TrySound:
Here "include" often have list of extensions and forgotten .tsx so such modules are left unchecked. Replaced with just "src" which consider all typescript extensions.

Ideally stop using tsc to build artifacts, switch to esbuild and remove complex configuration from all tsconfig.json.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/tsconfig.json
+++ b/packages/react-sdk/tsconfig.json
@@ -1,12 +1,7 @@
 {
   // Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
   "extends": "../../node_modules/@webstudio-is/tsconfig/base.json",
-  "include": [
-    "./src/**/*.ts",
-    "./src/**/*.tsx",
-    "./src/**/*.json",
-    "./@types/**/*.d.ts"
-  ],
+  "include": ["./src", "./src/**/*.json", "./@types/**/*.d.ts"],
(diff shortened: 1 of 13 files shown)


kof (approved):

packages/react-sdk/tsconfig.json
```diff
@@ -1,12 +1,7 @@
 {
   // Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
   "extends": "../../node_modules/@webstudio-is/tsconfig/base.json",
-  "include": [
-    "./src/**/*.ts",
-    "./src/**/*.tsx",
-    "./src/**/*.json",
-    "./@types/**/*.d.ts"
-  ],
+  "include": ["./src", "./src/**/*.json", "./@types/**/*.d.ts"],
```
Andarist, line 4: can't we simplify this one to something like this?
```suggestion
  "include": ["./src", "./@types"],
```

Andarist (approved):

packages/react-sdk/tsconfig.json
TrySound: Looks like json is not part of default extensions. I get errors.

PR #478, opened 2022-11-19
don't display empty unitless dropdown

natlusyht:
don't display chevron if there are no items to display in the dropdown and fix the bg on chevron's/unit text-field suffix button.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -61,17 +61,12 @@ const useHandleOnChange = (
     }
 
     // We want to switch to unit mode if entire input is a number.
-    if (isNumericString(input)) {
+    if (isNumericString(input) && isValid(property, input + "px")) {
       if (value.type === "unit" && String(Number(input)) !== input) return;
       onChange?.({
         type: "unit",
         // Use previously known unit or fallback to px.
-        unit:
-          valueRef.current.type === "unit"
-            ? valueRef.current.unit
-            : isValid(property, input + "px")
-            ? "px"
-            : "number",
+        unit: valueRef.current.type === "unit" ? valueRef.current.unit : "px",
         value: Number(input),
       });
       return;
@@ -131,11 +126,7 @@ const useScrub = ({
     const scrub = numericScrubControl(scrubRef.current, {
       initialValue: valueRef.current.value,
       onValueInput(event) {
-        onChangeRef.current({
-          type,
-          unit,
-          value: event.value,
-        });
+        if (inputRef.current) inputRef.current.value = String(event.value);
         setIsInputActive(true);
         inputRef.current?.blur();
       },
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -29,12 +29,9 @@ import {
 } from "react";
 import { useIsFromCurrentBreakpoint } from "../use-is-from-current-breakpoint";
 import { useUnitSelect } from "./unit-select";
-import { isValid } from "../parse-css-value";
+import { isValid, isNumericString } from "../parse-css-value";
 
 const unsetValue: UnsetValue = { type: "unset", value: "" };
-
-const isNumericString = (input: string) =>
-  String(input).trim().length !== 0 && isNaN(Number(input)) === false;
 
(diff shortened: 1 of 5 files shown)


kof (approved):

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -303,21 +303,24 @@ export const CssValueInput = ({
...
+        {isKeywordValue ? (
+          <TextFieldIconButton
+            {...getToggleButtonProps()}
+            state={isOpen ? "active" : undefined}
+          >
+            <ChevronDownIcon />
+          </TextFieldIconButton>
+        ) : isUnitValue ? (
+          unitSelectElement
+        ) : null}
```
kof, line 321: we can simplify it at this point: first handle keyword value, then unit value, because we have now nested stuff, will cause confusion

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -59,7 +58,7 @@ const useHandleOnChange = (
     }
 
     // We want to switch to unit mode if entire input is a number.
-    if (isNumericString(input)) {
+    if (isNumericString(input) && isValid(property, input + "px")) {
```
istarkov, line 61: What if its not valid with current unit `unit = valueRef.current.type === "unit" ? valueRef.current.unit : "px"`
Here we validate with fallback and later call onChange with other unit

istarkov: Also would be good to have some explanation why px is fallback unit

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -124,11 +123,7 @@ const useScrub = ({
     const scrub = numericScrubControl(scrubRef.current, {
       initialValue: valueRef.current.value,
       onValueInput(event) {
-        onChangeRef.current({
-          type,
-          unit,
-          value: event.value,
-        });
+        if (inputRef.current) inputRef.current.value = String(event.value);
```
istarkov, line 126: We already have inputRef.current null check, if `ts` can't detect that probably would be better to define at top
`const inputElement  = inputRef.current`

Also, I think in case of any ref is null inside effect you need to at least `warnOnce` or even `throw new Error ` instead of just `return ""`

apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
```diff
@@ -19,6 +19,12 @@ export const getFinalValue = ({
   if (currentValue?.value === "inherit" && inheritedValue !== undefined) {
     return inheritedValue;
   }
+  if (
+    currentValue?.type !== "unit" &&
+    isNumericString(String(currentValue?.value))
+  ) {
+    return { value: Number(currentValue?.value), type: "unit", unit: "number" };
```
istarkov, line 26: I see that the `unit` definition is `export const Unit = z.union([z.enum(units), z.literal("number")]);` so we have all possible `css` values and add our own "unit" "number",  probably better if it will have just `type="number"`
instead of type="unit" unit="number"
Otherwise for me it looks like mixing different concepts

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
istarkov: Also we have `toValue` https://github.com/webstudio-is/webstudio-designer/blob/0b37ccdefe7dcf692a5efe5eba33a51adc47ec71/packages/css-engine/src/core/to-value.ts#L12

apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
istarkov: Not a single case, there are integer and number data types. Some are have sense only being positive etc
Probably its ok to be in unit.

`order`, `z-index` https://developer.mozilla.org/en-US/docs/Web/CSS/integer
`flex-grow` is [float >=0](https://developer.mozilla.org/en-US/docs/Web/CSS/number)

istarkov (approved):

kof (approved):

PR #483, opened 2022-11-20
Fix upload errors

### LINKED ISSUE #479: Weird bug with image upload
With one specific image. Will attach in DMs, since contains personal info.

```
 Error: malformed content-disposition header: mismatched quotations in `form-data; name="image"; filename="store.google.com_orderdetails_&hl=de (1).png"`
 ```

### PR DESCRIPTION
**Pull request by istarkov**
Cherry-picked from #444 
closes #479

- [x] - Fixes Cyrillic, spaces etc
- [x] - `a=b.png` fails

Finally: 
Remix `unstable_parseMultipartFormData` is broken, https://github.com/remix-run/remix/issues/4443
request.formData on remix is broken too

2 ways to fix:
-  see comment [here](https://github.com/webstudio-is/webstudio-designer/issues/479#issuecomment-1321504440) - is to use undici Request.
- temporary sanitize file name on a client

Sanitize instead of encoding allows to use non-urlencoded URLs like [https://assets-dev.webstudio.is/иван_copy_637b183a8ede199b0b2ae5d8.png](https://assets-dev.webstudio.is/иван_copy_637b183a8ede199b0b2ae5d8.png) what every SEO would like a lot

image

BTW as of now we use <https://github.com/webstudio-is/webstudio-designer/blob/ed6dfa92b8818074813a60a45246507629efe810/packages/asset-uploader/src/utils/get-asset-path.ts#L23> what encode filename anyway

Probably better to just join `origin` and `pathname` instead of using URL.

I decided to duplicate sanitize instead of reusing it as it should go some day from designer code.

### CODE REVIEW

File: packages/asset-uploader/src/targets/s3/upload.ts [hunk 1]
```diff
@@ -86,7 +86,8 @@ const uploadHandler = async ({
     Body: data,
     ContentType: file.contentType,
     Metadata: {
-      filename: file.filename || "unnamed",
+      // encodeURIComponent is needed to support special characters like Cyrillic
+      filename: "meta", // encodeURIComponent(file.filename) || "unnamed",
```
rpominov, line 90: I think we should either add a comment about why we preserve the old code (`encodeURIComponent`) or delete it

rpominov: Also, is it ok to loose `file.filename` here?

istarkov: Comment exists, "meta" was to test

rpominov (approved):

TrySound (approved):

apps/designer/app/designer/shared/assets/sanitize-s3-key.ts
```diff
@@ -0,0 +1,26 @@
...
+const specialRegexps = [/[\x00-\x1F\x7F]+/g, /[&$@=;/:+\s,?]+/g];
+
+// eslint-disable-next-line no-control-regex
+const avoidRegexps = [/[\x80-\xFF]+/g, /[\\{^}%`\]'"“”>[~<#|]+/g];
+
+const allRegexps = [...specialRegexps, ...avoidRegexps];
+
+const REPLACE_CHAR = "_";
+
+export const sanitizeS3Key = (str: string): string => {
```
kof, line 24: should this function be in asset-uploader package?

kof: oh you have it twice in the PR now???

packages/asset-uploader/src/utils/sanitize-s3-key.ts
```diff
@@ -0,0 +1,24 @@
...
+const specialRegexps = [/[\x00-\x1F\x7F]+/g, /[&$@=;/:+\s,?]+/g];
+
+// eslint-disable-next-line no-control-regex
+const avoidRegexps = [/[\x80-\xFF]+/g, /[\\{^}%`\]'"“”>[~<#|]+/g];
+
+const allRegexps = [...specialRegexps, ...avoidRegexps];
+
+const REPLACE_CHAR = "_";
+
+export const sanitizeS3Key = (str: string): string => {
```
kof, line 22: please add some tests with different example strings that this is fixing

apps/designer/app/designer/shared/assets/sanitize-s3-key.ts
istarkov: I wrote in the commit description and in duplicate comments that it's temporary, and since remix will fix it we will just remove it from UI side. No need to share tmp things

kof: what's the problem with sharing it until it's not fixed? it can be imported from the package just like any other function. its a non-trivial function with lots of logic hidden in the regexes, it needs to have tests etc, what stops you from reusing it? Just the fact it will be removed from the designer means you can remove the usage of it from the  designer.

istarkov: What package I should put it in? I always prefer duplication instead of referencing other packages.
I really don't like the idea of putting the same piece of code into a shareable place. Here is obviously temporary function wich is not needed to be in public interface of any package.

packages/asset-uploader/src/utils/get-asset-path.ts
```diff
@@ -8,18 +8,24 @@ const fsEnv = FsEnv.parse(process.env);
 export const getAssetPath = (asset: DbAsset) => {
   if (asset.location === Location.FS) {
     const splitPath = fsEnv.FILE_UPLOAD_PATH.split("public");
-    return path.join("/", splitPath[splitPath.length - 1], asset.name);
+    const locUrl = new URL(
```
kof, line 11: what's locUrl, localUrl? locationUrl? try to not use abbreviations as much as possible

packages/asset-uploader/src/utils/sanitize-s3-key.test.ts
```diff
@@ -0,0 +1,46 @@
+import { sanitizeS3Key } from "./sanitize-s3-key";
+
+describe("sanitizeS3Key", () => {
```
kof, line 3: nice!

kof (approved):
one naming comment, otherwise great!

PR #488, opened 2022-11-21
use RemixLink in Link

### LINKED ISSUE #457: Use Remix's Link component to navigate without full reload
Our link component in sdk needs to use Remix's link to take advantage of the router

https://remix.run/docs/en/v1/api/remix#link

https://github.com/remix-run/remix/blob/2248669ed59fd716e267ea41df5d665d4781f4a9/packages/remix-react/components.tsx#L489-L516

https://github.com/remix-run/react-router/blob/f3d3e05ec00c6950720930beaf74fecbaf9dc5b6/packages/react-router-dom/index.tsx#L344-L400

rpominov:
Fixes #457

TODO:

- [x] `?mode=preview` should be preserved when clicking on a link while in preview mode
- [x] Make sure absolute URLs work as before

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/custom-components/index.ts
+++ b/apps/designer/app/canvas/custom-components/index.ts
@@ -1,0 +1,8 @@
+import { customComponents } from "@webstudio-is/react-sdk";
+import { wrapLinkComponent } from "./link";
+
+export default {
+  ...customComponents,
+  Link: wrapLinkComponent(customComponents.Link),
+  RichTextLink: wrapLinkComponent(customComponents.RichTextLink),
+};
--- a/apps/designer/app/canvas/custom-components/index.ts
+++ b/apps/designer/app/canvas/custom-components/index.ts
@@ -1,8 +1,8 @@
-import { customComponents } from "@webstudio-is/react-sdk";
+import { customComponents as baseCustomComponents } from "@webstudio-is/react-sdk";
 import { wrapLinkComponent } from "./link";
 
-export default {
-  ...customComponents,
-  Link: wrapLinkComponent(customComponents.Link),
-  RichTextLink: wrapLinkComponent(customComponents.RichTextLink),
+export const customComponents = {
+  ...baseCustomComponents,
+  Link: wrapLinkComponent(baseCustomComponents.Link),
+  RichTextLink: wrapLinkComponent(baseCustomComponents.RichTextLink),
 };
--- a/apps/designer/app/canvas/custom-components/link.tsx
+++ b/apps/designer/app/canvas/custom-components/link.tsx
@@ -8,15 +8,25 @@ import type {
 import { forwardRef } from "react";
 import { preserveSearchBuildParams } from "~/shared/router-utils";
 
-// href starts with http:, mailto:, tel:, etc.
-const isAbsoluteUrl = (href: string) => /^[a-z]+:/i.test(href);
+const isAbsoluteUrl = (href: string) => {
+  try {
+    new URL(href);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
 
 export const preserveBuildParams = (href: string, sourceSearch: string) => {
   if (isAbsoluteUrl(href)) {
     return href;
   }
 
-  const url = new URL(href, "http://localhost");
+  const url = new URL(
+    href,
+    // doesn't affect anything, just need some base to use URL
+    "http://localhost"
+  );
 
   preserveSearchBuildParams(
     new URLSearchParams(sourceSearch),
--- a/apps/designer/app/canvas/custom-components/link.tsx
+++ b/apps/designer/app/canvas/custom-components/link.tsx
@@ -1,0 +1,48 @@
+import { useLocation } from "@remix-run/react";
+import type {
+  ElementRef,
+  ComponentProps,
+  ForwardRefExoticComponent,
+  RefAttributes,
+} from "react";
+import { forwardRef } from "react";
+import { preserveSearchBuildParams } from "~/shared/router-utils";
+
+// href starts with http:, mailto:, tel:, etc.
+const isAbsoluteUrl = (href: string) => /^[a-z]+:/i.test(href);
+
+export const preserveBuildParams = (href: string, sourceSearch: string) => {
+  if (isAbsoluteUrl(href)) {
+    return href;
+  }
+
+  const url = new URL(href, "http://localhost");
+
+  preserveSearchBuildParams(
+    new URLSearchParams(sourceSearch),
+    url.searchParams
+  );
+
+  const search = url.searchParams.toString();
+
+  return `${url.pathname}${search === "" ? "" : `?${search}`}`;
+};
+
+type Props = Omit<ComponentProps<"a">, "href"> & { href?: string };
+
+type Ref = ElementRef<"a">;
+
+export const wrapLinkComponent = (
+  BaseLink: ForwardRefExoticComponent<Props & RefAttributes<Ref>>
+) => {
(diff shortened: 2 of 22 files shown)


rpominov:
> ?mode=preview should be preserved when clicking on a link while in preview mode

This is tricky actually. `react-sdk` should know about `mode` URL parameter that designer uses. 

I may be able to solve this by changing how router converts `to` to `href` if that's configurable. But if that doesn't work, we probably need to do something similar to how we're going to change props for the image in #444

rpominov:
> I may be able to solve this by changing how router converts `to` to `href` if that's configurable.

Don't see any way to do this

rpominov:
- [ ] Doesn't work in preview deployment for some reason. Looking...

packages/react-sdk/src/components/link.tsx
```diff
@@ -1,13 +1,51 @@
...
+    return href;
+  }
+
+  const mode = new URLSearchParams(currentSearch).get("mode");
+
+  if (mode !== "preview" && mode !== "edit") {
+    return href;
+  }
+
+  const [path, search] = href.split("?");
```
istarkov, line 23:
```js
const url = new URL(href, fakeBaseIfNotAbsolute)
url.searchParams.set()
return isAbs ? url.href : url.path...
```

rpominov: Do you think it's bug prone to `href.split("?")`? Because otherwise current way seem simpler to me

packages/react-sdk/src/components/link.tsx
```diff
@@ -1,13 +1,57 @@
...
+
+// @todo: Ideally this code should live in designer app.
+// Sdk should not know how routing is implemented in designer.
+// We can move it there later when we have a way of customizing components per environment.
+const preserveBuildParams = (href: string, currentSearch: string) => {
+  if (isAbsoluteUrl(href)) {
+    return href;
+  }
+
+  const [path, search] = href.split("?");
```
istarkov, line 17: please don;t what if user will pass `/ddd?ddd/?d=1` are u sure you will parse it same as browser

istarkov:
I think it's +- the same as an Image component. In the case of the Image component, it can optimize only if images are hosted on our/Cloudflare domain with specific settings. Here `Link` logic is applied only on canvas.

I think in both cases we can pass Components from the top, and override "native" components. 
In both cases instead of this https://github.com/webstudio-is/webstudio-designer/blob/39d61f318bd335a4bb3895b11ebe77b06a37e056/packages/react-sdk/src/tree/wrapper-component.tsx#L40 will be something like `useComponent(instance)` where we can place any logic of selecting "right" component from native or overrides.

rpominov:
Yeah, I also think it's the same. 

`useComponent(instance)` sounds good to me, and it can read from a context provider.

Two questions:
 - @kof do you agree on this approach?
 - @istarkov do you want to do this? Or I can do it

rpominov:
Discussed on a call. Summary:

- We don't want to swap component at the designer app side, because designer will assume that SDK uses React. That's bad because we want to be able to run a Vue SDK for example.
- For that reason we decided to have a dev mode in SDK. In canvas we'll run SDK in dev mode, and in published sites in prod mode.
- It's ok that dev mode code in SDK will be aware of designer implementation details
- We don't want dev mode SDK code to appear in bundle in prod sites

TrySound (approved):

packages/react-sdk/src/components/link.tsx
```diff
@@ -1,13 +1,56 @@
...
+  if (isAbsoluteUrl(href)) {
+    return href;
+  }
+
+  const url = new URL(href, "http://localhost");
+
+  const currentSearchParams = new URLSearchParams(currentSearch);
+
+  const mode = currentSearchParams.get("mode");
+  if (mode === "preview" || mode === "edit" || mode === "published") {
```
kof, line 22: we need to have a single source of truth for those modes since they are used in multiple places now and if we need to rename or add, it would be much simpler to figure out

packages/react-sdk/src/components/link.tsx
```diff
@@ -1,13 +1,56 @@
...
 
+const isAbsoluteUrl = (href: string) => {
+  // http:, mailto:, tel:, etc.
+  return /^[a-z]+:/i.test(href);
+};
+
+// @todo: Ideally this code should live in designer app.
+// Sdk should not know how routing is implemented in designer.
+// We can move it there later when we have a way of customizing components per environment.
+const preserveBuildParams = (href: string, currentSearch: string) => {
```
kof, line 12: this function could use some tests I think

kof:
@istarkov we have an option here to move this stuff  into remix folder to keep all remix specific stuff in one place and later that stuff to a remix-app package, but we would need here the registration so he can override the components/link

packages/react-sdk/src/components/link.tsx
```diff
@@ -1,13 +1,29 @@
...
+
+export const renderLink = (
+  href: string,
+  props: Omit<ComponentProps<"a">, "href">,
+  ref: React.ForwardedRef<HTMLAnchorElement>
+) => {
+  return isAbsoluteUrl(href) ? (
+    <a {...props} href={href} ref={ref} />
+  ) : (
+    <RemixLink {...props} to={href} ref={ref} />
```
kof, line 17: since we now hvae src/spp/custom-components in sdk package, we should clearly separate react components and remix components, so we can later move the app-level components and logic out of SDK

In this case it means the link link components in  src/components can stay dumb and we wrap it into that component that uses RemixLink

Does this make sense to you?

rpominov: Yeah, was also thinking about this

apps/designer/app/canvas/custom-components/index.ts
```diff
@@ -0,0 +1,8 @@
+import { customComponents } from "@webstudio-is/react-sdk";
+import { wrapLinkComponent } from "./link";
+
+export default {
```
istarkov, line 4: mb `export const customComponent = {}` remember @TrySound don't like default exports

apps/designer/app/canvas/custom-components/link.tsx
```diff
@@ -0,0 +1,48 @@
...
+  ElementRef,
+  ComponentProps,
+  ForwardRefExoticComponent,
+  RefAttributes,
+} from "react";
+import { forwardRef } from "react";
+import { preserveSearchBuildParams } from "~/shared/router-utils";
+
+// href starts with http:, mailto:, tel:, etc.
+const isAbsoluteUrl = (href: string) => /^[a-z]+:/i.test(href);
```
istarkov, line 12: inetresting that `new URL('hello:world');` works and defines hello as protocol. Always thought that '//' is needed.
probably safe way is 
```js
try {
  new URL(href); 
  return true
}
catch { return false }
```
but have no strong opinion

apps/designer/app/canvas/custom-components/link.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const preserveBuildParams = (href: string, sourceSearch: string) => {
+  if (isAbsoluteUrl(href)) {
+    return href;
+  }
+
+  const url = new URL(href, "http://localhost");
```
istarkov, line 19: I would write that localhost here has no sense and used as fake host just to use URL

istarkov (approved):

PR #491, opened 2022-11-21
Pages update for Edge version

andrasbacsai:
Related [PR](https://github.com/webstudio-is/webstudio-saas/pull/7).

# Features
- Added new rest API to gather all project details in one request.
- Rewritten `loadProject` function in `http-client` to use the new endpoint.
- Moved some dependencies of the new `CSS Engine` to `@webstudio-is/project` to be usable in the edge version.

# Fixes
- Added metadata to published sites.

# Internal
- Removed package tests.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/http-client/src/index.ts
```diff
@@ -1,80 +1,129 @@
...
     }
 
-    const pagesData = await Promise.all(
+    if (urls.length === 0) {
+      throw new Error("No pages to fetch");
+    }
+    const data = await Promise.all(
       urls.map(async (url) => {
         const res = await fetch(url.url);
         if (res.ok) {
```
kof, line 93: I am not quite sure but I think you are serially awaiting here, not fetching in parallel

packages/http-client/src/index.test.ts
```diff
@@ -1,39 +1,39 @@
 import { loadProject } from "./index";
 
-const existingProjectId = "398d3918-9b00-4a43-8ab6-de4e151ae98a";
```
kof, line 3: I think tests are not reflecting all the new cases we have added to the client

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,129 @@
...
+    if (homePage) {
+      const { id } = homePage;
+      const treeUrl = new URL(`/rest/tree/${projectId}/${id}`, baseUrl);
+      const propsUrl = new URL(`/rest/props/${projectId}/${id}`, baseUrl);
+      const breakpointsUrl = new URL(`/rest/breakpoints/${projectId}`, baseUrl);
+      const cssUrl = new URL(`/rest/css/${projectId}/${id}`, baseUrl);
       urls.push({
-        url: new URL(`/rest/tree/${projectId}/${page.id}`, baseUrl),
-        path: page.path,
+        page: { ...homePage },
```
kof, line 32: why do you spread instead of just using? Is it mutating it or something?
```suggestion
        page: homePage,
```

apps/designer/app/shared/css-utils/generate-css-text.ts
```diff
@@ -0,0 +1,44 @@
+import { json } from "@remix-run/node";
+import { db } from "@webstudio-is/project/server";
+import { utils } from "@webstudio-is/project";
+import { loadCanvasData } from "~/shared/db";
+import { createCssEngine } from "@webstudio-is/css-engine";
+import { idAttribute } from "@webstudio-is/react-sdk";
+import { addGlobalRules } from "~/canvas/shared/styles";
+import { BuildParams } from "../router-utils";
```
kof, line 8: use type hint always, we will automate the check through linter later

```suggestion
import type { BuildParams } from "../router-utils";
```

packages/http-client/src/index.d.ts
```diff
@@ -4,17 +4,18 @@ import type {
   UserProp,
 } from "@webstudio-is/react-sdk";
 
-export type IncludeTypes = "tree" | "props" | "breakpoints" | "pages";
-export type Includes<T> = { [include in IncludeTypes]: T };
 export type Page = {
   id: string;
   root: [string, string];
 };
 export type Project = {
```
kof, line 11: Maybe `BuildData`?

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,129 @@
...
+      throw new Error(`Project ${projectId} needs to be published first`);
+    }
+    const { pages, homePage } = listOfPages;
+    if (homePage) {
+      const { id } = homePage;
+      const treeUrl = new URL(`/rest/tree/${projectId}/${id}`, baseUrl);
+      const propsUrl = new URL(`/rest/props/${projectId}/${id}`, baseUrl);
+      const breakpointsUrl = new URL(`/rest/breakpoints/${projectId}`, baseUrl);
+      const cssUrl = new URL(`/rest/css/${projectId}/${id}`, baseUrl);
       urls.push({
```
kof, line 31: I think the entire block for building the fetch request would be a bit nicer to read if you had a fetch utility and then just would just compose it directly

```ts
await Promise.all([
  fetchData('project', {projectId, pageId}),
  ...
])
```

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,129 @@
...
+            ...gatheredPages[path],
+            page: page.page,
+            [type]: page.data,
+            breakpoints: breakpoints?.data || {},
           };
         }
       }
     }
+    return {
+      ...gatheredPages,
```
kof, line 121: also spreading, seems unnecessary

apps/designer/app/shared/router-utils/build-params.ts
```diff
@@ -2,6 +2,27 @@
...
+      pathname: string;
+      pageId?: string;
+    }
+  | {
+      projectDomain: string;
+      mode: BuildMode;
+      pathname: string;
+      pageId?: string;
+    }
+  | undefined;
```
rpominov, line 18: I wouldn't add `| undefiend` here, but instead use `BuildParams | undefined` where `BuildParams` is used. Because `undefined` is not a correct params object

packages/http-client/src/index.d.ts
```diff
@@ -4,17 +4,18 @@ import type {
...
-  tree: BaseInstance & { errors: string };
-  props: Array<UserProp> & { errors: string };
-  breakpoints: Array<Breakpoint> & {
-    errors: string;
+  [include in string]: {
```
rpominov, line 12: This is equivalent to just `[string]`, no?

packages/http-client/src/index.d.ts
```diff
@@ -4,17 +4,18 @@ import type {
...
+    page: Page;
+    tree: BaseInstance & { errors: string };
+    props: Array<UserProp> & { errors: string };
+    breakpoints: Array<Breakpoint> & {
+      errors: string;
+    };
```
rpominov, line 18: From the consumer of REST api point of view, is it useful to get a response where you get all the data, but a `{ errors: string }` in one of the fields? Maybe the response type should be either all data or an error?

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,129 @@
...
 
-  if (include.pages && listOfPages.pages && listOfPages.pages.length > 0) {
-    for (const page of listOfPages.pages) {
+    const baseUrl = new URL(apiUrl);
+    const pagesUrl = new URL(`/rest/pages/${projectId}`, baseUrl);
+    const listOfPages = await (await fetch(pagesUrl)).json();
+    if (listOfPages === undefined || listOfPages.errors) {
+      throw new Error(`Project ${projectId} needs to be published first`);
+    }
+    const { pages, homePage } = listOfPages;
```
rpominov, line 24: To avoid some code duplication we could probably do something like `const pages = [listOfPages.homePage, ...listOfPages.pages]`, and then handle only `pages` in one block

packages/http-client/src/index.ts
rpominov: Looks like parallel to me, but it actually can be a bad idea to send all these requests at once. If we assume there can be 100s of pages, we might be sending 300 requests at once here.

Not sure it should be a priority now, but we should probably implement a queue where only a fixed number of requests is handled in parallel. Maybe there's a library for that.

I'd actually consider using https://github.com/graphql/dataloader , as it would allow to write code as if we're fetching everything serially instead of building a `urls` array, but under the hood it would send requests in parallel. This would make code more straightforward and with better TypeScript types. And it has a `maxBatchSize` setting.

apps/designer/app/routes/rest/css.$projectId.$pageId.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const projectId = params.projectId ?? undefined;
+    const pageId = params.pageId ?? undefined;
+
+    if (projectId === undefined) {
+      throw json("Required project id", { status: 404 });
+    }
+
+    if (pageId === undefined) {
+      throw json("Required page id", { status: 404 });
+    }
```
rpominov, line 18: With these error messages, we should probably use `400`. Same with `Required project info` below

packages/http-client/src/index.ts
rpominov: After thinking more: probably a bad idea with dataloader actually. Not a good fit for the problem

andrasbacsai: Yes, that is a valid problem (100s pages). 
A proper solution would be to use a single endpoint dedicated to fetching all project details and sending a response.

andrasbacsai: I go this way.

andrasbacsai:
Redesigned the whole process flow. One endpoint gathers all data.

@kof & @rpominov, please take a look again.

TrySound (approved):

packages/http-client/src/index.d.ts
```diff
@@ -4,17 +4,20 @@ import type {
...
+        [string]: {
+          page: Page;
+          tree: BaseInstance;
+          props: Array<UserProp> | [];
+          breakpoints: Array<Breakpoint> | null;
+          css: string;
+        };
+      };
+    }
+  | Error;
```
rpominov, line 23: `Error` seem wrong here:

 1. We shouldn't use native Error type probably unless it's in context of try..catch or something
 2. Same as with `undefined` before, this is wrong semantically: a error is not a `Project`

packages/http-client/src/index.d.ts
```diff
@@ -4,17 +4,20 @@ import type {
   UserProp,
 } from "@webstudio-is/react-sdk";
 
-export type IncludeTypes = "tree" | "props" | "breakpoints" | "pages";
```
rpominov, line 7: Just noticed. Why do we have a `.d.ts` file that we edit manually? All this code probably should be in `index.ts`.

But I might be missing something I'm new to TS myself, just seem strange compared to how we do it everywhere else.

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,29 @@
...
-            ...(pages[page.path] as Record<string, Page>),
-            [page.type]: page.data,
-            breakpoints,
-          };
-        }
-      }
+    return project;
+  } catch (error: unknown) {
+    if (error instanceof Error) {
+      return error;
```
rpominov, line 25: It's a strange API to return a Error from a function

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,29 @@
...
     }
+    return new Error("Unknown error");
```
rpominov, line 27: This is bad that we ignore the caught error if it's not an instance of Error. We should log / re-throw it anyway because it can contain insight of what went wrong.

packages/http-client/src/index.ts
kof: there are some tiny abstractions out there that allow you to nicely limit the parallelization level, e.g. limit it to 20 pages in parallel or something ... even if you add an endpoint that does all the work on the node server, is it actually making it faster? This would only be the case if you loading more data from the database for each request than you need.

andrasbacsai: It is not about how fast data loads but the number of requests going back and forth. With the single endpoint, it is only one.

apps/designer/app/routes/rest/project.$projectId.tsx
```diff
@@ -0,0 +1,97 @@
...
+        {
+          status: 500,
+        }
+      );
+    }
+    pagesDetails["/"] = {
+      page: homePage,
+      tree,
+      props: await db.props.loadByTreeId(homePage.treeId),
+      css: await generateCssText({
```
kof, line 51: hmm, now every rest call would be generating the CSS text and also delivering CSS as part of the tree, kindof duplicated data quite a bit, I think we should probably only do that as an option?

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,46 @@
...
+import type { BaseInstance, UserProp } from "@webstudio-is/react-sdk";
+import type { Breakpoint } from "@webstudio-is/css-data";
+
+export type Page = {
+  id: string;
+  root: [string, string];
+};
+export type Project = {
+  pages: {
+    [path: string]: {
```
kof, line 11: ultimately this data structure isn't great I think, ultumately you want pages to be an array:

```
pages: [{path, page, tree, props etc}]
```

rpominov (approved):

apps/designer/app/routes/rest/project.$projectId.tsx
kof: I also just checked you are loading the build in generateCssText function and you do the same here  ... you will produce here lots of unnecessary queries to the database

kof: I think that you broke the fundamental idea behind the API: allowing user to decide how they render anything on the canvas.

API was supposed to deliver all the data so that user can compile whatever they need: any CSS they need, any components they need.

In the attempt to optimize the data fetching you changed the very premise of the api.

kof: Now even if we add an option for CSS either being a string or part of the tree, you don't have all the data user needs to build their CSS: e.g. breakpoints.

packages/http-client/src/index.ts
andrasbacsai: For me, both are fine. So I will make it ultimate 😎.

apps/designer/app/routes/rest/project.$projectId.tsx
kof: I think you need to step back and only deliver data, without compiling anything. Compilation step should be still up to the consumer of the data and we need to take great care that those functions are easy to use and fast.

packages/http-client/src/index.ts
kof: generally for apis its not great to have collections expressed as hashes. Hashes is an optimization for certain access patterns, that should be done on the consumer site when necessary

kof (approved):

packages/http-client/src/index.ts
```diff
@@ -1,80 +1,46 @@
...
+      page: Page;
+      tree: BaseInstance;
+      props: Array<UserProp> | [];
+      breakpoints: Array<Breakpoint> | null;
+      css: string;
```
kof, line 16: didn't you remove it from api?

PR #493, opened 2022-11-21
Migrate packages to native ESM

TrySound:
Our designer bundle heavily bloated with not analyzable commonjs generated by typescript.

The biggest problem comes from bundled icons library which esbuild cannot tree shake because of cjs.

Another issue is broken code splitting. Packages in one piece are appear in chunks and always loaded even with dynamic import and React.lazy.

To solve this I migrated each package to native node es modules. Storybook and jest work almost without changes. Though it's not possible to run remix in esm project (while still esm packages can be bundled for ssr). And as a result designer jest failed because of cjs tests requires from esm packages. As a solution I generated additional commonjs artifacts in lib/cjs.

du -ck public/build
2336
1736

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [x] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/project/package.json
```diff
@@ -28,7 +30,14 @@
...
     "typescript": "4.7.4"
   },
-  "main": "lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib/cjs/index.cjs",
+      "default": "./lib/index.js"
+    },
+    "./server": "./server.js"
```
Andarist, line 39: q: why this doesnt have a require condition? 🤔

packages/project/package.json
```diff
@@ -28,7 +30,14 @@
     "tsup": "^6.1.3",
     "typescript": "4.7.4"
   },
-  "main": "lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib/cjs/index.cjs",
+      "default": "./lib/index.js"
```
Andarist, line 37: With those set of conditions `default` seems to be redundant. I don't exactly remember the very first rules around conditions in node before they were adjusted but I don't think that you have to worry about some old node@12 versions here.

packages/asset-uploader/server.js
```diff
@@ -1 +1 @@
-module.exports = require("./lib/index.server");
+export * from "./lib/index.server";
```
Andarist, line 1: This doesn't adhere to "native ESM" requirement - you need an extension here.

packages/icons/package.json
```diff
@@ -37,7 +39,11 @@
     "@radix-ui/react-icons": "^1.1.0",
     "@webstudio-is/css-vars": "*"
   },
-  "main": "lib/index.js",
+  "exports": {
```
Andarist, line 42: This creates problems and those packages, with the current output, are not actually loadable at all - with `exports` you need explicit extension.

```ts
// test.mjs
import * as Icons from "@webstudio-is/icons";

console.log(Icons);
```

The snippet above fails with:
> Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/webstudio-designer/packages/icons/lib/legacy' imported from /webstudio-designer/packages/icons/lib/index.js

And the same thing happens for:
```js
// test.cjs
console.log(require("@webstudio-is/icons"));

```

packages/asset-uploader/package.json
```diff
@@ -29,7 +30,10 @@
     "typescript": "4.7.4",
     "zod": "^3.17.3"
   },
-  "main": "lib/index.js",
+  "exports": {
```
Andarist, line 33: Note that TypeScript won't understand `exports`. To make it understand them (while there is a chance that you don't want it to understand them right now) you'd have to use `moduleResolution: node16`. This in turn would **require** you to also add a `types` condition here - in that mode TS doesn't read from the top-level fields (`package.json#types`) if `exports` are present. Enabling this would also require you to use explicit extensions within `src`.

packages/css-data/package.json
```diff
@@ -26,7 +27,11 @@
   "peerDependencies": {
     "zod": "^3.19.1"
   },
-  "main": "lib/index.js",
+  "exports": {
+    "import": "./lib/index.js",
+    "require": "./lib/cjs/index.cjs",
```
Andarist, line 32: To avoid dual package hazard in bundlers (although not in esbuild) you could also have a `module` condition here:
```suggestion
    "module": "./lib/index.js",
    "import": "./lib/index.js",
    "require": "./lib/cjs/index.cjs",
```

packages/jest-config/index.js
```diff
@@ -9,12 +9,12 @@ module.exports = {
       require.resolve("babel-jest"),
```
Andarist, line 9: I've missed this change in the previous PR (https://github.com/webstudio-is/webstudio-designer/commit/47a7bd6e4baefe0351ca80b1985ca626121bed86#diff-835ed62c13e91354b349566613127e466f4f597630588cc90d9ba5c3b8f56ad0R9) - why esbuild has been replaced with Babel here? 🤔

packages/project/package.json
TrySound: Caz not used in tests. I added dual modules support only for tested code.

packages/project/package.json
```diff
@@ -4,12 +4,14 @@
   "description": "Webstudio Project",
   "author": "Webstudio <github@webstudio.is>",
   "homepage": "https://webstudio.is",
+  "type": "module",
   "scripts": {
     "typecheck": "tsc --noEmit",
-    "test": "jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
     "checks": "yarn typecheck && yarn lint && yarn test",
     "dev": "tsup --watch",
```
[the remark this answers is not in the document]
TrySound: They did different things before too. tsup bundles code. tsc produce transpiled modules. But yeah, ideally have single tool. But probably we can worry about this later. As only tests will be affected.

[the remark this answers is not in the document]
TrySound: Eventually we would not need watch node if we switch to source code as entry point.

packages/asset-uploader/server.js
TrySound: I bet it's true for all modules. Though it's fixable with flag (or loader in the future). We can figure this out later as well.

packages/asset-uploader/package.json
TrySound: Is it newer typescript requirement? Caz now everything works fine.

packages/jest-config/index.js
TrySound: Because that esbuild plugin unconditionally run babel/plugin-transform-modules-commonjs

packages/css-data/package.json
TrySound: Looks like not a problem with remix right now. It always resolves modules. Ideally enable modules in designer as well.

packages/project/package.json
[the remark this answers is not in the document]
kof: are we making yarn dev or yarn test slower by introducing more artifacts?

packages/asset-uploader/server.js
kof: ohh, really? we are using export * all over the place

TrySound: It's about explicit extension in import source "./lib/index.server.js" etc. To fix it need probably to switch from tsc but let's leave it for another PR.

packages/project/package.json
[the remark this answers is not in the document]
TrySound: yep, tsc is slow and now it's slower but only for a few packages, not all.

packages/css-data/package.json
TrySound: Had to add old school "module" to make webpack4 in storybook work

packages/icons/package.json
TrySound: Oops, I guess designer storybook with webpack 5 just caught this issue

packages/project/package.json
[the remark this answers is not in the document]
Andarist: My main concern related to this right now is that the `dev` script doesn't actually build all of the required artifacts. So this script becomes sensitive to what other scripts you might run. Like, it might be enough when you just work on the app, but it might not be enough for whatever needs CJS artifacts (jest?). It's raising a mental overhear around this because whoever runs this script needs to know that this is the case - so the overall experience is not seamless.

packages/asset-uploader/package.json
Andarist: Not exactly - it works right now because you don't use `moduleResolution: node16`. But because you don't use it TS doesn't actually look into `package.json#exports` at all. It sticks to reading from top-level keys **and** file system layout. If you'd switch to `moduleResolutions: node16` then it would stop working because you have no `types` condition (**unless** the requested file has a sibling `.d.ts` file - which might be the case here cause you build those artifacts using `tsc`, and even in `dev` TS writes to the same dir as `tsup`, this is kinda accidental though).

My main concern here is the `/server` entrypoint. It might work because there is a file called like this at the root of the module but this won't work with dual conditions etc (could be fixed with `pkg/server/package.json` though).

So, I'm not as much concerned here about how it works **right now** but I'm worried that the fact that it works is a little bit accidental and that it might require further tweaking in the future

packages/css-data/package.json
Andarist: > Looks like not a problem with remix right now. It always resolves modules.

That's specific to this repo - because you don't explicitly use `require` anywhere (so you always end up with the import condition). For potential external consumers, this wouldn't necessarily be always true.

packages/jest-config/index.js
Andarist: It actually calls it conditionally:
https://github.com/aelbore/esbuild-jest/blob/daa5847b3b382d9ddf6cc26e60ad949d202c4461/src/index.ts#L28-L33

But ye, this is quite... something :o Was this creating any issues though?

packages/asset-uploader/server.js
Andarist: `tsc` should be able to output explicit extensions with `moduleResolution: node16` - just be cautious about this change for reasons outlined [here](https://github.com/webstudio-is/webstudio-designer/pull/493#discussion_r1029055092). Note that this, most likely, would require you to add explicit extensions to every relative import in the codebase (annoying but worthwhile in the long run)

packages/jest-config/index.js
TrySound: You are right. Didn't check format option.

packages/asset-uploader/package.json
TrySound: Alright, let's do this in separate PR

kof (approved):
@TrySound please create a followup issue for things @Andarist mentioned

Issue #238, opened 2022-08-22
Optimize published site bundle


kof:
Currently its heavily bundling a lot of unnecessary code that is only used in design mode.

TrySound:
Most of published code is react and remix. Though one chunk still has more code than we expected.

A few icons still appear in bundle. This means we need to split components and their meta. Ivan is working on this I think.

All components are appear in bundle. So I think the right way would be to statically compute all these components. Though overall I think this issue is solved.

<img width="1005" alt="image" src="https://user-images.githubusercontent.com/5635476/204834154-f5d2af27-8a2e-4f7f-bdd0-bdc2b20ad273.png">

kof:
regarding icons, please create an issue with a proposal on how to provide icons from SDK components while keeping them out of the published site bundle

PR #495, opened 2022-11-21
Get rid from remix scripts in published site

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/238

Remix is app level. App should not leak into reusable packages. Here scripts were burried somewhere deep and affected unexpected part of the app.

Here moved out of react-sdk only remix scripts. Ideally move out everything else as well.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests


kof:
Hmm, Remix is app level in the designer, but if sdk is rendering using Remix, it is also part of the published site. A full name for that react sdk is actually remix-react-sdk (should probably rename)

It seems like you are moving things from one place into another, but in the end, in both cases, the same would end up rendered on the canvas. What do I miss?

rpominov:
Might be a good idea to make `react-sdk` framework agnostic actually, so people could do something similar to https://github.com/webstudio-is/webstudio-landing but not use Remix. 

Or maybe someone has a big project where a different framework is used and they want to use Webstudio to build just a couple of pages...

Not sure this should be a priority right now though, as it will make things harder like: https://github.com/webstudio-is/webstudio-designer/pull/488

TrySound:
> Might be a good idea to make react-sdk framework agnostic actually

Yep, this is my intention. Remix does nothing useful in sdk. Only mess with component responsibility.

TrySound:
> but in the end, in both cases, the same would end up rendered on the canvas. What do I miss?

You miss published site which does not need any of this

kof:
Not sure why you say this, published site also needs scripts and scroll restoration

kof:
As discussed, we are rendering a remix app in the worker and will soon be able to build a new worker per project, so we don't need to turn it into a static site.

PR #502, opened 2022-11-23
Image

istarkov:
- [x] - Image optimization component moved into standalone package
- [x] - generate-arg-types moved into standalone package - one issue with current yarn https://github.com/yarnpkg/yarn/issues/4964 (you need to delete root node_modules, then yarn and yarn build to make generate-arg-types to work)
- [x] - move json into `__generated__` folder, (allow import json from __generated__ folders only)
- [x] - Icons moved all gen into `__generated__`
- [x] - css-data generated content moved to `__generated__`


packages/image/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "target": "ES2020",
+    "strict": true,
+    "baseUrl": ".",
+    "rootDir": "./src",
+    "outDir": "./lib"
+  },
+  // this is auto-generated by update-ts-project-references.ts
+  "references": [
+    {
+      "path": "../generate-arg-types"
```
istarkov, line 16: We must not generate references on packages without exports/module/etc.

packages/image/src/image-dev.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+  title: "Components/ImageDev",
+} as ComponentMeta<typeof ImagePrimitive>;
+
+type ImageProps = ComponentProps<typeof ImagePrimitive>;
+
+/**
+ * In case you need to test img with real cloudflare trasforms
+ * set  USE_CLOUDFLARE_IMAGE_TRANSFORM = true
+ **/
+const USE_CLOUDFLARE_IMAGE_TRANSFORM = true;
```
istarkov, line 22: false

rpominov (commented):
> allow import json from generated folders only

Why do we want this? Why not allow to import json from anywhere?

istarkov:
> > allow import json from generated folders only
> 
> Why do we want this? Why not allow to import json from anywhere?

No use-case I can imagine you would want to import hand written json ;)

rpominov (approved):

packages/generate-arg-types/src/index.ts
```diff
@@ -1,9 +1,15 @@
...
 import path from "path";
 import { z } from "zod";
 import { withCustomConfig } from "react-docgen-typescript";
 import fg from "fast-glob";
+
 import fs from "fs-extra";
-import { propsToArgTypes } from "../src/arg-types/utils";
+import { propsToArgTypes } from "./arg-types";
+
+const GENERATED_FILES_DIR = "__generated__";
```
kof, line 12: this is clear enough, no need to make it also bad looking one haha
```suggestion
const GENERATED_FILES_DIR = "generated";
```

kof (approved):

packages/image/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  // Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
+  "extends": "../../node_modules/@webstudio-is/tsconfig/base.json",
+  "include": ["src", "./src/**/__generated__/*.json"],
```
kof, line 4: since this is only for image package, I wouldn't do it, because this would be unexpected. It should be either fo all packages/apps or for none I think, then it's consistent, but doing it for all is probably not reasonable idk

kof: whatyou will likely see is someone wasting hours because importing a json in one package works and in another one doesn't, until they realize it is disabled in that package ... we need to strive for consistency across packages

so I am very much against doing it in one package

istarkov: now allowed only in 2 packages - Image and react-sdk and in both restricted to generated as no real use cases someone would want to import non generated json. This setup forces all to put new json into generated. Imo current is consistent

packages/generate-arg-types/src/index.ts
istarkov: same as default in folder for generated files in relay. Imo underscores pointing that its not usual folder like in modern folder based routers signs on files like + $ etc are used

packages/generate-arg-types/src/index.ts
```diff
@@ -1,9 +1,15 @@
+#!/usr/bin/env -S node --experimental-specifier-resolution=node
```
TrySound, line 1: Do we need a separate package for this? Maybe generate only inside sdk?

istarkov: and how to generate in image

packages/image/src/image.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return (
+      <img {...imageProps} {...imageAttributes} decoding="async" ref={ref} />
+    );
+  }
+);
+
+Image.defaultProps = {
+  src: "",
+  loading: "lazy",
+};
```
TrySound, line 36: defaultProps is legacy and may be eventually be removed in favour of destructuring defaults

istarkov: We know there is another PR exists, will fix all defaultProps there

packages/generate-arg-types/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "@webstudio-is/generate-arg-types",
+  "version": "0.0.1",
+  "description": "Generate components property types",
+  "author": "Webstudio <github@webstudio.is>",
+  "homepage": "https://webstudio.is",
+  "type": "module",
+  "scripts": {
+    "typecheck": "tsc --noEmit",
+    "checks": "yarn typecheck && yarn lint",
+    "build": "rm -rf .bin tsconfig.tsbuildinfo && tsc && chmod +x .bin/index.js",
```
TrySound, line 11: I would not build .bin at all. We have tsx loader which does all work.
NODE_OPTIONS='--loader=tsx' svgo-jsx svgo-jsx.config.ts

packages/generate-arg-types/src/index.ts
TrySound, line 1: --loader tsx can be put here

istarkov: Super!

packages/generate-arg-types/package.json
```diff
@@ -28,7 +27,7 @@
     "!*.test.*"
   ],
   "bin": {
-    "generate-arg-types": "./.bin/index.js"
+    "generate-arg-types": "./src/index.ts"
```
TrySound, line 30: And from .gitignore

Issue #494, opened 2022-11-21
Tokens Manager


kof:
Based on this user story https://github.com/webstudio-is/webstudio-designer/issues/422

As a user, I need a way to define and apply atomic tokens to the style panel.

## Concept

The user opens a tokens picker panel that is positioned next to the style panel.
When the tokens picker is open you fundamentally need to have selected the "destination" and the "source".  
The destination is any style value (can be a property name, value button-menu, select, CSS Value Input, or anything else we use). The source is the token item from the tokens picker.

You can select a destination first, and a source second. You can select source first, and destination second. You can drag the source onto the destination (probably for later).

Once both are selected, a token value (the actual value, e.g. #000 color) applies to the style panel and canvas. The label or the button for that label receives token color coding (green?). The actual value in the data is not just a value, it's a CSS variable.
From now on, this variable can be changed in the tokens manager and reflect automatically on all consumers.

## Todo
- [ ] Decide what button will open the tokens picker
- [ ] `ctrl+t` will open the same tokens picker
- [ ] Open floating window
- [ ] How do we position that window?
- [ ] Tokens picker renders collapsible sections like style panels and lists of tokens inside of each panel
- [x] Each panel header has a + button to add a new token
- [x] Each section represents a token group. Groups can be predefined.
- [x] Each token will have a menu button (like fonts picker) and menu will have delete and edit items
- [x] Edit item opens n inline dialog with: name input, CSS value input, description textarea, cross-icon button
- [x] Changes in the tokens editor live-preview just like style panel changes and save on enter/blur
- [x] ~Data structure we save tokens in is defined by the [spec](https://design-tokens.github.io/community-group/format/) ~ - it's bad, we will support the official format for importing/exporting but won't use it internally
- [ ] fetch and render tokens on the edge
- [ ] Need to figure out if/how to copy tokens/bindings that are used on a component that we copy/pasted
- [ ] Do we want to enable copying any value from style panel so it can be pasted into token editor?

taylornowotny:
CMD + T in a browser window opens a new tab. Seems like a common shortcut that we might not want to overwrite. Any other shortcut ideas?

kof:
Simply `t` without cmd? or option+t or ctrl+t 

Need to also check if those are not used in some popular way by figma and webflow.

taylornowotny:
> Simply `t` without cmd? or option+t or ctrl+t
> 
> Need to also check if those are not used in some popular way by figma and webflow.

[...]

PR #507, opened 2022-11-24
Primitive tokens

kof:
https://github.com/webstudio-is/webstudio-designer/issues/494

This PR Renders this UI and, when adding tokens - renders CSS variables on the canvas: It's behind a feature flag.

Screenshot 2022-11-25 at 15 51 00

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/shared/styles.ts
```diff
@@ -190,6 +216,14 @@ const useUpdateStyle = () => {
   });
 };
 
+const useUpdateTokens = () => {
+  useSubscribe("updateTokens", (updatedTokens) => {
+    store.createTransaction([designTokensContainer], (tokens) => {
+      tokens.splice(0, tokens.length, ...updatedTokens);
```
rpominov, line 222: This means patch objects will always contain full rewrite of all tokens, right? Not sure if this matters now, but probably will cause problems with collaboration for example.

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    if (errors.hasErrors && event.type === "submit") {
+      setErrors(errors);
+    }
+
+    if (errors.hasErrors === false && event.type === "submit") {
+      onChangeComplete({ ...data, group, type } as DesignToken);
+      setIsOpen(false);
+    }
+  };
```
rpominov, line 98: This function seem to do anything useful only when `event.type === "submit"`. Should we even call it on change?

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -0,0 +1,178 @@
...
+        return (
+          <CollapsibleSection
+            label={group}
+            key={group}
+            rightSlot={
+              <TokenEditor
+                group={group}
+                type={type}
+                onChangeComplete={(token) => {
+                  const nextTokens = [...tokens, token];
```
rpominov, line 169: It looks like `token` will be duplicated in `nextTokens` on edit, as it probably already exists in `...tokens`.

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -0,0 +1,178 @@
...
+                  publish({ type: "updateTokens", payload: nextTokens });
```
rpominov, line 170: I'd consider passing `token` instead of `nextTokens` here, which would allow to make patches more granular in `useUpdateTokens` (as I mentioned there)

apps/designer/app/canvas/shared/styles.ts
kof: I think I made a mistake by wanting to update all tokens instead of one

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: rewrote

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: agree

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: yep, correct, editing existing token not implemented yet, added a todo

rpominov (approved):

PR #508, opened 2022-11-24
Improve packages auto publish

TrySound:
Current solution often fails with

> You cannot publish over the previously published versions

Here I added small script which compares committed version and version on npm and publish if do not match.

All scripts are run by turbo. So publishing new package is just changing package script now.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


istarkov (approved):

istarkov:
And if we want to publish standalone packages we can add:
```
on:
  workflow_dispatch:
    inputs:
      package:
        type: choice
        description: 'Select package to deploy'
        options:
        - react-sdk
        - project
        - blabla
        - ...
        required: true
```

andrasbacsai (approved):

publish-to-npm.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
kof, line 1: I was trying to keep all scripts in bin folders so far, any reason this has to be separate?

publish-to-npm.sh
```diff
@@ -0,0 +1,13 @@
...
+
```
kof, line 2: this script is small and readable so far, but it won't stay like this over time I am afraid, even now I would rewrite it in ts, so we all understand what it does clearly

I personally am a complete bash noob.

Another point is if this was merged, what stops us from merging another bash script later? and another one?

TrySound: I'm bash noob as well but simple scripts can be more expressive than ts. Since ts is not  shell language. And everything googleable.

> what stops us from merging another bash script later?

If those will be simple scripts I wouldn't worry.
This one is specifically for github actions which are bash as well by default.

Issue #487, opened 2022-11-21
Can't edit Link's text


rpominov:
1. Add a Link component
2. Double-click to start editing the text
3. Delete all text
4. Start entering new text

After you enter the first letter, it doesn't allow you to enter the rest

TrySound:
Maybe we need to distinct block link and inline link?

TrySound:
Space cannot be entered inside button because of default behaviour

TrySound:
Form button is edited with separate panel, without inline editor

rpominov:
> Maybe we need to distinct block link and inline link?
 
Sounds like a good idea, especially because afaik it's the only component that can be inline, but has `isInlineOnly: false`. If there won't be any such components, it can simplify code in some places.

TrySound:
> Another case is when cursor before or after text, input does not work for some reason

Yep, this case is fixed when I add display: block. inline-block doesn't work either.

TrySound:
display: table works and preserves inline style!

TrySound:
Funny thing, this is not lexical but contenteditable behaviour
https://codesandbox.io/s/shy-cookies-0sx1qe?file=/src/index.js

kof:
Webflow works with a link and display inline

TrySound:
Maybe we don't need Link component at all. It's same text block with link but introduce bugs like this and breaks links inside text block when double click on them.

kof:
We are going to use a Link component from Remix that comes from the router for example. User may have other needs as well in a custom sdk

TrySound:
This is fine. I mean do we need a separate Link element in left panel if text with link is the same?

Issue #397, opened 2022-11-06
Eslint rule for 2 new cases


kof:
Let's add an eslint rule for 2 new cases

1. == (only allowed to compare to null)
2. define some min length for variables min length, to avoid using one-char  variables or types like T, I think mostly we just need to check 2+ chars

_Originally posted by @kof in https://github.com/webstudio-is/webstudio-designer/pull/378#discussion_r1014797222_

istarkov:
+ Rule to not use `import React` if jsx only 
and `import * as React` in case of React.useBlabla etc is used.

TrySound:
named imports are fine, just forbid `import React from 'react'` which prevents mangling.
Btw can be done with https://eslint.org/docs/latest/rules/no-restricted-imports

andrasbacsai:
Adding the following rules to `@webstudio-is/eslint-custom-rules`:
```js
{
    eqeqeq: ["error", "always", { null: "ignore" }],
    "id-length": ["error", { min: 2 }],
}
```
And add `plugin:react/jsx-runtime` to extends for React 17+ new JSX transform.

PR #514, opened 2022-11-25
Remove paragraphs from text editor

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/487

Before lexical created paragraphs when press enter which we tried to visually represent as line break by removing margins and converted to line break to instances.

This could lead to a few issues.

1. When paragraph block is inside `<a>` nothing can be typed in the beginning and in the end of paragraph

2. Paragraphs can be styled by users which can break text editor visually

Here I overwritten default Enter command and insert line break instead of paragraph. To protect from external sources like paste from clipboard added transform to merge all paragraphs if move than one.

We will still have one paragraph in the root with display: inline style. Though it will cause less damage to editor with user styles.

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/features/text-editor/text-editor.tsx
+++ b/apps/designer/app/canvas/features/text-editor/text-editor.tsx
@@ -1,4 +1,14 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useLayoutEffect } from "react";
+import {
+  KEY_ENTER_COMMAND,
+  INSERT_LINE_BREAK_COMMAND,
+  COMMAND_PRIORITY_EDITOR,
+  RootNode,
+  ElementNode,
+  $createLineBreakNode,
+  $getSelection,
+  $isRangeSelection,
+} from "lexical";
 import { LinkNode } from "@lexical/link";
 import { LexicalComposer } from "@lexical/react/LexicalComposer";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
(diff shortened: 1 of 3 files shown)


apps/designer/app/canvas/features/text-editor/text-editor.tsx
```diff
@@ -39,6 +49,61 @@ const AutofocusPlugin = () => {
...
+    // flatten pasted paragraphs
+    const removeNodeTransform = editor.registerNodeTransform(
+      RootNode,
+      (node) => {
+        // merge paragraphs into first with line breaks between
+        if (node.getChildrenSize() > 1) {
+          const children = node.getChildren();
+          let first;
+          for (let i = 0; i < children.length; i += 1) {
+            const p = children[i];
```
kof, line 81: please use full names: paragraph, index

kof: it's readable here, but it won't be somewhere else, so lets set a standard

kof: also please add an eslint plugin for this https://github.com/webstudio-is/webstudio-designer/issues/397

apps/designer/app/canvas/features/text-editor/text-editor.tsx
```diff
@@ -54,20 +119,24 @@ export const TextEditor = ({
...
+  const [paragraphClassName] = useState(() => nanoid());
   const [italicClassName] = useState(() => nanoid());
 
-  // initially instance styles are applied to all nodes
-  // so not necessary to use layout effect
-  useEffect(() => {
+  useLayoutEffect(() => {
     const engine = createCssEngine();
+    // reset paragraph styles and make it work inside <a>
+    engine.addPlaintextRule(`
```
kof, line 128: didn't you just removed paragraphs?

kof: oh that's for our paragraph instance

kof: no its not, ok it seems like you are keeping the lexical paragraph and only replacing pagraphs when new line is hit

TrySound: Yep, one paragraph is lexical requirement. Even in plain text editor.

kof (approved):

apps/designer/app/canvas/features/text-editor/text-editor.tsx
```diff
@@ -39,6 +49,61 @@ const AutofocusPlugin = () => {
...
+                for (const child of paragraph.getChildren()) {
+                  first.append(child);
+                }
+                paragraph.remove();
+              }
+            }
+          }
+        }
+      }
+    );
```
rpominov, line 96: How does this work? How do we tell whether pasted content contains paragraphs? I would expect something like `isParagraph(node)` somewhere in this code.

TrySound: With default configuration root contains only paragraphs. Lexical converts headings to paragraphs on paste for example. So he we just check root children which are all paragraphs.

rpominov: Got it. Might be a good idea to add this as a code comment

rpominov (approved):

kof:
Ideally this complex thing would have some tests, but I don't know if that's easy enough to test, maybe we could look  how they test lexical internally and do the same thing?

kof:
@TrySound should we merge this now since we are releasing?

PR #516, opened 2022-11-25
Ability to replace SDK components

rpominov:
Whats done:
- [x] - Added getComponent, getComponentMeta (old getComponenMeta renamed to getComponentMetaProps)
- [x] - Added `registerComponents` (for now able only override existing components)
- [x] - Renamed react-sd `remix` folder to `app`
- [x] - At the `app` folder added `custom-components/Image` wrapper component, which reads assets etc, and use `webstudio/image`
- [x] `getParams`, `setParams` added at `react-sdk/app/` as they are app level

TODO:
- [x] Create "normal" `useParams` to be able to use data.params  https://github.com/webstudio-is/webstudio-designer/blob/c05f40818beaac8a811fe0a97ba49e30f01f5045/packages/react-sdk/src/app/custom-components/image.tsx#L15-L23
- [x] Reuse props meta from imported image - created issue https://github.com/webstudio-is/webstudio-designer/issues/521


packages/react-sdk/src/components/index.ts
```diff
@@ -1,15 +1,43 @@
...
+  Subscript,
+  Button,
+  Input,
+  Form,
+  Image,
+} as const;
+
+export type ComponentName = keyof typeof components;
+
+export const componentNames = Object.keys(components) as ComponentName[];
```
rpominov, line 37: This should be a function too: `export const getComponentNames`

packages/react-sdk/src/components/meta.ts
```diff
@@ -58,6 +58,6 @@ type MetaProp =
 
 type MetaProps = Record<string, MetaProp>;
 
-// Probably better instead of JSON to generate ts with `as const`
```
rpominov, line 61: Deleted this code comment by mistake. Still applies

kof (approved):
🙏

packages/react-sdk/src/app/custom-components/image.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    const loader = useMemo(() => {
+      if (asset == null) return null;
+      if (asset.location === "REMOTE")
+        return loaders.cloudflareImageLoader(params);
+      return loaders.localImageLoader();
+    }, [asset, params]);
+
+    if (asset === null || loader == null)
+      return <WebstudioImage {...props} ref={ref} />;
```
istarkov, line 40: return sdk image here

kof: Should we use full  Webstudio prefix or "WsImage" I am on one hand against abbreviations, but on the other hand its so long lol

packages/react-sdk/src/components/index.ts
istarkov: We can't add new components because of typing issues, so now doesn't matter

packages/react-sdk/src/components/meta.ts
istarkov: https://github.com/webstudio-is/webstudio-designer/issues/521

apps/designer/app/canvas/shared/use-track-selected-element.ts
```diff
@@ -91,18 +91,18 @@ export const useTrackSelectedElement = () => {
       // It's the second click in a double click.
       if (event.detail === 2) {
         const component = dataset.wsComponent as Instance["component"];
-        if (component === undefined || component in components === false) {
+        if (component === undefined /*|| component in components === false*/) {
```
rpominov, line 94: I commented out this temporarily. Forgot to tell you. Should have merge with comment

PR #518, opened 2022-11-26
Primitive tokens

kof:
https://github.com/webstudio-is/webstudio-designer/issues/494

https://user-images.githubusercontent.com/52824/205440707-e1fe79a0-a527-4810-9d76-99bba10b1745.mp4

- [x] create and save new token
- [x] load tokens
- [x] update token
- [x] delete token
- [x] undo/redo

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


TrySound (approved):

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -79,108 +104,230 @@ const TokenEditor = ({
...
+              />
               {isNew && (
                 <Button type="submit" variant="blue">
                   Create
                 </Button>
               )}
             </Flex>
           </form>
-          <PopoverHeader title={isNew ? "New Token" : name} />
+          <PopoverHeader title={"name" in token ? token.name : "New Token"} />
```
TrySound, line 193: What is seed and why it doesn't have a name? Maybe make it undefined instead of optional? Then you can use nullish coalescing.

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -79,108 +104,230 @@ const TokenEditor = ({
...
+    event?.preventDefault();
+    if (formRef.current === null) return;
+    const data = getData(formRef.current);
+    const nextErrors = validate(tokens, data, isNew);
     setErrors(nextErrors);
 
     if (nextErrors.hasErrors === false) {
-      onChangeComplete({ ...data, group, type } as DesignToken);
-      setIsOpen(false);
+      onChangeComplete({ ...token, ...data } as DesignToken);
```
TrySound, line 123: Why cast to DesignToken?

apps/designer/app/canvas/shared/design-tokens.ts
```diff
@@ -0,0 +1,20 @@
...
+import { publish } from "~/shared/pubsub";
+
+export const useInitializeDesignTokens = (designTokens: Array<DesignToken>) => {
+  const [, setDesignTokens] = useDesignTokens();
+  const ref = useRef(false);
+  if (ref.current === false) {
+    ref.current = true;
+    setDesignTokens(designTokens);
+  }
+};
```
istarkov, line 13:
```js
// called once 
useState(() => setDesignTokens(designTokens));
```
As I remember has no issue vs refs has

apps/designer/app/designer/shared/design-tokens-manager/hooks.ts
```diff
@@ -0,0 +1,14 @@
+import { useSubscribe } from "~/shared/pubsub";
+import { useDesignTokens } from "~/shared/nano-states";
+import type { DesignToken } from "@webstudio-is/design-tokens";
+
+declare module "~/shared/pubsub" {
```
istarkov, line 5: I would create a single d.ts file and extend the pubsub interface there, easy to view in a single place what pubsub supports. 
Same for example is doing svelte-kit in `app.d.ts`

(same for above `declare module`)

istarkov (approved):

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: because data is not necessary a correct token, so not sure how to handle it, either I will have to cast at getData or here, its basically serialization of the form

apps/designer/app/canvas/shared/design-tokens.ts
kof: that's a neat trick

kof: used it in the places where I used ref before

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -79,108 +104,230 @@ const TokenEditor = ({
...
+    return (
+      <TokenEditor
+        token={token}
+        isOpen={isOpen}
+        trigger={trigger}
+        onChangeComplete={(token) => {
+          const foundToken = findByName(tokens, token.name);
+          const previousTokenName = token.name;
+          const isNew =
+            foundToken === undefined || previousTokenName === undefined;
```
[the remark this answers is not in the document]
kof: I am actually advocating to always use undefined and remove the need for null checks, by using == you imply supporting both null and undefined, but that's not the case here and neither would I want any of our apis to expose interfaces with null, but rather actually try to contain them in the local usages where we forced to use null with dom/react apis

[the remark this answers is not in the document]
kof: reason name is not set is because its the `token` it references is TokenSeed, basically token is {group, type, name, value, descript}, seed is {group, type} ... group and type are predefined, the rest is user defined.

would you structure it differently?

apps/designer/app/designer/shared/design-tokens-manager/hooks.ts
kof: We thought about it and went for a distributed approach, the idea is that each "feature" or even package declares it's events and yet they are all typechecked globally, so they can't override each other

Problem with having them centrally is 2:
1. it's going to be many over time
2. it's going to have dependencies to all possible modules to define types. It would turn into a god-module that depends on everything

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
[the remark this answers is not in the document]
kof: > Why can't we force for the token to always have a defined name?

We could have make the Token to always have all the fields with empty strings, but token name has to be unique and can't be empty, so it would not match DesignToken schema if I run it over it

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: yeah undefined could be an option, doesn't have a name because the token has group and type predefined but the rest is user-defined

kof: is it bad though the way it is, using in operator? What do you see as a problem here?

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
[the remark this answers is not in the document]
kof: separated seed from token so that there is no semantic confusion

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: separated seed from token so that there is no semantic confusion an no need for these checks

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
kof: moved casting to where we get the data, so that its obvious why we need t cast, does that make sense now?

apps/designer/app/designer/shared/design-tokens-manager/hooks.ts
TrySound: Alternatively we can make isolated pubsub for each feature like each nano-state

kof: Not sure how this could look like, but if you have an idea, let's create an issue

TrySound (approved):

packages/design-tokens/README.md
```diff
@@ -0,0 +1,3 @@
+# Webstudio Project
+
+Project related functionalities.
```
rpominov, line 3: Probably need to change

packages/prisma-client/prisma/schema.prisma
```diff
@@ -68,7 +68,8 @@ model Build {
   isDev  Boolean // exctly one is true per project
   isProd Boolean // at most one is true per project (none if not published)
 
-  breakpoints Breakpoints?
+  breakpoints  Breakpoints?
+  designTokens DesignTokens?
```
rpominov, line 72: Need to make sure `designTokens` are copied to build when we publish. The code is here: https://github.com/webstudio-is/webstudio-designer/blob/fb946ab2966dfeb3ee84c2fcbd52e43f5dbfaa9b/packages/project/src/db/build.ts#L194-L266

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
```diff
@@ -1,186 +1,189 @@
...
+  >();
+  const { render: renderMenu, isOpen: isMenuOpen } = useMenu({
+    selectedIndex,
+    onSelect: setSelectedIndex,
+    onDelete: () => {
+      const { name } = tokens[selectedIndex];
+      publish({ type: "deleteToken", payload: name });
+      const updatedTokens = [...tokens];
+      deleteTokenMutable(updatedTokens, name);
+      setTokens(updatedTokens);
```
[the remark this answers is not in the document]
kof: we need a generally better api for syncing nano states between iframe and main too, I have consistently used nano state containers the way it is right now, but I agree that we need to make it auto update, so not sure if I should start now with the changes

apps/designer/app/designer/shared/design-tokens-manager/token-editor.tsx
```diff
@@ -0,0 +1,179 @@
...
+    name.push("Name is already taken");
+  }
+  if (String(data.value).trim() === "") value.push("Value is required");
+
+  return {
+    name,
+    value,
+    hasErrors: name.length !== 0 || value.length !== 0,
+  };
+};
```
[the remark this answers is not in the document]
kof: we use the DesignToken Schema for the db, but I agree it's not 100% the same, it doesn't do trim when checking the length and it doesn't check if name is unique, I will add a todo for this, because we are haven't finalized the design yet, it might all change

packages/prisma-client/prisma/schema.prisma
kof: yes, though as long as we don't go to production with this, we should be fine, right?

kof: ah, yes true, I misunderstood

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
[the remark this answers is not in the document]
rpominov: > I have consistently used nano state containers the way it is right now

Ah, I didn't notice before. I thought we do auto updates everywhere

> but I agree that we need to make it auto update, so not sure if I should start now with the changes

Yeah, not sure also

apps/designer/app/canvas/shared/design-tokens.ts
```diff
@@ -0,0 +1,17 @@
+import type { DesignToken } from "@webstudio-is/design-tokens";
+import { useEffect, useState } from "react";
+import { useDesignTokens } from "~/shared/nano-states";
+import { publish } from "~/shared/pubsub";
+
+export const useInitializeDesignTokens = (designTokens: Array<DesignToken>) => {
+  const [, setDesignTokens] = useDesignTokens();
+  // Trick to set it only once.
+  useState(() => setDesignTokens(designTokens));
```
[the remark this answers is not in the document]
rpominov: Btw. In other places where you've replaced `useEffect` with `useState` (breakpoints.ts etc.), it will run on server now, but is it ok? I think we've started to use `useEffect` specifically because we didn't want it to run on the sever as it would write into a shared global variable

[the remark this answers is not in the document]
rpominov: Sorry, misread the diff. Nowhere, my mistake.

apps/designer/app/designer/shared/design-tokens-manager/token-editor.tsx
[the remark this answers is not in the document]
kof: rewrote the entire validation

apps/designer/app/designer/shared/design-tokens-manager/design-tokens-manager.tsx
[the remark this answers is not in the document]
kof: keepping for now

apps/designer/app/designer/shared/design-tokens-manager/hooks.ts
kof: @istarkov ?

packages/project/src/db/build.ts
```diff
@@ -215,52 +231,38 @@ export async function create(
...
-  const breakpointsValues = (await db.breakpoints.load(sourceBuild.id)).values;
+  const breakpointsValues = sourceBuild
+    ? (await db.breakpoints.load(sourceBuild.id)).values
+    : db.breakpoints.createValues();
+
+  await prisma.$transaction(async (client) => {
+    const pages =
+      sourceBuild === undefined
+        ? await createPages(breakpointsValues, client)
+        : await clonePages(sourceBuild.pages);
```
rpominov, line 248: Should pass `client` into `clonePages` too?

kof:
@rpominov fixed all comments and added tokens cloning to the build

rpominov (approved):
👍

apps/designer/app/designer/shared/design-tokens-manager/hooks.ts
kof: resolving this conversation for now, lets have a separate issue or a thread

PR #519, opened 2022-11-26
Deploy per-PR Storybooks on Chromatic

Andarist:
This publishes all available Storybooks to Chromatic - this happens per push so you will always be able to check out the current state of things in those Storybooks.

Since you have a lot of Storybooks... this isn't quite convenient. I plan to combine all of those Storybooks into a single one (a single storybook that will reference all of those smaller ones). I'd like to do that in a follow-up PR though - let's now focus on getting this up and running.


.github/workflows/publish.yml
```diff
@@ -1,24 +1,27 @@
...
-          cache: "yarn"
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - uses: ./.github/actions/ci-setup
+
       - run: yarn
       - run: yarn turbo run build
       - run: yarn turbo run publish-to-npm
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
Andarist, line 27: In a way, this reverts the change from [this commit](https://github.com/webstudio-is/webstudio-designer/commit/44e5c8bc77c6b8dd60dec7bc2d7c8f099cbdac0a). I think that this should be fine but @TrySound has changed this for a reason - so perhaps there is something that I'm missing here. 

I'd like to avoid passing this token to the `ci-setup` step as that would make this token available to all workflows. From the security PoV it's better to scope this as much as possible - and we can't scope it much further than a single step :p

.github/workflows/publish.yml
```diff
@@ -1,24 +1,27 @@
 name: Publish packages on NPM 📦
+
 on:
   push:
     branches:
       - main
+
 jobs:
   publish:
+    timeout-minutes: 20
```
Andarist, line 10: I added this to all workflows - I had some rogue actions in the past that were running for a couple of hours, it's IMHO better to put some reasonable limit upfront and fail the action sooner than in a couple of hours :P

.github/actions/ci-setup/action.yml
```diff
@@ -0,0 +1,15 @@
+name: CI setup
```
Andarist, line 1: Using composite action I've extracted the common setup+install steps into a single block. You already had some small differences in those steps within your workflows.

.github/workflows/main.yml
```diff
@@ -1,27 +1,22 @@
...
-      - name: Setup Node.js ${{ matrix.node }}
-        uses: actions/setup-node@v2
-        with:
-          node-version: ${{ matrix.node }}
-          cache: "yarn"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/ci-setup
```
Andarist, line 18: Even though I was able to move setup+install to a single `ci-setup` action, we still need to have checkout as a separate step. This can't be put in the `ci-setup` because `ci-setup` is local, it's in this repo - and the CI runner starts without any repo files downloaded. So we first need to fetch them (with `actions/checkout`) and only then we can start using our local `ci-setup` action

.github/workflows/publish.yml
Andarist: A notable difference between what I have here and what you had there before is that I've removed `registry-url` input from `actions/setup-node`. I don't think it's needed as that's the default registry anyway. And since I don't use this input the action itself no longer creates `.npmrc` as [this whole thing](https://github.com/actions/setup-node/blob/41acaa2e85a9c20b0bcaff1397645d5f33cab6ac/src/authutil.ts) gets skipped (this `.npmrc` shouldn't be required here anyway).

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        project:
+          [designer-app, css-engine, design-system, icons, image, react-sdk]
+        include:
+          - project: designer-app
+            chromaticToken: ${{ secrets.CHROMATIC_DESIGNER_APP_TOKEN }}
```
Andarist, line 17: Lol, GitHub is complaining about this line... it seems that perhaps secrets cant be used in the matrix. I need to investigate this.

Andarist: Ok, I think that I've solved this here: https://github.com/webstudio-is/webstudio-designer/pull/519/commits/8c5ad32c74e9b954f13feccafe796cd66508d949

.github/actions/ci-setup/action.yml
```diff
@@ -0,0 +1,15 @@
...
+  steps:
+    - name: Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16
+        cache: "yarn"
+
+    - name: Install Dependencies
+      run: yarn install --frozen-lockfile --ignore-scripts
+      shell: bash
```
TrySound, line 15: Isn't bash default? Probably no need to specify

.github/actions/ci-setup/action.yml
TrySound: What's the point of reusing a few lines of code? You don't even need some of those lines. This way log will be good enough

```
    - uses: actions/setup-node@v3
      with:
        node-version: 16
        cache: "yarn"
    - run: yarn install --frozen-lockfile --ignore-scripts
```

.github/workflows/publish.yml
TrySound: All is needed. I tried all possible combinations. registry-url is needed and token no setup-node is needed.

kof (approved):
I have no idea what's happening 🤣

kof:
Btw is this waiting for us?

.github/workflows/publish.yml
Andarist: This still baffles me a little - I maintain Changesets, a tool responsible for versioning and publishing packages. To the best of my knowledge, what I have in this PR should work. However, there are some nuances, and it's easy to miss things. It's also not the part of the codebase that I often revisit/touch so it's not like I exactly know all of the details.

I'll try to do a few test runs in a separate kitchen sink project to test it out and get a better feeling of the problem.

.github/actions/ci-setup/action.yml
Andarist: Maybe something has changed but I had to add it explicitly in the past:
https://github.com/statelyai/xstate/compare/b8b4a79a7e53b2f79b102f9f12de16a72b728f02..f5485bdea0c4d9d257fd04c0fe4907720eb7e8af

.github/actions/ci-setup/action.yml
Andarist: I don't feel super strong about this. It's a strategy I have used in the past in some other repos and I just used it here as well, since you already had different actions doing this thing slightly differently. So to keep things consistent, I've moved this to shared action.

bin/publish-to-npm.sh
```diff
@@ -9,5 +9,5 @@ echo Current $npm_package_name $NPM_VERSION
 
 if [[ $PKG_VERSION != $NPM_VERSION ]]; then
   echo Publishing $npm_package_name $PKG_VERSION
-  npm publish --access public --registry 'https://registry.npmjs.org/'
+  npm publish --access public
```
Andarist, line 12: @TrySound i've copy-pasted all of the publish-related setups from this branch into my playground repo and it published packages successfully from the CI, just got the email from the npm:
> Successfully published andarist-changeset-action-test-a@0.0.3

I'd say that we can roll with the changes in this PR and if this ends up creating any problems, then I could investigate this further. WDYT?

Andarist: Oh, sorry - I've misinterpreted some results. Still rechecking what's wrong here.

Andarist: Ok, I've fixed the problem. It seems that `.npmrc` is the easiest way to make this work. I've always believed that it can be done solely based on the env vars but I've failed to do so right now. So I just ended up creating this manually before calling the publish script

TrySound (approved):

PR #522, opened 2022-11-26
Use serialized loader types

TrySound:
New remix enforces types to be serialized.
Prisma output dates as Date instance and this is not customizable. Had to fix only 2 cases.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [x] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/src/utils/format-asset.ts
+++ b/packages/asset-uploader/src/utils/format-asset.ts
@@ -19,6 +19,7 @@ export const formatAsset = (asset: DbAsset): Asset => {
   if (isFont) {
     return {
       ...base,
+      createdAt: base.createdAt.toISOString(),
       format: asset.format as FontFormat,
       meta: FontMeta.parse(JSON.parse(asset.meta)),
     };
@@ -26,6 +27,7 @@ export const formatAsset = (asset: DbAsset): Asset => {
 
   return {
     ...base,
+    createdAt: base.createdAt.toISOString(),
     meta: ImageMeta.parse(JSON.parse(asset.meta)),
   };
 };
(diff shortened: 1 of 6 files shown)


packages/asset-uploader/src/utils/format-asset.ts
```diff
@@ -19,13 +19,15 @@ export const formatAsset = (asset: DbAsset): Asset => {
   if (isFont) {
     return {
       ...base,
+      createdAt: base.createdAt.toISOString(),
```
kof, line 22: is this really the only way we can deal with dates? its going to be everywhere? any remix discussion on this?

kof (approved):

packages/asset-uploader/src/utils/format-asset.ts
TrySound: It's json. Remix just added proper type checking. I'd say it's on prisma to fix but I saw in issue they are not gonna change this. So I guess we either need layer between loaders and prisma or add these hacks.

kof: yeah, the place where we convert Date to json is the place where we have to do it and that's basically loaders

TrySound: The issue is that we reuse code for server and client. So I did conversion earlier to make all parts work.

istarkov: https://github.com/Rich-Harris/devalue

sveltekit switched on this

istarkov (approved):

PR #524, opened 2022-11-27
Fix accidentally commented code at #516

istarkov:
https://github.com/webstudio-is/webstudio-designer/pull/516#pullrequestreview-1194934498


packages/react-sdk/src/components/index.ts
```diff
@@ -45,7 +45,16 @@ let registeredComponents: RegisteredComponents | null = null;
 
 export type ComponentName = keyof typeof components;
 
-export const componentNames = Object.keys(components) as ComponentName[];
+const componentNames = Object.keys(components) as ComponentName[];
+
+export const getComponentNames = (): ComponentName[] => {
```
istarkov, line 50: I think we should use "use" instead of "get" as we assume that they are called inside React components
after init cc @kof

rpominov (approved):

PR #529, opened 2022-11-27
Add a composed Storybook

Andarist:
This PR builds on top of the work done in https://github.com/webstudio-is/webstudio-designer/pull/519 and introduces a root-level storybook that just aggregates all of the per-package storybooks into one. 

This allows you to have a single, convenient, URL to visit and browse all of those existing storybooks at once. To access this storybook you need to click on the `Details` link attached to the "Storybook Publish: !webstudio-is/webstudio-designer" status check.

The current URL from this branch is available [here](https://638affb61acca1e593c6e558-dksaophpxa.chromatic.com/) so you can take a look at how this looks in practice.


kof:
@Andarist @TrySound yes, ideally one config to run them all, its truly too much configs right now for creating a package

Andarist:
Dunno, I don't feel like this amount of config is too much - and you don't add new packages on the daily basis. I was also planning to create a doc with all steps required to add configs related to this in your GitHub actions etc.

It's definitely possible to have a single Storybook for the whole monorepo but this raises some questions - like where do you keep those stories, what if you need to customize something for one of them? My philosophy is that a monorepo is an implementation detail and I try to isolate packages as much as I can - there is some glue on the top to make it work nicely in tandem but on principle I like when I can just pull a directory out of a monorepo into a separate project and have it set up there in minutes (with minimal changes).

For me, it's also about traceability - when you split things into packages properly then you end up with much nicer logs/reports on the CI. And it's much easier to know in what area something has changed. 

There are certainly tradeoffs here to each solution but, overall, I think that this is a **better** and more future-proof setup. Is it more complex because of that? Definitely, can't argue with that. I don't think though that it's that complex. 

I think that what Turbo (and other projects like that) teaches us - is that there is a good value in having **boundaries**. They allow for more things - each unit becomes configurable on its own and it becomes cacheable. At this size of the project, I don't think the caching is that big of a problem, especially since this is an OSS project so you have unlimited GitHub Action minutes available to you. I very much appreciate having those boundaries in place though.

If you prefer to merge those things into one - so be it, although that should probably be done in a separate PR. I don't have to agree with the decision to work on its implementation 😉

kof:
can't we have a central config that we import from the local one and reexport? like we do with jest, so if we need for some reason something custom for one package - we can, but the baseline is managed from a central config

Andarist:
Yes, that can be done - I got the impression though that you both were asking about a different thing. Just like in the case of Jest - the config can be shared but also just like in the case of Jest, I personally prefer having a separate config for each package (even if it just uses the shared config).

Moving everything into a single Storybook (with a truly single config) would be like removing all jest configs in favor of one, central, one.

kof:
@Andarist we just want to simplify creating and maintaining packages

kof:
We need a guideline somewhere on how to add a package with all this stuff lol. I would be already lost tbh. But besides of that it seems we are on the same page with keeping the config like with jest centrally and only reexporting it from the package, right?

Andarist:
> We need a guideline somewhere on how to add a package with all this stuff lol. I would be already lost tbh. 

Yeah, I'll add guidelines around Storybook - I just want to land this PR first, so I could use URLs in this doc that would refer to non-detached commits and stuff (I'm using squash & merge on GitHub).

> But besides of that it seems we are on the same page with keeping the config like with jest centrally and only reexporting it from the package, right?

I think so. I got the impression that Bogdan wants to move everything into a single Storybook though. Since this PR is approved - I'll take the liberty to land this. If any of you have any doubts or thoughts though - let's keep discussing this. I want to make sure that you are satisfied with the prepared setup,

TrySound:
> I got the impression that Bogdan wants to move everything into a single Storybook though. 

I do. Caz it's big overhead for project setup to install all this beside configuration.
image

kof:
> I do. Caz it's big overhead for project setup to install all this beside configuration.

this is all going to be part of a single package, so we won't have to add them every time to each new package manually

PR #532, opened 2022-11-27
Replace flags with component type

TrySound:
Boolean flags may have a lot of combinations which can lead to unexpected bugs. Better approach is to categorize nodes with specific behaviours.

Here I added 5 types

- container - can accept other components with dnd
- control - usually form controls like inputs, without children
- embed - images, videos or other embeddable components, without children
- rich-text - can be edited on double click
- rich-text-child - formatted text fragment, not listed in components list

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/shared/tree/index.tsx
+++ b/apps/designer/app/designer/shared/tree/index.tsx
@@ -15,20 +15,24 @@ const instanceRelatedProps = {
   getItemPath: utils.tree.getInstancePath,
   getItemPathWithPositions: utils.tree.getInstancePathWithPositions,
   canLeaveParent(item: Instance) {
-    return getComponentMeta(item.component).isInlineOnly !== true;
+    const { type } = getComponentMeta(item.component);
+    return (
+      type === "container" ||
+      type === "control" ||
+      type === "embed" ||
+      type === "rich-text"
+    );
   },
   canAcceptChild(item: Instance) {
-    return getComponentMeta(item.component).canAcceptChildren;
+    const { type } = getComponentMeta(item.component);
+    return type === "container";
   },
   getItemChildren(item: Instance) {
-    const component = getComponentMeta(item.component);
+    const { type } = getComponentMeta(item.component);
 
     // We want to avoid calling .filter() unnecessarily, because this is a hot path for performance.
     // We rely on the fact that only content editable or inline components may have `string` children.
-    if (
-      component.isContentEditable === false &&
-      component.isInlineOnly === false
-    ) {
+    if (type === "container" || type === "control" || type === "embed") {
       return item.children as Instance[];
     }
 
--- a/apps/designer/app/designer/shared/tree/index.tsx
+++ b/apps/designer/app/designer/shared/tree/index.tsx
@@ -16,12 +16,7 @@ const instanceRelatedProps = {
   getItemPathWithPositions: utils.tree.getInstancePathWithPositions,
   canLeaveParent(item: Instance) {
     const { type } = getComponentMeta(item.component);
-    return (
-      type === "container" ||
-      type === "control" ||
-      type === "embed" ||
-      type === "rich-text"
-    );
+    return type !== "rich-text-child";
   },
   canAcceptChild(item: Instance) {
     const { type } = getComponentMeta(item.component);
@@ -31,7 +26,7 @@ const instanceRelatedProps = {
     const { type } = getComponentMeta(item.component);
 
     // We want to avoid calling .filter() unnecessarily, because this is a hot path for performance.
-    // We rely on the fact that only content editable or inline components may have `string` children.
+    // We rely on the fact that only rich-text or rich-text-child components may have `string` children.
     if (type === "container" || type === "control" || type === "embed") {
       return item.children as Instance[];
     }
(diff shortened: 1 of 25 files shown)


packages/react-sdk/src/components/component-type.ts
```diff
@@ -1,58 +1,30 @@
...
-        val.children === undefined ||
-        val.children.every((child) => typeof child !== "string"),
-      {
-        message:
-          "Components that can accept children via drag&drop are not allowed to have `string` children",
-        path: ["children"],
-      }
-    )
+  z.object({
+    type: z.enum([
```
istarkov, line 17: plz add here what you write in pr comment

istarkov (approved):

apps/designer/app/designer/shared/tree/index.tsx
```diff
@@ -15,20 +15,24 @@ const instanceRelatedProps = {
   getItemPath: utils.tree.getInstancePath,
   getItemPathWithPositions: utils.tree.getInstancePathWithPositions,
   canLeaveParent(item: Instance) {
-    return getComponentMeta(item.component).isInlineOnly !== true;
+    const { type } = getComponentMeta(item.component);
+    return (
+      type === "container" ||
+      type === "control" ||
+      type === "embed" ||
+      type === "rich-text"
```
kof, line 23: isn't this basically type !== 'rich-text-child'?

packages/project/src/shared/tree-utils/find-closest-non-inline-parent.ts
```diff
@@ -7,7 +7,13 @@ export const findClosestNonInlineParent = (
...
-  return path.find(
-    (item) => getComponentMeta(item.component).isInlineOnly === false
-  );
+  return path.find((item) => {
+    const type = getComponentMeta(item.component).type;
+    return (
+      type === "container" ||
+      type === "control" ||
+      type === "embed" ||
+      type === "rich-text"
```
kof, line 16: same here

kof (approved):

apps/designer/app/designer/shared/tree/index.tsx
TrySound: Yes, but when/if new type is added we will need to catch all such cases. It's also easier to understand when behaviour is applied rather than when it's not.

kof: true, though I still wonder if it would be better to simplify this, maybe by adding a utility function? just wondering if its gonna be easy to overlook one usage, especially when there are multiple of them, once we need to add another case

packages/project/src/shared/tree-utils/find-closest-non-inline-parent.ts
rpominov: I think Oleg is right. We call `findClosestNonInlineParent` under this condition:

```
if (getComponentMeta(instance.component).type === "rich-text-child")
```

I think it would be good to have the opposite condition inside the function. Otherwise we might get some confusing behaviour when a new type is added if we forget to add it here.

apps/designer/app/designer/shared/tree/index.tsx
```diff
@@ -15,20 +15,24 @@ const instanceRelatedProps = {
...
-    const component = getComponentMeta(item.component);
+    const { type } = getComponentMeta(item.component);
 
     // We want to avoid calling .filter() unnecessarily, because this is a hot path for performance.
     // We rely on the fact that only content editable or inline components may have `string` children.
-    if (
-      component.isContentEditable === false &&
-      component.isInlineOnly === false
-    ) {
+    if (type === "container" || type === "control" || type === "embed") {
```
rpominov, line 35: Need to update code comment to say: "... only rich-text or rich-text-child components may have `string` children."

Or maybe "We rely on the fact that components of the following types cannot have `string` children." (would be more inline with the condition in `if`)

apps/designer/app/designer/shared/tree/index.tsx
rpominov: So, when a new type is added, unless this code is updated it won't be able to live parent by default. Not sure, but I think default should be opposite.

Also, whatever we decide to keep here, it should match what we do here: https://github.com/webstudio-is/webstudio-designer/blob/9ed4d4a18bad75c92d71315c5106b13127eefc96/apps/designer/app/canvas/shared/use-drag-drop.ts#L181-L182

packages/react-sdk/src/components/component-type.ts
```diff
@@ -1,58 +1,37 @@
...
-      (val) =>
-        val.canAcceptChildren === false ||
-        val.children === undefined ||
-        val.children.every((child) => typeof child !== "string"),
-      {
-        message:
-          "Components that can accept children via drag&drop are not allowed to have `string` children",
-        path: ["children"],
-      }
-    )
```
rpominov, line 57: Not super important, but maybe we should keep this one

TrySound: We should probably change it to string later and make it work only for rich-text

PR #539, opened 2022-11-28
Use sources as type deifnitions

TrySound:
Right now our build and CI take a lot of time. 1:40min - 2min on CI and 1min locally. We spend time on vercel, on main CI, on publish and locally.

At the moment in packages anything outside of src folder is not checked because of rootDir.

We have to generate "references" field on precommit.

All these issues can be avoided if remove .d.ts generation completely and specify sources in "types" field. This way we get zero build time for types and can switch to other build tool (like esbuild) for .js generation and get almost instant build on vercel, on main CI and on publish.

Downsides

- it can be slower (actually not really caz we delete .tsbuildinfo cache anyway and we can still generate it with incremental flag)
- it contraditcs recommended practices for published packages (we can generate .d.ts in publish script only and use "types" as source internally)

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


kof:
I have seen a few more downsides in that article, are you aware of them?

https://moonrepo.dev/docs/guides/javascript/typescript-project-refs#what-about-not-using-project-references-and-only-using-source-files

kof:
To me 2 min is not such a huge problem right now, I am more concerned with the complexity and amount of problems/breaking things and having to sink dev time into understanding some complex scenario that broke.

If I understand correctly the way we have it now is the most straight forward way for typescript as it treats our internal packages same way it treats every other installed package and it's the recommended and supported way from typescript team.

I would personally want to revisit this once we actually have serious perf issues that we need to fix.

TrySound:
Yep, I saw this article

> Loading declaration files are much faster than source files.

Not much, we still delete ts cache and do check of sources in editor

> You'll lose all the benefits of TypeScript's incremental caching and compilation. TypeScript will consistently load, parse, and evaluate these source files every time. This is especially true for CI environments.

Again, we delete cache

> When using package.json workspaces, bundlers and other tools may consider these source files "external" as they're found in node_modules. This will require custom configuration to allow it.

typescript is the only tool using "types", works perfectly

> It breaks consistency. Consistency with the npm ecosystem, and consistency with how packaging and TypeScript was designed to work. If all packages are internal, then great, but if you have some packages that are published, you now have 2 distinct patterns for "using packages" instead of 1.

We can setup .d.ts generation specifially for published package, but doesn't matter until before we go beta, we probably won't expose much to user. Only types and cli may be enough.

kof:
We delete cache only when we call build, this is a bit of a separate story, improving the `build` command performance. We rarely have to call build, most of the time we just use `yarn dev` - this one has to be really fast

TrySound:
With this references option we still consume sources, not .d.ts files. They are not regenerated. And tsserver does not "consistently load, parse, and evaluate these source files every time". It's true only for build.

TrySound:
My indention to make build as fast as dev.

kof (approved):

istarkov (approved):
hope it will be better

PR #541, opened 2022-11-28
Move defaults into ws

istarkov:
meta props are moved to *.ws.ts
Component removed from *.ws.ts

This allows to merge ws props with registered components props
`registerComponents` now are 
-   registerComponents(customComponents);
-   registerComponentsMeta(customComponentsMeta);

_(Splitting will allow for code splitting to work.)_

Removal of `Component` (and then `Icon` prop) would allow serializing *.ws.ts metadata between iframe and app designer

Next PR:

Move Icon into base64 or like.

publish/subscribe ws meta of registered components, so it can be reused on the Designer side
I need this to pass props of registered components to Designer.


packages/react-sdk/src/components/index.ts
```diff
@@ -58,34 +96,24 @@ export const getComponentNames = (): ComponentName[] => {
...
   return registeredComponents != null && name in registeredComponents
-    ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      (registeredComponents[name]!
-        .Component as typeof components[Name]["Component"])
-    : components[name].Component;
+    ? (registeredComponents[name] as typeof components[ComponentName])
+    : components[name];
+};
+
+export const getComponentMetaProps = (name: ComponentName): MetaProps => {
```
istarkov, line 115: Better to merge all props if defaultValue non null then prop will be in result. Having that we mostly overwrite components.

packages/image/src/image.tsx
```diff
@@ -25,14 +35,19 @@ export const Image = forwardRef<ElementRef<typeof defaultTag>, ImageProps>(
     }) ?? { src: imagePlaceholderSvg };
 
     return (
-      <img {...imageProps} {...imageAttributes} decoding="async" ref={ref} />
+      <img
+        {...imageProps}
+        {...imageAttributes}
+        decoding={decoding}
+        loading={loading}
```
kof, line 42: shoulnd't these be overwritable by props?

istarkov: its props just with defaults ;) so will be overrided

packages/react-sdk/src/components/index.ts
```diff
@@ -96,3 +151,9 @@ export const getComponent = <Name extends ComponentName>(
 export const registerComponents = (components: RegisteredComponents) => {
   registeredComponents = components;
 };
+
+export const registerComponentsMeta = (
+  componentsMeta: RegisteredComponentsMeta
+) => {
+  registeredComponentsMeta = componentsMeta;
+};
```
kof, line 159: registry implementation should probably be a separate module, so that there is no mixture of definition and registry itself. In the end registry will be completely reusable for all sdk's, but definitions code is specific to each sdk.

kof (approved):

Issue #92, opened 2022-06-27
Session maintained when user no longer exists 


SaraVieira:
When a user is deleted or we drop the database if we are still logged in their session stays and we get an error for a Postgres constrain since it's trying to create a project with a user id that does not exist 

To reproduce 
1. login
2. drop database
3. go to dashboard, create project

Possible solution

When loading up the dashboard check if the user is logged in and if the user exists, if the user does not exist logout the user and redirect to the login page

kof:
Done?

PR #542, opened 2022-11-28
Check authenticated user against db

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/92

We store user session in storage and checked its existance to check if user is logged in. Though where we clear db or for example delete user this session becomes stale so we should check if user exists in db too.

## Before requesting a review

- [x] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/routes/dashboard/index.tsx
+++ b/apps/designer/app/routes/dashboard/index.tsx
@@ -8,7 +8,7 @@ import {
 import { Dashboard, links } from "~/dashboard";
 import { db } from "@webstudio-is/project/server";
 import { ensureUserCookie } from "~/shared/session";
-import { authenticator } from "~/services/auth.server";
+import { findAuthenticatedUser } from "~/services/auth.server";
 import { zfd } from "zod-form-data";
 import { type Project, User as DbUser } from "@webstudio-is/prisma-client";
 import { designerPath, loginPath } from "~/shared/router-utils";
@@ -31,7 +31,7 @@ export const action: ActionFunction = async ({ request }) => {
   const { project: title } = schema.parse(await request.formData());
 
   const { userId, headers } = await ensureUserCookie(request);
-  const authenticatedUser = await authenticator.isAuthenticated(request);
+  const authenticatedUser = await findAuthenticatedUser(request);
   try {
     const project = await db.project.create({
       title,
@@ -47,7 +47,7 @@ export const action: ActionFunction = async ({ request }) => {
 };
 
 export const loader: LoaderFunction = async ({ request }) => {
-  const user = await authenticator.isAuthenticated(request);
+  const user = await findAuthenticatedUser(request);
   if (!user) {
     return redirect(loginPath({}));
   }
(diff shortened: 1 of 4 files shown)


rpominov (approved):

kof (approved):

apps/designer/app/routes/dashboard/index.tsx
```diff
@@ -31,7 +31,7 @@ export const action: ActionFunction = async ({ request }) => {
   const { project: title } = schema.parse(await request.formData());
 
   const { userId, headers } = await ensureUserCookie(request);
-  const authenticatedUser = await authenticator.isAuthenticated(request);
+  const authenticatedUser = await findAuthenticatedUser(request);
```
istarkov, line 34: And later even if user is null we create project anyway?

apps/designer/app/routes/rest/project/clone.$domain.ts
```diff
@@ -30,7 +30,7 @@ const ensureProject = async ({
  */
 export const loader: LoaderFunction = async ({ request, params }) => {
   if (params.domain === undefined) return { errors: "Domain required" };
-  const user = await authenticator.isAuthenticated(request);
+  const user = await findAuthenticatedUser(request);
```
istarkov, line 33: Later same strange logic about `generatedUserId`

apps/designer/app/routes/dashboard/index.tsx
TrySound: The action will fail with db constraint. Though it will not be called this way caz 
 dashboard redirects to login page.

apps/designer/app/routes/rest/project/clone.$domain.ts
TrySound: I guess same, db fails

apps/designer/app/routes/dashboard/index.tsx
istarkov: if db fails then types are wrong

apps/designer/app/routes/rest/project/clone.$domain.ts
rpominov: This might be to make this work:

At https://webstudio.is/ you can click "try", and if you're not authenticated, you get authenticated automatically with a new user created.

kof: We could actually stop generating users. This was done initially because we didn't have login functionality. We can expect them to signup now with google or github. I think we should remove the auto-generated users logic entirely. Also it is:
1. confusing once you had created a site under a demo user but you never saw the dashboard, so you don't know you can go bak there to actually login and save the sites
A better flow would be to let user signup first, then show them the dashboard, allow them to start from a template (/clone) or a blank project
2. it may be dangerous to keep generating projects without signup when we launch public beta, could be bots or just a wave of users generating too many of them

kof: Lets wait though until we get the new design for dashboard with  rewriting this logic, so we can do the new flow completely.

kof:
@TrySound is this blocked?

TrySound:
> Lets wait though until we get the new design for dashboard with rewriting this logic, so we can do the new flow completely.

So you want this to be merged now? I was confused

kof:
Sorry, I wasn't clear: I meant rewriting the entire flow logic should be postponed until we work on the new dashboard, but fixing the bug can be merged now

PR #543, opened 2022-11-28
Replace tsc with esbuild to build packages

TrySound:
Build of whole monorepo without cache takes 1min or more on my laptop. tsc is very slow tool for this. Also having rootDir in tsconfig prevents checking outside of src folder.

Here I added "scripts" package with "build-package" bin. It does both esm and cjs builds. --watch flag support completely replaces tsup for us.

Now full rebuild of monorepo without cache takes only 10s including prisma-client, http-client and designer builds.

## Before requesting a review

- [x] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/scripts/build-package.ts
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env -S node --experimental-specifier-resolution=node --loader=tsx
```
kof: we have been running bin scripts always with tsx so far using tsx from scripts, I would love a consistency

packages/scripts/build-package.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const entryPoints: string[] = [];
+const assets: string[] = [];
+
+await totalist("./src", (rel) => {
+  if (rel.endsWith(".ts") || rel.endsWith(".tsx")) {
+    if (rel.includes(".test.") || rel.includes(".stories.")) {
+      return;
+    }
+    entryPoints.push(join("src", rel));
```
Andarist, line 19: out of curiosity - why not flat-bundling? to cover for `/server`?

TrySound: Bundling breaks code splitting. All bundlers do it based on modules.

packages/scripts/build-package.ts
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env tsx
```
kof, line 1: lets do it this way everywhere then?

kof: all bin/*

Andarist (approved):

apps/designer/package.json
```diff
@@ -6,7 +6,7 @@
     "build": "remix build",
     "dev": "remix dev",
     "start": "remix-serve build",
-    "typecheck": "tsc -b",
+    "typecheck": "tsc",
```
Andarist, line 9: nice, I've noticed that this was a leftover here but had no idea that it could impact the watch behaviour like this

kof (approved):

istarkov (approved):

PR #545, opened 2022-11-29
Drop lost-pixel dependency

TrySound:
We use lost-pixel only as github action. Having it in dependenies adds ~270MB in node_modules and breaks vercel CI. With this change we can remove hack around odiff.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -28,7 +28,6 @@
     "@storybook/testing-library": "^0.0.13",
     "@webstudio-is/scripts": "*",
     "@webstudio-is/jest-config": "*",
-    "lost-pixel": "^2.23.0",
     "typescript": "4.7.4"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


Andarist (approved):

kof:
Just for understanding, how is ci able to run it if its nowhere in dependencies?

kof:
cc @d-ivashchuk

kof (approved):

d-ivashchuk:
> Just for understanding, how is ci able to run it if it's nowhere in dependencies?

this is easy, we build the image that runs everything, and dependency is needed only for ts types in the config. Definitely should not add 270MB 😆

d-ivashchuk (approved):

Andarist:
> Just for understanding, how is ci able to run it if its nowhere in dependencies?

GitHub Actions are "self-contained". They can contain everything that they need to execute - as the whole actions is being downloaded before it's executed. So as long as that action contains this tool, you don't actually need to depend on it directly

PR #549, opened 2022-11-30
Fix that images with asset had value inside db

istarkov:
ref #196

- [x] UserProps now [discriminated union by type](https://github.com/webstudio-is/webstudio-designer/blob/img_asset/packages/react-sdk/src/user-props/schema.ts#L10-L31)

- [x] Data migrations done


packages/react-sdk/src/user-props/schema.ts
```diff
@@ -1,26 +1,46 @@
...
+    value: z.union([z.string(), z.boolean(), z.number()]),
+    required: z.optional(z.boolean()),
+  }),
+  z.object({
+    id: z.string(),
+    prop: z.string(),
+    asset: Asset,
+    required: z.optional(z.boolean()),
+  }),
+]);
```
rpominov, line 43: Why we should move Zod definition of data structures that defined in SDK and going to be part of SDKs public API?

But I agree something like `UserDbProp` should be moved

apps/designer/app/designer/features/props-panel/control.tsx
```diff
@@ -174,13 +174,15 @@ type ControlProps = {
   component: Instance["component"];
   userProp: UserProp;
   onChangePropValue: (value: string | boolean, asset?: Asset) => void;
+  onChangePropAsset: (asset: Asset) => void;
```
kof, line 177: why not accepting Asset as a value in onChangePropValue? I just noticed that onChangePropValue also got a second param asset, which is not a good design

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -127,9 +125,39 @@ export const usePropsLogic = ({
...
+
+    const updatedProps = nextUserProps.filter((item) => item.id === id);
+
+    updateProps(updatedProps);
+  };
+
+  const handleChangePropValue: HandleChangePropValue = (id, value) => {
+    const nextUserProps = produce((draft: Array<UserProp>) => {
+      const index = draft.findIndex((item) => item.id === id);
+      const val = draft[index];
```
kof, line 148: instead of using abbreviations just to not have a namespace conflict, we can use a more specific names, e.g. 
```suggestion
      const draftValue = draft[index];
```

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -161,6 +194,8 @@ export const usePropsLogic = ({
     updateProps(updatedProps);
   };
 
+  // HandleChangePropAsset
```
kof, line 197: this is not useful :)

packages/image/src/image.tsx
```diff
@@ -22,7 +22,7 @@ export const Image = forwardRef<ElementRef<typeof defaultTag>, ImageProps>(
     ref
   ) => {
     // Temporary set to false, to support previous image behaviour
-    const DEFAULT_OPTIMIZE = false;
+    const DEFAULT_OPTIMIZE = true;
```
kof, line 25: What does this do? maybe a better inline comment for "why"

packages/project/src/db/props.ts
```diff
@@ -49,23 +50,23 @@ export const loadByTreeId = async (treeId: Tree["id"]) => {
...
+      if ("assetId" in dbProp) {
+        const { assetId, ...userProp } = dbProp;
+        const asset = assetsMap.get(assetId);
+
+        if (asset) {
+          userProps.push({
+            ...userProp,
+            asset,
+          });
+        } else {
```
kof, line 62: try to avoid using `else`, they always increase indentation and usually making it harder to read for someone who got used to early returns style.

In this case we can just use `continue;` before

packages/react-sdk/src/app/custom-components/image.tsx
```diff
@@ -28,12 +28,17 @@ export const Image = forwardRef<ElementRef<typeof defaultTag>, Props>(
...
+    let src = props.src;
 
-    return (
-      <WebstudioImage {...props} loader={loader} optimize={true} ref={ref} />
-    );
+    if (asset != null) {
+      src = asset.path;
+    }
+
+    if (asset === null || loader == null) {
```
kof, line 37: shouldn't loader be either undefined or defined? why do we need to support null for it? I would just make loader optional and that's it, trying to not have to support both null and undefined as much as possible.

We would mostly only need null when using dom API and we can keep those usages local and not expose it to consumers

packages/react-sdk/src/user-props/schema.ts
```diff
@@ -1,26 +1,46 @@
...
 
-const UserPropDb = z.object({
-  id: z.string(),
-  prop: z.string(),
-  value: z.union([z.string(), z.boolean(), z.number()]),
-  assetId: z.optional(z.string()),
-  required: z.optional(z.boolean()),
-});
-export type UserPropDb = z.infer<typeof UserPropDb>;
+// @todo move zod out of sdk
```
kof, line 4: btw what was the reasoning for moving zod out of SDK?

kof: So far my realization was that zod schema's should stay in the place where we fetch the data and potentially transform/stringify/parse it. Basically keep zod at the edge where we receive the data

kof (changes requested):
left some comments, looking good though, mostly about aligning the code style

packages/react-sdk/src/user-props/use-user-props.ts
```diff
@@ -15,10 +16,15 @@ export const useUserProps = (instanceId: Instance["id"]) => {
...
-    }, {} as UserProps);
+    const result: UserProps = {};
+
+    for (const userProp of propsData.props) {
+      if ("value" in userProp) {
+        result[userProp.prop] = userProp.value;
+      }
+    }
+
+    return result;
```
rpominov, line 27: Probably a proper way to fix `useUserProps` is to make it support props that refer to an asset too, rather than ignoring them

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -127,9 +125,39 @@ export const usePropsLogic = ({
...
+        return;
       }
+
+      draft[index] = {
+        id: val.id,
+        prop: val.prop,
+        required: val.required,
+        value,
+      };
     })(userProps);
```
rpominov, line 161: I'd move logic like this into `/shared` somewhere. Similar to this https://github.com/webstudio-is/webstudio-designer/blob/55280b8348a8a12cdabdd15eced6fa0aa2223a0c/apps/designer/app/shared/props-utils/delete-prop.ts#L3-L13

rpominov (commented):
In general assets support in props is added kinda awkwardly? Couldn't it be just another value type?

In DB: `value: string | boolean | number | {assetId: string}`
In SDK `value: string | boolean | number | Asset` or ` ... number | {asset: Asset}`

rpominov:
Do we need to migrate old data in the database to the new slightly different format?

kof:
@istarkov I agree with @rpominov on this

> In general assets support in props is added kinda awkwardly. Couldn't it be just another value type?

istarkov:
Finally after talking with Oleg, conclusion - my way makes UI much harder if we will have other objects. Writing `handlePropChange{ObjectName}` would be a disaster at some moment. Moving to value of discriminated Union type.

kof:
@istarkov btw StyleValue is basically already modeled well, that's what we need for props value too

kof:
in fact we may need at some point to support a style value as a prop value

apps/designer/app/designer/features/props-panel/use-props-logic.ts
istarkov: I'm not getting what's shared here. And why to move handlers into shared

packages/react-sdk/src/user-props/use-user-props.ts
istarkov: We did that with props transform then decided to move that logic into the component. As the image has better knowledge of what it needs from assets like our Image uses 2 asset props.

apps/designer/app/designer/features/props-panel/use-props-logic.ts
rpominov: I meant we could share something like `setPropValueMutable(allProps: AllUserProps, name: ..., value: ...): void`. But after you've updated the code maybe it doesn't make as much sense as the code became trivial

packages/prisma-client/prisma/migrations/20221201075120_user-props/migration.ts
```diff
@@ -0,0 +1,139 @@
...
+              id: prop.id,
+              prop: prop.prop,
+              required: prop.required,
+              type: "asset",
+              value: prop.assetId,
+            });
+            continue;
+          }
+
+          if (prop.asset_id != null) {
```
kof, line 80: uhh, this also exists)))

packages/project/src/db/props.ts
```diff
@@ -1,16 +1,50 @@
...
+  }),
+  z.object({
+    ...baseUserProps,
+    type: z.literal("boolean"),
+    value: z.boolean(),
+  }),
+  z.object({
+    ...baseUserProps,
+    type: z.literal("asset"),
+    // In database we hold asset_id
```
kof, line 39: assetId

kof (approved):
looks good!

packages/prisma-client/prisma/migrations/20221201075120_user-props/migration.ts
```diff
@@ -0,0 +1,139 @@
...
+    // Uncomment to see the queries in console as the migration runs
+    // log: ["query", "info", "warn", "error"],
+  });
+  return client.$transaction(
+    async (prisma) => {
+      const allInstanceProps = await prisma.instanceProps.findMany({});
+
+      for (const instanceProps of allInstanceProps) {
+        const rawProps = JSON.parse(instanceProps.props);
+        const props = OutdatedProps.parse(rawProps);
```
rpominov, line 58: This may throw when we run migration in production as it's possible that we have invalid data there.

I thinks we should do one of these:
1. Import prod db locally and try applying the migration
2. Wrap this in try..catch and delete row or something so migration would complete anyway
3. Parse all rows in a separate loop before the main loop where you write to the DB. Catch all parse errors and collect them in an array. Before the second loop, if array is not empty, throw an error with a message that would help us decide how to fix prod DB (Migration will fail, but with a useful error)

rpominov (approved):

packages/prisma-client/prisma/migrations/20221201075120_user-props/migration.ts
istarkov: Start testing locally, 
- [ ] - copy props
```
CREATE EXTENSION dblink;

INSERT INTO "public"."InstanceProps"("id","instanceId","treeId","props")
SELECT * FROM dblink (
'dbname=postgres port=5432 host=prodhost user=postgres password=prodpasswd',
'select id, "instanceId", "treeId", "props" from "InstanceProps"'
) AS t(
    id text,
    "instanceId" text,
    "treeId" text,
    props text
)
ON CONFLICT DO NOTHING;
```

- [ ] Test

istarkov: Thank you, really had timeout on production data

rpominov (approved):

PR #550, opened 2022-12-01
Spacing UI

### LINKED ISSUE #421: New spacing UI
Current UI was made without proper design and we now have CSS value input we want to use in it.

New design is still in tweaking, but it will looks something like this:
Screenshot 2022-11-08 at 10 10 17

https://www.figma.com/file/EfzZ2w94KpmUviZJSBgxCd/Designer-UI?node-id=1142%3A20479

## Todo

- [x] use svg from figma to render the 3d-shaped control.
- [x] hover and click of controls has the non-rect form, so we need to render svg as component and work with it's paths/shapes for event handling
- [x] use scrub primitives we already have to add scrub to the controls
- [x] each control should have a cursor showing the right direction (see wf)
- [x] click on the control should open the CSS value input as demonstrated [here](https://github.com/webstudio-is/webstudio-designer/pull/401 )
- [x] CSS value input also has scrub built in and should also work when opened
- [x] Rendering the value in control 
   - [x] when value is too long, we need to start rendering unit on the next line.
   - [ ] when value is in px, we don't show the px in the control
   - [ ] when value is still too long after unit is on the next line, we should center the value so that it grows like wf does this.
   
   
 Here is a prototype of the  value input based on the current spacing UI [see PR](https://github.com/webstudio-is/webstudio-designer/pull/401 )
     

https://user-images.githubusercontent.com/52824/200816742-108bed81-54f3-4a31-9e2d-e68e59b9f09f.mp4

---------------

https://www.figma.com/file/EfzZ2w94KpmUviZJSBgxCd/Designer-UI?node-id=3024%3A44415&t=pMvbZ6AkTH7iyihi-4
image

rpominov:
Implementation of new spacing UI. Closes #421 

## Not in this PR

- https://github.com/webstudio-is/webstudio-designer/issues/665
- https://github.com/webstudio-is/webstudio-designer/issues/666
- https://github.com/webstudio-is/webstudio-designer/issues/667
- https://github.com/webstudio-is/webstudio-designer/issues/668

## Code Review

(Please ignore behaviour when you hit `esc` in input, will fix it here after we merge https://github.com/webstudio-is/webstudio-designer/pull/661)

  - test it on preview
  - test it on preview (everything should work unless listed in "Not in this PR" above)
  - have a look at code where CSSInputValue is used or changed (just in case)
  - do a detailed review if you have time

## Before requesting a review

- [x] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


istarkov:
Just for fun with named lines, html more redable, template syntax is disaster ;-) but I know some layouts where template syntax much more readable

https://codepen.io/istarkov/pen/dyKggMq

```html
<div class="grid">
  <div class="bg" style="grid-column: margin;grid-row: margin-top">mt</div>
  <div class="bg" style="grid-column: padding;grid-row: padding-start">pt</div>
  <div class="bg" style="grid-column: content;grid-row: content">center</div>
  <div class="bg" style="grid-column: padding;grid-row: padding-bottom">pb</div>
  <div class="bg" style="grid-column: margin;grid-row: margin-bottom">mb</div>
  
  <div class="bgy" style="grid-row: margin;grid-column: left-margin">ml</div>
  <div class="bgy" style="grid-row: padding;grid-column: left-padding">pl</div>
  <div class="bgy" style="grid-row: padding;grid-column: right-padding">pr</div>
  <div class="bgy" style="grid-row: margin;grid-column: right-margin">mr</div>
</div>
```

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -193,7 +200,7 @@ type CssValueInputProps = {
    */
   keywords?: Array<KeywordValue>;
   onChange: (value: CssValueInputValue) => void;
-  onChangeComplete: (value: StyleValue) => void;
+  onChangeComplete: (value: StyleValue, reason: ChangeReason) => void;
```
istarkov, line 203: Just wrote to Oleg in DM that for esc we need to know the reason of called events.
Moreover seems like I need to distinguish blur and enter too, so it looks like a good change.

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+const PopoverContentStyled = styled(PopoverContent, {
+  minWidth: 0,
+  minHeight: 0,
+  width: 88,
+  border: "1px solid $colors$slate8",
+  borderRadius: "$borderRadius$7",
+  background: "$colors$gray2",
+  padding: "$spacing$5",
+  boxShadow: "0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15)",
```
istarkov, line 75: Usually most box shadows are same, and should be in tokens too?

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+        if (nextOpen === false) {
+          onClose();
+        }
+      }}
+    >
+      <PopoverTrigger asChild>
+        <Trigger />
+      </PopoverTrigger>
+      <PopoverPortal>
+        <PopoverContentStyled hideArrow sideOffset={-24}>
```
istarkov, line 107: Looks like you have same for offset https://github.com/webstudio-is/webstudio-designer/pull/659

apps/designer/app/designer/features/style-panel/sections/spacing/layout.tsx
```diff
@@ -0,0 +1,302 @@
+import { styled, useId } from "@webstudio-is/design-system";
+import type { ComponentProps } from "react";
+import type { HoverTagret, SpacingStyleProperty } from "./types";
+import { spacingPropertiesNames } from "./types";
+
+const VALUE_WIDTH = 34;
```
istarkov, line 6: If someday we will decide to chaange width of `aside` we will get that internal container size is fixed?

image

Mb calculations should start from aside theme value width? As looks like would be hard to find constants.
(IMHO any non-percentage width set inside the layout with known size would be an issue some day)

apps/designer/app/designer/features/style-panel/sections/spacing/value-text.stories.tsx
```diff
@@ -0,0 +1,53 @@
+import type { ComponentStory, ComponentMeta } from "@storybook/react";
+import React from "react";
+import { SpacingLayout } from "./layout";
+import { ValueText as ValueTextCommponent } from "./value-text";
+
+const Template = (
+  args: Pick<
+    React.ComponentProps<typeof ValueTextCommponent>,
```
istarkov, line 8: Commponent -> Component

apps/designer/app/designer/features/style-panel/sections/spacing/value-text.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export default {
+  title: "ValueText",
+  component: Template,
+  argTypes: {
+    source: { control: "select", options: ["set", "unset", "inherited"] },
+    value: {
+      control: "select",
+      options: {
+        "0": { type: "unit", value: 0, unit: "number" },
```
istarkov, line 35: probably use [parseCssValue](https://github.com/webstudio-is/webstudio-designer/blob/e78919cc999320f2d42d65e4ee43605d1a308966/apps/designer/app/designer/features/style-panel/shared/parse-css-value.ts#L28) not a big deal anyway

istarkov (approved):

apps/designer/app/designer/features/style-panel/sections/spacing/layout.tsx
rpominov: I used base values as chosen in Figma:

Screenshot 2022-12-20 at 19 45 56

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
rpominov: 24 here is half of the width of the popover basically. It shouldn't change and if changed can just change here as well. Not a big deal IMO

taylornowotny:
My feedback:

- First of all, great work! The proportions and styles look nearly perfect. Adding a value with the pop-up text field works really well.

- can we show a pointer instead of the arrow cursor when the value is hovered?

- There should not be any 'set' styles on the Body. If you want to disable margin changes on the Body element then just hide the value on the margins and disable them. Screenshot 2022-12-20 at 11 05 41 AM

- "preset" styles like on a new div should have the "preset" label style. Added an example to the component docs for reference.
Screenshot 2022-12-20 at 11 09 44 AM
Screenshot 2022-12-20 at 11 15 11 AM

-the left and right triangles appear lighter than the top & bottom triangles but this is opposite of the design
Screenshot 2022-12-20 at 11 17 57 AM
Screenshot 2022-12-20 at 11 19 11 AM

kof:
> can we show a pointer instead of the arrow cursor when the value is hovered?

just the value, but not the scrubbable surface, right?

taylornowotny:
> > can we show a pointer instead of the arrow cursor when the value is hovered?
> 
> just the value, but not the scrubbable surface, right?

Yes that's what I mean

kof:
also you want not the pointer cursor but the normal, default cursor on the value

taylornowotny:
Few more things - we shouldn't allow invalid values. The input does have an error state when I add "1-" for example, but you can still add the invalid value. If an invalid value is applied and the user closes the input , the value should revert to what it was before the value was changed. 

Screenshot 2022-12-20 at 11 27 47 AM

If I click a value and then press ESC before making changes, the value changes to "unset". This happens consistently regardless of what the initial value was.
Screenshot 2022-12-20 at 11 30 40 AM

taylornowotny:
> also you want not the pointer cursor but the normal, default cursor on the value

ok agreed

kof:
> The input does have an error state when I add "1-" for example, but you can still add the invalid value. If an invalid value is applied and the user closes the input , the value should revert to what it was before the value was changed.

This isn't just about spacing, its a decision that we need to make for all values with custom input. We have been discussing it today.

With linked editor in mind and with the fact that we may not 100% be correct about some value, a better option is to show  that we think the value is invalid clearly but allow saving, because a user might know something we don't know.

E.g. with linked CSS editor the goal is to provide full power of CSS, but that means we might have certain values in the style panel that originates in the linked editor and both might think its an invalid value.

On the other hand what's the value of reverting to the previous valid value? Its not like the user didn't see its invalid, we can show clearly. So reverting is a form of hand-holding

kof:
> If I click a value and then press ESC before making changes, the value changes to "unset". This happens consistently regardless of what the initial value was.

Separate bug in all values, already being worked on here https://github.com/webstudio-is/webstudio-designer/pull/661

kof:
> "preset" styles like on a new div should have the "preset" label style. Added an example to the component docs for reference.

preset state needs to be added to the CssValueInputIconButton in CssValueInput and then we will refactor the inputs according to the new design and move those to the ds

rpominov:
Confused about cursor. 
- Do we want just a different cursor or ability to select and copy the text?
- Should click work on the entire surface, or just on the value?
- Should scrub work on the entire surface?

rpominov:
@taylornowotny 

> First of all, great work! The proportions and styles look nearly perfect. Adding a value with the pop-up text field works really well.

Thanks!

> can we show a pointer instead of the arrow cursor when the value is hovered?

- [ ] Will fix, but need some clarification as I wrote above

> There should not be any 'set' styles on the Body. If you want to disable margin changes on the Body element then just hide the value on the margins and disable them.

We define margins for body here. @kof why do we need them?
https://github.com/webstudio-is/webstudio-designer/blob/a4c0e5c070c51e1c6ca8a63b3023829a490e1114/packages/react-sdk/src/components/body.ws.tsx#L6-L25

In any case, don't think we should fix it in this PR. Control just shows styles that are there.

Another problem is that they look like `set` instead of `preset`, more on that just bellow

> "preset" styles like on a new div should have the "preset" label style. Added an example to the component docs for reference.

We don't have good way to distinguish between styles' types (set / preset / inherited / etc) currently. In any control, not just spacing. Created a thread here: https://discord.com/channels/955905230107738152/1055011309487407194 

> the left and right triangles appear lighter than the top & bottom triangles but this is opposite of the design

- [ ] will fix

> we shouldn't allow invalid values. The input does have an error state when I add "1-" for example, but you can still add the invalid value. If the user attempts to apply an invalid value, the value should revert to what it was before it was changed.

As @kof said, we might have to allow invalid values. Discussed here: https://discord.com/channels/955905230107738152/1054691114164174848 . TLDR: CSS is complex and supports a lot of things as values, there can be a value that our code can't understand but browser will understand, and user may want to set such a value.

> If I click a value and then press ESC before making changes, the value changes to "unset". This happens consistently regardless of what the initial value was

As @kof said this is being worked on in #661 .
- [ ] Once that PR is merged, I'll make sure that it's works as expected in spacing too

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
rpominov: We currently don't have boxShadows in `stitches.config.ts`, all of them defined in code like here. 

In Figma it also seem to be unique:
Screenshot 2022-12-21 at 11 31 20

Maybe we should start defining them in `stitches`, but not sure I should do it in this PR.

kof:
> We define margins for body here. @kof why do we need them?

We are resetting the body margins to 0, they are non-0 by default in browsers

kof:
@rpominov the styles we set in the presetStyle are supposed to be of "preset" type, not "set", because "set" is only by the user

rpominov:
> We are resetting the body margins to 0, they are non-0 by default in browsers

Got it

> @rpominov the styles we set in the presetStyle are supposed to be of "preset" type, not "set", because "set" is only by the user

Yes, I wrote above:

Screenshot 2022-12-21 at 12 44 50

kof:
> Do we want just a different cursor or ability to select and copy the text?

just the cursor on the value, to indicate you can click not only drag

> Should click work on the entire surface, or just on the value?

yes, still entire surface is clickable, this is better than forcing them to point exactly at the value

> Should scrub work on the entire surface?

definitely

kof:
I just tested it, it looks and works amazingly.

One thing I would probably improve is making popover "modal" (it already supports this).

@taylornowotny I remember we discussed it a while ago and I don't remember the outcome, but from my first testing it felt like I want to click it away with the first click but instead I end up clicking something else every time.

Maybe its just me, but I think we want to make all popovers behave consistently for the sake of user knowing upfront what's gonna happen, consistency.

rpominov:
> One thing I would probably improve is making popover "modal" (it already supports this).

I tried modal at first, but to me it was very annoying to have to click twice on e.g.`margin-left` to open popover there when I have popover already open for another property

kof:
> I tried modal at first, but to me it was very annoying to have to click twice on e.g.margin-left to open popover there when I have it already open for another property

Yeah, I think its because you wanted to open another input right away, I assumed people will want to just close the current one and do something else, but I may be wrong too, not sure what's the user expectation is here, lets keep it like this for now and we can show it to some users and ask them what they would expect

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+  );
+};
+
+// trigger is used only for positioning
+const Trigger = styled("div", { position: "absolute", width: 0, height: 0 });
+
+const PopoverContentStyled = styled(PopoverContent, {
+  minWidth: 0,
+  minHeight: 0,
+  width: 88,
```
kof, line 70: how did we end up with this one? can it use token?

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+  border: "1px solid $colors$slate8",
+  borderRadius: "$borderRadius$7",
+  background: "$colors$gray2",
+  padding: "$spacing$5",
+  boxShadow: "0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15)",
+  animationDuration: "200ms",
```
kof, line 76: I would try 150ms, its good but it feel like it should be a bit faster

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -184,6 +184,13 @@ export type IntermediateStyleValue = {
 
 type CssValueInputValue = StyleValue | IntermediateStyleValue;
 
+type ChangeReason =
+  | "enter"
+  | "blur"
+  | "unit-select"
+  | "combo-select"
```
kof, line 191: keyword-select?

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
kof: Might make sense to change the callback type from value to {value, reason} to keep the single-callback pattern

apps/designer/app/designer/features/style-panel/sections/spacing/tooltip.tsx
```diff
@@ -0,0 +1,51 @@
...
+  position: "absolute",
+  width: "100%",
+  height: "100%",
+  pointerEvents: "none",
+  top: 0,
+  left: 0,
+});
+
+const labels = {
+  paddingTop: "Padding-top",
```
kof, line 15: We just had a discussion with @taylornowotny to use space separated upper cased names: `Padding Top`, he will update the designs next

apps/designer/app/designer/features/style-panel/sections/spacing/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { StyleValue, StyleProperty } from "@webstudio-is/css-data";
+import { categories } from "@webstudio-is/react-sdk";
+
+export type StyleChangeEvent = {
```
kof, line 4: hmm, why do we need a separate type for spacing, isn't this something we already have for other things?

apps/designer/app/designer/features/style-panel/sections/spacing/value-text.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+          "100000px",
+          "auto",
+          "revert-layer",
+        ].map((value) => [value, parseCssValue("marginTop", value)])
+      ),
+    },
+  },
+} as ComponentMeta<typeof Template>;
+
+export const ValueText: ComponentStory<typeof Template> = Template.bind({});
```
kof, line 52: do you need to use the Template thing? storybook's latest format uses plain component export, e.g. combobox.stories

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
kof: we are going to hav tokens for those too, defined in figma tokens plugin and synced into the codebase, I guess for now we could do either way, might be easier to migrate if its in stitches config to see them all at once

kof (changes requested):
Looking great, some comments added

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
rpominov: Width of the popover from Figma.

Screenshot 2022-12-21 at 13 39 01

We don't have tokes for neither 88 nor 72 (input width).

apps/designer/app/designer/features/style-panel/sections/spacing/types.ts
rpominov: This type allowed me to unify nicely all change handlers in spacing. In other places we use two callbacks to distinguish between `isEphemeral` and we don't pass `property`.

Maybe we should use this type everywhere not sure

apps/designer/app/designer/features/style-panel/sections/spacing/value-text.stories.tsx
rpominov: I use Template to control `args` (to make it possible to change some of the props via Storybook UI and see how component behaves)

apps/designer/app/designer/features/style-panel/sections/spacing/types.ts
kof: Yeah, if anything consistency is important

kof: We got StyleUpdate and StyleUpdateOptions. 
Options is used locally in the style panel only, while StyleUpdate represents datastructure that goes beyond style panel, that's why it was separated

kof: if every panel changes these things it will be harder to maintain, so lets have one thing, whatever that is

apps/designer/app/designer/features/style-panel/sections/spacing/value-text.stories.tsx
kof: you can put args on the component directly, see text-editor.stories for example

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
kof: that means something is wrong with the design, @taylornowotny probably just didn't use the right spacing, I would use the closest thing we have in tokens to it, he will update when he can

rpominov: Switched to `$spacing$20` which is `80px`. I think it looks good too:

Screenshot 2022-12-21 at 14 18 34

apps/designer/app/designer/features/style-panel/sections/spacing/value-text.stories.tsx
rpominov: Template also changes render a lot. Take a look at what it does

apps/designer/app/designer/features/style-panel/sections/spacing/input-popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+const PopoverContentStyled = styled(PopoverContent, {
+  minWidth: 0,
+  minHeight: 0,
+  width: "$spacing$20",
+  border: "1px solid $colors$slate8",
+  borderRadius: "$borderRadius$7",
+  background: "$colors$gray2",
+  padding: "$spacing$5",
+  boxShadow: "0px 2px 7px rgba(0, 0, 0, 0.1), 0px 5px 17px rgba(0, 0, 0, 0.15)",
+  animationDuration: "150ms",
```
kof, line 76: this didn't help, the initial start is a bit too slow before it shows up, probably need to start with scale not 0 but something like 20-40, I will experiment locally a bit

rpominov: Yeah, please experiment yourself, looks all the same to me 😅

kof (approved):

apps/designer/app/designer/features/style-panel/sections/spacing/types.ts
rpominov: Using StyleUpdate and StyleUpdateOptions now

rpominov:
Everything except `esc` is fixed. Created separate issue https://github.com/webstudio-is/webstudio-designer/issues/670

Will merge

PR #551, opened 2022-12-02
Make CSSValueInput controllable,

istarkov:
CSSValueInput used useEffect and executed onChange etc events on its own. What in case of any small changes caused loops and strange behavior. Now should execute everything only as a result of user input.

`parseCssValue` - now stupid parse CSS value and someday will be used only inside text-control.
setProperty - now accepts StyleValue, the string param is deprecated, warning set.

CSSValueInput now not update styles during editing.

unit menu now not hiding during formulas etc

fixed bug with updating style on newly created items

fixed scrub bugs with unsubscribing

css-tree Adds 52101 gzip bytes. 
A lot of size is coming from here 
https://github.com/csstree/csstree/blob/97d7f3b95a3efd8f431af39986acc149dceefa59/lib/data.js
more than 1/3 of `esm` size (20+kb gz), and most huge are properties json we need and use. 

Doesn't look tree-shakeable, anyway can be loaded lazily as until editing we don't need to validate CSS vars.

Investigating: 
- [x] - Changing any style Combobox  with mouse cause multiple value updates to previous and current value (was a blur event issue)


@types/css-tree.d.ts
```diff
@@ -0,0 +1,21 @@
+declare module "css-tree" {
```
TrySound, line 1: why not use @types/css-tree?

kof: they are missing types, I also asked the same in dm, time to put a comment and a todo to submit a pr for them later

apps/designer/app/designer/features/style-panel/shared/parse-css-value.ts
```diff
@@ -1,28 +1,30 @@
+import * as csstree from "css-tree";
 import hyphenate from "hyphenate-style-name";
 import type { StyleProperty, StyleValue, Unit } from "@webstudio-is/css-data";
 import { units } from "@webstudio-is/css-data";
 
-const unitRegex = new RegExp(`${[...units, "number"].join("|")}`);
+const cssTryParseValue = (input: string) => {
+  try {
+    const ast = csstree.parse(input, { context: "value" });
```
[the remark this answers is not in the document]
kof: yes, the problem is this isn't crossbrowser yet, we can replace it some day.
additionally we may need it server-side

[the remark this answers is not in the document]
kof: Essentially only chrome has implemented it atm, because Tab Atkins works at google and probably could make that happen. Neither firefox nor safari have done it yet

@types/css-tree.d.ts
TrySound: Or get rid of css-tree eventually :)

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -29,7 +30,9 @@ import {
 } from "react";
 import { useIsFromCurrentBreakpoint } from "../use-is-from-current-breakpoint";
 import { useUnitSelect } from "./unit-select";
-import { isValid, isNumericString } from "../parse-css-value";
+import { parseCssValue } from "../parse-css-value";
+import { unstable_batchedUpdates } from "react-dom";
```
TrySound, line 34: Will need to retest when we bump to react 18

TrySound (approved):
Let's hope this will be simpler with time

apps/designer/app/designer/features/style-panel/controls/index.ts
```diff
@@ -5,4 +5,3 @@ export * from "./menu/menu-control";
 export * from "./font-family/font-family-control";
```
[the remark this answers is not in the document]
istarkov: Done

apps/designer/app/canvas/shared/styles.ts
```diff
@@ -225,16 +225,26 @@ const useUpdateTokens = () => {
...
-      cssEngine.render();
+    let rule = getRule(id, breakpoint.id);
+
+    if (rule === undefined) {
+      rule = addRule(id, { breakpoint: breakpoint.id, style: {} });
     }
+
     for (const update of updates) {
+      // This is possible on newly created instances, properties are not yet defined in the style.
+      if (rule.styleMap.get(update.property) == null) {
```
kof, line 236: == null is confusing, it will rever ever return a null, its only undefined

kof: You make it look like the api could return null or undefined

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -184,16 +160,28 @@ const useHandleKeyDown =
...
+    onKeyDown(event);
   };
 
+export type IntermediateStyleValue = {
+  type: "intermediate";
+  value: string;
+  unit?: Unit;
+};
+
+type CSSValueInputValue = StyleValue | IntermediateStyleValue;
```
kof, line 176:
```suggestion
type CssValueInputValue = StyleValue | IntermediateStyleValue;
```

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -227,16 +215,102 @@ type CssValueInputProps = {
...
+    }
+
+    const unit = "unit" in value ? value.unit ?? "px" : "px";
+    styleInput = parseCssValue(property, `${input}${unit}`);
+
+    if (styleInput.type !== "invalid") {
+      props.onChange(styleInput);
+      return;
+    }
+    */
```
kof, line 244: Normally I would say don't leave comments like that

apps/designer/app/designer/features/style-panel/shared/css-value-input/evaluate-math.ts
```diff
@@ -0,0 +1,14 @@
...
+    return undefined;
+  }
+  try {
+    // Eval is safe here because of the regex above
+    const result = eval(`(${expression})`);
+    if (typeof result === "number") {
+      return result;
+    }
+  } catch (err) {
+    return undefined;
```
kof, line 12:
```suggestion
    return;
```

apps/designer/app/designer/features/style-panel/shared/css-value-input/evaluate-math.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (/^[\d\s.+*/-]+$/.test(expression) === false) {
+    return undefined;
+  }
+  try {
+    // Eval is safe here because of the regex above
+    const result = eval(`(${expression})`);
+    if (typeof result === "number") {
+      return result;
+    }
+  } catch (err) {
```
kof, line 11:
```suggestion
  } catch (error) {
```

apps/designer/app/designer/features/style-panel/shared/css-value-input/evaluate-math.ts
```diff
@@ -0,0 +1,14 @@
+export const evaluateMath = (expression: string) => {
```
kof, line 1: this nice little function could have some unit tests :)

apps/designer/app/designer/features/style-panel/shared/css-value-input/unit-select.tsx
```diff
@@ -49,18 +50,31 @@ export const useUnitSelect = ({
...
 }: UseUnitSelectType) => {
   const [isOpen, setIsOpen] = useState(false);
+
   const renderUnits = useMemo(
     () =>
       value &&
       units
         .filter((unit) => {
+          if (value.type === "intermediate") {
+            if (value.unit != null) {
```
kof, line 60: thanks to this null check I went checking if unit could be null 🤷‍♂️

apps/designer/app/designer/features/style-panel/shared/css-value-input/unit-select.tsx
```diff
@@ -49,18 +50,31 @@ export const useUnitSelect = ({
...
+
           return isValid(property, toValue({ ...value, unit }));
         })
         .map((unit) => unitRenderMap.get(unit) ?? unit),
     [units, property, value]
   );
 
-  const renderValue = value && unitRenderMap.get(value.unit);
+  const renderValue =
+    value && value.unit != null ? unitRenderMap.get(value.unit) : undefined;
```
kof, line 77: could be just 
```suggestion
   value?.unit != null ? unitRenderMap.get(value.unit) : undefined;
```

kof (approved):
all comments are stylistic, logic seems fine

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -251,6 +255,21 @@ export const CssValueInput = ({
...
 
+    let styleInput = parseCssValue(property, value.value);
+
+    if (styleInput.type !== "invalid") {
+      props.onChange(styleInput);
+      return;
+    }
+
+    const unit = "unit" in value ? value.unit ?? "px" : "px";
+    styleInput = parseCssValue(property, `${value.value}${unit}`);
```
rpominov, line 271: Strange logic:
1. In the second `parseCssValue` we use `value.unit` if present, which makes me think that `value.unit` is potentially important
2. But in the first `parseCssValue` we ignore `value.unit`

apps/designer/app/canvas/shared/styles.ts
istarkov: Switched on has, as basically we are checking existance, also fixed PlaintextRule styleMap on  `new StylePropertyMap();`
https://github.com/webstudio-is/webstudio-designer/pull/551/commits/274e68843d39fa70b9492576363b817be01e049f
as it cause `any` in the return type of `get`

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
istarkov: See comments `// Probably value is already valid, use it` 

My idea is that user raw input has max priority, if it's valid, let's use it.
Then I just try to get what the user probably wanted if the raw input is invalid.

istarkov: After discussion, line-height will not work with this logic. Will write test for `value` conversion

rpominov (approved):

PR #553, opened 2022-12-03
Replace `UnionToTuple` with safer approach

Andarist:


apps/designer/app/designer/shared/design-tokens-manager/schema.ts
```diff
@@ -1,34 +1,29 @@
...
-  : [];
-
-type Type = typeof groups[number]["type"];
-type Group = typeof groups[number]["group"];
-
-// zod is super restrictive on enum type https://github.com/colinhacks/zod/issues/1118
-// so we convert union to tuple here its seriously safer than any as it will break on things like
-// ['a', 'b'] as UnionToTuple<Type> etc
-const types = groups.map((group) => group.type) as UnionToTuple<Type>;
-const groupNames = groups.map((group) => group.group) as UnionToTuple<Group>;
```
Andarist, line 23: This was fine here as those variables don't escape this module and are only used as inputs for zod. However, `UnionToTuple` is not safe in the generic sense - the order of the elements in the result is **unstable** and can't be relied upon.

Therefore, I've replaced this with a different approach - mostly just for educational reasons, to showcase how this kind of a problem can be approached in the future etc.

apps/designer/app/designer/shared/design-tokens-manager/schema.ts
```diff
@@ -1,34 +1,29 @@
...
+// zod expects readonly tuples for its enums
+const types = mapToProp(groups, "type");
+const groupNames = mapToProp(groups, "group");
 
 export const DesignToken = z.object({
   name: z.string(),
-  // @todo if https://github.com/colinhacks/zod/issues/1118
```
Andarist, line 27: I think that this issue was linked here incorrectly. The issue is about the fact that `enum` requires strings - but we have strings here so it wasn't really applicable to this issue here.

istarkov:
I haven't knew this ;-) never mapped array types through object syntax ;-) SUPER!!!
```
type X<T extends ReadonlyArray<any>> = {
    [k in keyof T]: T[k]
}

const z = [1,2] as const;

type Y = X<typeof z>
```

istarkov (approved):
Super!

kof:
I may be missing something, but it feels like its basically the same as just using a type cast at the end and be done

kof:
like all those types for what, if we can just do at the end of the expression as unknown as Array<Group>

Andarist:
> I may be missing something, but it feels like its basically the same as just using a type cast at the end and be done

Yes, kinda - if i’d have access to the type i’d like to cast then i would just do that. But it’s not directly available here and i need to derive it somehow at some point anyway. And this variant feels much simpler than the old one that was based on two complex conditional types

kof:
honestly, if this doesn't give us any value, then we should just typecast at the end of an expression, I asked for help with this one because I thought its possible to have accurate types and not have castings and anys. If the answer is no, I would just remove all of that complexity

Andarist:
This is what the "cast only" variant could look like:
https://github.com/Andarist/webstudio-designer/compare/0aea516571f88062fcb96b15cb49f6f230ac369b..3c2bf0dcfbfc95ea6eb7910d039ebed09ae4b2c8

It's just worth noting that I treat here a mapped array as a single element tuple - it works here but if somebody would copy-paste it from here and use it in a different context then they could get in trouble. In reality, those values are *not* single-element tuples.

That being said - I took a closer look at this now and it seems that this whole `DesignToken` type is prone to invalid combinations anyway. If I understand correctly, your `.type` and `.group` should always be in sync but there is nothing here that would enforce it. Perhaps if you'd drop `.group` (and derive the label's text from a separate map or something, using a util) then you wouldn't have such problems and you could end up with strict types.

kof:
@Andarist yeah, that's what I meant that is same safety but reduces complexity and I don't have to read those insane ts types

kof:
@Andarist fixed it here https://github.com/webstudio-is/webstudio-designer/pull/518/commits/662bfa580a45ace88ff81e6e68f05bb3219cd3f3

This was exactly what I didn't know, it needed the array to be a readonly tuple, that was the problem with zod!!!!!

PR #557, opened 2022-12-06
Remove a redundant install step in one of the workflows

Andarist:


.github/workflows/publish.yml
```diff
@@ -20,7 +20,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/ci-setup
 
-      - run: yarn
```
Andarist, line 23: This is already included in the `- uses: ./.github/actions/ci-setup` (and that is used one line above this).

TrySound: Does it break publish?

Andarist: I doubt it - it's something that I've noticed when investigating that publish failure.

TrySound (approved):

PR #559, opened 2022-12-06
Fixed publish by overriding the implicitly enforced yarn registry with the default one

Andarist:


bin/publish-to-npm.sh
```diff
@@ -9,5 +9,8 @@ echo Current $npm_package_name $NPM_VERSION
 
 if [[ $PKG_VERSION != $NPM_VERSION ]]; then
   echo Publishing $npm_package_name $PKG_VERSION
-  npm publish --access public
+  # even though `npm` is used here this bash script gets invoked through `yarn`
+  # yarn 1 provides an env variable overriding the default npm registry with their own proxy
+  # to keep the authentication working we need to override the registry url since we only configure the auth token for the default registry
```
Andarist, line 14: I was previously accidentally testing this with Yarn 3 - and it doesn't use this proxy registry. So in all of my tests in the kitchen sink repo the previous configuration was working fine. When I removed Yarn 3 from it and replaced it with Yarn 1, I immediately reproduced the issue.

TrySound (approved):

PR #561, opened 2022-12-07
Fix dependency mapping for blockLevelElements

kof:
For example property "clear"  was  not showing up in position section because of that. It is supposed to show when display is any block-level value

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/dependencies.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/dependencies.ts
@@ -75,7 +75,7 @@ export const dependencies: Dependencies = {
     notValues: ["none"],
   },
   blockLevelElements: {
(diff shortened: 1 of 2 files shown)


apps/designer/app/designer/features/style-panel/shared/dependencies.ts
```diff
@@ -75,7 +75,7 @@ export const dependencies: Dependencies = {
     notValues: ["none"],
   },
   blockLevelElements: {
-    property: "clear",
+    property: "display",
```
kof, line 78: this seems to have been just some early typo

PR #563, opened 2022-12-07
Fix scrub uses other instance value

istarkov:
The issue is if we select on canvas instance of one Box for example, and then other box, scrub had in the state value of first selected Box.

https://user-images.githubusercontent.com/5077042/206211694-19124e0b-5b90-4aef-8848-7ebe7f1d7bae.mov

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -91,6 +91,14 @@ const useScrub = ({
     }
 
     const scrub = numericScrubControl(scrubRefCurrent, {
+      // @todo: after this https://github.com/webstudio-is/webstudio-designer/issues/564
+      // we can switch back on using just initial value
+      //
+      // For now we are reusing controls for different selectedInstanceData,
+      // and the best here is to call useEffect every time selectedInstanceData changes
+      // and recreate numericScrubControl.
+      // Until we have decision do we use key properties for this or not,
+      // on of the solution to get value inside scrub is to use ref and lazy getter.
       // Getter to avoid recreating scrub on every value change
       getValue: () => {
         if (valueRef.current.type === "unit") {
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -96,8 +96,6 @@ const useScrub = ({
         if (valueRef.current.type === "unit") {
           return valueRef.current.value;
         }
-
-        return;
       },
       onValueInput(event) {
         setIsInputActive(true);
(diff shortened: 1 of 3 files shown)


apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -90,9 +90,15 @@ const useScrub = ({
...
-    const value = valueRef.current.value;
     const scrub = numericScrubControl(scrubRefCurrent, {
-      initialValue: value,
+      // Getter to avoid recreating scrub on every value change
+      getValue: () => {
+        if (valueRef.current.type === "unit") {
+          return valueRef.current.value;
+        }
+
+        return;
```
kof, line 100: you don't need this

packages/design-system/src/components/primitives/numeric-gesture-control.stories.tsx
```diff
@@ -18,7 +18,7 @@ const useNumericScrubControl = ({
     if (ref.current === null) return;
     ref.current.value = String(value);
     const { disconnectedCallback } = numericScrubControl(ref.current, {
-      initialValue: value,
+      getValue: () => value,
```
kof, line 21: do you intent to call it any time or only as an initial value? if only as initial, then we couldn't call it getInitialValue

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -94,6 +95,13 @@ export const numericScrubControl = (
         if (event.target && shouldHandleEvent?.(event.target) === false) return;
         // light touches don't register corresponding pointerup
         if (event.pressure === 0 || event.button !== 0) break;
+        const value = getValue();
+
+        // We don't support scrub on non unit values
+        if (value === undefined) return;
```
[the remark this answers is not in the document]
kof: or maybe even better instead of disabling it - destroy it completely, once its unit input again - initialize it again

rpominov (approved):
Agree with Oleg that using `getValue()` to also disable/enable scrub is kinda dirty and confusing. But otherwise looks good.

kof (approved):

kof:
We just discussed in dm, there is 2 cases when we need to reset the initial value:
1. when value type changes from unit to keyword
2. when selected instance changes

In the first case we could easily destroy it and reinitialize, but in the second case, value type doesn't change, so we would need to react to a value update, but reacting to the value update would have an effect on overall performance because value changes all the time and we have no way to differentiate between regular value update and selected instance change. 

We would have to memorize also based on the selected instance

istarkov:
The issue not even performance but vanilla events, having that we change value during “scrub dragging”  it will cause subscribe/unsubscribe of pointermove etc and loosing current state of scrub i.e that drag in progress, etc.

kof:
@istarkov another good point is collaboration, what if one user changes the unit while the other user is dragging? :)

istarkov:
> @istarkov another good point is collaboration, what if one user changes the unit while the other user is dragging? :)

Now it will just unsubscribe all handler and dragging will be ended, probably in collaboration scenario its not that bad

istarkov:
created issue https://github.com/webstudio-is/webstudio-designer/issues/564

PR #565, opened 2022-12-07
Upload each asset separately

kof:
This will make each file uploaded as a separate request, serially.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/designer/shared/assets/asset-upload.tsx
```diff
@@ -30,29 +30,31 @@ type AssetUploadProps = {
...
-        css={{ flexGrow: 1 }}
-      >
-        <Flex align="center" gap={1}>
-          <UploadIcon />
-          <Text>Upload</Text>
-        </Flex>
-      </Button>
-    </Flex>
+    <form onChange={onChange}>
+      <Flex css={{ flexGrow: 1 }}>
```
TrySound, line 36: Will it grow inside form?

apps/designer/app/designer/shared/assets/use-assets.ts
```diff
@@ -93,12 +101,28 @@ export const useAssets = (type: AssetType) => {
           toast.error(error.message);
         }
       });
+
+    for (const formData of formsData) {
+      submit<UploadData>(
+        formData,
+        {
+          method: "post",
```
istarkov, line 109: In all RFCs `POST` `GET` uppercased

kof: Yeah, it gets converted I believe in the end, but since it works, in the end it will become a mixture anyways, to enforce it more or less we probably need to abstract it away with a const or something or just ignore it and live with inconsistency

kof: checked remix docs, even they are inconsistent

kof: it should have never worked both ways to be consistent, I suggest just to ignore it

kof: in HTML lowercase is allowed https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form

kof: actually its not just allowed in html, its how its suggested

apps/designer/app/designer/shared/assets/asset-upload.tsx
kof: good catch, it has no need to grow it seems

TrySound (approved):

apps/designer/app/designer/features/sidebar-left/panels/assets/assets.tsx
```diff
@@ -2,20 +2,21 @@ import { ImageIcon } from "@webstudio-is/icons";
...
+import { Flex } from "@webstudio-is/design-system";
 
 export const TabContent = ({
   onSetActiveTab,
 }: {
   onSetActiveTab: (tabName: TabName) => void;
 }) => {
   return (
-    <>
+    <Flex direction="column" css={{ height: "100%" }}>
```
istarkov, line 13: `flex: 1` probably having its flex children?

apps/designer/app/designer/shared/assets/use-assets.ts
```diff
@@ -1,56 +1,89 @@
...
+
+export type UploadData = FetcherData<ActionData>;
+
+const toPreviewAssets = (
+  formsData: Array<FormData>
+): Promise<PreviewAsset[]> => {
+  const assets: Array<Promise<PreviewAsset>> = [];
+  for (const formData of formsData) {
+    for (const entry of formData) {
+      const file = entry[1] as File;
```
istarkov, line 51: as I remember `instanceof` operator should work with file
```js
var a= new File([23], '2323')
if (a instanceof File) {}
```

apps/designer/app/designer/shared/assets/use-assets.ts
```diff
@@ -1,56 +1,89 @@
...
+          }
 
-        resolve({
-          format: file.type.split("/")[1],
-          path: String(dataUri),
-          name: file.name,
-          id: ObjectID().toString(),
-          status: "uploading",
+          resolve({
+            format: file.type.split("/")[1],
```
istarkov, line 61: Is it always in a form like `image/jpg` etc? Is it documented?

istarkov (approved):

apps/designer/app/designer/features/sidebar-left/panels/assets/assets.tsx
kof: you mean to try flex: 1 instead of height?

istarkov: yes

apps/designer/app/designer/shared/assets/use-assets.ts
kof: got it

apps/designer/app/designer/shared/assets/use-assets.ts
kof: mime type? https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

Was so far with my tests, though not 100% sure, potentially need better logic

apps/designer/app/designer/features/sidebar-left/panels/assets/assets.tsx
kof: tried, didn't work

apps/designer/app/designer/shared/assets/use-assets.ts
kof: regular stuff like images and fonts works

PR #570, opened 2022-12-08
Able to use remote dev environments with IP address

andrasbacsai:
I'm using a remote development server and accessing it with an IP address. This minor fix allows me to use it.

### DIFF
--- a/apps/designer/app/shared/router-utils/build-params.ts
+++ b/apps/designer/app/shared/router-utils/build-params.ts
@@ -47,7 +47,11 @@ export const getBuildOrigin = (
       return `http://${host.split(".").pop()}`;
     }
     // If we're not on localhost, we're probably using ip address - useful for remote development
-    if (host.split(".").length === 4) {
+    if (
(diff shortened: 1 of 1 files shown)


istarkov (approved):

apps/designer/app/shared/router-utils/build-params.ts
```diff
@@ -42,8 +42,14 @@ export const getBuildOrigin = (
...
   // Local development special case
   const host = getRequestHost(request);
-  if (env.NODE_ENV === "development" && isLocalhost(host)) {
-    return `http://${host.split(".").pop()}`;
+  if (env.NODE_ENV === "development") {
+    if (isLocalhost(host)) {
+      return `http://${host.split(".").pop()}`;
+    }
+    // If we're not on localhost, we're probably using ip address - useful for remote development
+    if (host.split(".").length === 4) {
```
istarkov, line 50: Anyway better to check that its IP with regexp, as its not that rare to use domains like
app.blabla.mydomain.com for dev branches etc

andrasbacsai: Added regex.

Issue #482, opened 2022-11-20
Eslint rule for index modules


kof:
It should prevent importing specific modules from a folder that has an index module. Index is like a public interface. If any export is not exported from index - it's private.

This way we can export things from modules that are for reuse between other modules from the same folder but NOT as a public interface. Basically This is how we can have private modules.

kof:
Part of https://github.com/sindresorhus/eslint-plugin-unicorn/

andrasbacsai:
Do you mean for example to:
- prevent `import { prisma } from "@webstudio-is/prisma-client"` 
- allow `import { prisma } from "@webstudio-is/prisma-client/src/prisma"`?

kof:
When a folder has an index module, imports should never reach into the folder by using path, they are only allowed to import what's exported from that index. Index module is the public interface of that folder.

andrasbacsai:
I used `import/no-internal-modules` that prevents importing modules from sub-folders but only allows from index files. If this is not suitable, I think we need to write a custom eslint rule.

kof:
Is this allowing to use internal modules when no index.ts provided? otherwise sounds like what we need

andrasbacsai:
Unfortunately, not. We need to allow those internal modules.

PR #573, opened 2022-12-08
Eslint improvements

andrasbacsai:
[This](https://github.com/webstudio-is/webstudio-designer/issues/482) is more complicated to implement. 

I used `import/no-internal-modules` that prevents importing modules from sub-folders but only allows from index files. If this is not suitable, I think we need to write a custom eslint rule.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/canvas.tsx
+++ b/apps/designer/app/canvas/canvas.tsx
@@ -65,7 +65,9 @@ const useElementsTree = () => {
   const onChangeChildren: OnChangeChildren = useCallback(
     (change) => {
       store.createTransaction([rootInstanceContainer], (rootInstance) => {
-        if (rootInstance === undefined) return;
+        if (rootInstance === undefined) {
+          return;
+        }
 
         const { instanceId, updates } = change;
         utils.tree.setInstanceChildrenMutable(
@@ -80,7 +82,9 @@ const useElementsTree = () => {
   );
 
   return useMemo(() => {
-    if (rootInstance === undefined) return;
+    if (rootInstance === undefined) {
+      return;
+    }
 
     return createElementsTree({
       sandbox: true,
@@ -182,7 +186,9 @@ export const Canvas = ({ data }: CanvasProps): JSX.Element | null => {
   const isPreviewMode = useSubscribePreviewMode();
   const elements = useElementsTree();
 
-  if (elements === undefined) return null;
+  if (elements === undefined) {
+    return null;
+  }
 
   if (isPreviewMode || isDesignerReady === false) {
     return (
--- a/apps/designer/app/canvas/shared/breakpoints.ts
+++ b/apps/designer/app/canvas/shared/breakpoints.ts
@@ -33,7 +33,9 @@ const usePublishBreakpoints = () => {
 const useBreakpointChange = () => {
   useSubscribe("breakpointChange", (breakpoint) => {
     store.createTransaction([breakpointsContainer], (breakpoints) => {
-      const foundBreakpoint = breakpoints.find(({ id }) => id == breakpoint.id);
+      const foundBreakpoint = breakpoints.find(
+        ({ id }) => id === breakpoint.id
+      );
       if (foundBreakpoint) {
         foundBreakpoint.label = breakpoint.label;
         foundBreakpoint.minWidth = breakpoint.minWidth;
@@ -52,9 +54,11 @@ const useBreakpointDelete = () => {
     store.createTransaction(
       [breakpointsContainer, rootInstanceContainer],
       (breakpoints, rootInstance) => {
-        if (rootInstance === undefined) return;
+        if (rootInstance === undefined) {
+          return;
+        }
 
-        const index = breakpoints.findIndex(({ id }) => id == breakpoint.id);
+        const index = breakpoints.findIndex(({ id }) => id === breakpoint.id);
         if (index !== -1) {
           breakpoints.splice(index, 1);
         }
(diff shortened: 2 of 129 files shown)


apps/designer/app/canvas/shared/breakpoints.ts
```diff
@@ -33,7 +33,9 @@ const usePublishBreakpoints = () => {
 const useBreakpointChange = () => {
   useSubscribe("breakpointChange", (breakpoint) => {
     store.createTransaction([breakpointsContainer], (breakpoints) => {
-      const foundBreakpoint = breakpoints.find(({ id }) => id == breakpoint.id);
+      const foundBreakpoint = breakpoints.find(
+        ({ id }) => id === breakpoint.id
+      );
```
kof, line 38: why did this change?

apps/designer/app/canvas/canvas.tsx
```diff
@@ -182,7 +186,9 @@ export const Canvas = ({ data }: CanvasProps): JSX.Element | null => {
   const isPreviewMode = useSubscribePreviewMode();
   const elements = useElementsTree();
 
-  if (elements === undefined) return null;
```
kof, line 185: did the rule make it completely impossible to make single line if?

apps/designer/app/canvas/shared/use-track-hovered-element.ts
```diff
@@ -4,7 +4,7 @@ import { useRootInstance } from "~/shared/nano-states";
 
 // @todo: importing normally doesn't work in Jest for some reason
 // import { useDebounce } from "react-use";
-import useDebounce from "react-use/lib/useDebounce";
+import useDebounce from "react-use/lib/useDebounce"; // eslint-disable-line
```
kof, line 7: I think we already fixed it and  can import normally

packages/design-system/src/components/__DEPRECATED__/text.tsx
```diff
@@ -1,6 +1,6 @@
 import { styled } from "../../stitches.config";
 
-export const __DEPRECATED__Text = styled("div", {
+export const DEPRECATEDText = styled("div", {
```
kof, line 3: hmm it would be nice to support `__DEPRECATED_Text__` btw I just noticed it was supposed to have the underscores like I this

kof: Otherwise if its impossible then I would just call it `DeprecatedText` camel case

kof (approved):

apps/designer/app/routes/rest/pages.$projectId.ts
```diff
@@ -13,8 +13,8 @@ import { type FetcherData, makeFieldError } from "~/shared/form-utils";
 
 const nonEmptyString = z
   .string({
-    invalid_type_error: "Can't be empty",
-    required_error: "Can't be empty",
+    invalid_type_error: "Can't be empty", // eslint-disable-line
+    required_error: "Can't be empty", // eslint-disable-line
```
rpominov, line 17: Not a good idea to disable all rules. What if in the same line we'll break some other rule that we don't want to be disabled?

Same in other places in the diff.

packages/prisma-client/src/prisma.ts
```diff
@@ -1,5 +1,5 @@
 import { PrismaClient, Prisma } from "@prisma/client";
-import { PrismaClientKnownRequestError, Decimal } from "@prisma/client/runtime";
+const { PrismaClientKnownRequestError, Decimal } = Prisma;
```
rpominov, line 2: Is this also because of `import/no-internal-modules`? It seem like a bad idea to me to use it if it forces us to do this.

yarn.lock
```diff
@@ -9096,6 +9153,40 @@ escodegen@^2.0.0:
...
+    resolve "^1.20.0"
+
+eslint-module-utils@^2.7.3:
+  version "2.7.4"
+  resolved "https://registry.yarnpkg.com/eslint-module-utils/-/eslint-module-utils-2.7.4.tgz#4f3e41116aaf13a20792261e61d3a2e7e0583974"
+  integrity sha512-j4GT+rqzCoRKHwURX7pddtIPGySnX9Si/cgMI5ztrcqOPtk5dDEeZ34CQVPphnqkJytlc97Vuk05Um2mJ3gEQA==
+  dependencies:
+    debug "^3.2.7"
+
+eslint-plugin-import@^2.26.0:
```
TrySound, line 9171: This plugin has a lot of big dependencies

istarkov (approved):

apps/designer/app/canvas/shared/breakpoints.ts
andrasbacsai: It is changed from `==` to `===`. Or you asking the styling?

apps/designer/app/canvas/shared/use-track-hovered-element.ts
andrasbacsai: We did not. Tests fail in case of `{ useDebouce }`.

packages/prisma-client/src/prisma.ts
andrasbacsai: It forces, but we can make exceptions where possible like we do for the followings:
```js
 allow: [
          "**/server",
          "@lexical/react/*",
          "__generated__/*",
          "react-hot-toast/headless",
 ],
```

apps/designer/app/canvas/canvas.tsx
andrasbacsai: Several [options](https://eslint.org/docs/latest/rules/curly) are available to force it only in some instances. 
Forcing for every case is reasonable to have consistency.

yarn.lock
andrasbacsai: Yes, but we need it at the moment. It does not cover 100% of our needs, but close to it. Probably change it to a custom `eslint` rule later on.

PR #576, opened 2022-12-09
Fix media query specificity

kof:
We haven't been sorting the media queries correctly

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-engine/src/core/css-engine.test.ts
+++ b/packages/css-engine/src/core/css-engine.test.ts
@@ -5,7 +5,14 @@ const style0 = {
 } as const;
 
 const mediaRuleOptions0 = { minWidth: 0 } as const;
-const mediaId = "0";
+const mediaId0 = "0";
+
+const style1 = {
+  display: { type: "keyword", value: "flex" },
+} as const;
+
+const mediaRuleOptions1 = { minWidth: 300 } as const;
+const mediaId1 = "1";
 
 describe("CssEngine", () => {
   let engine: CssEngine;
@@ -14,7 +21,7 @@ describe("CssEngine", () => {
     engine = new CssEngine();
   });
 
-  test("use default media rule when there is no matching one registrered", () => {
+  test("use default media rule when there is no matching one registered", () => {
     engine.addStyleRule(".c", {
       style: style0,
       breakpoint: "x",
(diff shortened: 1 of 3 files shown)


istarkov:
Tested with bug we have in production, it solves it. But I think the media queries we generate look strange at least from my side.
Usually, if we are "mobile first" we don't have media query for mobile resolution

i.e. tailwind example
```
<div class="w-4 sm:w-6">ss</div>
```
generates.

```
.w-4 {
  width: 1rem;
}

@media (min-width: 640px) {
  .sm\:w-6 {
    width: 1.5rem;
  }
}
```

so mobile in the absense of `sm: lg:` applied everywhere 

We on mobile generate

```
@media all and (min-width: 360px) {
  [data-ws-id="root"] {
      flex-direction: var(--flexDirection-root, column);
      align-items: var(--alignItems-root, center);
      justify-content: var(--justifyContent-root, center);
      column-gap: var(--columnGap-root, normal);
      display: var(--display-root, flex);
  }
}
```

For me, it looks strange, what if the resolution is less than 360px?

packages/css-engine/src/core/css-engine.test.ts
```diff
@@ -35,25 +42,57 @@ describe("CssEngine", () => {
...
-      breakpoint: mediaId,
+      breakpoint: mediaId0,
     });
     // Default media query should allways be the first to have the lowest source order specificity
     expect(engine.cssText).toMatchInlineSnapshot(`
-      "@media all {
+      "@media all and (min-width: 0px) {
+        .c1 { color: blue }
+      }
+      @media all {
```
istarkov, line 55: In queries below `@media all {` is first, comment says that default should be first but here its 2nd

packages/css-engine/src/core/rules.ts
```diff
@@ -63,11 +63,25 @@ export type MediaRuleOptions = {
...
 export class MediaRule {
-  #options: MediaRuleOptions;
+  // Sort media rules by minWidth.
+  // Needed to ensure that more specific media rules are inserted after less specific ones.
+  // So that they get a higher specificity.
+  static sort(mediaRules: Iterable<MediaRule>) {
+    return Array.from(mediaRules).sort((a, b) => {
+      if (
+        a.options.minWidth === undefined ||
+        b.options.minWidth === undefined
```
istarkov, line 73: So it wouldn't sort but just reverse the below examples, looks like a bug as IMO both arrays after sorting should give same value
```
[{ options: { minWidth: undefined } }, { options: { minWidth: 1 } }]
// and 
[{ options: { minWidth: 1 } }, { options: { minWidth: undefined } }]
```

istarkov (changes requested):

packages/css-engine/src/core/css-engine.test.ts
kof: huh that's a bug, thank you

packages/css-engine/src/core/rules.ts
kof: added a test to ensure that

kof:
> For me, it looks strange, what if the resolution is less than 360px?

God question, I assumed its impossible, but if you are targeting a watch or something - yes, possible, our breakpoints are editable, so I think its a special case where a user might want to edit them?

I didn't set the minWidth from 0  there because the UI didn't make sense to go below  that narrow resolution.

packages/css-engine/src/core/rules.ts
```diff
@@ -65,11 +65,22 @@ export type MediaRuleOptions = {
 };
 
 export class MediaRule {
-  #options: MediaRuleOptions;
+  // Sort media rules by minWidth.
+  // Needed to ensure that more specific media rules are inserted after less specific ones.
+  // So that they get a higher specificity.
+  static sort(mediaRules: Iterable<MediaRule>) {
+    return Array.from(mediaRules).sort((ruleA, ruleB) => {
+      return (
```
istarkov, line 73: -Infinity - (-Infinity) = NaN
Looks like undefined behavior?

kof: oh amazing behavior from javascript, will switch to -Number.MAX_SAFE_INTEGER

istarkov (approved):

PR #578, opened 2022-12-09
Add new RGB Style, then fix typescript errors

istarkov:
The reason of use opaque types, as otherwise props like `value` reused everywhere.

Implemented same as CssValueInputLogic - during input editing nothing changed.
Error is shown on Enter click. Value resetted on blur.

### DIFF
--- a/apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
+++ b/apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
@@ -3,6 +3,7 @@ import { toValue } from "@webstudio-is/css-engine";
 import { getFinalValue } from "../../shared/get-final-value";
 import type { ControlProps } from "../../style-sections";
 import { ColorPicker } from "../../shared/color-picker";
+import { colord } from "colord";
 
 export const ColorControl = ({
   currentStyle,
@@ -10,21 +11,49 @@ export const ColorControl = ({
   setProperty,
   styleConfig,
 }: ControlProps) => {
-  const value = getFinalValue({
+  let value = getFinalValue({
     currentStyle,
     inheritedStyle,
     property: styleConfig.property,
   });
+
   if (value === undefined) {
     return null;
   }
+
   const setValue = setProperty(styleConfig.property);
+
+  if (value.type !== "rgb") {
+    // Support previously set colors
+    const colordValue = colord(toValue(value));
+
+    if (colordValue.isValid()) {
+      const rgb = colordValue.toRgb();
+      value = {
+        type: "rgb",
+        r: rgb.r,
+        g: rgb.g,
+        b: rgb.b,
+        alpha: rgb.a ?? 100,
+      };
+    } else {
+      // @todo what to show as default?
+      // Default to black
+      value = {
+        type: "rgb",
+        r: 0,
+        g: 0,
+        b: 0,
+        alpha: 100,
+      };
+    }
+  }
 
   return (
     <Flex align="center" css={{ gridColumn: "2/4" }} gap="1">
       <ColorPicker
         id={styleConfig.property}
-        value={toValue(value)}
+        value={value}
         onChange={(value) => {
           setValue(value, { isEphemeral: true });
         }}
--- a/apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
+++ b/apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
@@ -1,4 +1,5 @@
 import { Flex } from "@webstudio-is/design-system";
+import { toValue } from "@webstudio-is/css-engine";
 import { getFinalValue } from "../../shared/get-final-value";
 import type { ControlProps } from "../../style-sections";
 import { ColorPicker } from "../../shared/color-picker";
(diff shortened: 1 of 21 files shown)


apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
```diff
@@ -1,29 +1,59 @@
...
+    const colordValue = colord(toValue(value));
+
+    if (colordValue.isValid()) {
+      const rgb = colordValue.toRgb();
+      value = {
+        type: "rgb",
+        r: rgb.r,
+        g: rgb.g,
+        b: rgb.b,
+        alpha: rgb.a ?? 100,
```
kof, line 37: alpha is either percentage or 0-1, if you don't use percentage then maybe we should use 1?

apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
```diff
@@ -1,29 +1,59 @@
...
+
   if (value === undefined) {
     return null;
   }
+
   const setValue = setProperty(styleConfig.property);
 
+  if (value.type !== "rgb") {
+    // Support previously set colors
+    const colordValue = colord(toValue(value));
```
kof, line 28: I guess this is fine, but we could also write a migration script and then remove it?

kof: also for now I would add an example of a value we know is in the database

kof: a reader doesn't know it was previously saved as a keyword

istarkov: Will be hard to convert keyword to rgb, we will need to know all "color" properties

apps/designer/app/designer/features/style-panel/shared/color-picker.tsx
```diff
@@ -1,5 +1,8 @@
 import { useState } from "react";
+import { colord, type RgbaColor } from "colord";
 import { ColorResult, RGBColor, SketchPicker } from "react-color";
+import type { RGBValue } from "@webstudio-is/css-data";
```
kof, line 4: lets use strict camel case (forgot the proper name): RgbValue

packages/css-data/src/schema.ts
```diff
@@ -16,70 +16,86 @@ type CustomProperty = `--${string}`;
...
 export type KeywordValue = z.infer<typeof KeywordValue>;
 
-export const FontFamilyValue = z.object({
+const FontFamilyValue = z.object({
   type: z.literal("fontFamily"),
   value: z.array(z.string()),
 });
 export type FontFamilyValue = z.infer<typeof FontFamilyValue>;
 
+const RGBValue = z.object({
```
kof, line 44:
```suggestion
const RgbValue = z.object({
```

packages/css-engine/src/core/to-value.ts
```diff
@@ -39,5 +44,26 @@ export const toValue = (
...
+
+  if (value.type === "rgb") {
+    return `rgba(${value.r}, ${value.g}, ${value.b}, ${value.alpha})`;
+  }
+
+  // Will give ts error in case of missing type
+  assertUnreachable(value, `Unknown value type`);
+
+  // Will never happen
+  throw new Error("Unknown value type");
```
kof, line 68: hmm why do you need both, the above should already throw

packages/css-engine/src/core/to-value.ts
```diff
@@ -39,5 +44,26 @@ export const toValue = (
...
+  if (value.type === "invalid") {
+    return value.value;
+  }
+
+  if (value.type === "unset") {
+    return value.value;
+  }
+
+  if (value.type === "rgb") {
+    return `rgba(${value.r}, ${value.g}, ${value.b}, ${value.alpha})`;
```
kof, line 61: with your alpha as a value from 0-100 this wasn't supposed to work as you need to add % to it, but I would keep it this way and instead use numbers from 0 to 1, because they don't require %, wdyt?

apps/designer/app/designer/features/style-panel/controls/color/color-control.tsx
kof: actually no, there are more, but the list is not super long, our search shows you them all:
Screenshot 2022-12-09 at 15 32 47

kof: since we don't use shorthands, all color properties are easy to identify

kof: but its def. separate pr

packages/css-engine/src/core/to-value.ts
istarkov: ts don't know that function always throws
image

There is another exhaustive check variant exists:

```
  // @ts-expect-error check value is never
  value.type;
```
In that case, will be just a final throw, but such a check looks 100x dirtier for uninitiated

kof: yes, so keep the one that works with typescript, remove the second one?

Issue #29, opened 2022-05-02
Explain where the value comes from and allow resetting it


kof:
When clicking on the property name

- [x] show a popover that explains the property was defined: 
  - inherited from parents
  - set locally
  - token (bound to a token, this will come later once we have done tokens)
- [x] show a link to a full mdn article (we have the links from mdn data)
- [x] ideally show a short simple description (probably has to be written or copied from mdn partially) - lets leave it for later
- [x] reset button that will unset the local value and the value becomes either inherited or defined by default styles or browser styles

Since we don't have designs for it yet, I am attaching examples from webflow. We will have a new design later, it seems easy to change.

<img width="207" alt="Screenshot 2022-11-30 at 12 46 44" src="https://user-images.githubusercontent.com/52824/204800831-e32c27a2-07cf-4533-aee6-a6be2b035c82.png">

<img width="213" alt="Screenshot 2022-11-30 at 12 46 24" src="https://user-images.githubusercontent.com/52824/204800824-f2ab2d2c-88ef-413f-88ec-4a02baaa34e4.png">

kof:
- [ ] reset spacing properties

cc @taylornowotny it seems like we don't have a design for where to put the spacing property reset and also the value origins info

taylornowotny:
> cc @taylornowotny it seems like we don't have a design for where to put the spacing property reset and also the value origins info

Well spacing values don't inherit so there's no need to show the value origins. 

To reset the the property the user can option+click the value.

We may also need to show a clickable reset button. I can add this below the input that pops up. I'll add this to my to-do list under tooltips.
<img width="237" alt="Screenshot 2022-12-23 at 11 16 57 AM" src="https://user-images.githubusercontent.com/86495896/209396411-1c9b2326-7d29-4747-a6d0-d776c1ad1dd3.png">

taylornowotny:
In case it wasn't clear, I am going to re-design tooltips to address the needs outlined in this issue. I am tracking what we need, this is just on hold while I work on tokens

TrySound:
Implementation is ready

kof:
@taylornowotny this is waiting for design

taylornowotny:
Tracking

taylornowotny:
Design is done! #1730

PR #583, opened 2022-12-10
Add proprety reset button in popover

TrySound:

Screenshot 2022-12-10 at 12 03 54

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/shared/styles.ts
```diff
@@ -180,6 +180,12 @@ export const useCssRules = ({
       for (property in dynamicStyle) {
         rule.styleMap.set(property, dynamicStyle[property]);
       }
+      // delete previously rendered properties when reset
+      for (const property of rule.styleMap.keys()) {
+        if (dynamicStyle[property] == null) {
```
kof, line 185: it can never be null or can it? I saw in types it can't but reading this it looks like it can :)

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -38,6 +38,11 @@ const FlexChildSectionAlign = (props: RenderCategoryProps) => {
       <PropertyName
         property={sectionStyle.alignSelf?.styleConfig.property}
         label="Align"
+        onReset={() => {
+          sectionStyle.alignSelf.setProperty(
+            sectionStyle.alignSelf?.styleConfig.property
+          )({ type: "unset", value: "" });
```
kof, line 44: why do we need a value if we are unsettling?

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -1,39 +1,89 @@
...
+type PropertyProps = {
+  property: StyleProperty | StyleProperty[];
+  label: string;
+  onReset: () => void;
+};
+
+export const PropertyName = ({ property, label, onReset }: PropertyProps) => {
   const isCurrentBreakpoint = useIsFromCurrentBreakpoint(property);
+  const [isOpen, setIsOpen] = useState(false);
+  const isPopoverEnabled = isCurrentBreakpoint;
```
kof, line 28: Since we use "set" keyword in the design system, maybe we should call it isSet and the other options isCascaded and isInherited? If you agree I would rename it all including that function

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -1,39 +1,89 @@
...
+
+  const labelElement = (
+    <Label
```
kof, line 31: Would it make sense to create a PropertyLabel component that would have same props as the icon button for colors?

based on the latest update from Taylor (https://share.descript.com/view/POUPnkVZUre) we would be using prop color=preset|inherited|default

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -1,39 +1,89 @@
...
+  if (isPopoverEnabled) {
+    return (
+      <Flex align="center">
+        <Popover modal open={isOpen} onOpenChange={setIsOpen}>
+          <PopoverTrigger asChild aria-label="Show proprety description">
+            {labelElement}
+          </PopoverTrigger>
+          <PopoverPortal>
+            <PopoverContent align="start" onClick={() => setIsOpen(false)}>
+              <Flex align="start" css={{ padding: "$spacing$3 $spacing$4" }}>
```
kof, line 60: nit: we should use less shorthand properties and more longhand or something more practical for this case: `px: '$spacing$4', py: '$spacing$3'`

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: oh lol I just realized we are inside the PropertyName component and that's exactly what's happening, but its color aware itself, ignore my comment above

packages/css-engine/src/core/rules.ts
```diff
@@ -15,6 +15,14 @@ class StylePropertyMap {
...
     return this.#styleMap.has(property);
   }
+  keys() {
+    return this.#styleMap.keys();
+  }
+  delete(property: StyleProperty) {
+    this.#styleMap.delete(property);
+    this.#isDirty = true;
+    this.onChange?.();
+  }
```
kof, line 25: lets add a test for each

packages/project/src/shared/tree-utils/set-instance-style.ts
```diff
@@ -22,7 +22,11 @@ export const setInstanceStyleMutable = (
   }
 
   for (const update of updates) {
-    cssRule.style[update.property] = update.value;
+    if (update.value.type === "unset") {
+      delete cssRule.style[update.property];
+    } else {
+      cssRule.style[update.property] = update.value;
+    }
```
kof, line 29: if we keep early return coding style, in this case I would write  a `continue`
```suggestion
    if (update.value.type === "unset") {
      delete cssRule.style[update.property];
      continue;
    }
    cssRule.style[update.property] = update.value;
```

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: lets add `isFeatureEnabled('reset')` and add reset = false to feature-flags/flags so that its not on production for now until we get the design

on preview we can enable it by adding "reset" to the FEATURES env variable, same locally in .env

I am wondering if we should just have an * in preview to avoid the need to add and remove new flags in env and just always have all things enabled in preview

kof (changes requested):
mostly small changes

kof:
- [ ] we need a hover state for the property name when its set/inherited, it looks like hover is defined in [figma](https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=4%3A3274&t=e1fHzKZPDyeTnP7L-3) but no style change @taylornowotny

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -1,39 +1,89 @@
...
+                <Button onClick={onReset}>
+                  <UndoIcon /> &nbsp; Reset
+                </Button>
+              </Flex>
+              <Separator />
+              <Box css={{ padding: "$spacing$3 $spacing$4" }}>
+                <Text color="hint">
+                  Resetting will revert to initial or inherited value
```
kof, line 68: Since we are unsetting the property, we are essentially going back to:
- inherited value if there is one
- cascaded value if there is one (when it was defined on a different breakpoint that impacts the current breakpoint)
- preset value if there is one
- browser initial value (https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value)  

If we make it entirely correct, we need to actually check what we would be resetting to:

e.g. we could check if there is an inherited value and if there is: "Resetting will revert to inherited value" and do the same with cascaded, preset and browser initial. 

That's what webflow does, except that they still call both cascaded value and inherited value "inherited".

packages/project/src/shared/tree-utils/set-instance-style.ts
TrySound: I honestly don't like to put it in every condition. Early return works well when we have only return or continue alone. Though here is symmetrical logic I would expect to see on the same depth.

TrySound:
This PR addresses only logic, don't want to put here design changes

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: Yes, I left this for another PR caz reset logic alone requires a lot of changes

kof: Screenshot 2022-12-10 at 10 23 57
Screenshot 2022-12-10 at 10 24 07

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
TrySound: Because it was defined in types this way and some logic relied on "value" prop.
cc @istarkov

apps/designer/app/canvas/shared/styles.ts
TrySound: I never use `in` operator :)

kof:
- [ ] show where the value will come from when you reset

apps/designer/app/canvas/shared/styles.ts
kof: I think using === undefined is even better here

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
kof: lol looks like I defined it 2month ago, no idea what I was thinking since we are not using that value when we delete the property

kof: @istarkov if you don't see why we need that value, I would remove it

packages/project/src/shared/tree-utils/set-instance-style.ts
kof: You might be right about it, idk, I was just trying to make it all consistent and basically stop using `else`. I guess I am trying to avoid else all together because I got burned by complex if/else statements.

In this particular case either way its readable, so up to you, unless we ban  `else` by linter haha

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: Where FEATURES env passed from?

kof: environment variable, added asterix support here https://github.com/webstudio-is/webstudio-designer/pull/586

kof: will enable asterix on preview, so we never have to enable them manually there

kof: set it to asterix locally too, for convenience

kof: environment variable is used to override, but flags.ts file is used to set the default

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
istarkov: I think value was added to just remove errors on `String(style.value)` imo now can be safely removed as I think I changed all the places where it was used

istarkov:
Its same behavioir as I see for all sizes but here it looks even worse. We are showing value in `px` even if its not set.
Now see. Set for any `box` width: 14em

image


Then reload designer.

And do reset, instead of showing original value `360px` (which is not original) I see `196px`

image

So measurement of width done immediately before reset or like

istarkov:
I'm not sure that its a good idea to use `unset` to remove value. But it depends.
Usually, we use `unset` to remove a style from the element and to force it to use an initial or inherited style.

```html
<div class="root">
  <div class="a">A class</div>
  <div class="b">B class</div>  
</div>
```

```css
.a {
  color: red;
}

/* style I don't control but want to reset */
.b {
  color: red;
}

.root {
  color: yellow;
}

/* I want b to have root color */
.root .b {
  color: unset;
}
```

Using `unset` for real prop removal assumes that we always control all css properties. 
I just afraid of what if we really will need to set `unset` value, will it be `{type:  "keyword", value: "unset" }`

Also the more I look into the more I feel that its not StyleValue but some action.

kof:
What if instead of using type: unset, we add a function removeProperty, so we don't have to use setProperty at all for this, it makes it logical and removeProperty exists in the DOM as well on `style` object

TrySound:
How to represent it in "updates" list?

kof:
`StyleUpdate` will have to support some value type that means "delete"

e.g.

```
type StyleUpdate = {
  property: StyleProperty;
  value?: StyleValue;
};
```

e.g. if value === undefined its deletion, unless you really really want it to be distinct with null, I am fine with that in isolation

TrySound:
I don't like the idea of using optional syntax. Looks very easy to avoid it because not required.

kof:
Like I said we can use null for that if you want more expliciteness

istarkov (approved):
After check that where are no `unset` values in production db we can remove this 
https://github.com/webstudio-is/webstudio-designer/blob/b353cf1e9137803b59b192d9b55b672664b04c3a/packages/css-data/src/schema.ts#L61-L65

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -134,6 +137,18 @@ export const useStyleData = ({
     };
   };
 
+  const deleteProperty = (property: StyleProperty) => {
```
kof, line 140: removeProperty would be nicer since its called this way in the dom on style object

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -157,20 +172,33 @@ export const useStyleData = ({
...
       const nextStyle = updates.reduce(
-        (currentStyle, { property, value }) => {
-          // @todo
-          currentStyle[property] = value;
+        (currentStyle, update) => {
+          if (update.operation === "delete") {
+            delete currentStyle[update.property];
+          }
+          if (update.operation === "set") {
+            // @todo
```
kof, line 199: what's the todo? forgotten?

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
kof: actually, removeProperty is on style object, but delete is on styleMap, which is a newer API, so lets keep it

kof (approved):
cosmetic comments, looking good!

kof:
- [ ] Found a visual bug - when I click reset, the control for that value quickly jumps right before rendering the new value


https://user-images.githubusercontent.com/52824/207019134-486d5fee-2453-4810-b746-5526b00e74d2.mp4

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -157,20 +172,33 @@ export const useStyleData = ({
...
         }
 
-        updates.push({ property, value });
+        updates.push({ operation: "set", property, value });
       };
       return setValue;
     };
 
+    const deleteProperty = (property: StyleProperty) => {
+      if (currentStyle === undefined) {
```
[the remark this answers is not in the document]
TrySound: I don't know why we have this check. Just took from setProperty

TrySound:
Looks like the issue with input rendering. I would fix in separate PR

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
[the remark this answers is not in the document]
TrySound: I would look into it when we will rework state management

Issue #579, opened 2022-12-09
ESC doesn't work on units dropdown in CssValueInput


kof:
1. Any unit value in style panel
2. Open units dropdown
3. Hit ESC - should close the unit, same as with keywords dropdown

It works in other select boxes like font-weight for example. Radix UI handles it out of the box.

TrySound:
> It works in other select boxes like font-weight for example.

Looks like it doesn't

PR #584, opened 2022-12-10
Close selects with escape

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/579

Looks like current version of radix-ui select do not handle escape press but gives us event. Hard to say if this is fixed as they have poor versioning and no way to see previous version.

So in the fix I removed open and onOpenChange props from select in design system and used them to react on escape keydown event.

On unit select there is already external open state which I used the same way.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/select.tsx
+++ b/packages/design-system/src/components/select.tsx
@@ -1,8 +1,9 @@
+import type { ReactNode, Ref, ComponentProps } from "react";
+import { forwardRef, useState } from "react";
 import * as SelectPrimitive from "@radix-ui/react-select";
-import React, { ReactNode, Ref } from "react";
+import { CheckIcon, ChevronDownIcon, ChevronUpIcon } from "@webstudio-is/icons";
 import { Grid } from "./grid";
 import { Box } from "./box";
-import { CheckIcon, ChevronDownIcon, ChevronUpIcon } from "@webstudio-is/icons";
 import { styled } from "../stitches.config";
 
 const StyledTrigger = styled(SelectPrimitive.SelectTrigger, {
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/select.tsx
```diff
@@ -135,20 +136,18 @@ const SelectItemBase = (
...
-  React.ComponentProps<typeof StyledTrigger>,
+  ComponentProps<typeof StyledTrigger>,
   "onChange" | "value" | "defaultValue"
 > & {
   options: Option[];
   defaultValue?: Option;
   value?: Option;
   onChange?: (option: Option) => void;
-  open?: boolean;
-  onOpenChange?: (open: boolean) => void;
```
kof, line 151: now we can't control the component from outside if we need to

kof:
1. It works on typography font weight property right now on main and prod, why does it work without this fix?
2. I think radix always handled it correctly, we are just probably preventing it somewhere before it handles it
3. Please add a stories to select component and see what it does by default

TrySound:
Doesn't work for on main and prod in chrome

packages/design-system/src/components/select.tsx
TrySound: We may never need to. Unused features often just add complexity

kof:
Here is a demo how I test it, 

https://share.descript.com/view/lkJ4JLpIJTp

TrySound:
idk I was able to close it only with shift+escape. Tab doesn't work for me too

TrySound:
Why do we have css value on display block and replace it with select on display: flex?

kof:
please add a story, I am sure it works

kof:
> Why do we have css value on display block and replace it with select on display: flex?

Bug, it should be a select in both cases

TrySound:
Alright, works in stories. Found it's because of react-hotkeys-hook. It's used all over the project with different configuration. IMO we should get rid of it because too easy to break anything.

Andarist:
What happens with the focus when you press Esc? Is it correctly returned to some trigger element?

TrySound:
When I press nothing happens until disable all useHotkeys escape overrides

kof:
@TrySound why does it work in other cases? e.g. font-weight select?

TrySound:
Doesn't work for me. Idk could be race condition with refs.

kof:
damn, now I test and it seeems that it works with units too:
1. open units with mouse but don't change the cursor with the mouse - esc works right away
2. if you move the mouse and highlight another item, first esc doesn't work, second esc works, I assume first escape unsets the value that was supposed to be set be onHighlight, but we don't update value on highlight from units (which is inconsistent with autocomplete, so the esc behavior now isn't right), maybe we should actually preview the value on canvas onHighlight?

TrySound:
The issue here is specifically hotkeys and we use old version which doesn't support custom root at all.

kof:
Lets create an issue for refactoring hotkeys?

PR #585, opened 2022-12-10
Unify Storybook config

Andarist:
This PR moves most of the Storybook configs into a single, private, package that can be imported/required by your packages. It aims to keep the config the same for all of them as much as possible - so, for example, all of the packages are now using Webpack 5 in their Storybooks.


.github/actions/ci-setup/action.yml
```diff
@@ -11,5 +11,5 @@ runs:
         cache: "yarn"
 
     - name: Install Dependencies
-      run: yarn install --frozen-lockfile --ignore-scripts
```
Andarist, line 14: I had to remove this to execute `patch-package`

apps/designer/package.json
```diff
@@ -74,15 +74,7 @@
   "devDependencies": {
     "@babel/core": "^7.20.2",
     "@remix-run/dev": "^1.6.4",
-    "@storybook/addon-actions": "^6.5.9",
```
Andarist, line 77: I couldn't find any reference to this one (`@storybook/addon-actions`) being used anywhere in the codebase - so I didn't include it in the common config anyhow (nor in any of the ones using the common config). Should I bring it back?

patches/@storybook+builder-webpack5+6.5.14.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/node_modules/@storybook/builder-webpack5/templates/virtualModuleModernEntry.js.handlebars b/node_modules/@storybook/builder-webpack5/templates/virtualModuleModernEntry.js.handlebars
```
Andarist, line 1: Previously most of the packages were using webpack 4 while the Designer app was using webpack 5. I unified this and now everything is using webpack 5. However, I've encountered some hiccups with this setup and had to include 2 patches in this template~ to work around them. One of them is already available in the beta/next version of Storybook, but the other one isn't - it got rejected in favor of some other PR. However, that other PR is included in the version of the Storybook that you are using and it doesn't fix the issue that I bumped into. I expect that this issue might be fixed in the beta/next version of Storybook or that it will get fixed there in the future. I don't want to migrate to the "unstable" Storybook 7 yet - so I didn't recheck with that at the moment. 

To avoid using `patch-package` to work around those we could either:
- try to switch back to webpack 4 for all the existing storybooks
- keep using webpack 5 for the Designer app and keep using webpack 4 for packages

Let me know if using `patch-package` here is OK or if you prefer one of the other mentioned solutions.

packages/storybook-config/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@webstudio-is/storybook-config",
+  "private": true,
+  "version": "0.0.0",
```
Andarist, line 4: side note: your `@webstudio-is/jest-config` (and other packages like that) are currently public and published to npm. They don't have to be.

.github/actions/ci-setup/action.yml
kof: I don't know what the implications are @andrasbacsai

patches/@storybook+builder-webpack5+6.5.14.patch
kof: is webpack 5 giving us advantages there? I heard they are working on vite support ... personally I would go for simplicity - one webpack version for all

but that depends if its not super slow and what bug you had with wp5, can we live with the bug?

kof (approved):
I have no actual knowledge on how to judge this PR

patches/@storybook+builder-webpack5+6.5.14.patch
Andarist: > is webpack 5 giving us advantages there? 

Probably not. Just note that webdesigner app was already using webpack 5 so switching it to webpack 4 could also cost some work as things could break somehow unexpectedly. When unifying the webpack version I was facing a dilemma either way, so I just tried to use the newer version (webpack 5). Even though I had to hack something around - it didn't take too much time so I didn't have to cave and switch back to webpack 4 ;p

At this point, I don't see a strong reason to switch back to webpack 4 - as the setup here seems to be working. We can always revisit this too. I will also be reporting that one "unsolved" issue with webpack 5 to the Storybook team, so I hope that they will land a PR fixing it at some point.

> I heard they are working on vite support

They have it in their org but outside of their main monorepo: https://github.com/storybookjs/builder-vite . I could try to switch to that in a follow-up PR if you'd like that.

> but that depends if its not super slow and what bug you had with wp5, can we live with the bug?

I patched the bug - so I think that you can live with it for now. it's just that the patch that I used to work around it isn't yet a part of the "next" Storybook. Their maintainers already expressed an interest in taking a look at my repro case though - so there is a hope that they will fix this in their "core" sometime in the future. Note that those patches wouldn't immediately benefit you though. From what I understand they are working on their v7 and the current stable v6 doesn't receive a lot of updates.

kof: would be awesome if that works

kof:
@TrySound I think this waits for your ✅

TrySound:
Sorry, missed notifications. Patch package complicates migration to pnpm. Maybe easier to switch to beta with vite support?

Andarist:
@TrySound doesnt pnpm have its own builtin patch-package? I mean - migrating to Vite is also an option but perhaps that could be done later

PR #594, opened 2022-12-10
Setup custom hotkeys for sidebar tree

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/579

useHotkeys blocks escape for whole app which breaks radix select as well. Here I replaced it with onKeyDown on tree root which localized all interactions.

Ideally remove useHokeys with all its implicit logic everywhere else.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/tree/tree.tsx
+++ b/packages/design-system/src/components/tree/tree.tsx
@@ -353,6 +353,8 @@ const useKeyboardNavigation = <Data extends { id: string }>({
     }
     if (event.key === " ") {
       setIsExpanded(selectedItem, !getIsExpanded(selectedItem));
+      // prevent scrolling
+      event.preventDefault();
     }
     if (event.key === "ArrowUp") {
       const index = flatCurrentlyExpandedTree.indexOf(selectedItem.id);


packages/design-system/src/components/tree/tree.tsx
```diff
@@ -333,50 +335,45 @@ const useKeyboardNavigation = <Data extends { id: string }>({
...
+  const rootRef = useRef<HTMLDivElement | undefined>();
+
+  const handleKeyDown = (event: ReactKeyboardEvent) => {
+    // skip if nothing is selected in the tree
+    if (selectedItem === undefined) {
+      return;
+    }
+
+    // prevent scrolling
+    event.preventDefault();
```
istarkov, line 347: probably not every key cause scrolling? Why to prevent it

istarkov (approved):

packages/design-system/src/components/tree/tree.tsx
TrySound: useHotkeys added both preventDefault and stopPropagation but maybe you are right

packages/design-system/src/components/tree/tree.tsx
```diff
@@ -333,50 +337,46 @@ const useKeyboardNavigation = <Data extends { id: string }>({
...
+    }
+
+    if (event.key === "ArrowRight" && getIsExpanded(selectedItem) === false) {
+      setIsExpanded(selectedItem, true);
+    }
+    if (event.key === "ArrowLeft" && getIsExpanded(selectedItem)) {
+      setIsExpanded(selectedItem, false);
+    }
+    if (event.key === " ") {
+      setIsExpanded(selectedItem, !getIsExpanded(selectedItem));
```
istarkov, line 355: space is well known scrolling button

packages/design-system/src/components/tree/tree.tsx
```diff
@@ -265,7 +265,11 @@ export const Tree = <Data extends { id: string }>({
       onScroll={dropHandlers.handleScroll}
     >
       <Box
+        // styled should support passing { current: T | undefined } to ref
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore React.MutableRefObject cannot be passed to React.RefObject
```
kof, line 270: that's weird

kof (approved):
haven't tested, but generally I agree with the approach useHotKeys is more for global shortcuts, not for components

I would look why ts-ignore is needed and how to do it without it, I am sure it can be done without ts-ignore stuff

packages/design-system/src/components/tree/tree.tsx
rpominov: The error is correct, ref should be defined like this: `const rootRef = useRef<HTMLDivElement | null>(null);`

kof: that's normal yeah

kof: @TrySound please ask for help @Andarist in case of typescript issues

kof: fixed it https://github.com/webstudio-is/webstudio-designer/commit/1a66acb79c8c560b3a54b55512c2779991a68eff

PR #599, opened 2022-12-11
Add initial props

istarkov:
- [x] - Add inital props
- [x] - Use sorting from inital props
- [x] - prevent deleting/renaming props in initialProps
- [x] - do now allow rename props on existing props.
- [x] - Fix bug that incorrect prop value selected if prop is not in initial props
- [x] - Add tests

### DIFF
--- a/apps/designer/app/designer/features/props-panel/props-panel.tsx
+++ b/apps/designer/app/designer/features/props-panel/props-panel.tsx
@@ -118,6 +118,8 @@ type PropertyProps = {
   onDelete: (id: UserProp["id"]) => void;
   setCssProperty: SetProperty;
   currentStyle: Style;
+  required: boolean;
+  existingProps: string[];
 };
 
 const Property = ({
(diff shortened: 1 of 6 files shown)


TrySound:
Will be useful for links

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -217,6 +251,8 @@ export const PropsPanel = ({
     </Button>
   );
 
+  const existingProps = userProps.map((userProp) => userProp.prop);
```
kof, line 254: its not the initialProps?

apps/designer/app/designer/features/props-panel/use-props-logic.test.ts
```diff
@@ -453,4 +457,113 @@ describe("usePropsLogic", () => {
       ]
     `);
   });
+
+  test("should respect initialProps ordering", () => {
```
kof, line 461: my feeling is that testing hooks etc is always harder and more fragile than functions like getInitialProps

kof (approved):
works

apps/designer/app/designer/features/props-panel/props-panel.tsx
istarkov: Yes it's all props the user has on the screen. To avoid having 2 of the same props on screen.

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -60,19 +63,30 @@ const getRequiredProps = (
   const { component } = selectedInstanceData;
   const meta = getComponentMetaProps(component);
 
+  const props =
+    selectedInstanceData.props === undefined ||
+    selectedInstanceData.props.props.length === 0
+      ? []
+      : selectedInstanceData.props.props;
```
rpominov, line 70: if we just remove `|| selectedInstanceData.props.props.length === 0` will anything change?

Same below

apps/designer/app/designer/features/props-panel/use-props-logic.ts
```diff
@@ -115,16 +185,20 @@ export const usePropsLogic = ({
...
-      ...getRequiredProps(selectedInstanceData),
-    ],
-    "prop"
+  const [requiredProps] = useState<Array<UserProp>>(() =>
+    uniqBy(
+      [
+        ...getInitialProps(selectedInstanceData),
+        ...getRequiredProps(selectedInstanceData),
+        ...getPropsWithDefaultValue(selectedInstanceData),
+      ],
```
rpominov, line 194: If `getInitialProps` and `getRequiredProps` used only here, will it make sense to override with values from `selectedInstanceData.props` somewhere here rather than in each of these functions?

rpominov (commented):
A bit late, sorry :)

apps/designer/app/designer/features/props-panel/props-panel.tsx
```diff
@@ -128,34 +135,61 @@ const Property = ({
...
+              onChangePropName(name);
+            }
+          }}
+          onSubmit={(name) => {
+            if (existingProps.includes(name) === false) {
+              setError(undefined);
+              onChangePropName(name);
+              return;
+            }
+            setError(`Property "${name}" is already exists`);
```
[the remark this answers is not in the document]
istarkov: Error is possible as user can just type property

apps/designer/app/designer/features/props-panel/use-props-logic.ts
istarkov: I have feeling that everything in this logic will be changed, hard to say now. For now just prevented that props user added/removed after reload occured or moved

PR #602, opened 2022-12-12
Move `@webstudio-is/design-tokens` to `@webstudio-is/react-sdk` prod deps

Andarist:
This should fix the problem reported on this line: https://github.com/webstudio-is/webstudio-saas/actions/runs/3657427162/jobs/6181006430#step:6:52 , cc @andrasbacsai 

Even though [you only reach for a type](https://github.com/webstudio-is/webstudio-designer/blob/1a66acb79c8c560b3a54b55512c2779991a68eff/packages/react-sdk/src/tree/root.ts#L11), you still should declare this package among your dependencies - otherwise, it won't be installed by your consumers and thus the TS compilation will fail for them.

### DIFF
--- a/packages/react-sdk/package.json
+++ b/packages/react-sdk/package.json
@@ -35,7 +35,6 @@
     "@webstudio-is/generate-arg-types": "*",
     "@webstudio-is/jest-config": "*",
     "@webstudio-is/scripts": "*",
-    "@webstudio-is/design-tokens": "*",
     "babel-loader": "^8.2.5",
     "esbuild": "^0.14.25",
     "esbuild-node-externals": "^1.4.1",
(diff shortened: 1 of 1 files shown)


kof (approved):
damn, I added it as a dev dep knowing its only using a type ... this stuff is unpredictable

Andarist:
if your `src` imports a thing then it's likely not a dev dependency :p

kof:
types have always been dev/peer in my mind

Andarist:
That's a common "mistake" (I've done it on several occasions too).

You can get away with an optional peer dep but that's IMHO not a good DX (your consumers have to install the missing packages manually). 

Alternatively, you could juggle your packages and create something like `@webstudio-is/webdesigner-types`. You could keep all shared types there and put that as a dependency in all of your other packages.

Issue #603, opened 2022-12-12
CSS Display flex property become selected during input and then Enter has no effect


istarkov:
Bug also reproduced for all css-input fields with input text and menu (Enter fixed at #607)

https://user-images.githubusercontent.com/5077042/207059736-9497bdb3-f65d-4f2c-90f2-336eb38b02a2.mov

Everything below is moved here https://github.com/webstudio-is/webstudio-designer/issues/610 

When `flex` is typed in the menu it has 
<img width="61" alt="image" src="https://user-images.githubusercontent.com/5077042/207060180-ade708fd-e215-4a04-b695-0879a6e23bd7.png">
sign like it's already a current selection, what's not true.

![image](https://user-images.githubusercontent.com/5077042/207059790-04234c02-4b08-45ce-868f-c1f49fc790d9.png)

istarkov:
Must use same as others

kof:
I think this bug is not only for display property @TrySound, you will fix it for display by using a select control, but that's not solving the core of the issue here for other fields

istarkov:
I fixed Enter issue for other fields #607, but checkmark logic imo is wrong too

istarkov:
Check mark issue is moved here https://github.com/webstudio-is/webstudio-designer/issues/610

PR #607, opened 2022-12-12
Enter has no effect if menu is open and no value selected

istarkov:
Fixes Enter non-working part of #603, but not fixes Check mark showing logic.
The wrong value is used to compare the selected field, https://github.com/webstudio-is/webstudio-designer/blob/26d0d0c2a6b3fe21c2298e451ffea5adc350290e/packages/design-system/src/components/combobox.tsx#L203 as its compares with current item but not with already selected


TrySound (approved):

istarkov:
Check mark logic in all Combos must be discussed, and fixed in next PRs

rpominov (approved):

PR #612, opened 2022-12-12
Cache css

istarkov:
From MDN:
A modern best practice for static resources is to include version/hashes in their URLs, while never modifying the resources — but instead, when necessary, updating the resources with newer versions that have new version-numbers/hashes, so that their URLs are different. That's called the cache-busting pattern.

### DIFF
--- a/apps/designer/app/routes/$.tsx
+++ b/apps/designer/app/routes/$.tsx
@@ -5,6 +5,7 @@ import type {
   ErrorBoundaryComponent,
 } from "@remix-run/node";
 import { useLoaderData } from "@remix-run/react";
+import djb2a from "djb2a";
 import { InstanceRoot, Root } from "@webstudio-is/react-sdk";
 import { loadCanvasData } from "~/shared/db";
 import env, { type Env } from "~/env.server";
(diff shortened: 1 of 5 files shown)


apps/designer/app/routes/$.tsx
```diff
@@ -29,6 +30,11 @@ export const dynamicLinks: DynamicLinksFunction<CanvasData> = ({
 }) => {
   const searchParams = new URLSearchParams(location.search);
   searchParams.set("pageId", data.page.id);
+
+  // Break cache in case of css has changed
+  const cssHash = djb2a(JSON.stringify(data.tree));
```
kof, line 35: how much overhead would stringify be compared to having it upfront? if significant we could add that hash in project tree.ts loadById

kof: as in having the hash saved statically at publish time or so

kof: in this case we would be able to save both djb2a and stringify cost by simply doing this at publish time

kof: on the related note if we start publishing using a worker per project, CSS will become static, so hash can be made at publish time when we deploy the worker, so it would be done once anyways

Issue #608, opened 2022-12-12
Reset style property doesn't reflect in the control until reload


kof:
It resets the value correctly, but controls still render the old value until reload.

TrySound:
Which control eexactly, all of them?

kof:
I suspect you calcuated it before react actually applied the changes? Otherwise layout effect should help

Issue #600, opened 2022-12-12
When I click reset, the control for that value quickly jumps right before rendering the new value


kof:
- [ ] Found a visual bug - when I click reset, the control for that value quickly jumps right before rendering the new value

https://user-images.githubusercontent.com/52824/207019134-486d5fee-2453-4810-b746-5526b00e74d2.mp4

_Originally posted by @kof in https://github.com/webstudio-is/webstudio-designer/issues/583#issuecomment-1346213429_

TrySound:
All reset issues comes from the fact we compute values from dom instead of data which is much faster and bullet proof

kof:
One place where we can't compute from data with full certainty is when we don't have a value set in our data. We kinda need to know what browser value is.

PR #620, opened 2022-12-13
Fix property reset glitches

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/600 https://github.com/webstudio-is/webstudio-designer/issues/608

- wrapped getComputedStyle with requestAnimationFrame to recompute some styles correctly
- added default to color picker to avoid hiding while styles are recomputed
- added placeholder to select to avoid hiding while styles are recomputed

These issues comes from the fact we compute default styles from dom instead of data.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/shared/instance.ts
+++ b/apps/designer/app/canvas/shared/instance.ts
@@ -194,24 +194,27 @@ export const usePublishSelectedInstanceData = (treeId: Tree["id"]) => {
   const [allUserProps] = useAllUserProps();
 
   useEffect(() => {
-    // Unselects the instance by `undefined`
-    let payload;
-    if (instance !== undefined) {
-      const props =
-        allUserProps[instance.id] ??
-        utils.props.createInstanceProps({ instanceId: instance.id, treeId });
-      const browserStyle = getBrowserStyle(selectedElement);
-      payload = {
-        id: instance.id,
-        component: instance.component,
-        cssRules: instance.cssRules,
-        browserStyle,
-        props,
-      };
-    }
-    publish({
-      type: "selectInstance",
-      payload,
+    // some styles require nexxxt frame to provide new value
+    requestAnimationFrame(() => {
+      // Unselects the instance by `undefined`
+      let payload;
+      if (instance !== undefined) {
+        const props =
+          allUserProps[instance.id] ??
+          utils.props.createInstanceProps({ instanceId: instance.id, treeId });
+        const browserStyle = getBrowserStyle(selectedElement);
+        payload = {
+          id: instance.id,
+          component: instance.component,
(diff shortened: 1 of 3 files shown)


apps/designer/app/canvas/shared/instance.ts
```diff
@@ -194,24 +194,27 @@ export const usePublishSelectedInstanceData = (treeId: Tree["id"]) => {
...
-        cssRules: instance.cssRules,
-        browserStyle,
-        props,
-      };
-    }
-    publish({
-      type: "selectInstance",
-      payload,
+    // some styles require nexxxt frame to provide new value
+    requestAnimationFrame(() => {
```
kof, line 198: I think you might be fixing it in the wrong place. The core issue is that this logic doesn't depend on when the style updates were actually applied, we are probably dispatching the new instance before we actually set the styles, can you please verify this?

apps/designer/app/canvas/shared/instance.ts
```diff
@@ -193,6 +193,15 @@ export const usePublishSelectedInstanceData = (treeId: Tree["id"]) => {
   const [selectedElement] = useSelectedElement();
   const [allUserProps] = useAllUserProps();
 
+  // trigger style recomputing every time
+  // new updates for this node are sent from designer
+  const [styleKey, recomputeStyles] = useReducer((d) => d + 1, 0);
+  useSubscribe("updateStyle", ({ id }) => {
```
kof, line 199: still kindof not guaranteed that the styles have been applied when this even arrives, because if the logic delays the application, we won't have computed the style, I think you need something that will tell you "styles have been applied"

TrySound: We already have a lot of indirection here. Let's not make it worse.
I don't see other proper way to do this without huge refactoring.
We just need to switch to data and rely less on computed styles.

kof: I agree in general, though we have use cases we can't get rid of, they will keep reading from the dom, so we need to have a general plan on how we organize the reads like this after updates

This updateStyle is tricky because it doesn't set style directly, it triggers tree update and react cycle, though we don't know when a specific component is going to update exactly to read from it.

I think we could have a sync event triggered by each component update from useLayutEffect specifically to allow other places like this one to read after that event has happened. 

I am ok with merging this one, but we need to start thinking how to fix the problem in a good way for all cases that need to read from the dom.

kof (approved):

rpominov (approved):

PR #623, opened 2022-12-13
Fix images appears disappears during upload/delete.

### LINKED ISSUE #605: Delete on Image uploads is not working, delete some image during upload may cause other uploading images to disappear
- Just delete leaves images in a state.

image

- Uploading 2 images.

One image disappears during upload, then appears again.

https://user-images.githubusercontent.com/5077042/207064120-928d41e7-7cab-4252-bc23-5683df0580a5.mov

- Deleting the image during upload.

https://user-images.githubusercontent.com/5077042/207064533-a5f50c32-c38c-4b01-bc05-819f432d5175.mov

- I don't know how but I had images were duplicated like

image

But after the page reload I got 
image

All above looks like some race condition or/and incorrect multiple sources of truth merging

istarkov:
closes #605

`useAssets` now use Provider to be sure that state (server data load) updates will not be lost.

- [ ] Set random upload errors locally, and test that even with errors it will work.

Changed Client types from instersected server Asset types to include Asset i.e.
was 
```
type ReviewAsset = Asset & { status: 'uploading', ... }
```
now
```
export type UploadedClientAsset = {
  status: "uploaded";
  asset: Asset;
  preview: PreviewAsset | undefined;
};

export type UploadingClientAsset = {
  status: "uploading";
  asset: Asset | undefined;
  preview: PreviewAsset;
};

export type DeletingClientAsset = {
  status: "deleting";
  asset: Asset;
  preview: PreviewAsset | undefined;
};
```

As previous behavior was hard to manage, were mistakes like adding client-only props to the Asset type. It was hard to add additional props like having preview and asset simultaneously (will be needed in next PR to avoid image flicker)


apps/designer/app/designer/shared/assets/use-assets.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+  useEffect(() => {
+    if (action && stateAssets.length === 0) {
+      load(action);
+    }
+  }, [action, stateAssets.length, load]);
+
+  useEffect(() => {
+    if (data !== undefined) {
+      let nextAssets = [...stateAssets];
```
istarkov, line 146: Add test, can cause an infinite loop in case of incorrect implementation

apps/designer/app/designer/shared/assets/use-assets.tsx
```diff
@@ -0,0 +1,345 @@
...
+      });
+
+      if (deepEqual(nextAssets, stateAssets) === false) {
+        setAssets(nextAssets);
+      }
+    }
+  }, [data, stateAssets, setAssets]);
+
+  const handleDeleteAfterSubmit = (data: UploadData) => {
+    // remove from assets deleted assets
```
istarkov, line 185: remove comment

apps/designer/app/designer/shared/assets/use-assets.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+  const handleAfterSubmit = (previewAssetId: string) => (data: UploadData) => {
+    const assets = assetsRef.current;
+
+    if (action) {
+      load(action);
+    }
+
+    if (data.status === "error") {
+      // Remove preview asset from assets for now
```
istarkov, line 216: Add "wait for the load to fix it"

apps/designer/app/designer/shared/assets/use-assets.tsx
```diff
@@ -0,0 +1,345 @@
...
+    if (uploadedAsset === undefined) {
+      warnOnce(true, "An uploaded asset is undefined");
+      toast.error("Could not upload an asset");
+
+      // @todo better to create ErrorAsset type and show asset with error
+      setAssets(assets.filter((asset) => asset.id !== previewAssetId));
+      return;
+    }
+
+    // Optimistically update preview asset with uploaded asset
```
istarkov, line 238: Optimistically update the preview asset with the uploaded asset

apps/designer/app/designer/shared/assets/use-assets.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+  const handleDelete = (ids: Array<string>) => {
+    const formData = new FormData();
+    const assets = assetsRef.current;
+
+    const nextAssets = [...assets];
+
+    for (const id of ids) {
+      formData.append("assetId", id);
+      // Mark assets as deleting
```
istarkov, line 261: Optimistically

apps/designer/app/designer/shared/assets/update-state-assets.ts
```diff
@@ -0,0 +1,56 @@
...
+        nextAssets[sameIndex] = serverAsset;
+      }
+      continue;
+    }
+
+    // Assets array were empty or somebody loaded in parallel
+    nextAssets.push(serverAsset);
+  }
+
+  // Remove non-preview assets that are not in the data
```
istarkov, line 37: data => serverAssets

apps/designer/app/designer/shared/assets/update-state-assets.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+
+    if (sameIndex !== -1) {
+      if (nextAssets[sameIndex].status !== "deleting") {
+        nextAssets[sameIndex] = serverAsset;
+      }
+      continue;
+    }
+
+    // Assets array were empty or somebody loaded in parallel
```
istarkov, line 33: Assets  => stateAssets

apps/designer/app/designer/shared/assets/update-state-assets.ts
```diff
@@ -0,0 +1,56 @@
+import deepEqual from "fast-deep-equal";
+import type { Asset } from "@webstudio-is/asset-uploader";
+import { DeletingAsset, PreviewAsset } from "./types";
+
+type StateAssets = Asset | PreviewAsset | DeletingAsset;
```
rpominov, line 5: Strange name `StateAssets`. What does prefix `State` means? And why plural Asset**s**?

Names that would make more sense to me: `AssetWithState`, `LocalAsset`, `ClientSideAsset`, `ClientAsset`

packages/asset-uploader/src/schema.ts
```diff
@@ -43,7 +43,7 @@ const BaseAsset = z.object({
   location: Location,
   createdAt: z.string(),
   path: z.string(),
-  status: z.literal("uploaded").optional(),
```
rpominov, line 46: This is a transform right? In other words this is possible now `equal(x, BaseAsset.parse(x)) === false`?

I think we should not use transforms in Zod types that are used to define TypeScript types. This is very confusing to me, because now we have two types to think about: the input type and the output type. Just one of the many questions:

Which of the two we store data in DB in?

If in the output type then we can't parse when reading from DB: `BaseAsset.parse(dataFromDatabaseInOutputFormat)` — wrong.

If in the input format, writing to DB would have to look something like this:
`const dontUsetThis = BaseAsset.parse(dataFromClientInInputFormat); db.write(dataFromClientInInputFormat)`

rpominov (commented):
Still looking some comments so far

packages/asset-uploader/src/schema.ts
istarkov: `equal(x, BaseAsset.parse(x)) === false` not possible as `status` now not optional type

kof:
> I have changed the color of DeletingAnimation otherwise was hard to get what's happening during visual tests.

is it red in figma? if not you have to ask @taylornowotny for a loading animation that corresponds deletion. Also i think deletion should be just optimistic and not require any loading indication at all.

packages/asset-uploader/src/schema.ts
istarkov: The issue is that optional `updated` makes code much harder. Making it default will not break anything but make types much more easier to use.

kof:
- [ ] loading indicator doesn't go away any more

kof (approved):
logic seems better now, please remove red loading indicator and make deletion optimistic

packages/asset-uploader/src/schema.ts
rpominov: Talked in dm. Takeaway:

1. Transforms in shared reusable Zod types are bad because you have to think about `In` and `Out` types.
2. But if `Out extends In` it's not as bad, because at least you can still validate data in both formants using the `Zod` schema. This is the case here, so tolerable.

istarkov: Here is the other issue, that field `status:"uploaded"` has no sense at server data at all. And seems like occurred there by mistake. Because types are reused, it was added in typescript type, then ts type was transformed into Zod. 
I narrowed it down from optional to required. Making the issue even worse.

kof: agree, yes, it was supposed to be only for the client

apps/designer/app/designer/shared/assets/use-assets.tsx
```diff
@@ -0,0 +1,316 @@
...
+    // Optimistically update preview asset with the uploaded asset
+    const nextAssets = [...assets];
+
+    const index = nextAssets.findIndex(
+      (nextAsset) => nextAsset.id === previewAssetId
+    );
+
+    if (index !== -1) {
+      nextAssets[index] = { ...uploadedAsset, status: "uploading" };
+    }
```
rpominov, line 218: What are we updating? Isn't `status=uploading` at this point? And we are setting it to `uploading` again? Just updating data, but is it necessary?

rpominov (commented):
Everywhere we do:

```js
const copy = [...arr];
const index = copy.findIndex(condition);
if (index !== -1) {
      copy[index] = newValue;
    }
```

We could probably do:

```js
const copy = arr.map(item => condition(item) ? newValue : item)
```

apps/designer/app/designer/shared/assets/use-assets.tsx
istarkov: Looking on this I found the other bug, at this point load can already deliver asset with real id, and we are getting duplicate here if it is. (this code was to sync previewId and id) after discussion we decided to generate uuid on the client, otherwise race conditions like this is hard to test/find and detect.

packages/asset-uploader/src/schema.ts
istarkov: Removed

apps/designer/app/designer/shared/assets/use-assets.tsx
istarkov: Now id is generated on a client, should be no race effect, comment has changed

packages/asset-uploader/src/utils/uuid-handler.ts
```diff
@@ -0,0 +1,20 @@
...
+  let id: string | undefined = undefined;
+  if (part.name === idsFormDataFieldName) {
+    // 36 is the length of a UUID in bytes
+    const idBuffer = new Uint8Array(36);
+    let offset = 0;
+
+    for await (const chunk of part.data) {
+      // Will throw in case of buffer overflow
+      idBuffer.set(chunk, offset);
+      offset += chunk.length;
```
kof, line 14: we already have toUint8Array utility

packages/design-system/src/components/progress-radial.tsx
```diff
@@ -1,17 +1,31 @@
...
+const ColorCircle = styled("circle", {
+  color: "$gray12",
+  variants: {
+    variant: {
+      red: {
+        color: "$red11",
+      },
+    },
+  },
+});
```
kof, line 23: todo

kof (approved):
commented a few things approving in advance, let me know if you want me to test again

Issue #598, opened 2022-12-11
Outline is lost after changing tag property on Box, or src on new Image


istarkov:
To see outline again I need to click other object and then current.

https://user-images.githubusercontent.com/5077042/206920240-fb962e1f-5abe-4b0b-96c3-24c37e60936c.mov

TrySound:
Similar bug with style panel
https://github.com/webstudio-is/webstudio-designer/issues/628

PR #627, opened 2022-12-14
Track selected rect close to element

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/598

When tracking selected element from the root it's hard to get current element which can be changed with props.

One more issue is storing element in state doubles amount of renders and in selected instance logic we have a chain of state -> effect -> state -> effect cases.

Here I store element in ref so we avoid rerender when set element to state. Then tracked element rect is not written to state and sent to designer directly.

Since the effect is unmounted along with element we do not need to use mutation observer to track parent changes and can just recompute rect on remount.

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts
```diff
@@ -0,0 +1,83 @@
...
+          type === "updateProps" ||
+          type === "deleteProp" ||
+          type === "insertInstance" ||
+          type === "deleteInstance" ||
+          type === "reparentInstance" ||
+          type === "updateStyle" ||
+          type.startsWith("previewStyle:")
+        ) {
+          publishSelectedRect(element);
+        }
```
istarkov, line 58: may be add at least runtime dev check that if new type will be added we need to recheck this code?

apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    // instance props may change dom element
+  }, [instanceElementRef, instanceProps]);
+
+  const onScrollEnd = useCallback(() => {
+    const element = instanceElementRef.current;
+    if (element === undefined) {
+      return;
+    }
+    publishSelectedRect(element);
```
istarkov, line 75: we dont track selection during scroll? Dont see code hiding selection

istarkov (approved):

apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts
TrySound: It's somewhere else, not sure where

kof (approved):

apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts
TrySound: These are not all events so probably doesn't make sense. We eventually may find another solution. Maybe track atom state instead of listening events.

rpominov:
You've broke what we've talked about after all:

https://user-images.githubusercontent.com/825702/207798709-99cc66fe-b345-4985-a954-115d579b1133.mov

rpominov (commented):
Looks good otherwise

TrySound:
Gotcha, now I see the case. Though it can be replaced with more straightforward previous and next siblings checks. Will do this in separate PR.

rpominov (approved):

PR #633, opened 2022-12-15
Fix performance issue with scrub when used inside popover

rpominov:
https://discord.com/channels/955905230107738152/1052545317922406440

> so the problem is useScrub wants to remove focus from input: https://github.com/webstudio-is/webstudio-designer/blob/c70d9ff21a64da12a0562d046246a5e359d556e9/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx#L111

> but popover wants to keep focus inside content while it's open (not configurable): https://github.com/radix-ui/primitives/blob/81b25f4b40c54f72aeb106ca0e64e1e09655153e/packages/react/popover/src/Popover.tsx#L261-L263

> and they fight

## Before requesting a review

- [x] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -79,8 +79,9 @@ const useScrub = ({
 
   // make sure caret doesn't jump around as you scrub
   useLayoutEffect(() => {
-    if (isInputActive) {
-      inputRef.current?.select();
+    const input = inputRef.current;
+    if (isInputActive && input) {
+      input.selectionStart = input.selectionEnd = input.value.length;
     }
   });
 
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -26,8 +26,6 @@ import {
   useCallback,
   useEffect,
   useRef,
-  useState,
-  useLayoutEffect,
 } from "react";
 import { useIsFromCurrentBreakpoint } from "../use-is-from-current-breakpoint";
 import { useUnitSelect } from "./unit-select";
@@ -63,7 +61,6 @@ const useScrub = ({
 ] => {
   const scrubRef = useRef<HTMLDivElement | null>(null);
   const inputRef = useRef<HTMLInputElement | null>(null);
-  const [isInputActive, setIsInputActive] = useState(false);
 
   const onChangeRef = useRef(onChange);
   const onChangeCompleteRef = useRef(onChangeComplete);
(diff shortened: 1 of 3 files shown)


apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -77,6 +77,13 @@ const useScrub = ({
   const type = valueRef.current.type;
   const unit = type === "unit" ? valueRef.current.unit : undefined;
 
+  // make sure caret doesn't jump around as you scrub
+  useLayoutEffect(() => {
+    if (isInputActive) {
+      inputRef.current?.select();
```
kof, line 83: This is going to select the text every time we scrub, the previous behavior was nicer since you could see the text without the selection, which makes the text less readable.

Was there no other way? At this point I am not sure which one is worse the jumping caret next to the number or the always selected text.

kof: What if we focus some element inside the CssValueInput component that is not the input itself while we drag? This way we still don't have to fight popover focus and we don't have to keep the focus in the input.

kof: Doing so we would also need to prevent calling the onBlur event and return the focus to the input once scrubbing is done

rpominov: > Was there no other way? At this point I am not sure which one is worse the jumping caret next to the number or the always selected text.

I thought it looks good actually, didn't try to avoid it. Will try then

kof: we had lots of back and forth on that behavior, but take a look at figma, the way they do it is really great, except they don't allow scrub on the input itself, which makes them always need an icon or otherwise there is no scrub.

kof: they also support an option key on inputs, which then activates scrub directly on the input, but I bet most people don't know that because its not easy to discover

rpominov: I move caret to the end now, same as Figma

kof: We had an extensive prior discussion about this and found it better to not have that carret there. 
Yes figma does that when drag starts on an input field using option key.
But also figma doesn't do that when dragging starts on the icon for fields with the icon.

We made a prior decision to not have that caret when dragging, because it looks cleaner, less annoying.

Have you tried focusing some other element inside the CssValueInput container? Did you exhaust all options to fix it while keeping the behavior we decided on?

rpominov: Focusing other element should work. Didn't try it as I thought moving to the end is fine and it's easier to do. 

Will try focusing other element.

kof: works like a charm

kof (approved):

Issue #628, opened 2022-12-14
Box with changed tag looses style information


TrySound:
https://user-images.githubusercontent.com/5635476/207690567-fa4d9a93-437d-4a6f-8f1f-2290f07ed957.mov

PR #635, opened 2022-12-15
Remove treeId from props updates

TrySound:
Will help with https://github.com/webstudio-is/webstudio-designer/issues/628

treeId is stored in every props update which complicates data flow and requires passing it everywhere. Here I instead pass it inside of transaction.

This allowed us pass undefined in props data when select instance so we no longer need to create empty instance props. Updates will create it lazily.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/shared/props-utils/update-props.ts
+++ b/apps/designer/app/shared/props-utils/update-props.ts
@@ -4,12 +4,13 @@ import {
 } from "@webstudio-is/react-sdk";
 
 export const updateAllUserPropsMutable = (
+  treeId: string,
   allUserProps: AllUserProps,
-  { instanceId, propsId, treeId, updates }: UserPropsUpdates
+  { instanceId, propsId, updates }: UserPropsUpdates
 ) => {
   if (instanceId in allUserProps === false) {
     allUserProps[instanceId] = {
-      id: propsId,
+      id: propsId ?? crypto.randomUUID(),
       instanceId,
       treeId,
       props: [],
--- a/apps/designer/app/shared/props-utils/update-props.ts
+++ b/apps/designer/app/shared/props-utils/update-props.ts
@@ -1,3 +1,4 @@
+import { v4 as uuid } from "uuid";
 import {
   type AllUserProps,
   type UserPropsUpdates,
@@ -10,7 +11,7 @@ export const updateAllUserPropsMutable = (
 ) => {
   if (instanceId in allUserProps === false) {
     allUserProps[instanceId] = {
-      id: propsId ?? crypto.randomUUID(),
+      id: propsId ?? uuid(),
       instanceId,
       treeId,
       props: [],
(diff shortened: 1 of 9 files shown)


apps/designer/app/shared/props-utils/update-props.ts
```diff
@@ -4,12 +4,13 @@ import {
...
 
 export const updateAllUserPropsMutable = (
+  treeId: string,
   allUserProps: AllUserProps,
-  { instanceId, propsId, treeId, updates }: UserPropsUpdates
+  { instanceId, propsId, updates }: UserPropsUpdates
 ) => {
   if (instanceId in allUserProps === false) {
     allUserProps[instanceId] = {
-      id: propsId,
```
istarkov, line 12: In my uderstanding if we start to generate Id here, there must be some place there we generated it before, but I dont see the removed line with previous generated Id

istarkov (approved):

apps/designer/app/shared/props-utils/update-props.ts
```diff
@@ -4,12 +4,13 @@ import {
...
+      id: propsId ?? crypto.randomUUID(),
```
rpominov, line 13: We use `import { v4 as uuid } from "uuid";` in other places to generate UUIDs. Not sure which is better, but we should use one or the other everywhere consistently

rpominov (approved):

apps/designer/app/shared/props-utils/update-props.ts
TrySound: We just talked with Ivan about it. randomUUID is supported everywhere though safari only since 15. So we could migrate if that's ok.

kof: checked, what we use is exactly that + fallback https://github.com/uuidjs/uuid/blob/main/src/v4.js

turns out support is really great https://caniuse.com/?search=randomUUID

@TrySound can you switch all of them over? this or separate PR doesn't matter

kof (approved):

apps/designer/app/shared/props-utils/update-props.ts
kof: actually only makes sense to switch if native version works in all environments where we might use it: node, workers, browsers

TrySound: Looks like not supported by node globally. Only with crypto module. Though added to cloudflare last year. Will switch later then.

kof: So you would go with uuid package then?

apps/designer/app/shared/props-utils/update-props.ts
TrySound: Here It's still generated. I just removed this method call in one place https://github.com/webstudio-is/webstudio-designer/blob/main/packages/project/src/shared/props/create-instance-props.ts#L13

apps/designer/app/shared/props-utils/update-props.ts
TrySound: Yes for now. We can wait node 20 or switch to simple polyfill later.

PR #636, opened 2022-12-15
Scrub fixes

### LINKED ISSUE #626: Scrub can stuck when moved over right edge of browser window
1. Scrub any CSS input (e.g. width)
2. Move cursor all the way to the left border of the browser window
3. Move slowly over the border
4. Cursor gets stuck at the border and you can no longer move it

https://github.com/webstudio-is/webstudio-designer/blob/984e8902ac3a8d708071af57c519aad0e90611af/packages/design-system/src/components/primitives/numeric-gesture-control.ts#L125

The if condition fails when `offset` is `0`

rpominov:
- Fixed #626
- Limited movement of cursor to just one coordinate. Not sure it's a good idea, but currently in `main` we move both coordinates for "horizontal" but only one for "vertical", so I've made it consistent.

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/primitives/numeric-gesture-control.ts
+++ b/packages/design-system/src/components/primitives/numeric-gesture-control.ts
@@ -37,7 +37,7 @@ export type NumericScrubOptions = {
 type NumericScrubState = {
   value: number;
   cursor?: SVGElement;
-  offset: number;
+  offset?: number;
   velocity: number;
   direction: string;
   timerId?: ReturnType<typeof window.setTimeout>;
@@ -60,7 +60,7 @@ export const numericScrubControl = (
     // We will read value lazyly in a moment it will be used to avoid having outdated value
     value: -1,
     cursor: undefined,
-    offset: 0,
+    offset: undefined,
     velocity: direction === "horizontal" ? 1 : -1,
     direction: direction,
     timerId: undefined,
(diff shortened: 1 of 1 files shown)


kof (changes requested):
it doesn't get stuck any more, which is nice!
but now you broke the vertical movement, which is necessary to avoid covering the input at small manipulations, cursor should keep moving freely

kof:
Btw since it was not clear why we need to allow vertical movement, please add a good inline comment, so that we don't attempt to delete it again

rpominov:
Should we allow free movement in both vertical and horizontal mode? 

Same reason applies for vertical: in the spacing UI cursor can obscure the value when you change e.g. `paddingTop`.

kof:
> Should we allow free movement in both vertical and horizontal mode?

Yes! Movement is free, values change only depending on one direction. So in horizontal mode, my vertical movement doesn't change the value, but I can move freely in any mode.

rpominov:
Got it. Fixed.

kof:
Now it seems that you broke infinite scrub - try to go out of the window, it won't come back on the other side for me.

kof:
Lets add testing scenarios inline in that module, so that everyone touching it knows what to test
1. 2d movement
2. ininite scrub
3. focus behavior
4. selection behavior

rpominov:
I think it's better to document intent of the code rather that writing testing instructions in code comments.

> 2d movement

Already wrote comment

> ininite scrub

Will write

> 3. focus behavior
> 4. selection behavior

These two are in a different module (CssValueInput), and in different PR. Will add comments there.

rpominov:
Fixed infinite scrub and added comment.

kof (approved):
its perfect now

Issue #621, opened 2022-12-13
Smallest breakpoint should start from 0


kof:
I used the lower bound of the breakpoint (min-width) as a way to set the initial canvas size. But that is a problem if someone resizes the screen on a published site below the smallest media query and looses all styling.

I think the smallest media query should start from 0, but the initial canvas size needs a different logic for decididing what size to start from.

E.g. it could start somewhere in the middle between the smallest media query and the next media query.

E.g. if the smallest one has min-width: 0px and the next one has min-width 768px, we could start at 384 which is in the middle between the two
Another option is to start at the upper bound of the first non-0 media query but before the next one, e.g. if the second mq is min-width 768, then we use 767px to stay on the previous media query at it's max size

taylornowotny:
The only thing I'm sure of is that default canvas size should be at the **lowest** width of the breakpoint.

Scenario: User switches to tablet breakpoint. The canvas is at the halfway size of the breakpoint. User adjusts the sizes of their elements to fit the canvas. Then if the screen width is reduced within that same breakpoint, those sizes may be too large and break the layout.

There is no negative consequence if the screen gets wider, but if it gets narrower than the developer is expecting, then it can cause unintended layout issues. This is is a problem that I have encountered many times in Webflow and it's why I always make sure I'm working on the smallest size of each breakpoint.

Also, it doesn't make sense from the user's perspective to have a breakpoint that goes to 0. There are no screens with a width of 0. If someone can break a layout by narrowing their browser window to an unreasonable size, then that's ok. No problem with that. I think our minimum size should be somewhere around 360 or 420px, but customizable of course.

kof:
The problem I have seen is that our lowest default mq starts at 360 and I managed to resize the browser under 360 and get the fully unstyled site, which should be not possible. I literally lost everything: colors, fonts etc. 

If developer doesn't optimize for under 360 - its fine, but at least the styles for 360 should apple to everything below

taylornowotny:
Yea totally. But it should not be possible to reduce the size of the smallest breakpoint beyond ~360 or so. If users want to design for smaller sizes they can customize their breakpoints.

kof:
>  to reduce the size of the smallest breakpoint

Do you mean it should not be possible to resize the breakpoint beyond 360 or reduce the size the canvas beyond 360?

taylornowotny:
It should not be possible to reduce the canvas below the minimum size & the
default min size should be ~360

[...]

PR #639, opened 2022-12-15
Start the smallest breakpoint at 0 

kof:
Start the smallest breakpoint at 0  and use the upper bound to set the canvas size

fixes https://github.com/webstudio-is/webstudio-designer/issues/621

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests


kof:
@taylornowotny please try it out, note its only visible in new projects

apps/designer/app/designer/features/breakpoints/use-next-breakpoint.ts
```diff
@@ -0,0 +1,30 @@
...
+  const [breakpoints] = useBreakpoints();
+
+  const sortedBreakpoints = useMemo(
+    () => utils.breakpoints.sort(breakpoints),
+    [breakpoints]
+  );
+
+  return useMemo(() => {
+    if (selectedBreakpoint === undefined) {
+      return;
```
TrySound, line 20: Return undefined explicitly?

kof: we don't need to do that, why?

kof: return; is a equally good, same thing, quite clear etc

kof: actually a benefit over other values, since you don't have to type anything and I find it very explicit

taylornowotny:
Why are we using the highest width for each breakpoint? I thought we agreed that the lowest width was the correct canvas size becuase you can screw up a layout by working on a canvas that is wider than the min width of the breakpoint.

kof:
Alright, media query starts from 0, minWidth for canvas starts at 360. Now there is no way to test the canvas on lower than 360px resolution, but media query will work the same way on under 360 ...

I guess we can live with that. The only other option is to separate the resize capability with the default minimum width for canvas. Meaning you would be able to resize to lower than 360 but by default it would still start at 360

rpominov:
User still can change it and loose all styles on some screen size?

Screenshot 2022-12-16 at 13 18 30

kof:
If user edits the breakoint value - they keep the same breakpoint, but if they delete a breakpoint - they loose styles on that breakpoint. This is something we will need to address later. Right now we just warn about it.

rpominov:
No, I mean what is user change 0 -> 300. Will all styles be lost in the real site on screen size 299?

kof:
> No, I mean what if user change 0 -> 300. Will all styles be lost in the real site on screen size 299?

Yes, just like right now in prod. All those styles would be on a `@media (min-width: 300)` so they would simply not apply

rpominov:
"No styles applied" is something user never wants, ideally we should not allow this to happen I think. Maybe not in this PR, just wanted to point this out.

kof:
> "No styles applied" is something user never wants, ideally we should not allow this to happen I think. Maybe not in this PR, just wanted to point this out.

Yeah, maybe a warning, because forbidding them to use legal CSS values is kinda bad too

rpominov (approved):

taylornowotny:
Look good to me now!

PR #641, opened 2022-12-15
Migrate to pnpm for package management

TrySound:
- setup corepack for pnpm and forbid yarn usage with "engines" field
- converted yarn lockfile into pnpm lockfile with `pnpm import` command
- replaced yarn with pnpm in scripts and github actions, had to use custom cache to support corepack

This PR is only migration. Will use new pnpm features in other PRs.

```diff
du -ckd1 node_modules (yarn)
- 1125032

du -ckd1 node_modules/.pnpm
+ 857456
```

Install time (including cache downloads) in github actions reduced from 42s to 23s

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [X] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


istarkov (approved):

.github/actions/ci-setup/action.yml
```diff
@@ -8,8 +8,22 @@ runs:
       uses: actions/setup-node@v3
       with:
         node-version: 16
-        cache: "yarn"
+
+    - run: corepack enable
```
istarkov, line 12: Do we need to run this locally?

apps/designer/package.json
```diff
@@ -9,7 +9,7 @@
     "typecheck": "tsc",
     "lint": "eslint . --ext .ts,.tsx --max-warnings 0",
     "test": "NODE_OPTIONS=--experimental-vm-modules jest",
-    "checks": "yarn lint && yarn typecheck && yarn prettier --check && yarn test",
+    "checks": "pnpm lint && pnpm typecheck && pnpm prettier --check && pnpm test",
```
istarkov, line 12: all can be run in parallel as looks independent

.github/actions/ci-setup/action.yml
TrySound: If you don't have pnpm istalled locally

apps/designer/package.json
TrySound: Yes, will propose some solution

kof (approved):
🎉

andrasbacsai:
LGTM! Will it break all existing preview deployments in Vercel as we need to change the build command from `yarn` to `pnpm`?

andrasbacsai (approved):
LGTM! 
Will it break old branches preview deployments in Vercel as we need to change the build command from `yarn` to `pnpm`?

TrySound:
Yes, it will. Though we not merging yet this because of storybook setup which need to be additionally configured.

kof:
@TrySound do you wanna let @Andarist help us with storybook and pnpm here?

kof:
This PR hanging is no good, we need to merge it

package.json
```diff
@@ -31,20 +24,26 @@
...
     "prettier": "2.7.1",
+    "tsx": "^3.9.0",
     "turbo": "1.3.1",
     "typescript": "4.7.4"
   },
   "engines": {
     "node": "16",
-    "yarn": "1.22"
+    "pnpm": "7.18",
+    "yarn": "This project is configured to use pnpm"
```
Andarist, line 35: out of curiosity - what this is for? 🤔

Andarist (approved):

package.json
TrySound: It will show the message when user try to run yarn without corepack enabled

.npmrc
```diff
@@ -0,0 +1 @@
+node-linker=hoisted
```
Andarist, line 1: image

🤣 

This kinda sums up most of my experience with non-classic `node_modules` 😅 

out of curiosity - what kind of an effect does it have on install times with pnpm?

apps/designer/package.json
```diff
@@ -22,6 +22,7 @@
     "@lexical/react": "^0.6.0",
     "@lexical/selection": "^0.6.0",
     "@lexical/utils": "^0.6.0",
+    "@prisma/client": "4.3.1",
```
Andarist, line 25: q: why is this needed in the `designer`? Shouldn't it reach to the `@prisma/client` only through `@webstudio-is/prisma-client`?

Andarist: For some other reasons it's likely that you have to use this in the `@webstudio-is/prisma-client`:
```prisma
generator client {
  output          = "../src/generated-client"
}
```

and perhaps that would fix your problems here. I have a draft for this change locally but didn't yet have time to finish it up.

TrySound: See, remix bundles all its packages because of esm. Though with pnpm all packages dependencies are no longer available in designer. So prisma/client was a check. So I decided to use hoisted until remix issues are solved.

TrySound: If we use custom output remix will try to bundle generated client

PR #642, opened 2022-12-15
Use resizer when we are rendering the thumbnails in image manager

### LINKED ISSUE #546: Image Manager thumbnails are using original image
We are not using resizer when we are rendering the thumbnails in image manager

istarkov:
closes #546
fixes image flicker on upload
fixes that uploading animation stopped and not properly centered

## Before requesting a review

- [x] made a self-review
  - [x] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

- [x] Check that images in Image manager are resized
- [x] Check that not flick of image on upload happens

## Before merging

- [ ] added tests


apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
```diff
@@ -98,7 +90,8 @@ export const ImageThumbnail = ({
 
   const isUploading = status === "uploading";
 
-  const src = useImageWithFallback({ path });
+  // Must be the same as the Thumbnail width "$spacing$19"
+  const imageWidth = getInt(theme.spacing[19].value) ?? 64;
```
kof, line 94: hmm, I wonder if there is a better way to reuse the value and keep it in sync

istarkov: same but havent found

rpominov: Put `const THUMBNAIL_SIZE = theme.spacing[19].value` at the top of the file and use it in both places?

istarkov: yep, even if its not a token but it uses token value anyway

rpominov, line 94: I think it's safe to assume `theme.spacing[19].value` is a number or use some simple check, `getInt()` seem like an overkill

apps/designer/app/designer/shared/image-manager/image.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const loader = useMemo(() => {
+    if (remoteLocation) {
+      return loaders.cloudflareImageLoader({
+        resizeOrigin: env.RESIZE_ORIGIN,
+      });
+    }
+
+    return loaders.localImageLoader();
+  }, [remoteLocation]);
```
kof, line 60: sounds like you could make useLoader hook outside of the component

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
```diff
@@ -116,19 +109,24 @@ export const ImageThumbnail = ({
...
+          if (renderableAsset.status === "uploaded") {
+            onChange?.(renderableAsset);
           }
         }}
-      />
+      >
+        <Image
+          renderableAsset={renderableAsset}
+          alt={description}
+          width={imageWidth}
```
kof, line 127: dunno if there is value in passing the width vs letting the image know it

apps/designer/app/designer/shared/image-manager/image.tsx
```diff
@@ -0,0 +1,73 @@
...
+  position: "absolute",
+  width: "100%",
+  height: "100%",
+  objectFit: "contain",
+
+  // This is shown only if image was not loaded and broken
+  // from spec:
+  // The pseudo-elements generated by ::before and ::after are contained by the element's formatting box,
+  // and thus don't apply to replaced elements such as <img>, or to <br> elements
+  // But broken image is replaced element so this style is applied
```
istarkov, line 24: this part is not in spec but I works across all browsers, will fix comment

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
kof: also you don't need a safe getInt here,since we know its not gonna be something else, so you could just use parseFloat()

apps/designer/app/designer/shared/image-manager/image.tsx
rpominov, line 24: Should the comment say "But broken image is **not** a replaced element"? Otherwise there's a contradiction with the previous text

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
kof: I was almost suggesting to reuse getInt implementation from the image package, which is also an option, but I would not copy paste getInt

apps/designer/app/designer/shared/image-manager/image.tsx
istarkov: Locally still possible that some images are remote and are not

rpominov (approved):

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
istarkov: width is needed for optimisation

apps/designer/app/designer/shared/image-manager/image.tsx
```diff
@@ -0,0 +1,73 @@
+import { useMemo } from "react";
+import { styled } from "@webstudio-is/design-system";
+import { Image as WebstudioImage, loaders } from "@webstudio-is/image";
+import env from "~/shared/env";
+import type { RenderableAsset } from "../assets";
+import brokenImage from "~/shared/images/broken-image-placeholder.svg";
+
+type ImageProps = {
+  renderableAsset: RenderableAsset;
```
kof, line 9: I guess you could call the property asset, since to the image its always a renderable asset and we have a type there for clarity

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
istarkov: its not a number, its 64px now

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
kof: actually no, you are correct with passing, since the size is defined in Thumbnail component

apps/designer/app/designer/shared/image-manager/image.tsx
istarkov: asset.asset then makes me cry, bad property name

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
kof: just thought about it - define a constant locally and use it in all 3 places, ditch the theme in this case, its not that important, just add a comment that the value should be one of the values in the theme, because of the spacing grid compliance

kof (changes requested):
a few comments, looking good otherwise

apps/designer/app/designer/features/style-panel/controls/font-weight/font-weight-control.tsx
```diff
@@ -2,7 +2,7 @@ import { Select } from "@webstudio-is/design-system";
 import { FontWeight, fontWeights } from "@webstudio-is/fonts";
 import { toValue } from "@webstudio-is/css-engine";
 import { useMemo } from "react";
-import { useAssets } from "~/designer/shared/assets";
+import { useAssetContainers } from "~/designer/shared/assets";
```
kof, line 5: I guess we could have still called it `useAsset` as a name, because asset is still in there, it doesn't dictate that you can receive asset container object

apps/designer/app/designer/shared/assets/types.ts
```diff
@@ -5,39 +5,28 @@ type PreviewAsset = Pick<
...
   asset: Asset;
-  preview: PreviewAsset | undefined;
 };
 
-export type UploadingClientAsset = {
+export type UploadingAssetContainer = {
   status: "uploading";
-  asset: Asset | undefined;
-  preview: PreviewAsset;
+  asset: Asset | PreviewAsset;
```
kof, line 15: btw since we call it container, maybe asset could be "value" but its up to you, not sure

apps/designer/app/designer/shared/image-manager/image.tsx
kof: Just better logic organization. I always try to pull all the logic out of component into a hook next to it.
const useLoader = (assetContainer) => {
  ...
}

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
```diff
@@ -59,55 +41,41 @@ const ThumbnailContainer = styled(Box, {
 });
 
 const Thumbnail = styled(Box, {
-  width: "$spacing$19",
+  width: "100%",
   height: "$spacing$19",
```
kof, line 45: since you are not using your constant THUMBNAIL_WIDTH nothing stops these to mismatch, I would use it

kof (approved):

apps/designer/app/designer/shared/assets/types.ts
istarkov: In some places it is used like `const { asset } = container`;  that looks better than `value`
In some `assetContainer.asset` that looks worse than `assetContainer.value`
Imo asset is better for search etc, as `value` is too wide a keyword

apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx
istarkov: Here width is coming from parent `<Grid columns={3} gap={2}>` we must not use `width` in 99% of cases. This `width` caused a bug with `Animation` wasn't centered. I will add other comment for `const` or even remove it and add comment to `Image` prop

istarkov: Changed comment onto
https://github.com/webstudio-is/webstudio-designer/blob/7abe851a8c0e9d3ca2a9ec9aea14859c0043f6ae/apps/designer/app/designer/shared/image-manager/image-thumbnail.tsx#L97-L98

PR #643, opened 2022-12-16
Resolve tsconfig package

TrySound:
While migrating to pnpm I got an issue with our extends relying on node_modules. pnpm have only symlinks and do not have hoisted node_modules in the root.

Here I switched extends to package name as all tools can resolve it and tweaked remix tsconfig to extend from packages folder instead of node_modules.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/tsconfig.json
+++ b/apps/designer/tsconfig.json
@@ -1,6 +1,6 @@
 {
   // Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
-  "extends": "../../node_modules/@webstudio-is/tsconfig/remix.json",
+  "extends": "../../packages/tsconfig/remix.json",
   "include": [
     "remix.env.d.ts",
     "**/*.ts",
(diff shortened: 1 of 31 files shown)


apps/designer/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   // Extends can't use node's resolution rn https://github.com/dividab/tsconfig-paths/issues/209#issuecomment-1177261091
-  "extends": "../../node_modules/@webstudio-is/tsconfig/remix.json",
+  "extends": "../../packages/tsconfig/remix.json",
```
kof, line 3: maybe we can actually use the proper package name now? 
```suggestion
  "extends": "@sebstudio-is/tsconfig/remix.json",
```

kof (approved):

apps/designer/tsconfig.json
TrySound: Nope, remix still can't resolve it

Issue #630, opened 2022-12-15
Hovered instance outline doesn't update on scroll


rpominov:
1. Add enough boxes for canvas to have scroll
2. Scroll while hovering an instance
3. After you've scrolled hovered outline doesn't move with the rest of content

https://user-images.githubusercontent.com/825702/207799759-f3ef678c-db0f-48e2-b3c1-11a7d2506421.mov

rpominov:
@TrySound maybe you could take this as well?

PR #650, opened 2022-12-17
Update hovered rect when scroll ends

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/630

Simplified hovered logic to single effect, removed unnecessary nano-state container and fixed hovered outline rect when scroll ends.

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests


apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,101 @@
...
+      },
+      onScrollEnd() {
+        if (hoveredElement !== undefined) {
+          publishHover(hoveredElement);
+        }
+      },
+    });
+
+    return () => {
+      window.removeEventListener("mouseover", handleMouseOver);
```
istarkov, line 95: I know that only the "capture" setting in options matters to removeEventListener() anyway I think if you created an event listener with options object you must pass it to removeEventListener too (what if anyone adds capture: true to options)

apps/designer/app/shared/dom-hooks/use-scroll-state.ts
```diff
@@ -41,6 +41,22 @@ type UseScrollState = {
   onScrollEnd?: () => void;
 };
 
+export const subscribeScrollState = ({
```
istarkov, line 44: van be reused below in useScrollState fully?

istarkov: Above seems line emitter.emit("scrollStart"); is executed on every scroll?

apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (element !== undefined) {
+          publishHover(element);
+        }
+      }
+    );
+
+    // remove hover outline when scroll starts
+    // and show it with new rect when scroll ends
+    const unsubscribeScrollState = subscribeScrollState({
+      onScrollStart() {
```
istarkov, line 81: I have feeling that it executed many times because of bug in scroll subsription below

apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,101 @@
...
+        payload: element.getBoundingClientRect(),
+      });
+      publish({
+        type: "hoverInstance",
+        payload: {
+          id: instance.id,
+          component: instance.component,
+        },
+      });
+    }, 50);
```
istarkov, line 40: why debounce 50 and not 150 or 200? Is it needed at all for scroll start/end?

istarkov (approved):
good but plz check scroll subscriber, and probably you don't need debounce at all

apps/designer/app/canvas/hovered-instance-connector.ts
TrySound: Typescript errors when pass passive

apps/designer/app/canvas/hovered-instance-connector.ts
TrySound: It's not for scroll. Allows to avoid glitchy delay on mouseover and skip fast hovers

apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,101 @@
...
+          publishHover(hoveredElement);
+        }
+      },
+    });
+
+    return () => {
+      window.removeEventListener("mouseover", handleMouseOver);
+      window.removeEventListener("mouseout", handleMouseOut);
+      unsubscribeNavigatorHoveredInstance();
+      unsubscribeScrollState();
```
istarkov, line 98: publishHover.cancel() in unsubscribe too

apps/designer/app/shared/dom-hooks/use-scroll-state.ts
```diff
@@ -22,8 +22,10 @@ if (typeof window === "object") {
   let isScrolling = false;
 
   emitter.on("scroll", () => {
+    if (isScrolling === false) {
```
istarkov, line 25: still no guaranty that scrollStart will be earlier vs scroll ;) depends on emitter code

apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,103 @@
+import { useEffect } from "react";
+import debounce from "lodash.debounce";
```
kof, line 2: we have been using useDebounce from react-use so far because it also handles cleanup correctly, I wonder if you used this because you didn't know that we have something else

apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,103 @@
+import { useEffect } from "react";
```
kof, line 1: lots of new logic written in this file without taking it out of react and only using minimal amount of react specific code as we discussed we need to do on canvas

I wonder if we actually agreed on the approach, because if now I would like to hear a strategy how we can get to a cross-framework canvas otherwise :)

kof (approved):
besides of adding more react code on canvas 👍

apps/designer/app/canvas/hovered-instance-connector.ts
TrySound: I used this because we useDebounce complicates code and deeply integrates with react which we don't need as there is no state or callbacks here.

apps/designer/app/canvas/hovered-instance-connector.ts
TrySound: Alright, though it's same logic as before + scroll fix

kof: since you already know what to test there, its simple right now for you to migrate

apps/designer/app/canvas/hovered-instance-connector.ts
kof: Don't you need to clear debounce in effect cleanup? in case component unmounts before debounce calls the callback?

What's your take on using lodash vs lodash.something?

kof: I like lodash.something because it allows us to say which lodash functions we want to use in the application and control the dependency, so that we don't start using every lodash function over time

On the other hand it may result in multiple versions of lodash core in the system and the need to maintain the versions of each function

istarkov: see publishHover.cancel(); for clear

kof: Maybe we need our own utils package where specific lodash functions can be exported?

apps/designer/app/canvas/hovered-instance-connector.ts
```diff
@@ -0,0 +1,108 @@
...
+  const publishHover = debounce((element: Element) => {
+    const id = element.getAttribute(idAttribute) ?? undefined;
+    if (id === undefined) {
+      return;
+    }
+    const instance = utils.tree.findInstanceById(rootInstance, id);
+    if (instance === undefined) {
+      return;
+    }
+    // @todo combine hover events
```
kof, line 22: why actually? rect can change without instance change, so they are independent ...

TrySound: Logic says they were always send together. Then we can use only one to update hovered state and slightly simplify interaction with designer.

apps/designer/app/canvas/hovered-instance-connector.ts
TrySound: `lodash.*` are deprecated. I'd prefer to ditch lodash eventually. It adds too much unnecessary code to bundle.

apps/designer/app/canvas/hovered-instance-connector.ts
kof: I think the logic will change later with the multiplayer if instance rect can change under your mouse lol

apps/designer/app/canvas/hovered-instance-connector.ts
kof: oh, why do you add lodash.debounce then instae of importing from lodash directly?

kof: we aren't even using it as it seems

kof: but i would still say that we need a common utilities belt so maybe having one as a first step and exporting certain lodash functions is a good idea?

TrySound: I will see later if we can use some alternatives

PR #652, opened 2022-12-18
Check mark in Combo Boxes can be lost or be incorrect in case of input changes

### LINKED ISSUE #610: Check mark in Combo Boxes can be lost or be incorrect in case of input changes
Reproduce:
Set width to auto, and click Enter.
image

Enter "a" into the input.
image

`auto` must have a Check mark as on the previous screenshot, but it doesn't.

Or if we enter `fit-content`
image
It now shows Check mark but should not.

The issue is here 
https://github.com/webstudio-is/webstudio-designer/blob/35af8358e9538b44ed5b7916b04713013ef28aca/packages/design-system/src/components/combobox.tsx#L203

We need to detect "selected" item based on saved value and not on current.

istarkov:
closes #610

## Before requesting a review

- [ ] made a self-review
  - [ ] conceptual
  - [ ] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/props-panel/props-panel.tsx
+++ b/apps/designer/app/designer/features/props-panel/props-panel.tsx
@@ -65,6 +65,7 @@ const Combobox = ({
   } = useCombobox({
     items: itemsProp,
     value,
+    selectedItem: value,
     itemToString: (item) => item ?? "",
     onItemSelect,
   });
--- a/apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
+++ b/apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
@@ -23,7 +23,7 @@ export const TextControl = ({
 
   const setValue = setProperty(styleConfig.property);
 
-  const [currentValue, setCurrentValue] = useState<
+  const [intermediateValue, setIntermediateValue] = useState<
     StyleValue | IntermediateStyleValue
   >();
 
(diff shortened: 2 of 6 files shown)


istarkov:
Probably better to move "currentState" inside `CssValueInput` in that case public API of `CssValueInput` would be a little bit simpler vs current. 
```{
 value,
 onChange,
 onPreview
}```

instead of current 
```
{
 value,
 selectedItem,
 onChange,
 onChangeComplete,
 onPreview
}
```

Now we have that each Component must have 2 value states:
1. Global which we change on Enter or on Blur
2. Intermediate state used for editing which is different from Global as we need to hold additional things like Highlighted items, intermediate value, color input, color Picker value etc...

And It's not good that internal type of components is leaked outside. Probably we don't want the same component to be edited simultaneously as it will need to share all that internal state that's hard.

WDYT?

rpominov (approved):

rpominov:
> Probably better to move "currentState" inside CssValueInput

Noticed only after approved. Good idea

kof:
We already started standardizing handlers for preview vs final on this language:

onChange - called on each change, can be frequent
onChangeComplete - final change, can't be frequent

So I don't want to add `onPreview` callback and actually we need to make other things that have similar situations use the same semantics

kof:
Having intermediate state outside of the control and passing it to it has an advantage - you can coordinate the logic between multiple controls for intermediate value changes.

E.g. flex has row and column gap controls. Each has a scrub. When scrubbing one in the linked mode, they both should change simultaneously.

Currently its not right.

istarkov:
`onPreview` came from the fact that `highlightedIndex` is controlled inside `Downshift` and we really has no reason to share it between components. On any +- hard components some state will always be "internal"
Otherwise, we need to move up everything - `useFilter` we use inside Combobox (has state),
highlightedIndex - not need to be shared, has state.  
Fixing api surface for internal components will not make life easier.

kof:
We could have 2 values over props: value, intermediateValue. Both can be matched to whatever is used internally

istarkov:
Renamed.

Issue #651, opened 2022-12-18
Make findInstanceById fast for large trees


kof:
To make findInstanceById fast, we need a flat map {instanceId: instance}.
We have 2 options: either index the tree on the client and server or make the tree flat: [{id, component, children: ['id1', 'id2'] ...}]
Making it flat will probably make it easier to work with the tree, since we don't need to react to every update of the tree and update the index.

kof:
Btw this change would make immer patches in case of parenting automatically small, so there is no value of an instance with children containing complete instances, resulting in a potentially large tree inside the patch.

kof:
Also if we make the structure look like this: `[{id, component, children: ['id1', 'id2'] ...}]` we still need an index and updating it on each deletion and creation of an instance, if we make the structure look like this: `{instanceId: instance}` then we don't have to maintain an index at all.

TrySound:
We have 3 variants of normalized tree

An object
```
{
  [instanceId]: Instance
}
```

Pros
- straight forward and fast loopup

Cons
- order is not guaranteed so need to store root id somewhere

Array

```
Instance[]
```

Pros
- order is preserved so we can consider first element as root

Cons

- .find by instance id can be slow

Map

```
Map<instanceId, Instance>
```

Pros
- fast lookup
- order is guaranteed so we can consider first element as root

Cons

- stored as array in db so requires conversion on client
- root should be additionally extracted from original array

istarkov:
I'm for Array.
We can always use `map` or `lazyMap` to reference array instances by id. 
Array with some restrictions can be used for collaborative editing  https://blog.excalidraw.com/building-excalidraw-p2p-collaboration-feature/ what is much more simple vs any crdt implementations.


Having in mind collaboration we can never delete items from the local array and because of that supporting `map` or `lazyMap` for array access is easy too. We don't need to manage deletes, so once instance is referenced through a map it can be where forever. 

Arrays also more easy types in DB, to do migrations etc. They can be even put in tables.

TrySound:
Marking items as deleted would give us additional flexibility with arrays. Instead of storing list of ids in instance.children we could store list of indexes. This way arrays lookup becomes very fast and we do not need to convert it to map and have an access to root as first element.

TrySound:
To avoid cleanup necessity and prevent document bloating we can just set null or undefined instead of marking as null.

kof:
[...]

PR #653, opened 2022-12-18
Normalize tree in db

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/651

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [x] detailed
  - [ ] with testing

## Test cases

## Before merging

- [ ] added tests


packages/prisma-client/prisma/migrations/20221218185551_normalized_tree/migration.ts
```diff
@@ -0,0 +1,73 @@
+import { PrismaClient } from "./client";
+
+type Instance = {
+  id: string;
+  component: any;
```
kof, line 5: any?

kof: ah its a migration lol

packages/prisma-client/prisma/migrations/20221218185551_normalized_tree/migration.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+const normalizeTree = (rootInstance: Instance) => {
+  const normalizedTree: NormalizedInstance[] = [];
+  const traverse = (
+    instance: Instance,
+    normalizedInstance: NormalizedInstance
+  ) => {
+    for (const child of instance.children) {
+      if (typeof child === "string") {
```
kof, line 24: should we also make child always an instance, no matter if its a string?

packages/prisma-client/prisma/migrations/20221218185551_normalized_tree/migration.ts
```diff
@@ -0,0 +1,73 @@
...
+        normalizedInstance.children.push(child);
+      } else {
+        const normalizedChild = {
```
kof, line 27: Wait, isn't the point to have children as ids?

packages/react-sdk/src/db/instance.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      createNormalizedInstance("3", []),
+      createNormalizedInstance("4", []),
+    ])
+  ).toMatchInlineSnapshot(`
+    {
+      "children": [
+        {
+          "children": [
+            "text",
+            {
```
kof, line 89: instance id is supposed to be here instead, right?

packages/prisma-client/prisma/migrations/20221218185551_normalized_tree/migration.ts
TrySound: We could distinct like this
```
| { type: 'text', value: '' }
| { type: 'instance', children: [] }
```

packages/react-sdk/src/db/instance.test.ts
TrySound: It's denormalizeTree test

packages/prisma-client/prisma/migrations/20221218185551_normalized_tree/migration.ts
kof: yep

kof:
So this PR is about normalizing the db but keeping the client data as-is as a first step, right? I wish I knew that before I misunderstood certain things

packages/project/src/db/tree.ts
```diff
@@ -42,10 +47,11 @@ export const loadById = async (
     return null;
   }
 
-  const root = JSON.parse(tree.root);
+  const root = denormalizeTree(JSON.parse(tree.root));
   Instance.parse(root);
```
kof, line 51: btw we should use the return value from .parse call everywhere, I didn't know before but it would allow skipping the data that isn't defined by the schema

kof: I didn't know it initially

packages/react-sdk/src/db/instance.ts
```diff
@@ -32,3 +32,90 @@ export const Instance = z.lazy(
...
+      } else {
+        const normalizedChild = {
+          id: child.id,
+          component: child.component,
+          cssRules: child.cssRules,
+          children: [],
+        };
+        // push returns length
+        const index = normalizedTree.push(normalizedChild) - 1;
+        normalizedInstance.children.push(index);
```
kof, line 67: we need to deeply discuss consequences of using index instead of ids

TrySound: It was a fair point about conflicts between clients, ids will work better.

packages/prisma-client/prisma/migrations/20221218185551_normalized_tree/migration.ts
TrySound: Done https://github.com/webstudio-is/webstudio-designer/pull/654

rpominov (commented):
Not sure we should do it right now. As @kof pointed out in https://github.com/webstudio-is/webstudio-designer/issues/651#issuecomment-1357412461 DB representation is not as important if it's going to differ from client-side one anyway.

kof:
@rpominov I still think that a long-term format will be more array-like than a tree for various reasons. We can hold off or we can migrate, doesn't matter, we will probably have more migrations in the future as well, since we also need to move cssRules to a separate array

kof (approved):
lets use the ids instead of index

TrySound:
@rpominov Migration performance is yet to be solved though I don't want to combine all these changes in single PR and migration caz it will affect everything, won't be reviewable and introduce a lot of bugs for sure.

TrySound:
Do we run data migrations on fresh db?

rpominov:
> Do we run data migrations on fresh db?

When setting up from scratch locally? Yes.

TrySound:
Do we have any data in scratch db? Why run data migration?
If we run data migration once on prod we should never run it again.

rpominov:
> Do we have any data in scratch db? Why run data migration?

Migrations can create data in theory. But yes, currently we could not run them. But doesn't make a difference either way, so they just run normally.

TrySound:
No data -> no problem 😄

PR #654, opened 2022-12-18
Convert text children to nodes in tree

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/651

We want to normalize our tree structure and need to distinct text children with ids. Now instead of string its respresented as
```
{
  type: 'text',
  value: string
}
```

This also gives us a little type safety because typeof child === 'string' ignores when child cannot be string.

## Before requesting a review

- [ ] made a self-review
  - [x] conceptual
  - [x] detailed
  - [x] with testing

## Test cases

## Before merging

- [ ] added tests


kof (approved):
looking good!

kof:
btw. this migration was already executed on dev right now, right?

apps/designer/app/canvas/features/text-editor/text-editor.stories.tsx
```diff
@@ -99,23 +99,23 @@ export const Basic: ComponentStory<typeof TextEditor> = ({ onChange }) => {
           instance={utils.tree.createInstance({
             component: "TextBlock",
             children: [
-              "Pragraph you can edit ",
+              { type: "text", value: "Pragraph you can edit " },
```
istarkov, line 102: typo Pragraph

packages/prisma-client/prisma/migrations/20221218211129_tree_text/migration.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+export default () => {
+  const client = new PrismaClient({
+    // Uncomment to see the queries in console as the migration runs
+    // log: ["query", "info", "warn", "error"],
+  });
+  return client.$transaction(
+    async (prisma) => {
+      const previousTrees = await prisma.tree.findMany();
```
istarkov, line 49: On prod we have 8855 Trees. IMHO this will definitely will timeout. Have you tried copy prod db and try locally?

istarkov (approved):
IMO migration could timeout

packages/prisma-client/prisma/migrations/20221218211129_tree_text/migration.ts
kof: btw what do you think we could do to make it work with very large tables? Some kind of batching?

TrySound: sql would be fast if stored with native types instead of json

TrySound: Btw where all these pages comes from? From test cases?

kof: sql is not the only problem, you are fetching the entire table there and in the real world it will be many gigabytes some day that you are trying to put into memory

kof: > Btw where all these pages comes from? From test cases?

in prod - yeah, from people clicking on "try" button and we clone for them the landing project

packages/prisma-client/prisma/migrations/20221218211129_tree_text/migration.ts
```diff
@@ -0,0 +1,64 @@
...
+      const update = previousTrees.map((tree) => {
+        return {
+          id: tree.id,
+          root: JSON.stringify(convertTree(JSON.parse(tree.root))),
```
rpominov, line 53: I'm not 100% sure all trees in prod DB are valid / in the current format. Might be a good idea to parse with zod. But maybe I'm overthinking.

rpominov (approved):

rpominov:
> btw. this migration was already executed on dev right now, right?

Will be executed after merge to `main`

kof:
I think its already being exectuted since I opened an old project and saw the errors that indicate this has already run a migration on preview db

rpominov:
This one works fine https://webstudio-designer-git-main-webstudio-is.vercel.app/designer/9fcbbbd6-f684-4868-89f1-3a039b2c24cf .

You might have caught something else.

Or, if you've opened preview of this PR, then error is expected, but because migration is **not** applied.

packages/prisma-client/prisma/migrations/20221218211129_tree_text/migration.ts
TrySound: > sql is not the only problem, you are fetching the entire table there and in the real world it will be many gigabytes some day that you are trying to put into memory

Yes, I know. Solution is not the best. Didn't expect we have 8k trees already.
Does prisma have any tools for this?

istarkov: Increase timeout maybe, what time now does it take?

istarkov: One of options to update as a batch is to use raw query and `unnest`, most raws allows to pass array as params
```
  UPDATE blabla AS t SET
    name = v.new_name
    blabbla = v.new_blabla
  FROM unnest(?::text[], ?::text[]) AS v(id, new_name, new_blabla)
  WHERE t.id = v.id
```

kof: We probably need a pagination helper that would allow to fetch X amount of items from the large table, then another batch and another batch

rpominov: Yeah, lots of ways to write migrations more efficiently if we use raw SQL in some ways. But I couldn't convince @kof that we should just use raw SQL when makes sense 😅

istarkov: In case of db.raw don't understand arrays as params:
you can inject array into string.
i.e. instead of this `?::uuid[]` inject `array['id1', 'id2', '...']` etc but you need to quote each value, so better to use escape lib like 
https://www.npmjs.com/package/sqlstring

rpominov: Also, sending 8000 `.update()` queries shouldn't take more than 5 minutes in prod with current setup, probably 1-2 minutes. So we don't have to overthink this particular migration too much. Should work as is.

kof: > Yeah, lots of ways to write migrations more efficiently if we use raw SQL in some ways. But I couldn't convince @kof that we should just use raw SQL when makes sense 😅

I just know that complex data migrations using just SQL would limit everyone. Only @istarkov and a few other people on the planet know how to write those for queries that require json parse and various other js manipulations.

without this migration tool we have now, even this migration here wouldn't be already done (I bet it took less than a few hours to write)

For speed of development this is huge and we will find ways to make them run well. That raw SQL stuff for complex migrations is not an option right now. Only if we make the entire data completely normalized and the query becomes human readable.

TrySound: $ yarn migrations --dev migrate
yarn run v1.22.19
$ /Users/bogdancadkin/host/webstudio-designer/node_modules/.bin/migrations --dev migrate
You're about to apply the following migration(s):

  - 20221218211129_tree_text

? Continue? yes
Starting 20221218211129_tree_text
Tree 621536ed2e8214a621cbdf5e cannot be converted
Tree 628513c2bd1a04feefcd32c4 cannot be converted
Tree 628513d2bd1a04feefcd32c6 cannot be converted
Tree 628513d6bd1a04feefcd32c8 cannot be converted
20221218211129_tree_text done in 69.886s
✨  Done in 72.93s.

TrySound: > Tree 621536ed2e8214a621cbdf5e cannot be converted

These messages are from errored rows

packages/prisma-client/prisma/migrations/20221218211129_tree_text/migration.ts
TrySound: Wrapped with try/catch and found a few cases

packages/prisma-client/prisma/migrations/20221218211129_tree_text/migration.ts
rpominov: > I just know that complex data migrations using just SQL would limit everyone. Only @istarkov and a few other people on the planet know how to write those for queries that require json parse and various other js manipulations.

Yeah. Doing it completely in an SQL file would be tough. But now that we have .ts migrations, I think using some raw SQL in the .ts files is fine. E.g. to combine many `update` queries into one as @istarkov have suggested above. And the SQL he posted is very readable to me, so that makes at least two :)

kof: ah, yeah, that can be done from time to time if need, I thought the discussion is about generally writing migrations with ts

PR #659, opened 2022-12-19
Position floating panel without hacks

kof:
## Description

We have been calculating the position of the floating panel using a relative container in the ancestors.
I was about to add that floating panel to the left side (then it turned out its not needed yet) and diddn't want to add that hack  also to the sidepanel on the left.

Now we don't rely on that parent container to be positioned relatively, but we still need to know what container we position floating panel relative to, so we are using a ref over context.

## Steps for reproduction

1. select an instance
2. typography > click font family value
3. fonts picker should be positioned right next to the right panel

## Code Review

  - conceptual review architecture

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


apps/designer/app/designer/shared/floating-panel/floating-panel.tsx
```diff
@@ -0,0 +1,78 @@
...
+}): [MutableRefObject<HTMLButtonElement | null>, number] => {
+  const triggerRef = useRef<HTMLButtonElement | null>(null);
+  const { container } = useContext(FloatingPanelContext);
+  if (isOpen === false) {
+    // Prevent computation if the panel is closed
+    return [triggerRef, 0];
+  }
+  const containerRect = container?.current?.getBoundingClientRect();
+  const triggerRect = triggerRef.current?.getBoundingClientRect();
+  return [triggerRef, (triggerRect?.left ?? 0) - (containerRect?.left ?? 0)];
```
TrySound, line 24: Accessing dom in render is not a good idea. Refs can still not be set. Use effects

apps/designer/app/designer/shared/floating-panel/floating-panel-provider.tsx
```diff
@@ -0,0 +1,21 @@
+import { createContext, RefObject } from "react";
+
+export const FloatingPanelContext = createContext<{
```
TrySound, line 3: Do we want to expose context?

kof: nope, removing from index

apps/designer/app/designer/shared/floating-panel/floating-panel.tsx
kof: done

TrySound (approved):

Issue #658, opened 2022-12-19
Stop overwriting browser cmd+(+-) keyboard shortcut


kof:
Problem with overwriting it is that it can't be used to zoom the interface itself, which is an important accessibility feature.
Figma decided to break it and I copied it initially, but I regret it.
The solution is to just support +- without modifiers and release cmd++-  to the browser.

kof:
Figma also supports +- without modifiers in addition to with modifiers

taylornowotny:
Sounds perfect to me 👍

PR #660, opened 2022-12-19
Use +- keys for zoom without cmd

kof:
## Description

This enables to keep browser zoom behavior for accessibility and zooming on canvas happens using +- keys without modifiers.

Closes https://github.com/webstudio-is/webstudio-designer/issues/658

## Steps for reproduction

1. press - on canvas and in designer for zoom out
2. press + for zoom in

## Code Review

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/apps/designer/app/shared/shortcuts/index.ts
+++ b/apps/designer/app/shared/shortcuts/index.ts
@@ -1,14 +1,1 @@
-export const shortcuts = {
-  esc: "esc",
-  undo: "cmd+z, ctrl+z",
-  redo: "cmd+shift+z, ctrl+shift+z",
-  preview: "cmd+shift+p, ctrl+shift+p",
-  delete: "backspace, delete",
-  breakpointsMenu: "cmd+b, ctrl+b",
-  breakpoint: Array.from(new Array(9))
-    .map((_, index) => `cmd+${index + 1}, ctrl+${index + 1}`)
-    .join(", "),
-  zoom: "cmd+=, ctrl+=, cmd+-, ctrl+-",
-} as const;
-
-export const options = { splitKey: "+", keydown: true };
+export * from "./shortcuts";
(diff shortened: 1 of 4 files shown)


apps/designer/app/shared/shortcuts/index.ts
```diff
@@ -1,14 +1 @@
...
-  delete: "backspace, delete",
-  breakpointsMenu: "cmd+b, ctrl+b",
-  breakpoint: Array.from(new Array(9))
-    .map((_, index) => `cmd+${index + 1}, ctrl+${index + 1}`)
-    .join(", "),
-  zoom: "cmd+=, ctrl+=, cmd+-, ctrl+-",
-} as const;
-
-export const options = { splitKey: "+", keydown: true };
+export * from "./shortcuts";
```
TrySound, line 1: apps/designer/app/shared/shortcuts.ts?

kof: I just moved code to a separate file from index, since we use index files only for exporting

kof: index files shouldn't contain anything but imports/exports

kof: so that we never have to see index files in the editors when working with code

TrySound: Sure, I would just prefer to not use index.ts at all. Or in this case at least do not create folder and index.ts just for one module.

kof: Yeah, I don't love it but the parent folder when there is mixture of folders and files doesn't look consistent either, so I thought maybe folder is not that bad and in such cases we also need to use index, because otherwise we have longer paths than necessary.

kof: I think overall its a rare case when we use a folder but there is only one file, it should be an edge case

kof: it seems unnecessary, but it doesn't really hurt ... so for consistency reasons seems like a better choice

TrySound (approved):

Issue #656, opened 2022-12-19
CSSValueInput doesn't work as expected with esc key


istarkov:
As of now 1st ESC press do nothing, 2nd reset the field to an empty value.
Should be:
1. If the menu is open - close the menu
2. If the menu is not open reset to the initial control value

kof:
1st ESC closes dropdowns when they are open. Bug is when there is nothing to close, first ESC should revert back to the value that was last saved (non-intermediate value)

PR #661, opened 2022-12-20
Fixes ESC behavior in CssValueInput

kof:
## Description

Closes https://github.com/webstudio-is/webstudio-designer/issues/656

1. When ESC is pressed without dropdowns open - it should reset the value to the last known permanent value
2. When any drop down is open - first ESC press should close the dropdown, second - reset the value (as described in point 1)

## Steps for reproduction

Case 1

1. find a numeric CSS value input
2. change the value with some number
3. press ESC
4. number resets to the initial value

Case 2

1. find a keyword CSS value input
2. start typing
3. autocomplete list is shown
4. press ESC
5. list is closed
6. press ESC
7. value resets to initial value

Case 3

1. find a numeric CSS value input
2. type some number
3. click on unit button - list shows
4. press ESC - list closes
5. press ESC - value resets to initial value

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
+++ b/apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
@@ -8,6 +8,8 @@ import type { StyleValue } from "@webstudio-is/css-data";
 import { useState } from "react";
 import { Box, Tooltip } from "@webstudio-is/design-system";
 
+const emptyValue: StyleValue = { type: "string", value: "" };
+
 export const TextControl = ({
   currentStyle,
   inheritedStyle,
@@ -37,7 +39,7 @@ export const TextControl = ({
         <CssValueInput
           icon={icon}
           property={styleConfig.property}
-          value={value}
+          value={value ?? emptyValue}
           intermediateValue={intermediateValue}
           keywords={styleConfig.items.map((item) => ({
             type: "keyword",
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -276,18 +276,18 @@ export const CssValueInput = ({
         ? String(item.value)
         : toValue(item),
     onInputChange: (inputValue) => {
-      onChange(inputValue ?? unsetValue.value);
+      onChange(inputValue ?? toValue(props.value));
     },
     onItemSelect: (value) => {
-      onChangeComplete(value ?? unsetValue);
+      onChangeComplete(value ?? props.value ?? unsetValue);
     },
     onItemHighlight: (value) => {
       if (value == null) {
-        onHighlight(unsetValue);
+        onHighlight(props.value ?? unsetValue);
         return;
       }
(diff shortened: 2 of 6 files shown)


packages/design-system/src/components/combobox.tsx
```diff
@@ -168,6 +170,14 @@ export const useCombobox = <Item,>({
...
+    onIsOpenChange(value) {
+      // Don't set isOpen to true if there are no items to show
+      // because otherwise first ESC press will try to close it and only next ESC
+      // will reset the value. When list is empty, first ESC should reset the value.
+      const nextIsOpen = value.isOpen === true && filteredItems.length !== 0;
+      setIsOpen(nextIsOpen);
+    },
     defaultHighlightedIndex: -1,
     selectedItem: selectedItem ?? null, // Prevent downshift warning about switching controlled mode
     stateReducer,
```
istarkov, line 183: please fix few lines below `inputValue: value ? itemToString(value) : undefined,`
It can't be undefined, we don't get an error just because of our ts falgs.
If `inputValue` is controllable it must be a string

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -8,6 +8,8 @@ import type { StyleValue } from "@webstudio-is/css-data";
 import { useState } from "react";
 import { Box, Tooltip } from "@webstudio-is/design-system";
 
+const emptyValue: StyleValue = { type: "string", value: "" };
```
istarkov, line 11: Its initially invalid property value. If for example we will add validation not on save but on some other events we will immediately get that its invalid value.

I think we need to pass value undefined here. IT's ok.
And define it inside CSSValueInput as { type: "intermediate", value: "" } as intermediate type can be non valid

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -235,7 +232,7 @@ export const CssValueInput = ({
   onHighlight,
   ...props
 }: CssValueInputProps & { icon?: JSX.Element }) => {
-  const value = props.intermediateValue ?? props.value ?? unsetValue;
+  const value = props.intermediateValue ?? props.value;
```
istarkov, line 235: props.intermediateValue ?? props.value ?? { type: "intermediate", value: "" }

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -276,18 +273,18 @@ export const CssValueInput = ({
...
-        onHighlight(unsetValue);
+        onHighlight(props.value);
         return;
       }
       if (value.type !== "intermediate") {
-        onHighlight(value ?? unsetValue);
+        onHighlight(value ?? props.value);
       }
     },
   });
```
istarkov, line 290: Below we have code like 
```
isOpen && !menuProps.empty
```

Seems like with new controllable `isOpen` we don't need to check menuProps.empty

Issue #625, opened 2022-12-14
Outline doesn't move on margin change


rpominov:
1. Select a Box on canvas
2. Change its margin from style panel
3. Box moves but the selected instance outline does not move

TrySound:
I can check

rpominov:
Would be great

taylornowotny:
Is this what you're talking about?

https://user-images.githubusercontent.com/86495896/207951671-b01e9cdb-ced4-4681-998b-6a3655891725.mov

If so, I don't think the outline of an element should include margin. We should outline elements where the border would appear, and borders are not affected by margin.

kof:
Seems like we already fixed it on main https://share.descript.com/view/s5OvybNlW3Y

rpominov:
Not fixed. Worked for you because changing margin also changed size in your case.

The problem is that outline does not update if you move component somehow without changing size. Many ways to do this:

1. Change margin
2. Change float (also shown in video)
3. Change top/left with position: absolute
4. etc.

https://user-images.githubusercontent.com/825702/208041669-80a88c3c-0edf-4d60-9ead-7f935b614bcb.mov

TrySound:
Looks like fixed here https://github.com/webstudio-is/webstudio-designer/pull/637

kof:
@TrySound did you try @rpominov's repro?

TrySound:
Yes. Why? You reproduced something?

rpominov:
Yes. Almost fixed. But still reproducible when you scrub:

https://user-images.githubusercontent.com/825702/208363661-79d4b336-66e4-44bd-8d82-97a8d1efea87.mov

rpominov:
BTW, none of this seem to be needed now (unless you scrub): https://github.com/webstudio-is/webstudio-designer/blob/bb0e3153387f680a37fd7e1e6a4451888e70db94/apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts#L33-L69

TrySound:
Maybe hide outline when preview styles are send? We have more and more of cases where frequently changing values make outline laggy

kof:
Yeah, I agree we should hide it, here is another issue for hiding it when resizing canvas, feel free to add a case there or to change it right away

https://github.com/webstudio-is/webstudio-designer/issues/638

taylornowotny:
Wouldn't it be better to fix the lag rather than hide outlines?

[...]

PR #663, opened 2022-12-20
Hide outline when preview styles

TrySound:
## Description

Ref https://github.com/webstudio-is/webstudio-designer/issues/625

Frequently changed style values with for example scrub makes outline look laggy because of latency between iframe and designer (where outline is rendered).

Here I subscribed to previewStyle and send undefined to hide rect. We do the same when scroll canvas and cannot update outline fast.

## Steps for reproduction

1. select instance
2. use scrub to change for example top

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts
+++ b/apps/designer/app/canvas/features/wrapper-component/selected-instance-connector.ts
@@ -4,17 +4,40 @@ import { getBrowserStyle } from "@webstudio-is/react-sdk";
 import { publish, subscribe, subscribeAll } from "~/shared/pubsub";
 import { subscribeScrollState } from "~/shared/dom-hooks";
 
-const publishSelectedRect = (element: HTMLElement) => {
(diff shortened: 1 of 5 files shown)


kof:
I see a flickr https://share.descript.com/view/taP1qzTW54z

kof (changes requested):
seems like timing is not optimal, because I see flicker with outline showing up while scrubbing

https://share.descript.com/view/taP1qzTW54z

TrySound:
Right, it's resize observer. So we may not need it at all since we won't show outline for preview style annymore.

kof:
I think resize-observer could be useful additionally in case component changes it's size in response to some property change later on

TrySound:
Changed atom type. Now resize observer passively updates rect without controlling visibility state.

kof (approved):
Tested, looking good 👍

PR #671, opened 2022-12-21
Make left sidebar grow full height

kof:
## Description
Screenshot 2022-12-21 at 14 16 24

Previous changes with flex-direction: column broke the left sidebar height

## Steps for reproduction

1. click + button
2. see all components without scrollbars

## Code Review

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
+++ b/apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
@@ -45,7 +45,7 @@ export const SidebarLeft = ({ publish }: SidebarLeftProps) => {
   );
 
   return (
-    <Flex>
+    <Flex css={{ flexGrow: 1 }}>
       <SidebarTabs activationMode="manual" value={activeTab}>
         <SidebarTabsList>
           {enabledPanels.map((tabName: TabName) => (
--- a/apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
+++ b/apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
@@ -45,7 +45,7 @@ export const SidebarLeft = ({ publish }: SidebarLeftProps) => {
   );
 
   return (
-    <Flex css={{ flexGrow: 1 }}>
+    <Flex grow>
       <SidebarTabs activationMode="manual" value={activeTab}>
         <SidebarTabsList>
           {enabledPanels.map((tabName: TabName) => (
--- a/packages/design-system/src/components/flex.tsx
+++ b/packages/design-system/src/components/flex.tsx
@@ -3,7 +3,9 @@ import { styled } from "../stitches.config";
 export const Flex = styled("div", {
   boxSizing: "border-box",
   display: "flex",
-
+  // Fixes a bug where container doesn't want to grow.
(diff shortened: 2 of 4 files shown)


TrySound (approved):

apps/designer/app/designer/features/sidebar-left/sidebar-left.tsx
```diff
@@ -45,7 +45,7 @@ export const SidebarLeft = ({ publish }: SidebarLeftProps) => {
   );
 
   return (
-    <Flex>
+    <Flex css={{ flexGrow: 1 }}>
```
TrySound, line 48: Doesn't flex have some prop for this?

kof: unfortunately no and since flexGrow is a number, its difficult to add using stitches variants. 
i don't like this too, basically with variants you can either predefine all number values or make some kind of grow: true variant that sets to 1, maybe that's what we should do, because that's very popular thing to do

also maybe we should add min-height: 1 and min-width: 1 because that fixes also tons of use cases

TrySound: 0 fixes too, no?

kof: actually min-height: 0 and min-width 0 also doing the same trick

kof:
@rpominov adding you for review since this PR now touches a design system component

packages/design-system/src/components/flex.tsx
```diff
@@ -90,6 +92,22 @@ export const Flex = styled("div", {
         gap: "$spacing$20",
       },
     },
+    shrink: {
+      true: {
+        flexShrink: 1,
+      },
+      false: {
+        flexShrink: 0,
+      },
```
rpominov, line 101: What happens if you don't pass `shrink` at all, neither `false` nor `true`, just `<Flex>`?

- If neither `false` nor `true` is applied, then we're good. 
- If `false` is applied, we potentially break usage of `<Flex>` somewhere as we're setting `flexShrink:0` for all `Flex`es by default.

rpominov (commented):
Might be a good idea to add `shrink/grow` to Box as well, as it can also be used as Flex's child

packages/design-system/src/components/flex.tsx
kof: neither is applied

kof:
> Might be a good idea to add `shrink/grow` to Box as well, as it can also be used as Flex's child

agree

kof:
> Might be a good idea to add `shrink/grow` to Box as well, as it can also be used as Flex's child

its actually kindof needed everywhere, since anything can be a flex child .... looking at what is being used with flexShrink and flexGrow ....

kof:
Idk what to do about that, adding flex properties everywhere is a little weird

rpominov:
Yeah. I'd choose between:

1. Flex and Box as most common places where we might need this
2. Don't add at all even for the Flex

Just Flex is kinda weird too, as Box seem like the "default" flex child

PR #674, opened 2022-12-22
Fix orientation detection in D&D

### LINKED ISSUE #632: Horizontal layout is detected as vertical when using drag and drop
1. Add a Box
2. Set it's "Display" to "Flex"
3. **Set alignment to center (see video)**
4. Put another Box inside
5. Now drag yet another Box inside the original Box

Bugs:

- The drop placement line appears at the bottom side of the existing child Box, while it should appear at left / right side
- It's impossible to put the second child to the left relative to the existing child

https://user-images.githubusercontent.com/825702/207823766-fd4dba48-ed1f-4c17-b156-929e19bbf305.mov

rpominov:
## Description

Fixed #632

There was a bug in how probe is used. We used to detect orientation by how the probe collapses:
- 123x0: vertical
- 0x123: horisontal
- 0x0: vertical (just a guess as we have no idea) — what is happening in #632
- 123x123: vertical (also a guess)

Now, we first try to detect by how probe is oriented relative to the future sibling of the drop item, and fallback to the logic above if there's no sibling.

------------

Also fixed a small bug where we use `parent.children` directly instead of `getValidChildren(parent)`.

## Steps for reproduction

See https://github.com/webstudio-is/webstudio-designer/issues/632

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


rpominov:
@kof I was wrong before when I said in DM that we don't use probe. We do in this case, but not well enough. More about it above.

kof:
@rpominov nice, can't test though, vercel is getting yarn error

kof (approved):
approving without testing

kof:
rebased and tested, nice!

PR #673, opened 2022-12-21
ESC support, do not save if nothing has changed

istarkov:
closes #656
closes #644

## Description

1. What is this PR about (link the issue and add a short description)

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

- [ ] hi @kof, I need you to do
  - conceptual review (architecture, feature-correctness)
  - detailed review (read every line)
  - test it on preview

## Before requesting a review

- [ ] made a self-review
- [ ] added inline comments where things may be not obvious (the "why", not "what")

## Before merging

- [ ] tested locally and on preview environment (preview dev login: 5de6)
- [ ] updated [test cases](https://github.com/webstudio-is/webstudio-designer/blob/main/apps/designer/docs/test-cases.md) document
- [ ] added tests
- [ ] if any new env variables are added, added them to `.env.example` and the `designer/env-check.js` if mandatory

PR #680, opened 2022-12-22
Fix ESC to work at unit menu

### LINKED ISSUE #656: CSSValueInput doesn't work as expected with esc key

istarkov:
closes #656

Following test case at #673 didn't work:

Click unit px
Hover over items
Press ESC - instead of closing menu it resets input and only 2nd ESC closes the menu.

*The issue is Tooltip + UnitSelect.*

UnitSelect set 'focus" on Item on hover.
So hovering over the menu causes a lot of blur/focus events dispatched to the ancestors.
Tooltip at the top closes/opens very fast (not visible but internally it is). 
That cause https://github.com/radix-ui/primitives/blob/e861c16f20fe8e2e440983355edf2ec78cc2a51c/packages/react/dismissable-layer/src/DismissableLayer.tsx#L105 to think that Tooltip popover is now at the top layer. And prevent `onEscapeKeyDown` to be executed.

Current hack somehow improves the situation. But seems like better would be to disable the tooltip in case of focus/blur of the whole control. Not sure it's needed at all as we have the label.

## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/designer/features/style-panel/shared/css-value-input/unit-select.tsx
```diff
@@ -112,6 +112,10 @@ const StyledTrigger = styled(TextFieldIconButton, textStyles, {
   px: 3,
 });
 
+const handlePreventEvent = (event: React.FocusEvent<HTMLDivElement>) => {
```
kof, line 115:
```suggestion
const handlePreventDefault = (event: React.FocusEvent<HTMLDivElement>) => {
```

TrySound (approved):

kof (approved):
if possible check if we can throw out react-use useDebounce so we have a single thing to learn

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -29,10 +29,10 @@ export const TextControl = ({
   >();
 
   return (
-    <Tooltip
+    <EnhancedTooltip
```
kof, line 32: actually we need to use it for all controls, not only text-control

packages/design-system/src/components/tooltip.tsx
```diff
@@ -85,6 +99,102 @@ export const Tooltip = React.forwardRef(function TooltipWrapper(
...
+
+    const onOpenChange = (isOpen: boolean) => {
+      if (allowOpenRef.current) {
+        setOpen(isOpen);
+      }
+    };
+
+    // Wrap in provider so each control will use own hover delay
+    return (
+      <TooltipPrimitive.Provider>
```
kof, line 186: can't we have a provider used separately wrapped around entire style panel? so we can set the settings for timeout on it in a single place?

packages/design-system/src/components/tooltip.tsx
```diff
@@ -85,6 +99,102 @@ export const Tooltip = React.forwardRef(function TooltipWrapper(
   );
 });
 
+/**
+ * EnhancedTooltip has the following differences from the radix-ui
+ * 1. Don't show the tooltip if any click or key-down was made inside Tooltip.Trigger
+ * 2. Show Tooltip on focus after the delay
+ **/
+export const EnhancedTooltip = React.forwardRef(
+  function TextControlTooltipWrapper(
```
kof, line 108: doesn't really need a name

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
```diff
@@ -29,10 +29,10 @@ export const TextControl = ({
...
       content={styleConfig.label}
-      delayDuration={600}
-      disableHoverableContent={true}
+      delayDuration={1600}
+      disableHoverableContent={false}
```
kof, line 35: I think you don't want to set this up for each tooltip individually

packages/design-system/src/components/tooltip.tsx
istarkov: eslint display name rule, changed on Enhanced...

apps/designer/app/designer/features/style-panel/controls/text/text-control.tsx
istarkov: Yep. If `skipDelayDuration = 0` would work as expected.

packages/design-system/src/components/tooltip.tsx
kof: The inner function will not help react dev tools afaik, you still need displayName in the end and everywhere else linter was fine

kof (approved):

PR #681, opened 2022-12-22
Don't populate an instance on paste, because its already populated

kof:
## Description

Can't change width of a copied box 

## Steps for reproduction

1. Add a Box
2. Select the Box and copy it (cmd+c)
3. Select Body
4. Paste (cmd+v)
5. Select the copy you've just pasted
6. Try to change its width
7. Observe that even though the number in input changes the actual width of the Box on canvas stays the same

https://user-images.githubusercontent.com/825702/207803118-5b6fedd8-ded4-4a03-97b9-d4d6cff6710a.mov

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/canvas.tsx
+++ b/apps/designer/app/canvas/canvas.tsx
@@ -123,7 +123,10 @@ const useCopyPaste = () => {
       publish({ type: "deleteInstance", payload: { id: instance.id } });
     },
     onPaste: (instance, props) => {
-      publish({ type: "insertInstance", payload: { instance, props } });
+      publish({
+        type: "insertInstance",
+        payload: { instance, props, isPopulated: true },
+      });
     },
   });
 };
--- a/apps/designer/app/canvas/shared/instance.ts
+++ b/apps/designer/app/canvas/shared/instance.ts
@@ -36,6 +36,7 @@ declare module "~/shared/pubsub" {
       instance: Instance;
       dropTarget?: { parentId: Instance["id"]; position: number };
       props?: Array<UserProp>;
+      isPopulated?: boolean;
     };
     unselectInstance: undefined;
   }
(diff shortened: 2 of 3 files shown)


apps/designer/app/canvas/shared/instance.ts
```diff
@@ -81,35 +82,37 @@ export const useInsertInstance = ({ treeId }: { treeId: string }) => {
...
+    ({ instance, dropTarget, props, isPopulated }) => {
+      store.createTransaction(
+        [rootInstanceContainer, allUserPropsContainer],
+        (rootInstance, allUserProps) => {
+          if (rootInstance === undefined) {
+            return;
+          }
+          const populatedInstance = isPopulated
+            ? instance
+            : utils.tree.populateInstance(instance, breakpoints[0].id);
```
kof, line 96: only this line is changed

apps/designer/app/canvas/canvas.tsx
```diff
@@ -123,7 +123,10 @@ const useCopyPaste = () => {
       publish({ type: "deleteInstance", payload: { id: instance.id } });
     },
     onPaste: (instance, props) => {
-      publish({ type: "insertInstance", payload: { instance, props } });
+      publish({
+        type: "insertInstance",
+        payload: { instance, props, isPopulated: true },
```
TrySound, line 128: I would split events. When drop from components list need to pass only component name so instance could be created in subscription.

kof: currrent way of doing it gives you the benefit of configuring the instance where its being created, might be better long-term

kof: actually populate instance shouldn't be even called inside insertInstance handler, it should be called where a new instance was created and needs to be populated

TrySound (approved):

PR #684, opened 2022-12-23
Enable lost pixel for designer app

TrySound:
We have a lot of features in designer to make snapshot of.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):

TrySound:
Hi @d-ivashchuk. I'm trying to setup lost-pixel for another project and I'm not sure it works. Could you check if I'm doing something wrong?

d-ivashchuk:
hey @TrySound! couple of questions :D is this monorepo lost pixel usage now? would it be used across packages?

@kof we did agree that migrating to bleeding edge lost-pixel makes sense, we could do monorepo support & migration in this PR already. 

Final question - does the team relies on `git rebase`? we are having troubles with this flow as it messes up git history for the repo that we rely on in `lost-pixel-v1`. We are working on the fix, but currently if you are rebasing I would not recommend to switch to next version. 

Legacy lost-pixel that you are using now `gitbased.app` also has monorepo support so no issues, we can make it work as well in the current setup.

Thanks for the answers in advance!

kof:
@d-ivashchuk for some reason I thought we already switched, I think we could avoid rebasing, I almost never rebase personally, but we had a lengthy discussion around it with no particular outcome, because there are tradeoffs for both

kof:
Rebasing is only a problem in wwhich sense? it won't pickup the changes?

d-ivashchuk:
@kof we just agreed that we want to switch :D rebasing will destroy your baseline approvals, so if you will want to create new PR -> approve -> merge normally, it will be ok. nothing that can't be remedied with this. It's different story for teams that **_just_** rebase, we have troubles with that now. but should be working as well, should not be a blocker. hopefully @chriskalmar will find the fix in the nearest future.

TrySound:
We do squash into single commit before merging and rebase PRs from main.

d-ivashchuk:
> We do squash into single commit before merging and rebase PRs from main.

this could be a trouble for now :/ would still be interesting to see, but I don't want to break your current lost-pixel flows just to see that it does not work. give us some time to figure out the problem with squashes + rebases and then we will safely migrate you to the new version of the platform :D 

for now if you can't avoid rebasing - let's stick to file based lost-pixel. I will send you some examples of how the monorepo mode works! but first  I did not see the confirmation for this:

> is this monorepo lost pixel usage now? would it be used across packages?

Did I get this right?

kof:
@d-ivashchuk yes, its always been a monorepo, it was just used before in one package and now we added app support and later other packages as well

d-ivashchuk:
perfect! I can't promise it today because relatives came and I need to entertain them for a while, but over the weekend I will help you set the project up for `gitbased` monorepo, and then as soon as we figure rebases we will migrate you to next :D

TrySound:
We would not have the issue with current setup if we had single storybook setup for whole monorepo.

.github/workflows/vis-reg-tests.yml
```diff
@@ -10,12 +10,24 @@ jobs:
...
           S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
           S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
           LOST_PIXEL_CONFIG_DIR: packages/design-system
+
+      - name: Lost Pixel for Designer
+        uses: lost-pixel/lost-pixel@v2.23.0
+        env:
+          S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
+          S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
+          LOST_PIXEL_CONFIG_DIR: apps/designer
```
d-ivashchuk, line 33:
```suggestion
      - name: Lost Pixel for Design System
        uses: lost-pixel/lost-pixel@v2.23.0
        env:
          S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
          S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
          LOST_PIXEL_CONFIG_DIR: packages/design-system

      - name: Lost Pixel for Designer
        uses: lost-pixel/lost-pixel@v2.23.0
        env:
          S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
          S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
          LOST_PIXEL_CONFIG_DIR: apps/designer
          
      - name: Lost Pixel
        uses: lost-pixel/lost-pixel@v2.24.1
        with:
          FINALIZE: true
```

d-ivashchuk: Happy new year guys!

Please tell me when you are ready to work on this & try this out. Then I'd need to flick the monorepo support for you manually & you will need to accept these changes :D

TrySound: Happy new year! Let's try this

.github/workflows/vis-reg-tests.yml
```diff
@@ -10,12 +10,31 @@ jobs:
...
+          
+      - name: Lost Pixel
+        uses: lost-pixel/lost-pixel@v2.24.1
+        env:
+          LOST_PIXEL_CONFIG_DIR: apps/designer
+        with:
+          FINALIZE: true
```
d-ivashchuk, line 40:
```suggestion
      - name: Lost Pixel
        uses: lost-pixel/lost-pixel@v2.24.1
        env:
          LOST_PIXEL_CONFIG_DIR: apps/designer
          S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
          S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
        with:
          FINALIZE: true
```

d-ivashchuk (commented):
we need to add this(still an after effect of us using legacy version, once we switch to the latest lost-pixel will be far less boilerplate here)

packages/design-system/lostpixel.config.ts
```diff
@@ -1,12 +1,11 @@
-import type { CustomProjectConfig } from "lost-pixel";
 import * as process from "process";
 
-export const config: CustomProjectConfig = {
+export const config = {
   storybookShots: {
     storybookUrl: "packages/design-system/storybook-static",
   },
   lostPixelProjectId: "cl8a8xmx714904901m9oce4vhqc",
```
d-ivashchuk, line 7:
```suggestion
  lostPixelProjectId: "clckoyuix110903101s75d4s71fc",
```

kof:
Seems like we are stuck here? the question is what blocks us, we have all the help we can get from Lostpixel authors

kof:
closing this one since @andrasbacsai has worked on it separately

PR #689, opened 2022-12-23
Fix hover outline

TrySound:
Found an issue when hover a few elements fast and then move cursor out of iframe hover gets stuck on some element. This is because we use combination of mouseover and mouseout events.

So when we move out of one instance to another it looks like this

over
out
over
out

Which may lead to race conditions. So here out is canceled by every hover and delayed a little to always emit last.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/hovered-instance-connector.ts
+++ b/apps/designer/app/canvas/hovered-instance-connector.ts
@@ -35,12 +35,14 @@ const startHoveredInstanceConnection = (rootInstance: Instance) => {
   let mouseOutTimeoutId: undefined | ReturnType<typeof setTimeout> = undefined;
 
   const handleMouseOver = (event: MouseEvent) => {
-    const element = event.target;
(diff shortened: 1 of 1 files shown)


kof:
Maybe we should use mouseenter and mouseleave events instead of hover and out?

kof (approved):

TrySound:
They do not bubble and we listen on window. So we won't be able to catch when element is hovered/unhovered.

kof:
Oh  I just realized we don't even check if the element is an instance, has data-ws-id attribute, it could be something component internal that we can't focus, not an instance!

TrySound:
Will add closest

TrySound:
Do we have anything to test such case with?

kof:
> Do we have anything to test such case with?

not atm, but soon, I would just fake it for now once

PR #694, opened 2022-12-24
Compute style property source

TrySound:

Rewritten inherited style logic and fixed inheriting from all parents not only one. Added cascaded style logic to provide additional insight where the value comes from.

Screenshot 2022-12-24 at 21 32 58
Screenshot 2022-12-24 at 21 33 03
Screenshot 2022-12-24 at 21 33 19
Screenshot 2022-12-24 at 21 33 35

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/designer/features/style-panel/shared/style-source.ts
```diff
@@ -0,0 +1,128 @@
...
+    if (child.type === "instance") {
+      if (child.id === instanceId) {
+        // I am your father
+        parents.push(parentInstance);
+        break;
+      }
+      const foundParents = findParents(child, instanceId);
+      // add parent if descendants matched
+      if (foundParents.length !== 0) {
+        parents.push(parentInstance, ...foundParents);
```
istarkov, line 92: I think we can `break` here too as obviously all other children are not ascendants/parents of your node

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -220,6 +251,8 @@ export const useStyleData = ({
 
   return {
     currentStyle,
+    setStyle: cssRule?.style ?? ({} as Style),
```
istarkov, line 254: setStyle looks like function name but its not

kof:
I will test it today, @TrySound looks like we should add unit tests to all the utility functions?

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
TrySound: Stupid word

kof: should we rename all "set" usages to "local"?

istarkov:
Now we have:
1. currentStyle - merge of browserStyle and current breakpoint style
2. inheritedStyle - merge of inherited styles from parent and other than the self breakpoint
3. cascadedStyle - styles inherited from other breakpoints

browserStyle contains inheritedStyle + cascadedStyle, but the source of browserStyle is browser computed styles.

At least would be good to add VSCode comments to each property, we return from `useStyleData`.

The other thing that I already have task where I need to know parent `currentStyle` or at least
parent `setStyle` + `inheritedStyle` + `cascadedStyle`

For me would be much easier if all that calculation would be done using the other hook, like

```js
const useStylesFromTree = (instanceId) => {
  ...
  return {
     style: InstanceCSSRulesForVreakpoint,
     inheritedStyle,
     ...etc
  }
}

```

apps/designer/app/designer/features/style-panel/shared/style-source.ts
```diff
@@ -0,0 +1,128 @@
...
+          breakpointId: breakpoint.id,
+          value,
+        };
+      }
+    }
+  }
+  return cascadedStyle;
+};
+
+const findParents = (parentInstance: Instance, instanceId: Instance["id"]) => {
```
istarkov, line 80: Also I think you must use utils.tree "find" methods.
So when they would be optimized you will automatically get fast code.
i.e. utils.tree.findParentInstance

kof: absolutely, there should be no external tree traversing functions other than the once in tree utils so we know in a central place what to optimize for

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
TrySound: localStyle, ownStyle, currentStyle, selectedStyle (for breakpoint and instance)

kof: @TrySound are those suggestions for the same thing?

TrySound: Yes

apps/designer/app/designer/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -82,14 +93,24 @@ const FlexChildSectionAlign = (props: RenderCategoryProps) => {
...
 
 const FlexChildSectionSizing = (props: RenderCategoryProps) => {
-  const { createBatchUpdate, currentStyle, sectionStyle } = props;
+  const {
+    createBatchUpdate,
+    currentStyle,
+    setStyle,
+    cascadedStyle,
+    inheritedStyle,
+    sectionStyle,
```
istarkov, line 102: All for has similar names but different types. 
Like setStyle is really { property: StyleValue }
cascadedStyle is container `{ property: { breakpointId, value: StyleValue } }`
inheritedStyle is container `{ property: { instanceId, value: StyleValue } }`
sectionStyle is container `{property: RenderPropertyProps;}`

So nothing here except `setStyle` is Style ;-)

apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
```diff
@@ -1,6 +1,6 @@
 import type { Style, StyleValue, StyleProperty } from "@webstudio-is/css-data";
 import { toValue } from "@webstudio-is/css-engine";
-import type { InheritedStyle } from "./get-inherited-style";
+import type { InheritedStyle } from "./style-source";
```
istarkov, line 3: I have a feeling that  `getFinalValue` is wrong. If we already injected `browserStyles` into currentStyle at `useStyleData` why we didn't do that there too and use some special `getFinalValue` ?

Logic like 
```
  if (currentValue?.type !== "unit" && isNumericString(toValue(currentValue))) {
    return {
      value: Number(toValue(currentValue)),
      type: "unit",
      unit: "number",
    };
  }
  if (
    currentValue &&
    currentValue.type === "unit" &&
    isValid(
      property,
      toValue({ ...currentValue, value: Math.abs(currentValue.value) })
    ) === false
  ) {
    return { value: String(currentValue.value), type: "invalid" };
  }
```
also looks useless.

Why I'm writing this here, that without inheritedStyle looks like everything is needed is 
`currentStyle` and `getStyleSource`

# MAIN IDEA
Why not just return `getStyleSource: property => source` at `useStyleDate`
Having that we don't use other styles

PR #698, opened 2022-12-26
Drop inherited style from style panel

TrySound:
The logic does not work correctly and not really used because we never set values to "inherit".

Later new logic will be added to show the source of values.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
```diff
@@ -9,19 +8,12 @@ export const isNumericString = (input: string) =>
...
   currentStyle: Style;
-  inheritedStyle: InheritedStyle;
   property: StyleProperty;
 }): StyleValue | undefined => {
   const currentValue = currentStyle[property];
-  const inheritedValue =
-    property in inheritedStyle ? inheritedStyle[property].value : undefined;
-  if (toValue(currentValue) === "inherit" && inheritedValue !== undefined) {
-    return inheritedValue;
-  }
```
istarkov, line 24: The whole code below seems like useless too, and controls dependent.
Best IMO would be or remove or set `warnOnce` the code path should not occur there.

istarkov (approved):
I haven't knew that computed styles never contain inherit too.

apps/designer/app/designer/features/style-panel/shared/get-final-value.ts
TrySound: Will leave it to you

kof (approved):

PR #702, opened 2022-12-26
Access style config by property

TrySound:
Since style config is hardcoded in module scope we can access it by property instead of passing style config everywhere.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
I know there was a reason for passing it, but don't remember what was it, so it might come back

istarkov (approved):

PR #704, opened 2022-12-26
Add style info with source data

TrySound:
This is a step toward https://github.com/webstudio-is/webstudio-designer/issues/29

When merge browser style and local styles we can specify the source of the style. Later we can add additional info about inherited and cascaded styles.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -2,8 +2,8 @@ import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
 import { useMemo } from "react";
 
 export type StylePropertyInfo = {
-  source: "local" | "external";
   value: StyleValue;
+  local?: StyleValue;
 };
 
 export type StyleInfo = {
@@ -21,14 +21,13 @@ export const useStyleInfo = ({
     const styleInfoData: StyleInfo = {};
     for (const [property, value] of Object.entries(browserStyle)) {
       styleInfoData[property as StyleProperty] = {
-        source: "external",
         value,
       };
     }
     for (const [property, value] of Object.entries(localStyle)) {
       styleInfoData[property as StyleProperty] = {
-        source: "local",
         value,
+        local: value,
       };
     }
     return styleInfoData;
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -11,10 +11,10 @@ export type StyleInfo = {
 };
 
 export const useStyleInfo = ({
-  breakpointStyle,
+  localStyle,
   browserStyle,
 }: {
-  breakpointStyle: Style;
+  localStyle: Style;
   browserStyle: Style;
 }) => {
   const styleInfoData = useMemo(() => {
@@ -25,14 +25,14 @@ export const useStyleInfo = ({
         value,
       };
     }
-    for (const [property, value] of Object.entries(breakpointStyle)) {
+    for (const [property, value] of Object.entries(localStyle)) {
       styleInfoData[property as StyleProperty] = {
         source: "local",
         value,
       };
     }
     return styleInfoData;
-  }, [browserStyle, breakpointStyle]);
+  }, [browserStyle, localStyle]);
 
   return styleInfoData;
 };
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -1,13 +1,13 @@
 import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
 import { useMemo } from "react";
 
-export type StylePropertyInfo = {
+export type StyleValueInfo = {
   value: StyleValue;
   local?: StyleValue;
 };
 
 export type StyleInfo = {
-  [property in StyleProperty]?: StylePropertyInfo;
+  [property in StyleProperty]?: StyleValueInfo;
 };
 
 export const useStyleInfo = ({
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -1,0 +1,38 @@
+import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
+import { useMemo } from "react";
+
+export type StylePropertyInfo = {
+  source: "local" | "external";
+  value: StyleValue;
+};
+
+export type StyleInfo = {
+  [property in StyleProperty]?: StylePropertyInfo;
+};
+
+export const useStyleInfo = ({
+  breakpointStyle,
+  browserStyle,
+}: {
+  breakpointStyle: Style;
+  browserStyle: Style;
+}) => {
+  const styleInfoData = useMemo(() => {
+    const styleInfoData: StyleInfo = {};
+    for (const [property, value] of Object.entries(browserStyle)) {
+      styleInfoData[property as StyleProperty] = {
+        source: "external",
+        value,
+      };
+    }
+    for (const [property, value] of Object.entries(breakpointStyle)) {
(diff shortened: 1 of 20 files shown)


apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -0,0 +1,38 @@
...
+export type StylePropertyInfo = {
+  source: "local" | "external";
+  value: StyleValue;
+};
+
+export type StyleInfo = {
+  [property in StyleProperty]?: StylePropertyInfo;
+};
+
+export const useStyleInfo = ({
```
istarkov, line 13: I think filename must me use-style-info

istarkov (approved):

apps/designer/app/designer/features/style-panel/shared/style-info.ts
TrySound: 1. locks with single hook export
2. we often provide non-hook api

I prefer to name files as types rather than utilities of these types

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -0,0 +1,38 @@
+import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
+import { useMemo } from "react";
+
+export type StylePropertyInfo = {
+  source: "local" | "external";
```
kof, line 5: the opposite of local would be remote

kof: or otherwise it would be internal/external

apps/designer/app/designer/features/style-panel/shared/style-info.ts
kof: I call them by the function name if its the only exported function

kof: this helps preventing adding more functions to it, but sometimes it makes sense to put more functions into the same module, in which case `style-info` kind of naming makes more sense

apps/designer/app/designer/features/style-panel/shared/style-info.ts
kof: The problem with "generalizing" the source type early on like you do here is that if you wanted to know a more specific source (inherited/cascaded etc) somewhere later, closer to the edge - you can't.

I would much rather have the source very presize on the datatype level and use a utility function where you need to use it that would convert it to the simplistic type like local/remote.

This way you would have both - presize data and generalized type.

kof (changes requested):
lets discuss where we define the source in precise form and where we start generalizing it. In my opinion you are generalizing too early.

TrySound:
> In my opinion you are generalizing too early.

This is info I actually need but yeah switched to different approach. Now it's style.local - if defined then "local" value.

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -0,0 +1,37 @@
+import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
+import { useMemo } from "react";
+
+export type StylePropertyInfo = {
+  value: StyleValue;
```
kof, line 5: this value is the `computed`, right? it seems for the consumers, it's now unclear what the value is,  because you don't see how its defined

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -0,0 +1,37 @@
+import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
+import { useMemo } from "react";
+
+export type StylePropertyInfo = {
```
kof, line 4: Seems like `StyleValueInfo` or maybe better `StyleValueData` since its actual data

apps/designer/app/designer/features/style-panel/shared/style-info.ts
TrySound: It's computed from browser and local style. Later it will be computed from inherited and cascaded

kof: computed from the browser always results in the final value, when all things come together, right?

TrySound: Yes, but we don't wanna always use it because browser computes in pixels.

TrySound: So everything we can find in data will give final result and then fallback to computed from browser

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -58,13 +59,10 @@ export const useStyleData = ({
...
 
-  const currentStyle = useMemo<Style>(
-    () => ({
-      ...selectedInstanceData?.browserStyle,
-      ...breakpointStyle,
-    }),
-    [selectedInstanceData?.browserStyle, breakpointStyle]
-  );
+  const currentStyle = useStyleInfo({
+    localStyle: breakpointStyle,
```
kof, line 63: can't we rename it everywhere consistently?

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -58,13 +59,10 @@ export const useStyleData = ({
...
+    browserStyle: selectedInstanceData?.browserStyle ?? {},
```
kof, line 64: selectedInstanceData can only be undefined if there is no selected instance, in this case we shouldn't have have to call useStyleInfo at all

kof: having a default there makes this call and the logic inside useStyleInfo confusing because we can't have an empty browserStyle actually

apps/designer/app/designer/features/style-panel/shared/style-info.ts
kof: sure, but then we shouldn't be using computed as a value at all?

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
TrySound: Can be done later. Don't want to put all change here.

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
TrySound: It's just the check is done later and since this is a hook we can't call it after check. So I'd say good for now. Later we will improve this.

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -0,0 +1,37 @@
...
+  browserStyle: Style;
+}) => {
+  const styleInfoData = useMemo(() => {
+    const styleInfoData: StyleInfo = {};
+    for (const [property, value] of Object.entries(browserStyle)) {
+      styleInfoData[property as StyleProperty] = {
+        value,
+      };
+    }
+    for (const [property, value] of Object.entries(localStyle)) {
```
kof, line 27: we figured it out: this is a temporarily overriding value, we will start using a combination of local and default (normal/auto etc) as a value, lets add this as a comment

kof (approved):

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
kof: yeah I know, but it's still confusing to have a default object there, I think maybe having undefined inside useStyleInfo is more logical?

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
kof, line 63: localStyle can't be also available if instance wasn't selected ... weird

TrySound: It's casted in old code

Issue #649, opened 2022-12-17
Review the decision on how to show computed style values in style panel


istarkov:
Currently, we show computed values for elements. 
For example instead of `width: auto` we show width px size.
<img width="225" alt="image" src="https://user-images.githubusercontent.com/5077042/208242946-4e714d45-fe00-49d4-bdb6-25927e917337.png">

This is confusing, for example, "Webflow" shows "Auto" with "placeholder color" in case of `width` is not set. Same for other props.
<img width="240" alt="image" src="https://user-images.githubusercontent.com/5077042/208243003-ab1f76be-78ae-4805-ab05-c8eb7de6c14f.png">

If I change the "width" property even on the same (default) value, I see its value with the ordinary `input color` simultaneously with the label background color. _(see width input font color)_

<img width="246" alt="image" src="https://user-images.githubusercontent.com/5077042/208243035-1f80fc45-fc86-4f3d-a188-5be6bc5fe637.png">

In case I need `scrub` in Webflow I click the `alt` button and scrub works with the "computed value".

I think we also need to show "real" values instead of computed ones, and use placeholders (colors) for undefined values (as IMO only color label selection is not enough _I don't have in my mind that label color somehow affects input behavior_)

kof:
@taylornowotny I had an idea to show a computed value when option/alt is pressed and mouse is over the input and otherwise show the value that we have saved, not the computed one

taylornowotny:
If a value is "auto" then why would a user need to know the computed value? Surely that value will be different on different screens, so how would the computed value be useful?

kof:
There are countless situations when you debug a problem, where you need to understand what the actual value is that CSS engine has computed. This is the only way to get to the truth of the problem. It the source of truth.

kof:
When you scrub, even when the value was auto before, you also start from a current computed value

kof:
We could also think of a debug panel that would show all actual values and more

taylornowotny:
Ok just asking becuase I have never been in that situation personally

I agree with 

> I think we also need to show "real" values instead of computed ones, and use placeholders (colors) for undefined values 

Except placeholder is not used for undefined values. It is used for default values, ie. values that have not been applied by the user. For example, a div with no width or height applied has a width and height of auto naturally. This is represented by placeholder text. If a user sets the width to Auto deliberately, then it is no longer placeholder text. The difference in the input + the color of the label tells you that there is a width of auto applied to the element. 

[...]

PR #712, opened 2022-12-27
Fallback to cascaded and inherited when value not set

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/29 https://github.com/webstudio-is/webstudio-designer/issues/649

Now instead of showing computed styles which come from parent instances or other breakpoints we can show actual value we set there.

### Inherited styles

1. add Box to canvas
2. change font size on Body to 2rem
3. check Box font size

### Cascaded styles

1. add Box to canvas
2. change font weight on Box to 600
3. select tablet breakpoint
4. check Box font weight

## Code Review

  - test it on preview
- [x] hi @istarkov, I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -15,29 +15,29 @@ import {
   useSelectedInstanceData,
 } from "~/designer/shared/nano-states";
 
-type Cascaded = {
+type CascadedValueInfo = {
   breakpointId: string;
   value: StyleValue;
 };
 
-type CascadedInfo = {
-  [property in StyleProperty]?: Cascaded;
+type CascadedProperties = {
+  [property in StyleProperty]?: CascadedValueInfo;
 };
 
-type Inherited = {
+type InheritedValueInfo = {
   instanceId: string;
   value: StyleValue;
 };
 
-type InheritedInfo = {
-  [property in StyleProperty]?: Inherited;
+type InheritedProperties = {
+  [property in StyleProperty]?: InheritedValueInfo;
 };
 
 export type StyleValueInfo = {
   value: StyleValue;
   local?: StyleValue;
-  cascaded?: Cascaded;
-  inherited?: Inherited;
+  cascaded?: CascadedValueInfo;
+  inherited?: InheritedValueInfo;
 };
 
 export type StyleInfo = {
@@ -99,7 +99,7 @@ export const getCascadedInfo = (
       styles.set(rule.breakpoint, rule.style);
     }
   }
-  const cascadedStyle: CascadedInfo = {};
+  const cascadedStyle: CascadedProperties = {};
   for (const breakpointId of cascadedBreakpointIds) {
     const style = styles.get(breakpointId);
     if (style !== undefined) {
@@ -123,7 +123,7 @@ export const getInheritedInfo = (
   instanceId: string,
   cascadedAndSelectedBreakpoints: string[]
 ) => {
-  const inheritedStyle: InheritedInfo = {};
+  const inheritedStyle: InheritedProperties = {};
   const parents = utils.tree.getInstancePath(rootInstance, instanceId);
   for (const parentInstance of parents) {
     // skip current element
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -122,20 +122,20 @@ export const getInheritedInfo = (
   cascadedAndSelectedBreakpoints: string[]
 ) => {
   const inheritedStyle: InheritedProperties = {};
-  const parents = utils.tree.getInstancePath(rootInstance, instanceId);
-  for (const parentInstance of parents) {
+  const ancestors = utils.tree.getInstancePath(rootInstance, instanceId);
+  for (const ancestorInstance of ancestors) {
     // skip current element
-    if (parentInstance.id === instanceId) {
+    if (ancestorInstance.id === instanceId) {
       continue;
     }
     const cascadedStyle = getCascadedInfo(
-      parentInstance.cssRules,
+      ancestorInstance.cssRules,
       cascadedAndSelectedBreakpoints
     );
     for (const [property, cascaded] of Object.entries(cascadedStyle)) {
       if (cascaded !== undefined && inheritableProperties.has(property)) {
         inheritedStyle[property as StyleProperty] = {
-          instanceId: parentInstance.id,
+          instanceId: ancestorInstance.id,
           value: cascaded.value,
         };
       }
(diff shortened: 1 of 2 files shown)


apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -23,31 +57,164 @@ export const getStyleSource = (
...
+/**
+ * extract all styles from breakpoints
+ * affecting current view
+ */
+export const getCascadedInfo = (
+  cssRules: CssRule[],
+  cascadedBreakpointIds: string[]
+) => {
+  const styles = new Map<string, Style>();
+  for (const rule of cssRules) {
```
istarkov, line 97: Just optimization, `cssRules` probably already contains all rules sorted by breakpoints (not sure)
in that case, we can just check that `cascadedBreakpointIds.includes(rule.breakpoint)` and fill `cascadedStyle` directly here.

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -23,31 +57,164 @@ export const getStyleSource = (
...
+ * extract all inheritable styles from ancestor instances
+ * including active breakpoints
+ */
+export const getInheritedInfo = (
+  rootInstance: Instance,
+  instanceId: string,
+  cascadedAndSelectedBreakpoints: string[]
+) => {
+  const inheritedStyle: InheritedInfo = {};
+  const parents = utils.tree.getInstancePath(rootInstance, instanceId);
```
istarkov, line 127: `ancestors` instead of parents looks better

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -23,31 +57,164 @@ export const getStyleSource = (
...
+  for (const parentInstance of parents) {
+    // skip current element
+    if (parentInstance.id === instanceId) {
```
istarkov, line 130: should be first or last always, may be slice it before the "for" cycle?

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -23,31 +57,164 @@ export const getStyleSource = (
...
+      continue;
+    }
+    const cascadedStyle = getCascadedInfo(
+      parentInstance.cssRules,
+      cascadedAndSelectedBreakpoints
+    );
+    for (const [property, cascaded] of Object.entries(cascadedStyle)) {
+      if (cascaded !== undefined && inheritableProperties.has(property)) {
+        inheritedStyle[property as StyleProperty] = {
```
istarkov, line 139: From the code above it looks like every inheritable property must be in cascadedStyle, but cascaded doesn't include the current breakpoint style. What if it's inherited from the current and not included in any cascaded?

istarkov: I got, `cascadedAndSelectedBreakpoints`

istarkov: Probably better to split as everywhere we see cascadedBreakpoints, but here not. May be pass selected as an argument and to merge it here in function? Anyway ok, just little bit confusing after previous functions ;-)

apps/designer/app/designer/features/style-panel/shared/style-info.ts
TrySound: Potentially it will be insertion order. So I would keep it and refactor once we migrate to normalized styles.

apps/designer/app/designer/features/style-panel/shared/style-info.ts
TrySound: Just thought this way will be more optimized than slice

istarkov (approved):
Super readable for such logic!!!

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -23,31 +57,164 @@ export const getStyleSource = (
...
+/**
+ * find all breakpoints which may affect current view
+ */
+export const getCascadedBreakpointIds = (
+  breakpoints: Breakpoint[],
+  selectedBreakpointId?: string
+) => {
+  const sortedBreakpoints = breakpoints
+    .slice()
+    .sort((a, b) => a.minWidth - b.minWidth);
```
kof, line 77: we have a breakpoints.sort utility, its not hard to implement, but over time, the sorting logic might become more complex if we add maxWidth or any other breakpoint params

apps/designer/app/designer/features/style-panel/shared/style-info.ts
kof: yep, its the right term

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -1,9 +1,43 @@
...
+import { properties } from "@webstudio-is/css-data";
+import { utils } from "@webstudio-is/project";
+import type { Instance } from "@webstudio-is/react-sdk";
+import { useBreakpoints, useRootInstance } from "~/shared/nano-states";
+import {
+  useSelectedBreakpoint,
+  useSelectedInstanceData,
+} from "~/designer/shared/nano-states";
+
+type Cascaded = {
```
kof, line 18: Trying to find a more precise name for this, CascadedValueContainer is very correct and very ugly

kof: Maybe CascadedInfo

kof: or CascadedValueInfo

TrySound: I don't think we will add "ValueContaine" to every computed value. This is an ugly part tbh.

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -1,9 +1,43 @@
...
+  breakpointId: string;
+  value: StyleValue;
+};
+
+type CascadedInfo = {
```
kof, line 23: how about `CascadedProperties`?

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -1,9 +1,43 @@
...
+  [property in StyleProperty]?: Cascaded;
+};
+
+type Inherited = {
```
kof, line 27: same here InheritedValueInfo

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -1,9 +1,43 @@
...
+  instanceId: string;
+  value: StyleValue;
+};
+
+type InheritedInfo = {
```
kof, line 32: and InheritedProperties

kof (approved):
added a bunch of naming suggestions

apps/designer/app/designer/features/style-panel/shared/style-info.ts
kof: something that is semantically not clear is long-term worse than something that has a long name

PR #713, opened 2022-12-28
Show remote style source in UI

TrySound:
Added remote (ex inherited) highlight for controls.
Migrated menu control to new icon button.

Screenshot 2022-12-28 at 11 17 00
Screenshot 2022-12-28 at 11 17 36
Screenshot 2022-12-28 at 11 18 36
Screenshot 2022-12-28 at 11 24 46

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/controls/menu/menu-control.tsx
+++ b/apps/designer/app/designer/features/style-panel/controls/menu/menu-control.tsx
@@ -41,13 +41,20 @@ export const MenuControl = ({
       };
     })
     .filter((item) => item.icon);
+  let variant: "default" | "set" | "inherited" = "default";
+  if (styleSource === "local") {
+    variant = "set";
+  }
+  if (styleSource === "remote") {
+    variant = "inherited";
+  }
   return (
     <IconButtonWithMenu
+      variant={variant}
       icon={items.find(({ name }) => name === currentValue)?.icon}
       label={label}
       items={items}
       value={String(currentValue)}
-      isActive={styleSource === "local"}
       onChange={setValue}
       onHover={(value) => setValue(value, { isEphemeral: true })}
     />
--- a/apps/designer/app/designer/features/style-panel/shared/style-info.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/style-info.ts
@@ -49,6 +49,8 @@ export type StyleSource = "local" | "remote" | "preset";
 export const getStyleSource = (
   ...styleValueInfos: (undefined | StyleValueInfo)[]
 ): StyleSource => {
+  // show source to use if at least one of control properties matches
+  // so user could see if something is set or something is inherited
   for (const info of styleValueInfos) {
     if (info?.local) {
       return "local";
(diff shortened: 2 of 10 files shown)


apps/designer/app/designer/features/style-panel/controls/menu/menu-control.tsx
```diff
@@ -41,13 +41,20 @@ export const MenuControl = ({
       };
     })
     .filter((item) => item.icon);
+  let variant: "default" | "set" | "inherited" = "default";
+  if (styleSource === "local") {
+    variant = "set";
+  }
+  if (styleSource === "remote") {
+    variant = "inherited";
+  }
```
kof, line 50: We need to unify this logic and make all controls take a `color` prop

kof: we should also start using semantic colors for these (ds in figma has fully switched to semantic colors, we haven't migrated yet but we can start one by one)

TrySound: Color is too optimistic to call it. These are still variants with multiple colors. And there would be 2 blue for "set/local" and "active". We just need to switch to new terms.

TrySound: Anyway let's do this in other PR

kof: color prop is meant to be interpreted as a "color scheme", we researched all options there

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -54,6 +54,11 @@ export const getStyleSource = (
       return "local";
     }
   }
+  for (const info of styleValueInfos) {
```
istarkov, line 57: Having that `getStyleSource` of multiple properties is used in 1 place, may be remove `styleValueInfos` array from argument. As I'm not getting logic why do we return 'local' if any of the info is local,
then do the same with remote? Or at least comment on the function.

istarkov (approved):

apps/designer/app/designer/features/style-panel/shared/style-info.ts
TrySound: It's trade off for controls with multiple properties. We need to get at least one match

kof (approved):
still need to change to use a consistent color prop with the same semantic name in all controls

apps/designer/app/designer/features/style-panel/shared/style-info.ts
```diff
@@ -44,16 +44,23 @@ export type StyleInfo = {
   [property in StyleProperty]?: StyleValueInfo;
 };
 
-export type StyleSource = "local" | "preset";
+export type StyleSource = "local" | "remote" | "preset";
```
kof, line 47: maybe a better name should have been StyleOrigin here, so that `styleSource = {origin: 'local'}`

PR #714, opened 2022-12-28
Restrict assets uploading by MAX_ASSETS_PER_PROJECT env variable

istarkov:
Set at VERCEL env 
image

## Description

Restrict assets by MAX_ASSETS_PER_PROJECT env variable.

## Steps for reproduction

- [x] - set MAX_ASSETS_PER_PROJECT = 4 as env variable for Vercel

Try upload more than 4 files.
Get Error: "`Max ${count} assets per project reached`"

## Code Review

  - test it on preview
- [ ] hi @TrySound plz check code

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/.env.example
+++ b/apps/designer/.env.example
@@ -6,7 +6,7 @@ GH_CLIENT_ID=
 DEV_LOGIN=true
 
 # Restrictions
-MAX_ASSETS_PER_PROJECT=
+MAX_ASSETS_PER_PROJECT=50
 
 # relative to the public directory
 FILE_UPLOAD_PATH="uploads"
--- a/apps/designer/.env.example
+++ b/apps/designer/.env.example
@@ -5,6 +5,8 @@ GH_CLIENT_SECRET=
 GH_CLIENT_ID=
 DEV_LOGIN=true
 
+# Restrictions
+MAX_ASSETS_PER_PROJECT=
 
 # relative to the public directory
 FILE_UPLOAD_PATH="uploads"
@@ -32,7 +34,7 @@ PUBLISHER_TOKEN=
 # Will be detected automatically in development environments like localhost or preview deployment
 BUILD_ORIGIN=
 
-# If set to `true`, user-generated build will be available 
+# If set to `true`, user-generated build will be available
 # only at {project.domain} subdomain of {BUILD_ORIGIN}.
 # This is used in production deployment for security reasons.
 BUILD_REQUIRE_SUBDOMAIN=
--- a/packages/asset-uploader/src/db/create.ts
+++ b/packages/asset-uploader/src/db/create.ts
@@ -59,7 +59,7 @@ export const createAssetWithLimit = async (
        * it's probable that the user can exceed the limit a little bit.
        * So it can be a little bit strange that the limit is 5 but the user already has 7.
        **/
-      throw new Error(`Max ${count} assets per project reached`);
+      throw new Error(`The maximum number of assets per project is ${count}.`);
     }
 
     /**
@@ -88,7 +88,7 @@ export const createAssetWithLimit = async (
 
     const asset = await uploadAsset();
 
-    const size = asset.size || 0;
+    const size = asset.size ?? 0;
     const { id, meta, format, name, location } = asset;
 
     const dbAsset = await prisma.asset.update({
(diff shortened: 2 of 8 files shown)


packages/asset-uploader/src/db/create.ts
```diff
@@ -22,28 +26,91 @@ type Options =
...
+        size: 0,
+      },
+      select: {
+        id: true,
+      },
+    });
+
+    const asset = await uploadAsset();
+
+    const size = asset.size || 0;
```
TrySound, line 91: ??

TrySound (approved):

apps/designer/.env.example
```diff
@@ -5,6 +5,8 @@ GH_CLIENT_SECRET=
 GH_CLIENT_ID=
 DEV_LOGIN=true
 
+# Restrictions
+MAX_ASSETS_PER_PROJECT=
```
kof, line 9: we should have examples with some useful value

kof:
>  - set MAX_ASSETS_PER_PROJECT = 4 as env variable for Vercel
If max assets is 4, then it should allow uploading 4 and only error with 5

packages/asset-uploader/src/db/create.ts
```diff
@@ -22,28 +26,91 @@ type Options =
...
+      },
+    });
+
+    if (count >= env.MAX_ASSETS_PER_PROJECT) {
+      /**
+       * Here is right to write `Max ${MAX_ASSETS_PER_PROJECT}` but see the comment below,
+       * it's probable that the user can exceed the limit a little bit.
+       * So it can be a little bit strange that the limit is 5 but the user already has 7.
+       **/
+      throw new Error(`Max ${count} assets per project reached`);
```
kof, line 62:
```suggestion
      throw new Error(`The maximum number of assets per project is ${count}.`);
```

packages/asset-uploader/src/schema.ts
```diff
@@ -15,6 +15,10 @@ const maxSize = z
   .transform(toBytes)
   .default(MAX_UPLOAD_SIZE);
 
+export const Env = z.object({
+  MAX_ASSETS_PER_PROJECT: z.string().default("50").transform(Number.parseFloat),
```
kof, line 19: probably this should be just added to both S3Env and FsEnv just like MAX_UPLOAD_SIZE?

kof (approved):

packages/asset-uploader/src/schema.ts
istarkov: This Env is used in create, we can make it as a param if needed for `createAssetWithLimit`.  But not sure why

packages/asset-uploader/src/targets/s3/upload.ts
```diff
@@ -34,32 +38,36 @@ export const uploadToS3 = async ({
...
+      )
+    );
+
+    const imagesFormData = formData.getAll("image") as Array<string>;
+    const fontsFormData = formData.getAll("font") as Array<string>;
+    const ids = Ids.parse(formData.getAll(idsFormDataFieldName));
+
+    const assetsData = [...imagesFormData, ...fontsFormData]
+      .slice(0, MAX_FILES_PER_REQUEST)
+      .map((dataString, i) => {
```
kof, line 62:
```suggestion
      .map((dataString, index) => {
```

packages/asset-uploader/src/targets/s3/upload.ts
```diff
@@ -34,32 +38,36 @@ export const uploadToS3 = async ({
...
+        // @todo validate with zod
+        return { ...JSON.parse(dataString), id: ids[i] };
```
kof, line 64: looks a little suspicious that ids index would correspond to assets

istarkov: Yep, the only thing that we restrict arrays to `size 1` and we know that we pass the same arrays length=1 from the client.
Looks like the only way this not being suspicious is to pass Blob instead of the file from the client and parse id and file content from the Blob.

kof: or run some validation before using the index

istarkov: Added zod validation of the whole final object.

istarkov:
Seems like we need to restrict assets by type = 'font' | 'image'

PR #715, opened 2022-12-28
Show in popover where remote style comes from

TrySound:

Here added info in popover where cascaded and inherited styles come from.

Screenshot 2022-12-28 at 14 47 01
Screenshot 2022-12-28 at 14 47 07
Screenshot 2022-12-28 at 14 47 19
Screenshot 2022-12-28 at 14 47 47

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/property-name.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/property-name.tsx
@@ -18,8 +18,9 @@ import { UndoIcon } from "@webstudio-is/icons";
 import type { Instance } from "@webstudio-is/react-sdk";
 import { utils } from "@webstudio-is/project";
 import { isFeatureEnabled } from "~/shared/feature-flags";
-import { type StyleInfo, type StyleSource, getStyleSource } from "./style-info";
+import { useSelectedInstanceData } from "~/designer/shared/nano-states";
 import { useBreakpoints, useRootInstance } from "~/shared/nano-states";
+import { type StyleInfo, type StyleSource, getStyleSource } from "./style-info";
 
 const PropertyPopoverContent = ({
   properties,
@@ -32,8 +33,10 @@ const PropertyPopoverContent = ({
   styleSource: StyleSource;
   onReset: () => void;
 }) => {
-  const [rootInstance] = useRootInstance();
   const [breakpoints] = useBreakpoints();
+  const [rootInstance] = useRootInstance();
+  const [selectedInstanceData] = useSelectedInstanceData();
+  const selectedInstanceId = selectedInstanceData?.id;
 
   // @todo replace with normalized data access
 
@@ -47,21 +50,18 @@ const PropertyPopoverContent = ({
 
   const instancesMap = useMemo(() => {
     const instancesMap: Record<string, Instance> = {};
-    if (rootInstance === undefined) {
+    if (rootInstance === undefined || selectedInstanceId === undefined) {
       return instancesMap;
     }
-    for (const property of properties) {
-      const styleValueInfo = style[property];
-      if (styleValueInfo?.inherited) {
-        const { instanceId } = styleValueInfo.inherited;
-        const instance = utils.tree.findInstanceById(rootInstance, instanceId);
-        if (instance) {
-          instancesMap[instanceId] = instance;
-        }
-      }
+    const ancestors = utils.tree.getInstancePath(
+      rootInstance,
+      selectedInstanceId
+    );
+    for (const ancestor of ancestors) {
+      instancesMap[ancestor.id] = ancestor;
     }
     return instancesMap;
-  }, [rootInstance, styleSource]);
+  }, [rootInstance, styleSource, selectedInstanceId]);
 
   if (styleSource === "local") {
     return (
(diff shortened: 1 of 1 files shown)


istarkov (approved):

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -14,8 +15,126 @@ import {
...
+  const breakpointsMap = useMemo(() => {
+    const breakpointsMap: Record<string, Breakpoint> = {};
+    for (const breakpoint of breakpoints) {
+      breakpointsMap[breakpoint.id] = breakpoint;
+    }
+    return breakpointsMap;
+  }, [breakpoints]);
+
+  const instancesMap = useMemo(() => {
+    const instancesMap: Record<string, Instance> = {};
```
kof, line 52: why not actual Map?

kof: actually looking how its being used
1. I think this map seems like an unnecessary optimization because depth won't be hundreds of items, so you could just lookup the array 
2. do we really need useMemo if getInstancePath is gonna be fast? we plan to optimize it with the new tree abstraction we planed, so building it should be very fast and this content is only computed when visible ...

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -14,8 +15,126 @@ import {
...
+  onReset: () => void;
+}) => {
+  const [breakpoints] = useBreakpoints();
+  const [rootInstance] = useRootInstance();
+  const [selectedInstanceData] = useSelectedInstanceData();
+  const selectedInstanceId = selectedInstanceData?.id;
+
+  // @todo replace with normalized data access
+
+  const breakpointsMap = useMemo(() => {
```
kof, line 43: same here, I would think this is not needed optimization, both building a map and using useMemo, because its like 5 items most of the time

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -14,8 +15,126 @@ import {
...
+  return (
+    <Box css={{ px: "$spacing$4", py: "$spacing$3" }}>
+      {properties.map((property) => {
+        const styleValueInfo = style[property];
+
+        if (styleValueInfo?.cascaded) {
+          const { breakpointId } = styleValueInfo.cascaded;
+          return (
+            <Text key={property} color="hint">
+              {property} value is cascaded from{" "}
```
kof, line 119: looks weird `from{" "}`

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
```diff
@@ -14,8 +15,126 @@ import {
...
+            </Text>
+          );
+        }
+
+        if (styleValueInfo?.inherited) {
+          const { instanceId } = styleValueInfo.inherited;
+          return (
+            <Text key={property} color="hint">
+              {property} value is inherited from{" "}
+              {instancesMap[instanceId].component}
```
kof, line 130: same

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: We won't need this code at all after normalization

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: Will be removed eventually

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: Prettier splits lines and this space is necessary to not collapse jsx lines. This is the difference from html.

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: Formatting

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: more reasons to not have it now :)

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: I guess would be better to use a template string then and use interpolation? not sure whats recommended here

TrySound: Why better? jsx works fine. And anyway this part will be refactored once design is ready.

istarkov: Its a well known thing to add spaces like here, I would leave as is.

kof: idk, it looks really weird, a template literal would look so much better

kof (approved):

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
kof: I know, but we don't need that code, its optimizing nothing and its like 10-15LOC that does nothing but noise

apps/designer/app/designer/features/style-panel/shared/property-name.tsx
TrySound: Template literal always looks weirder for long strings because prettier doesn't have context
```
                  {`Resetting will change ${property} to cascaded ${toValue(
                    value
                  )} from ${breakpoint?.label}`}

```

Issue #465, opened 2022-11-18
Option names and select value do not match for the Box tag property at the properties panel. 


istarkov:
Option names and select value do not match for the Box tag property at the properties panel. 
_uppercase first letter_

https://user-images.githubusercontent.com/5077042/202745991-0a61b978-1a47-4e1f-b2d9-05ee8d2f9bba.mov

istarkov:
We have same behaviour for style `Display` 
flex/etc
<img width="230" alt="image" src="https://user-images.githubusercontent.com/5077042/207275583-fd82aee2-a050-4834-82bf-6117808b4e15.png">

<img width="176" alt="image" src="https://user-images.githubusercontent.com/5077042/207275720-d6745b5b-362c-423c-926c-fe4526fbb9dc.png">

kof:
cc @taylornowotny

istarkov:
From discussion:

@taylornowotny wrote:

My suggestion is this:
- Show an intuitive name in sentence case for property labels. The name should be as close to CSS as reasonable, but may be a different name and will not have hyphens.
- Show the CSS names in tooltips for properties and values, with correct syntax
- Use CSS names for values as much as possible, in sentence case with or without hyphens 

Why use sentence case? Capitalized first letters will be consistent with the rest of the UI and consistent with other UIs. We are going along with the user's expectations based on their past experience with other similar programs. Users can look up documentation all the same whether the first letter is capitalized or not. If we use correct CSS syntax outside of the context of a code window it look pretty weird to me. I realize that's not very objective, but in this case I believe my perspective is representative of our core user, a designer with limited code experience.

kof:
@istarkov we discussed this with @taylornowotny, the most correct notation is "Inline Flex" both words capitalized

PR #716, opened 2022-12-28
Fix css input (scrub, etc) Beautify options and input for css-value-input

istarkov:
## Description

ref #465 for css-value-input

Add beauty names to keywords, a little bit better search for keywords.

+ Fixes non-working scrub.
+ Fixes jumping cursor.

## Steps for reproduction

1.
Write `min-content` inside `width` field  - All options you see must have `Pascal Case`
Select value through menu - inputValue must be also in `Pascal Case`

2.
Edit width input, write `Min Content`, and click Enter - must be saved with `Min Content` value. 
(internally must be keyword with value=`min-content`) 

3.
Edit width input, write `min-content`, and click Enter - must be saved with `Min Content` value. 
(internally must be keyword with value=`min-content`) 

4. 
Write `min content`, click tab - must be saved with `Min Content` value. 
(internally must be keyword with value=`min-content`)

5. Edit input in any place of cursor it must not jump to the end.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -484,7 +484,7 @@ export const CssValueInput = ({
             }}
             onBlur={handleOnBlur}
             onKeyDown={handleKeyDown}
-            baseRef={scrubRef}
+            containerRef={scrubRef}
             inputRef={inputRef}
             name={property}
             state={value.type === "invalid" ? "invalid" : undefined}
--- a/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
+++ b/apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
@@ -32,6 +32,7 @@ import { parseIntermediateOrInvalidValue } from "./parse-intermediate-or-invalid
 import { toValue } from "@webstudio-is/css-engine";
 import { useDebouncedCallback } from "use-debounce";
 import type { StyleSource } from "../style-info";
+import { toPascalCaseNoDashes } from "./keyword-utils";
 
 // We increment by 10 when shift is pressed, by 0.1 when alt/option is pressed and by 1 by default.
 const calcNumberChange = (
@@ -228,20 +229,6 @@ const initialValue: IntermediateStyleValue = {
   value: "",
 };
 
-const keywordToLabel = (keyword: KeywordValue) => {
-  const { value } = keyword;
-
-  const label = value
-    .replace(/-/g, " ")
-    .split(" ")
-    .map((word) => {
-      return word[0].toUpperCase() + word.slice(1);
-    })
-    .join(" ");
-
-  return label;
-};
-
 /**
  * Common:
  * - Free text editing
@@ -317,7 +304,7 @@ export const CssValueInput = ({
     item === null
       ? ""
       : item.type === "keyword"
-      ? keywordToLabel(item)
+      ? toPascalCaseNoDashes(item.value)
       : item.type === "intermediate" || item.type === "unit"
       ? String(item.value)
       : toValue(item);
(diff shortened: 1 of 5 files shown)


apps/designer/app/designer/features/style-panel/shared/css-value-input/keyword-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+      return word[0].toUpperCase() + word.slice(1);
+    })
+    .join(" ");
+
+  return label;
+};
+
+/**
+ * Try to transform something from Pascal Case / Pascal Case / camelCase / camel Case to kebab-case
+ */
```
istarkov, line 19: In the future we can make search over all keywords we know, would be better as things like `mincontent` `alignleft` would work. (Now not)

apps/designer/app/designer/features/style-panel/shared/css-value-input/keyword-utils.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Beautify a CSS keyword by capitalizing the first letter of each word,
+ * and replacing dashes with spaces.
+ */
+export const toPascalCaseNoDashes = (keyword: string) => {
```
kof, line 5:
```suggestion
export const toPascalCaseWithoutDashes = (keyword: string) => {
```

kof: but I would just use toPascalCase, since its a specialized utility for keywords

kof: additionally we could have a cache, because its a limited set of keywords, it will be all cached at one point

packages/design-system/src/components/combobox.tsx
```diff
@@ -92,7 +94,7 @@ const defaultMatch = <Item,>(
   itemToString: (item: Item | null) => string
 ) =>
   matchSorter(items, search, {
-    keys: [(item) => itemToString(item)],
+    keys: [itemToString, (item) => itemToString(item).replace(/\s/g, "-")],
```
kof, line 97: wrong place to do this, combobox isn't aware of CSS specific stuff

packages/design-system/src/components/combobox.tsx
```diff
@@ -198,7 +200,6 @@ export const useCombobox = <Item,>({
     onInputValueChange({ inputValue, type }) {
       if (type === comboboxStateChangeTypes.InputChange) {
         filter(inputValue);
-        onInputChange?.(inputValue);
```
kof, line 201: I bet this will break something

kof: ah you moved it

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -466,7 +471,7 @@ export const CssValueInput = ({
             }}
             onBlur={handleOnBlur}
             onKeyDown={handleKeyDown}
-            baseRef={scrubRef}
+            containerRef={scrubRef}
```
kof, line 474: can you please make it first throw error from typescript?

kof (approved):
works!

apps/designer/app/designer/features/style-panel/shared/css-value-input/css-value-input.tsx
istarkov: The issue with getBlablaProps in downshift that each getter can have input which is going into out.
And they decided to use `any` type for getInputProps() Return type. Removal of spread in case of downshift could cause more serious issues and harder to find bugs vs we already had. 
That needs to reimplement all downshift types first. 
IMO can be done but not in this PR

PR #723, opened 2022-12-30
Keyboard support in spacing control

### LINKED ISSUE #666: Spacing keyboard interactions
See https://github.com/webstudio-is/webstudio-designer/issues/421#issuecomment-1312579096 and discussion after.

Summary:

- When moving focus with Tab entire spacing control acts as a single focus target (focus state design: https://github.com/webstudio-is/webstudio-designer/issues/421#issuecomment-1353622224 )
- When spacing control is focused, you can activate/focus individual items inside using arrow keys (see image in: https://github.com/webstudio-is/webstudio-designer/issues/421#issuecomment-1354327860) 
- When moving focus inside entire control keeps the visual focus state
- `Enter` opens input popover
- When popover open, `esc` and `enter` closes it, and returns focus to the last item, so one can hit `enter` again

rpominov:
## Description

Adds keyboard support to the Spacing control. Closes #666

## Steps for reproduction

1. Focus spacing control using TAB
2. Blue outline should appear and "Margin Top" should get highlighted
3. Arrow keys should change highlighted item
4. `Enter` should open popover for the highlighted item
6. Mouse navigation should still work when control is focused

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/sections/spacing/keyboard.ts
+++ b/apps/designer/app/designer/features/style-panel/sections/spacing/keyboard.ts
@@ -1,0 +1,102 @@
+import { useState } from "react";
+import type { FocusEvent, KeyboardEvent, MouseEvent } from "react";
+import type { SpacingStyleProperty } from "./types";
+
+const movementKeys = [
+  "ArrowUp",
+  "ArrowRight",
+  "ArrowDown",
+  "ArrowLeft",
+] as const;
+
+const movementMap = {
+  marginTop: ["marginBottom", "marginRight", "paddingTop", "marginLeft"],
+  marginRight: ["marginTop", "marginLeft", "marginBottom", "paddingRight"],
+  marginBottom: ["paddingBottom", "marginRight", "marginTop", "marginLeft"],
+  marginLeft: ["marginTop", "paddingLeft", "marginBottom", "marginRight"],
+  paddingTop: ["marginTop", "paddingRight", "paddingBottom", "paddingLeft"],
+  paddingRight: ["paddingTop", "marginRight", "paddingBottom", "paddingBottom"],
+  paddingBottom: ["paddingTop", "paddingRight", "marginBottom", "paddingLeft"],
+  paddingLeft: ["paddingTop", "paddingTop", "paddingBottom", "marginLeft"],
+} as const;
+
+export const useKeyboardNavigation = ({
+  onOpen,
+}: {
+  onOpen: (property: SpacingStyleProperty) => void;
+}) => {
+  const [activeProperty, setActiveProperty] =
+    useState<SpacingStyleProperty>("marginTop");
+
+  const [isActive, setIsActive] = useState(false);
+
+  const hadnleFocus = (event: FocusEvent<HTMLElement>) => {
+    if (event.currentTarget.matches(":focus-visible")) {
+      setIsActive(true);
+    }
+  };
+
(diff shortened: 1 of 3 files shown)


apps/designer/app/designer/features/style-panel/sections/spacing/keyboard.ts
```diff
@@ -0,0 +1,102 @@
...
+  marginRight: ["marginTop", "marginLeft", "marginBottom", "paddingRight"],
+  marginBottom: ["paddingBottom", "marginRight", "marginTop", "marginLeft"],
+  marginLeft: ["marginTop", "paddingLeft", "marginBottom", "marginRight"],
+  paddingTop: ["marginTop", "paddingRight", "paddingBottom", "paddingLeft"],
+  paddingRight: ["paddingTop", "marginRight", "paddingBottom", "paddingBottom"],
+  paddingBottom: ["paddingTop", "paddingRight", "marginBottom", "paddingLeft"],
+  paddingLeft: ["paddingTop", "paddingTop", "paddingBottom", "marginLeft"],
+} as const;
+
+export const useKeyboardNavigation = ({
```
kof, line 23: do you see any way to reuse the arrow keys logic with other components? e.g. flex control

rpominov: Not sure I understand. Do we need the same logic there? If so, I would copy paste at first, make all necessary changes there, and then would consider merging two similar components into one reusable.

kof: The same thing in all complex components is
1. tab to focus the most parent element
2. use arrow keys to navigate internal components

This is gonna be the same for all complex components

rpominov: Still would think about how to extract after we have at least two instances

kof: Problem is that its unlikely anybody will touch this logic if its not designed to be reusable, because they would have to test both components and will be afraid to break them, so we will unlikely reuse the logic if we don't make it reusable right away.

On the other hand, its difficult to make a good abstraction without actually reusing it, so maybe we should just do it with flex now so that this can be refactored into a reusable piece as long as its still fresh

kof (approved):
good work, posted a little demo https://twitter.com/oleg008/status/1610298974010839040

apps/designer/app/designer/features/style-panel/sections/spacing/keyboard.ts
rpominov: > Problem is that its unlikely anybody will touch this logic if its not designed to be reusable, because they would have to test both components and will be afraid to break them, so we will unlikely reuse the logic if we don't make it reusable right away.

That's solvable by either putting the same person (me) on implementing this in other components. Or by making a note in the issue about implementing this in other component.

> On the other hand, its difficult to make a good abstraction without actually reusing it

Yeah, that's basically my concern. You end up adding unnecessary complexity by making some parts unnecessarely flexible, at the same time you don't make other parts flexible enough.

Also, sometimes two simple components are easier to maintain than one reusable.

> so maybe we should just do it with flex now so that this can be refactored into a reusable piece as long as its still fresh

I'll create an issue

PR #724, opened 2022-12-30
Store preset style separately from tree

TrySound:
Before we populated preset style on instance creation. This does not let us to distinct preset style from local one.

As an option we could use preset style hardcoded in meta though this may affect user when we change defaults.

Here I added presetStyle column in Tree table with the list of styles for component.

```
Array<{
  component: string,
  property: string,
  value: StyleValue
}>
```

Will style controls as grey for such styles.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/canvas/shared/styles.ts
```diff
@@ -97,11 +94,28 @@ export const GlobalStyles = ({ assets }: { assets: Array<Asset> }) => {
...
 
+  const presetStyle = usePresetStyle();
+
+  useIsomorphicLayoutEffect(() => {
+    presetStyleEngine.clear();
+    const rules = new Map<string, StyleRule>();
+    for (const { component, property, value } of presetStyle) {
+      let rule = rules.get(component);
+      if (rule === undefined) {
+        rule = presetStyleEngine.addStyleRule(
```
kof, line 108: as an alternative to adding style rule for preset, we could be merging styles with the local styles, do you think this approach is better?

TrySound: It adds more work on every style update. Here it is done only once. Do you see any downside with using data-ws-component selector?

kof: only that I was hoping to delete all data attributes except of data-ws-id, but if this use case is super important we could do it, I am going to read more and test soon

kof:
Do we need to migrate existing projects?

```
Invalid `prisma.tree.findUnique()` invocation:
The column `Tree.presetStyle` does not exist in the current database.
```

kof:
Same error when trying to create a new project

```
Invalid `prisma.tree.create()` invocation:
The column `presetStyle` does not exist in the current database.
```

TrySound:
Migration not run in PR

kof:
> Migration not run in PR

Right, otherwise we would break all other previews, we really need branch dbs

apps/designer/app/canvas/shared/styles.ts
```diff
@@ -97,11 +94,28 @@ export const GlobalStyles = ({ assets }: { assets: Array<Asset> }) => {
...
-      tokensCssEngine.render();
-    }
+    tokensCssEngine.render();
   }, [tokens]);
 
+  const presetStyle = usePresetStyle();
+
+  useIsomorphicLayoutEffect(() => {
+    presetStyleEngine.clear();
+    const rules = new Map<string, StyleRule>();
```
kof, line 104: Why are we using a map here? it seems like its possible that we have multiple preset styles for the same component, but I don't see how this is conceptually possible

TrySound: Because single item of presetStyle is a declaration, not rule. Here I reuse rules to combine declarations into rules.

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,49 @@
...
+import { type ComponentName, getComponentMeta } from "../components";
+import { z } from "zod";
+
+export type PresetStyleItem = {
+  component: ComponentName;
+  property: StyleProperty;
+  value: StyleValue;
+};
+
+// @todo can't figure out how to make component and property to be enum
```
kof, line 12: @Andarist can you help here please?

TrySound: It's copy paste from instance type which has the same issue with component. Just didn't want to solve it here.

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,49 @@
...
+export const zPresetStyleItem = z.object({
+  component: z.string(),
+  property: z.string(),
+  value: SharedStyleValue,
+}) as z.ZodType<PresetStyleItem>;
+
+export const zPresetStyle = z.array(zPresetStyleItem);
```
kof, line 19: we have been using the convention for same-named type and zod schema, so before switching it, we need to discuss it and figure out if its worth it switching everywhere

TrySound: Typescript hover is broken and shows any when same name is used

packages/react-sdk/src/db/style.ts
Andarist: I can look into it in the next couple of days. Right now I’m on vacation

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type PresetStyle = z.infer<typeof zPresetStyle>;
+
+export const findMissingPresetStyle = (
```
kof, line 23: I was quite a bit confused about the purpose of this function, please correct me if I am wrong: this is collecting preset styles for all instances as they are being iterated over in a tree, so that only presets for the instances inside the tree are collected, correct?

TrySound: It find all possible components and by these components finds all preset styles.
This utility will probably go away when all tree data will be normalized.

packages/react-sdk/src/db/style.ts
kof: that's news to me, needs investigation, that wasn't a problem before

kof: tested, no problem there, its likely your ts was still parsing or something, try restarting ts in your editor

packages/react-sdk/src/db/style.ts
kof: `collectUsedPresetStyle` seems like a better name, works well with `collectUsedComponents`

kof: actually you are returning an array of presetStyle, so the name should be `collectUsedPresetStyles`

kof: please add a test for this function

apps/designer/app/canvas/shared/styles.ts
```diff
@@ -97,11 +94,28 @@ export const GlobalStyles = ({ assets }: { assets: Array<Asset> }) => {
       tokensCssEngine.addStyleRule(`:root`, { style });
     }
 
-    if (typeof document !== "undefined") {
-      tokensCssEngine.render();
-    }
```
kof, line 102: good catch, we don't actually need it since we moved it into that hook

packages/react-sdk/src/db/style.ts
TrySound: Reproduced this in the same module where both defined. Values are overriden by types. So when hover value I get any.

packages/react-sdk/src/db/style.ts
TrySound: > actually you are returning an array of presetStyle, so the name should be collectUsedPresetStyles

And array of used preset styles missing in passed list.

> please add a test for this function

Maybe not spend time if will be removed?

apps/designer/app/canvas/shared/styles.ts
kof: couldn't understand the logic from reading, had to log it, turned out presetStyle is a `presetStyles`, which is an array, semantics!

TrySound: So an array is "styles" but record used everywhere is "style"? What semantics?

packages/react-sdk/src/css/preset-style.ts
```diff
@@ -0,0 +1,52 @@
+import { useMemo, useState } from "react";
```
kof, line 1: I am wondering if all this presetStyle logic should be in SDK, seems like it should be in designer, wdyt?

packages/react-sdk/src/css/preset-style.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const presetStyleContainer = createValueContainer<PresetStyle>([]);
+
+export const useSetPresetStyle = (presetStyle: PresetStyle) => {
+  useState(() => {
+    presetStyleContainer.value = presetStyle;
+  });
+};
+
+export const usePresetStyle = () => {
```
kof, line 15:
```suggestion
export const usePresetStyles = () => {
```

kof: essentially everywhere where its not a style block like declaration for a single rule, please use `styles`

packages/react-sdk/src/css/preset-style.ts
```diff
@@ -0,0 +1,52 @@
...
+    for (const component of components) {
+      // skip if already stored in db
+      if (storedComponents.has(component)) {
+        continue;
+      }
+      const meta = getComponentMeta(component);
+      if (meta.presetStyle === undefined) {
+        continue;
+      }
+      for (const [property, value] of Object.entries(meta.presetStyle)) {
```
kof, line 39: not sure if it has a perf implication, but `for in` would not require creating entires arrays out of an object, at scale it would have like 2x perf implication at least, but not sure if it matters here

kof (changes requested):
Finished first pass, lets go with this preset rendering model

apps/designer/app/canvas/shared/styles.ts
kof: yes, foundationally `style` without s is like inline style property, it represents a single style block

packages/react-sdk/src/db/style.ts
kof: if its going away for sure then, yeah, unless you can write a higher level test that will stay and can confirm unchanged behavior once you refactor

packages/react-sdk/src/css/preset-style.ts
TrySound: It's the same as props. The state is used for rendering in canvas and published site.

kof: published site is using a generated static stylesheet, btw did you check if these styles land there?
canvas - can go to canvas side of the things inside designer

TrySound: Gotcha, added utility for generated css. I guess it can be moved indeed.

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,49 @@
+import { z } from "zod";
+import type { StyleProperty, StyleValue } from "@webstudio-is/css-data";
+import { SharedStyleValue } from "@webstudio-is/css-data";
+import { type ComponentName, getComponentMeta } from "../components";
```
Andarist, line 4:
```suggestion
import { type ComponentName, getComponentMeta, getComponentNames } from "../components";
```

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type PresetStylesItem = {
+  component: ComponentName;
+  property: StyleProperty;
+  value: StyleValue;
+};
+
+// @todo can't figure out how to make component and property to be enum
+export const zPresetStylesItem = z.object({
+  component: z.string(),
```
Andarist, line 14: Zod requires tuples with at least one element as an argument to `z.enum`:
```suggestion
  component: z.enum(getComponentNames() as [ComponentName]),
```

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,49 @@
...
+  property: z.string(),
```
Andarist, line 15: I'm not sure if you can make this an actual enum. Your `StyleProperty` includes a string index (`[custom: CustomProperty]: { /*...*/}`).

But perhaps smth like this would work:
```ts
z.union([z.enum(properties as [keyof properties]), z.string.regex(/^--/)]);
```

Note that you would have to access those [`properties`](https://github.com/webstudio-is/webstudio-designer/blob/c69c715bb7f16d762bc148702182476b2152e5d5/packages/css-data/src/schema.ts#L2) and plumb them correctly with this call site

TrySound: @kof Why do we have it in types? Shouldn't it be only for css engine?

packages/react-sdk/src/db/style.ts
TrySound: cool, thanks
Will fix in separate PR

packages/react-sdk/src/db/style.ts
kof: @TrySound not sure what you mean, please explain the full context

apps/designer/app/canvas/shared/styles.ts
```diff
@@ -69,23 +74,20 @@ export const useManageDesignModeStyles = () => {
 const helpersCssEngine = createCssEngine();
 const fontsAndDefaultsCssEngine = createCssEngine();
 const tokensCssEngine = createCssEngine();
+const presetStylesEngine = createCssEngine();
```
kof, line 77: please add a name after rebasing

apps/designer/app/canvas/canvas.tsx
```diff
@@ -174,6 +175,11 @@ export const Canvas = ({ data }: CanvasProps): JSX.Element | null => {
   useInitializeDesignTokens(data.designTokens);
   const assets = useAssets(data.assets);
   useAllUserProps(data.props);
+  useState(() => {
+    if (data.tree?.presetStyles) {
+      presetStylesContainer.value = data.tree.presetStyles;
+    }
+  });
```
kof, line 182: this should be probably a custom hook with a semantic name, that goes along with other things for presetStyles on canvas

TrySound: Or we could use single useState to populate data in atoms

kof: yeah, though keeping concerns in the specific modules makes it more maintainable long-term, downside - you have to jump into that file when reading

I think its generally a downside with organizing things well, the better it is organized, the more you have to jump, its a tradeoff for sure, but I don't think we can scale without organizing things semantically.

Inline stuff like this will grow in 3-6 month into a monster.

packages/react-sdk/src/db/style.ts
```diff
@@ -0,0 +1,53 @@
...
+  getComponentNames,
+} from "../components";
+
+export type PresetStylesItem = {
+  component: ComponentName;
+  property: StyleProperty;
+  value: StyleValue;
+};
+
+// @todo can't figure out how to make property to be enum
```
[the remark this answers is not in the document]
TrySound: No, we did not fix property enum because of additional complexity with css vars syntax

Issue #366, opened 2022-10-31
Units in style panel


kof:
Some properties have limitations on what units they accept:  e.g. some accept unitless number (e.g. line-height) some accept fr (e.g. grid-auto-columns)

## Todo

1. Make a list and see if it makes sense to just maintain specific differences for specific properties and otherwise use a default units list
2. Enable typing a unit even if its not in the list (since we don't show a full list of units anyways, but sometimes when you know what you need, you should still be able to use it by typing it)
3. Make sure we don't allow setting unsupported unit values for properties that can't use them. E.g. currently you can set overflow: 10px and there is no error outline, maybe runtime validation logic can detect this too

PR #726, opened 2023-01-03
Parse properties initial value

TrySound:

Parsed initial values to get correct unit instead of px for some properties. There is still edge case with unitless values I'm gonna fix in another PR.

Also bumped mdn-data and removed overrides fixed on mdn side.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-data/bin/mdn-data.ts
```diff
@@ -57,12 +37,62 @@ const normalizedValues = {
...
     unit: "px",
   },
-  // https://github.com/mdn/data/issues/556
-  // @todo remove once fixed
   "background-size": autoValue,
   "text-size-adjust": autoValue,
 } as const;
 
+const parseInitialValue = (property: string, value: string): StyleValue => {
+  // Our default values hardcoded because no single standard
```
kof, line 45: any links to learn from reg discrepancies in the standard?

packages/css-data/bin/mdn-data.ts
```diff
@@ -57,12 +37,62 @@ const normalizedValues = {
...
+  if (property in normalizedValues) {
+    return normalizedValues[property as keyof typeof normalizedValues];
```
kof, line 47: why do you need an assertion here? its a string by default and you already checked its in the object, I don't see a need to assert to the compiler its a specific property

packages/css-data/bin/mdn-data.ts
```diff
@@ -57,12 +37,62 @@ const normalizedValues = {
     value: 0,
     unit: "px",
   },
-  // https://github.com/mdn/data/issues/556
-  // @todo remove once fixed
   "background-size": autoValue,
   "text-size-adjust": autoValue,
 } as const;
 
+const parseInitialValue = (property: string, value: string): StyleValue => {
```
kof, line 44: this script starts to become more and more complex, I think its time to take out functions and add tests for them

packages/css-data/src/__generated__/properties.ts
```diff
@@ -3036,7 +3019,7 @@ export const properties = {
     inherited: false,
     initial: {
       type: "unit",
-      unit: "px",
+      unit: "s",
```
kof, line 3022: something is wrong here, it is supposed to be a number or %

kof: sorry, got confused by the diff

packages/react-sdk/src/css/categories.ts
```diff
@@ -10,7 +10,6 @@ const layout = [
   "justifyContent",
   "alignContent",
   // Grid
-  "placeContent",
```
kof, line 13: why removing it?

kof (approved):
Please move the complex functions to separate utils and add tests, since things get more and more complex

packages/css-data/bin/mdn-data.ts
TrySound: It's not my comment. Probably yours. Just moved the code into function.

packages/css-data/bin/mdn-data.ts
TrySound: 🤷

packages/css-data/bin/mdn-data.ts
TrySound: __generated__ here is the result of this script. Kinda snapshot test already.

TrySound:
Don't worry, I will remove half of this code eventually.

packages/css-data/bin/mdn-data.ts
kof: damn, I wish I had cr back then, now I have no idea what I was looking at

packages/css-data/bin/mdn-data.ts
kof: its a very complex snapshot, once it changes, you won't have any clue why :) and its easy to overlook

packages/react-sdk/src/css/categories.ts
TrySound: Yes, because property was removed the generated list of properties and types no longer allow it in categories.

packages/css-data/bin/mdn-data.ts
TrySound: Well, you can see from issues. Some properties initials were different from browsers. Some just didn't work for input because of multiple values.

PR #728, opened 2023-01-03
Use tokens from Figma in Stitches

rpominov:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/package.json
```diff
@@ -6,8 +6,8 @@
   "homepage": "https://webstudio.is",
   "type": "module",
   "scripts": {
-    "dev": "build-package --watch",
-    "build": "build-package",
+    "dev": "tsx ./src/figma-tokens/transform.ts && build-package --watch",
```
kof, line 9: scripts should go into bin folder like we did everywhere else

packages/design-system/src/figma-tokens/README.md
```diff
@@ -0,0 +1,23 @@
+# Figma Design Tokens
```
kof, line 1: I am not sure if keeping the file here is better than having a docs folder in the root, for discoverability purposes

kof (approved):
Looking great!
A few things regarding the organization of files can be improved.
Approving in advance

packages/design-system/src/figma-tokens/README.md
rpominov: Moved to `packages/design-system/documentation/figma-design-tokens.md`

packages/design-system/bin/transform-figma-tokens.ts
```diff
@@ -0,0 +1,108 @@
...
+  return text;
+};
+
+const deCapitalize = (text: string) => text[0].toLowerCase() + text.slice(1);
+
+const pathToName = (path: string[], type: string) =>
+  deCapitalize(trimPrefix(type, toCamelCase(path.join("-"))));
+
+const main = () => {
+  execSync(`token-transformer ${SOURCE_FILE} ${TMP_OUTPUT_FILE}`, {
```
TrySound, line 60: You can try to use api directly. The code may be a little simpler with it
https://github.com/tokens-studio/figma-plugin/blob/main/token-transformer/index.js#L7

packages/design-system/bin/transform-figma-tokens.ts
```diff
@@ -0,0 +1,108 @@
...
+const toCamelCase = (text: string) =>
+  text
+    .split(/[^a-z0-9]+/i)
+    .map((word) => {
+      if (word.length === 0) {
+        return "";
+      }
+      return word[0].toUpperCase() + word.slice(1);
+    })
+    .join("");
```
TrySound, line 45: We already use camelcase package in css-data

packages/design-system/bin/transform-figma-tokens.ts
rpominov: Yeah, but this is not official public API, so may stop working at any time

PR #731, opened 2023-01-03
Add a stylesheet name for easier debugging

kof:
## Description

Multiple style elements are easier to differentiate when they have some hint on the purpose

## Steps for reproduction

1. select style element on canvas, see data-webstudio attribute with a name

## Code Review

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


packages/css-engine/src/core/style-element.ts
```diff
@@ -1,12 +1,16 @@
...
+    this.#name = name;
+  }
   get isMounted() {
     return this.#element?.parentElement != null;
   }
   mount() {
     if (this.isMounted === false) {
       this.#element = document.createElement("style");
-      this.#element.setAttribute("data-webstudio", "");
+      this.#element.setAttribute("data-webstudio", this.#name);
```
TrySound, line 13: Why not something like data-ws-style similar to data-ws-id and data-ws-component?

kof: I remember we had a conversation that "ws" isn't meaningful and confusing with websockets so I called it webstudio and thought we should use webstudio full name everywhere 🤷‍♂️ but didn't make it consistent, I guess since this question is now here, we should make a consistent naming

TrySound (approved):

PR #734, opened 2023-01-03
Fix unitless initial values

TrySound:

There were many unitless initial values with "px" unit. Here I extracted all possible unit types from properties syntaxes. csstree provides a list of unit groups which in the future will let us provide correct list of units for css value input.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
nice

PR #735, opened 2023-01-04
Improve units in style panel

TrySound:

- generate grouped units for non length values
- show unit select when initial unitless

So we get all 3 points from the issue

- all properties has only supported units
- hidden but supported units still can be entered
- unsupported values are validated

Screenshot 2023-01-04 at 14 12 29
Screenshot 2023-01-04 at 14 14 54

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/designer/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
```diff
@@ -9,6 +9,8 @@ import { evaluateMath } from "./evaluate-math";
 import { units } from "@webstudio-is/css-data";
 import { toKebabCase } from "../keyword-utils";
 
+const unitsList = Object.values(units).flat();
```
kof, line 12: its not already flat?

packages/css-data/src/schema.ts
```diff
@@ -17,13 +17,22 @@ type CustomProperty = `--${string}`;
 
 export type AppliesTo = Properties[StyleProperty]["appliesTo"];
 
-const Unit = z.union([z.enum(units), z.literal("number")]);
+export type UnitGroup = keyof typeof units;
 
-export type Unit = z.infer<typeof Unit>;
+type UnitEnum = typeof units[UnitGroup][number];
+
+const zUnit = z.union([
```
kof, line 24: convention

kof (changes requested):
Good job, a few things commented

packages/css-data/src/schema.ts
TrySound: Undefined behavior :)

kof: yes, we discussed it, we need an expert opinion on this @Andarist it turns out typescript/or tooltip in the vscode UI shows any as a type when we use that "shadowing" feature from typescript where schema has the same name as a type.

1. what is the reason it shows us any there, is it a bug or by design? 
2. what can we do about it to avoid confusion?

kof: @TrySound until we figured it out please use a convention we use everywhere

kof (approved):

PR #747, opened 2023-01-05
Auth Zanzibar subset

istarkov:
## Description

Authorization.
A subset of Zanzibar paper https://research.google/pubs/pub48190/ 

The idea that this system can be changed on 3rd party implementation like https://www.osohq.com/ or https://www.ory.sh/docs/keto

Short idea:

A simple ACL (Access Control Lists) takes the form of “user U has relation R to object O”. 
More complex ACLs take the form of “set of users S has relation R to object O”,
where S is itself specified in terms of another object-relation pair. 
ACLs can thus refer to other ACLs:
Example:
All members of GroupX can read Project Y.

Can be implemented as ACL list of a Groups
- Alice member of GroupA
- Bob member of GroupA

And ACL list with reference:
- All members of GroupA have read access to Project Y

How it works.

Imagine you have relations like
- Alice member of GroupA
- Bob member of GroupA
- Alice is reader of ProjectA

We can think about of these relations like 
`Subject has Relation on Object`

Now we can write the rules above like table

| Object | Relation |Subject|
| --- | --- |---|
| GroupA | member | Alice |
| GroupA | member  | Bob |
| ProjectA | reader  | Alice |

For any fixed Object, Relation we have `SubjectSet`. 

example:
`GroupA#member = SubjectSet(Object = GroupA  relation = member)  =  [Alice, Bob]`

If we allow Subject to be SubjectSet we can say that writers of ProjectA are SubjectSet of Group members.

| Object | Relation |Subject|
| --- | --- |---|
| GroupA | member | Alice |
| GroupA | member  | Bob |
| ProjectA | reader  | Alice |
| ProjectA | writer  | GroupA#member |

What can be expanded to 

| Object | Relation |Subject|Source
| --- | --- |---|---|
| GroupA | member | Alice | direct |
| GroupA | member  | Bob | direct |
| ProjectA | reader  | Alice | direct |
| ProjectA | writer  | Alice |  GroupA#member |
| ProjectA | writer  | Bob |  GroupA#member |

Now we know that each writer has reader permissions. Instead of creating a new row Google suggests a rewrite system. For each row with a 'writer' relation, we will create 2 rows with a `writer` and `reader` relation. Example.

| Object | Relation |Subject|Source
| --- | --- |---|---|
| GroupA | member | Alice | direct |
| GroupA | member  | Bob | direct |
| ProjectA | reader  | Alice | direct |
| ProjectA | writer  | Alice |  GroupA#member |
| ProjectA | reader  | Alice |  GroupA#member |
| ProjectA | writer  | Bob |  GroupA#member |
| ProjectA | reader  | Bob |  GroupA#member |

To make the system less error-prone we add Namespace. Each namespace defines what relations are possible inside, and defined rewrites rules.
Example: 
- Group namespace defines only the `member` relation.
- Project namespace defines `reader`, `writer`, `owner` relations and rewrite rules `owner` -> [`owner`, `writer`, `reader`], `writer` -> [`writer`, `reader`]

This system is used at google and is called Zanzibar.

## Our implementation.

We implement subset of Zanzibar.

- The depth of the subset can expand is equal to 1.
- In our system subsets relation is always defined (no `any` relation).
- no `zookie`

## What our implementation solves.

For example, Alice wants to share her project `ProjectX` with another user who probably does not exist in our system.  She can use his email `bob@bob.com` for sharing. (email is some kind of global uid)

Let's implement this.

Initialy we have.

|Namespace| Object | Relation |Subject
| --- | --- |---|---|
| Project | ProjectXID | owner | AliceID

Now we want to add `bob@bob.com` permissions to Project. So later user with `bob@bob.com` would have `write` access to ProjectX. If user `bob@bob.com` exists in our system all is simple. Imagine it's not.

We add write permission to Project to any object which has `email` relation to `bob@bob.com`

|Namespace| Object | Relation |Subject
| --- | --- |---|---|
| Project | ProjectXID | owner | AliceID | 
| Project | ProjectXID | writer | User:bob@bob.com#email| 

And if Bob will register the table will be.

|Namespace| Object | Relation |Subject
| --- | --- |---|---|
| Project | ProjectXID | owner | AliceID |
| Project | ProjectXID | writer | User:bob@bob.com#email|
| User | bob@bob.com | email |  BobID |

And after expanding we will have.
|Namespace| Object | Relation |Subject
| --- | --- |---|---|
| Project | ProjectXID | owner | AliceID |
| Project | ProjectXID | writer | BobID | 
| User | bob@bob.com | email |  BobID |

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/authorize/package.json
+++ b/packages/authorize/package.json
@@ -1,0 +1,47 @@
+{
+  "name": "@webstudio-is/authorize",
+  "version": "0.19.0",
+  "description": "Webstudio Authorize API Interface",
+  "author": "Webstudio <github@webstudio.is>",
+  "homepage": "https://webstudio.is",
+  "type": "module",
+  "scripts": {
+    "typecheck": "tsc --noEmit",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
+    "checks": "pnpm typecheck && pnpm lint && pnpm test",
+    "dev": "build-package --watch",
+    "build": "build-package",
+    "lint": "eslint ./src --ext .ts,.tsx --max-warnings 0"
+  },
+  "dependencies": {
+    "@webstudio-is/prisma-client": "workspace:^",
+    "uuid": "^9.0.0",
+    "warn-once": "^0.1.1",
+    "zod": "^3.19.1"
+  },
+  "devDependencies": {
+    "@jest/globals": "^29.3.1",
+    "@types/uuid": "^8.3.4",
+    "@webstudio-is/jest-config": "workspace:^",
+    "@webstudio-is/scripts": "workspace:^",
+    "@webstudio-is/tsconfig": "workspace:^",
+    "jest": "^29.3.1",
+    "typescript": "4.7.4"
+  },
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib/cjs/index.cjs"
+    },
+    "./server": "./server.js"
+  },
+  "types": "src/index.ts",
+  "files": [
+    "lib/*",
+    "src/*",
+    "!*.test.*"
+  ],
+  "license": "MIT",
+  "private": false,
+  "sideEffects": false
+}
--- a/packages/authorize/package.json
+++ b/packages/authorize/package.json
@@ -7,7 +7,7 @@
   "type": "module",
   "scripts": {
     "typecheck": "tsc --noEmit",
-    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
     "checks": "pnpm typecheck && pnpm lint && pnpm test",
     "dev": "build-package --watch",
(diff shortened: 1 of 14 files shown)


istarkov:
# Implementation.
The system implemented as a single table, and 2 views. One view is for "expand" and the other for "rewrite" rules.

To check does UserX has `read` access to Project Y we need the following query

```sql
select * from "AuthApplyRewrites" where object=E'Project Alice' AND relation = 'reader' and subject='Bob';
```

To find all users with reader relation to Project.
```sql
select * from "AuthApplyRewrites" where object=E'Project Alice' AND relation = 'reader';
```
image

istarkov:
Api is like https://www.ory.sh/docs/keto/reference/rest-api
Typed for our needs. Probably would force to use User, Project  objects instead of ID for write API. To force id to be consistent.

Issue #696, opened 2022-12-25
Normalize tree data


TrySound:
Currently we store denormalized tree data in db as json text. This makes postgres useless when dealing with json.

To solve this we can normalize all data and store as array with flat (more or less) objects.

To simplify updates and querying we can colocate all data in single table while split into separate columns.

JSON data stores as text may bloat db very fast. Postgres jsonb is more optimal binary representation. It also supports indexing, which can be a significant advantage.

Postgres json has a lot of builtin utilities

https://www.postgresql.org/docs/15/functions-json.html

Prisma supports json as well

https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields

I propose the following Tree table structure

id: string

instances: jsonb
Array<{ type, id, component, children }>

styles: jsonb
Array<{ breakpointId, instanceId, type, value }>

props: jsonb
Array<{ instanceId, type, name, value, required }>

kof:
1. We need to figure out how to use jsonp without loosing multi-database abstraction from prisma (if that's even possible). We haven't decided yet if we need/want sqllite and prisma doesn't have jsonp for sqllite, so we would have to write raw queries for it

2. We need to fully understand how we are going to implement conflict resolution with offline sync, because depending on that we may need a different solution or even a separate database for these things. It seems that we are going to need an entire tree when we apply patches, so we would either have to do crdt on a database level or store the entire json somewhere in memory or otherwise have some kind of low-latency key-value storage that would quickly give us the full tree.

TrySound:
jsonB is binary json format to store in sql. Nothing more. It has support for direct updates without parsing json in js runtime. So it's not really a transport protocol, just builtin json manipulation on db side.

TrySound:
And looks like sqlite supports json as well though with a little different api
https://www.sqlite.org/json1.html

kof:
@TrySound both of your statements are correct but none has addressed the problems I mentioned

TrySound:
> We need to figure out how to use jsonp without loosing multi-database abstraction from prisma

If we normalize whole data, updating will be the least problem, in worst case we can parse json on server as now

>  It seems that we are going to need an entire tree when we apply patches

It's easier to change normalized tree

PR #749, opened 2023-01-06
Normalize styles in database

TrySound:

Here only migration and transforms for load and patch. Styles have the following type and stored in Tree.styles column.

```
Array<{
  breakpointId: string,
  instanceId: string,
  property: StyleProperty,
  value: StyleValue
}>
```

Did migration with cursor pagination with 1000 trees per transaction.

## Code Review

  - test it locally

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (child.type === "instance") {
+      convertTree(child, styles);
+    }
+  }
+};
+
+export default async () => {
+  const client = new PrismaClient({
+    // Uncomment to see the queries in console as the migration runs
+    // log: ["query", "info", "warn", "error"],
```
rpominov, line 55: BTW, another potential reason for `JSON.stringify` error is if you've run with this line uncommented

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+    for (const [property, value] of Object.entries(rule.style)) {
+      styles.push({
+        breakpointId: rule.breakpoint,
+        instanceId: instance.id,
+        property,
+        value,
+      });
+    }
+  }
+  instance.cssRules = [];
```
rpominov, line 44: Do want to keep `cssRules` property on instance still? Will it always be an empty array now?

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        for (const tree of trees) {
+          try {
+            const root = JSON.parse(tree.root);
+            const styles: Styles = [];
+            convertTree(root, styles);
+            tree.root = JSON.stringify(root);
+            tree.styles = JSON.stringify(styles);
+          } catch {
+            console.info(`Tree ${tree.id} cannot be converted`);
```
rpominov, line 88: Would be good to log the caught error as well, just in case

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+          }
+        }
+        await Promise.all(
+          trees.map(({ id, root, styles }) =>
+            prisma.tree.update({ where: { id }, data: { root, styles } })
```
rpominov, line 93: Not a strong opinion, but I'd run `tree.update()` in the loop above right after conversion if it's successful.

Because as is, we're updating trees which we've failed to convert so potentially corrupting them even further. Although the only way this can happen is if `JSON.stringify(styles)` throws, which shouldn't happen. So not a big deal.

rpominov (approved):
Minor coments

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
kof: I think this is here to satisfy typescript and keep potentially not migrated cssRules functional while we can keep running the migration for any amount of time .... basically this is how we would have to do it once we have production sites from customers

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+            id: "asc",
+          },
+          ...(cursor
+            ? {
+                skip: 1, // Skip the cursor
+                cursor: { id: cursor },
+              }
+            : null),
+        });
+        cursor = trees.at(-1)?.id;
```
kof, line 77: wooow, didn't know this method was added to array to explicitly allow negative indexes NICE!

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+        const chunkSize = 1000;
+        const trees = await prisma.tree.findMany({
+          take: chunkSize,
+          orderBy: {
+            id: "asc",
+          },
+          ...(cursor
+            ? {
+                skip: 1, // Skip the cursor
+                cursor: { id: cursor },
```
kof, line 73: very interesting!

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
```diff
@@ -0,0 +1,100 @@
...
+    // Uncomment to see the queries in console as the migration runs
+    // log: ["query", "info", "warn", "error"],
+  });
+
+  await client.$transaction(
+    async (prisma) => {
+      let cursor: undefined | string = undefined;
+      let hasNext = true;
+
+      while (hasNext) {
```
kof, line 63: maybe there is a way to generalize this type of chunking into a utility and make it dead-simple?

kof (approved):

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
TrySound: Yes, I will remove it when normalize tree

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
TrySound: If something will fail on parsing or conversion there will be old data here. So no issue.
Don't want it in the same for loop because may take ages to run sequential even though prisma doesn't send requests in parallel.

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
TrySound: Two primary reasons: negative access and type safety

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
TrySound: I'd say better be explicit to have a way to change query any time. Anyway, it would be copy pasted to keep migrations isolated.

packages/prisma-client/prisma/migrations/20230106000143_tree_styles_data/migration.ts
rpominov: Should be 0 difference between `Promise.all()` and loop from performance point of view.

In theory `Promise.all()` can be slower because Prisma has to manage an internal queue.

PR #750, opened 2023-01-06
Render static css from normalized styles

TrySound:

Gradually starting to migrate to new styles format. Here replaced getCssRules and getPresetStylesMap with getStyleRules and getPresetStyleRulese utilities.

Will reuse both at once on release.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/project/src/shared/styles/style-rules.test.ts
```diff
@@ -0,0 +1,109 @@
+import { test, expect } from "@jest/globals";
+import type { Styles } from "@webstudio-is/react-sdk";
+import { getStyleRules, getPresetStyleRules } from "./style-rules";
+
+test("get a list of style rules grouped by instance and breakpoint", () => {
```
kof, line 5: no describe? I thought we should always have a describe for output

kof (approved):

packages/project/src/shared/styles/style-rules.test.ts
TrySound: No, it's mocha requirement. jest allows to wrap with only "test"

kof: I mean technically its allowed, but I thought it would result in bad reporting, without proper encapsulation

TrySound: If file name is descriptive enough we may not need to wrap all tests.

```
style-rules.test.ts
  get a list of style rules grouped by instance and breakpoint
    getStyleRules
  something else
    getPresetRules
```

```
style-rules.test.ts
  get a list of style rules grouped by instance and breakpoint
  something else
```

PR #751, opened 2023-01-06
Update Button

### LINKED ISSUE #450: Clean and refactor <Button> to make it in sync with Figma
`<Button>` was originally copied from somewhere. It has variants that are not from our design system, and it doesn't have variants that we have in Figma.

- [x] Remove from `<Button>` everything that's not defined in Figma
- [x] Add variants from Figma
- [x] Add a `pending` variant (need it in the Pages UI) that:
   - adds "..." to the label (questionable as adding "..." is not the only change we wan't to make to the text, e.g. not "Create..." but "Creating...")
   - sets `cursor` to `wait`
   - sets `pointer-events: none`

rpominov:
## Description

New `Button` implementation based on the component in Figma. Closes #450

Component in Storybook: https://638245578b47d4399fba191a-cjnaykysmq.chromatic.com/?path=/story/library-button--button. The link may become outdated at some point, to get the latest one click "details" next to `Storybook Publish: @webstudio-is/design-system` at the bottom of this PR, and then find "Button" in the left menu:

Screenshot 2023-01-12 at 17 54 12

Note that you can control the "Configurable" demo using the controls at the bottom:

Screenshot 2023-01-12 at 18 01 51

## Code Review

  - test the component using the Storybook link above 

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


taylornowotny:
Awesome work! It looks perfect to me

taylornowotny (approved):

packages/design-system/src/components/button.stories.tsx
```diff
@@ -1,10 +1,94 @@
...
+      {variants.map((variant) => (
+        <ButtonComponent variant={variant} key={variant}>
+          {variant}
+        </ButtonComponent>
+      ))}
+    </Section>
+
+    <Section title="With Icon">
+      <ButtonComponent icon={<TrashIcon />}>Button</ButtonComponent>
+      <ButtonComponent icon={<TrashIcon />} iconPosition="right">
```
kof, line 50: do we need to call it icon? is there anything icon specific? or could it be just prefix/suffix?

packages/design-system/src/components/button.stories.tsx
```diff
@@ -1,10 +1,94 @@
...
+      </ButtonComponent>
+      <ButtonComponent icon={<TrashIcon />} />
+    </Section>
+
+    <Section title="Disabled">
+      <ButtonComponent disabled>Button</ButtonComponent>
+      <ButtonComponent disabled icon={<TrashIcon />}>
+        Button
+      </ButtonComponent>
+      <ButtonComponent disabled icon={<TrashIcon />} />
```
kof, line 61: can pending and disabled be at the same time? isn't it impossible state made possible?

kof (approved):
added some comments, approving in advance

packages/design-system/src/components/button.stories.tsx
rpominov: Nothing icon specific, could be prefix/suffix. 

Not completely sure it's a good idea. It encourages to use component in a way not intended by our design in Figma, which I'm also not sure whether it's a good or bad thing...

Will switch to `prefix/suffix` as I'm not sure either way and it's a bit more straightforward API than `icon + iconPosition`

packages/design-system/src/components/button.stories.tsx
rpominov: As it implemented currently `pending=true disabled=true` is equivalent to `pending=true disabled=false`, so although it's a weird state there's no problem.

Will keep as is, because an API that will not allow this state will be more confusing.

PR #754, opened 2023-01-07
Update styles without patches

TrySound:
Here introducing first updates without patches.
Basic idea is that we need only two operations "set" and "delete" which can do all necessary changes on normalized data.

Added new `/update` rest endpoint which accepts a list of messages and changes data.

Once of important points is type safety of such operations. We can run zod validation on server and not worry about unexpected data in json.

Client side is implemented similar to immerhin. Messages are queued in global array and sent to server every second.

Queue is temporary updated on canvas frame to reuse the same queue and sync state. When all is migrated the code can be just moved to designer.

Intentionally tried to add all logic in designer app caz not needed for published worker.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/shared/sync/use-sync.ts
+++ b/apps/designer/app/canvas/shared/sync/use-sync.ts
@@ -2,8 +2,9 @@ import { useInterval } from "react-use";
 import { sync } from "immerhin";
 import { enqueue } from "./queue";
 import type { Build } from "@webstudio-is/project";
-import { Tree } from "@webstudio-is/react-sdk";
-import { restPatchPath } from "~/shared/router-utils";
+import type { Tree } from "@webstudio-is/react-sdk";
+import { restPatchPath, restUpdatePath } from "~/shared/router-utils";
+import { syncMessages, useSubscribeMessages } from "~/shared/stores";
 
 export const useSync = ({
   treeId,
@@ -14,7 +15,8 @@ export const useSync = ({
 }) => {
   useInterval(() => {
     const entries = sync();
-    if (entries.length === 0) {
+    const messages = syncMessages();
+    if (entries.length === 0 && messages.length === 0) {
       return;
     }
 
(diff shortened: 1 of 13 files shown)


apps/designer/app/canvas/shared/sync/use-sync.ts
```diff
@@ -14,23 +15,34 @@ export const useSync = ({
 }) => {
   useInterval(() => {
     const entries = sync();
-    if (entries.length === 0) {
+    const messages = syncMessages();
```
istarkov, line 18: In case of `fetch` fail seems like messages used here will be lost.
Also using `useInterval` gives 0 order guarantee. I mean that 1 fetch can "patch" data later when 2nd.

IMO all such systems must not do any operations until the response from the previous has been received.

apps/designer/app/routes/rest/update.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  if (tree === null) {
+    return [];
+  }
+
+  return Styles.parse(JSON.parse(tree.styles));
+};
+
+export const updateStyles = async (messages: StylesMessage[]) => {
```
istarkov, line 18: Just curious about compute time of all of this. Zod.parse messages and styles, iterate over mesage and find style.

TrySound: What with compute time? I run zod parse only once right after request.json().

apps/designer/app/canvas/shared/sync/use-sync.ts
TrySound: Here I just mimic existing code. Agree not ideal. Though we need first figure out how to bypass current json limitations so we could do more atomic updates and get rid of this for good.

apps/designer/app/routes/rest/update.ts
TrySound: Ah, got it. You mean also parsing db data. This is done once too. And actual data is very simple comparing to denormalized tree. Should be faster.

PR #756, opened 2023-01-08
Move preview app to designer repo

TrySound:
We want to deploy preview with cloudflare pages to reduce amount of steps and automate deploy for all branches.

Here only moved code and integrated into monorepo.

## Code Review

- [ ] hi @kof and @istarkov, I need you to do

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/preview/app/env.ts
```diff
@@ -0,0 +1,6 @@
+import type { KVNamespace } from "@cloudflare/workers-types";
+
+export interface Env {
+  __STATIC_CONTENT: KVNamespace;
+  SITES: KVNamespace;
+}
```
kof, line 6: problem is we wanted to keep cloudflare specific logic completely outside of this repo

kof (changes requested):
one fundamental problem so far

apps/preview/app/env.ts
TrySound: It's just copy paste of whole workspace. It doesn't work yet.
Gonna integrate with pages in other PR.

TrySound: And reading from kv is not much of a logic. Though we gonna keep writing outside of it for sure.

kof: sure, but we need to get it to a state where webstudio designer doesn't depend on workers in any way

TrySound: cloudflare pages are workers. And anyway this is only an environment like node.js and this binding is an env variable. It can be developed locally too.

kof:
@TrySound @andrasbacsai does this PR still fit the current plan?

PR #757, opened 2023-01-08
Update normalized styles directly

TrySound:

Migrated updating and rendering to normalized styles. There are still a few cases to migrate so left legacy root instance updates.

Styles are updated with immerhin consistently with whole project.

Can be tested changing styles and deleting breakpoints.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/shared/breakpoints.ts
+++ b/apps/designer/app/canvas/shared/breakpoints.ts
@@ -6,6 +6,7 @@ import { deleteCssRulesByBreakpoint } from "~/shared/css-utils";
 import {
   breakpointsContainer,
   rootInstanceContainer,
+  stylesContainer,
   useBreakpoints,
 } from "~/shared/nano-states";
 import { publish } from "~/shared/pubsub";
@@ -52,8 +53,8 @@ const useBreakpointChange = () => {
 const useBreakpointDelete = () => {
   useSubscribe("breakpointDelete", (breakpoint) => {
     store.createTransaction(
-      [breakpointsContainer, rootInstanceContainer],
-      (breakpoints, rootInstance) => {
+      [breakpointsContainer, rootInstanceContainer, stylesContainer],
+      (breakpoints, rootInstance, styles) => {
         if (rootInstance === undefined) {
           return;
         }
@@ -64,6 +65,15 @@ const useBreakpointDelete = () => {
         }
 
         deleteCssRulesByBreakpoint(rootInstance, breakpoint.id);
+
+        // delete breakpoint styles
+        // reversed order to splice without breaking index
+        for (let index = styles.length - 1; index >= 0; index -= 1) {
+          const style = styles[index];
+          if (style.breakpointId === breakpoint.id) {
+            styles.splice(index, 1);
+          }
+        }
       }
     );
   });
--- a/apps/designer/app/canvas/shared/breakpoints.ts
+++ b/apps/designer/app/canvas/shared/breakpoints.ts
@@ -12,6 +12,7 @@ import {
 import { publish } from "~/shared/pubsub";
 import { addMediaRules } from "./styles";
 import { useSyncInitializeOnce } from "~/shared/hook-utils";
+import { filterMutable } from "~/shared/array-utils";
 
 export const useInitializeBreakpoints = (breakpoints: Array<Breakpoint>) => {
   const [, setCurrentBreakpoints] = useBreakpoints();
@@ -67,13 +68,7 @@ const useBreakpointDelete = () => {
         deleteCssRulesByBreakpoint(rootInstance, breakpoint.id);
 
         // delete breakpoint styles
-        // reversed order to splice without breaking index
-        for (let index = styles.length - 1; index >= 0; index -= 1) {
-          const style = styles[index];
-          if (style.breakpointId === breakpoint.id) {
-            styles.splice(index, 1);
-          }
-        }
+        filterMutable(styles, (style) => style.breakpointId !== breakpoint.id);
       }
     );
   });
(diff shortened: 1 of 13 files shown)


apps/designer/app/canvas/shared/breakpoints.ts
```diff
@@ -64,6 +65,15 @@ const useBreakpointDelete = () => {
...
+
+        // delete breakpoint styles
+        // reversed order to splice without breaking index
+        for (let index = styles.length - 1; index >= 0; index -= 1) {
+          const style = styles[index];
+          if (style.breakpointId === breakpoint.id) {
+            styles.splice(index, 1);
+          }
+        }
       }
```
kof, line 77: should be a utility like deleteCssRulesByBreakpoint with a test

apps/designer/app/canvas/shared/styles.ts
```diff
@@ -250,6 +272,37 @@ const useUpdateStyle = () => {
       }
       utils.tree.setInstanceStyleMutable(rootInstance, id, updates, breakpoint);
     });
+
+    store.createTransaction([stylesContainer], (styles) => {
+      const instanceId = id;
+      const breakpointId = breakpoint.id;
+      for (const update of updates) {
+        const matchedIndex = styles.findIndex(
```
kof, line 280: could be a useful reusable styles utility, no?

apps/designer/app/shared/nano-states/nano-states.ts
```diff
@@ -23,6 +23,16 @@ export const useSetPresetStyles = (presetStyles?: PresetStyles) => {
   });
 };
 
+export const stylesContainer = createValueContainer<Styles>([]);
+export const useStyles = () => useValue(stylesContainer);
+export const useSetStyles = (styles?: Styles) => {
+  useState(() => {
+    if (styles) {
```
kof, line 30: what happens if styles is undefined? don't we want to unset?

apps/designer/app/shared/nano-states/nano-states.ts
```diff
@@ -23,6 +23,16 @@ export const useSetPresetStyles = (presetStyles?: PresetStyles) => {
   });
 };
 
+export const stylesContainer = createValueContainer<Styles>([]);
+export const useStyles = () => useValue(stylesContainer);
+export const useSetStyles = (styles?: Styles) => {
+  useState(() => {
```
kof, line 29: this is where I did the research last time and ended up reverting the usage of useState for things like this, because its a side effect, and ended up with creating this hook useSyncInitializeOnce

packages/project/src/db/styles.ts
```diff
@@ -0,0 +1,30 @@
+import { type Patch, applyPatches } from "immer";
+import { prisma } from "@webstudio-is/prisma-client";
+import { Styles } from "@webstudio-is/react-sdk";
+
+const loadStylesByTreeId = async (treeId: string) => {
```
kof, line 5: I used Tree["id"] so that in case we want to change that id at some point, we have all pointers referenced, I don't know if it makes much sense but it also reads quite well, because reader knows its not just any string, its the string used in the Tree id and can easily find the algorithm used for generating that id


I am pointing this out because its a general question of how we treat such strings  from types perspective, cc @istarkov  and @rpominov

kof (approved):
commented on a number of small things

apps/designer/app/canvas/shared/breakpoints.ts
TrySound: This is basically filterMutable, can be extracted in utilities if you want

apps/designer/app/canvas/shared/styles.ts
TrySound: You mean findIndex? We may end up with some kind of indexes to speedup access. But it's a different story.

apps/designer/app/shared/nano-states/nano-states.ts
TrySound: This is just type guard. It should not be possible.

apps/designer/app/canvas/shared/styles.ts
kof: no, the whole block

apps/designer/app/shared/nano-states/nano-states.ts
kof: if it should be impossible, then types should not define it as optional, but if its possible then maybe you wanna allow setting an undefined

apps/designer/app/canvas/shared/breakpoints.ts
kof: yeah

packages/project/src/db/styles.ts
rpominov: I also like to define as `Tree["id"]` for documentations purposes / to show intent. But no strong opinion

apps/designer/app/canvas/shared/styles.ts
```diff
@@ -250,6 +272,37 @@ const useUpdateStyle = () => {
       }
       utils.tree.setInstanceStyleMutable(rootInstance, id, updates, breakpoint);
     });
+
+    store.createTransaction([stylesContainer], (styles) => {
```
rpominov, line 276: We're calling `store.createTransaction()` twice in a row. I think this results in 2 HTTP requests. Can we call it once line this?

```
store.createTransaction([rootInstanceContainer, stylesContainer], 
(rootInstance, styles) => { ... })
```

/cc @kof as I only vaguely understand how `createTransaction` works.

apps/designer/app/canvas/shared/styles.ts
TrySound: It's gonna be used only here. Where do you want to reuse it?

apps/designer/app/canvas/shared/styles.ts
TrySound: We put transactions in queue and send all at once.
I split because previous block will be completely removed in coming PRs.

rpominov (approved):

PR #758, opened 2023-01-09
Synchronize canvas styles with designer

TrySound:
To get rid of cssRules in instances we need to have an access to normalized styles.

Here I grouped all styles state management into styles store.

The idea is the same whole project use. All state management happen in canvas and resulting state is sent to designer.

Ideally do all changes on designer side and send to canvas. This will be easy to achieve once we have whole state management grouped into single place and all interaction done via messages.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/shared/stores/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+
+type DeleteStylesMessage = {
+  store: "styles";
+  operation: "delete";
+  breakpointId: string;
+  instanceId: string;
+  property: StyleProperty;
+};
+
+export type StylesMessage = SetStylesMessage | DeleteStylesMessage;
```
kof, line 30: "StylesMesasge" as a name is confusing in different contexts, how about `StyleUpdateOperation`

kof: generally, "message" confuses me, in this case its a terminology from the underlying protocol, not high-level business level naming

apps/designer/app/shared/stores/messages.ts
```diff
@@ -0,0 +1,10 @@
+import type { StylesMessage } from "./styles";
+
+type Message = StylesMessage;
+export type Messages = Message[];
```
kof, line 4: UpdateOperations ?

apps/designer/app/shared/stores/styles.ts
TrySound: We already use operation as field for set and delete. Maybe just an StylesUpdate = SetStylesUpdate | DeleteStylesUpdate?

apps/designer/app/canvas/shared/breakpoints.ts
```diff
@@ -66,11 +68,26 @@ const useBreakpointDelete = () => {
         }
 
         deleteCssRulesByBreakpoint(rootInstance, breakpoint.id);
-
-        // delete breakpoint styles
-        filterMutable(styles, (style) => style.breakpointId !== breakpoint.id);
       }
     );
+
+    const messages: Update[] = [];
```
kof, line 74: nit: would call it updates

apps/designer/app/shared/stores/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export type StylesUpdate = SetStylesUpdate | DeleteStylesUpdate;
+
+declare module "~/shared/pubsub" {
+  export interface PubsubMap {
+    loadStyles: Styles;
+  }
+}
+
+export const updateStyles = (styles: Styles, update: StylesUpdate) => {
```
kof, line 38: this is mutating, we used convention updateStylesMutable when we mutate things

apps/designer/app/shared/stores/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  if (update.operation === "delete" && matchedIndex !== -1) {
+    styles.splice(matchedIndex, 1);
+  }
+};
+
+export const useInitStyles = (styles: Styles) => {
+  // set initial styles value
```
kof, line 67: obvious comment, not providing the "why"

apps/designer/app/shared/stores/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+  useSyncInitializeOnce(() => {
+    stylesContainer.value = styles;
+  });
+
+  // styles are maintained on canvas side
+  // always send latest state to designer
+  useEffect(() => {
+    publish({
+      type: "loadStyles",
```
kof, line 76: loadStyles sounds like something that is done once, maybe we need to change the semantics if this happens periodically

kof: question: is this going to send the entire styles array every time a single property has been changed in style panel? on a large site, this is going to send a large payload?

apps/designer/app/shared/stores/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+      });
+    });
+  }, []);
+
+  // subscribe to styles updates
+  useEffect(() => {
+    return subscribe("update", (updates) => {
+      store.createTransaction([stylesContainer], (styles) => {
+        for (const update of updates) {
+          if (update.store === "styles") {
```
kof, line 92: lets talk about concept of stores you are now introducing. Initially nano-states were the same thing as stores, so what's the difference here.

I know we want to switch to nanostores, but this change is not doing it ....

kof: was it necessary to have an update specify the store `update.store` instead of just subscribing to "updateStyles" and setting the value on the nano state - this is how this has been done before, so I would like to understand what exactly we are changing and why

apps/designer/app/shared/stores/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+      });
+    });
+  }, []);
+};
+
+// synchronize with canvas styles
+export const useSubscribeStyles = () => {
+  useEffect(() => {
+    return subscribe("loadStyles", (styles) => {
+      stylesContainer.dispatch(styles);
```
kof, line 105: since you are in react here you could as well use `const [, setStyles] = useStyles()`

apps/designer/app/shared/stores/styles.ts
TrySound: Yes, same way implemented loadRootInstance.
We could send to designer and canvas same messages and compute new state in both instead of sending whole state. Though for now there is a problem with undo/redo which are tied to immerhin store which exist only in canvas at the moment.

apps/designer/app/shared/stores/styles.ts
TrySound: update can include updates for single transaction. With events we'll get multiple transactions.

apps/designer/app/shared/stores/styles.ts
TrySound: I would get rid of this dispatch in hooks at all and used only messages.

apps/designer/app/shared/stores/styles.ts
TrySound: > lets talk about concept of stores you are now introducing. Initially nano-states were the same thing as stores, so what's the difference here.

The idea is to group whole logic related to updating specific store/state. Caz it's getting messy with other nano states where all updates happen.

Once we rely only on specific messages we can add data indexing (some maps) for fast access.

Issue #759, opened 2023-01-09
Size section of the style panel


taylornowotny:
You'll find these designs in the Size Section page in the Webstudio Design Docs file.
<img width="224" alt="Screenshot 2023-01-08 at 11 07 59 PM" src="https://user-images.githubusercontent.com/86495896/211369468-396e6bf1-8879-4dbe-9eb8-08161f1812f4.png">

It's a pretty simple section for now. Should work exactly like Webflow's Size section.
<img width="329" alt="Screenshot 2023-01-08 at 11 07 35 PM" src="https://user-images.githubusercontent.com/86495896/211369471-a5262dea-dcf6-4d5c-81a4-96b922c0682d.png">

- [x] Object Fit is will be added to this section soon, but I need some time to make icons for this. I will make a separate issue for the object fit property when that's done.

TrySound:
Object position looks weird
<img width="227" alt="image" src="https://user-images.githubusercontent.com/5635476/228025345-d3580334-6207-4d1d-8a1d-b81c61278b45.png">

kof:
how do you get it, I have it full size

kof:
oh we broke it on the current main, prod still looks good

PR #760, opened 2023-01-10
Improve size section layout according to design

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/759

Spacing a little not clear so wrapped with common grid

Screenshot 2023-01-10 at 16 55 40

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


taylornowotny:
The defaults are not correct. Please use the default values from the design.
Min Width & Min Height have a default of 0px.
Width & Height have a default of Auto. (Do not show the computed value, show "Auto").

Default values on a new box in prod:
Screenshot 2023-01-10 at 8 47 54 AM

Design:
Screenshot 2023-01-10 at 8 49 40 AM

TrySound:
Changed title. This is only layout.

Though it's not clear how we implement keywords support. auto and non are no the only supported keywords. Are we not gonna show dropdown button?
image

taylornowotny:
> Changed title. This is only layout.
> 
> Though it's not clear how we implement keywords support. auto and non are no the only supported keywords. Are we not gonna show dropdown button? image

Ok, the layout looks great!

We will consider the keywords question. But for now yes, do not show a dropdown menu only the unit menu.

TrySound:
Btw this is not true if we follow css. Initial value is auto. 0 behaves a little differently

> Min Width & Min Height have a default of 0px.

https://developer.mozilla.org/en-US/docs/Web/CSS/min-width#formal_definition

kof:
> Btw this is not true if we follow css. Initial value is auto. 0 behaves a little differently

webflow doesn't actually set min- values to 0 but shows 0 in the UI, its actually wrong, @taylornowotny do you know any practical implication there?

kof:
E.g. for max-width wf shows correct: 'none', it seems to me we should stay with correct browser values

taylornowotny:
> Btw. there is another problem with min-width/min-height or actually a solution related to it - in nested flex layouts we use min-width: 0; to ensure the box respects layout, its a hack as far as I know but it seems to be the best solution that exists and we started adding it to all components in our own design system.
> 
> https://makandracards.com/makandra/66994-css-flex-and-min-width
> 
> So if we don't have any other solution, we should just set min-width in our presetStyle for all elements.

Sounds like this is the reason for min width & min height to have a default value of 0.

kof:
> Sounds like this is the reason for min width & min height to have a default value of 0.

I checked, even though webflow shows in the UI 0, they don't apply it as a preset. So webflow is kinda wrong twice, I am not sure why they do it.

taylornowotny:
Are we in agreement to use 0 for min-width & min-height default value for the flexbox reason?

I am not aware of any reason to do it any particular way. 0 is the same as "None" so I have no preference there. I have never had or heard any complaints about Webflow's size section.

rpominov (approved):

PR #762, opened 2023-01-11
Synchronize canvas store with designer

TrySound:
Here used immerhin subscribe method to send patches from canvas to desiger instead of uploading full states when their value is changed.

Syncrhonization is still one-way. Need more changes in immerhin to make two-way patches work.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/shared/sync/sync-stores.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const containers = new Map<string, ValueContainer<any>>([
+  ["breakpoints", breakpointsContainer],
+  ["root", rootInstanceContainer],
+  ["styles", stylesContainer],
+  ["props", allUserPropsContainer],
+  ["designTokens", designTokensContainer],
+]);
```
kof, line 38: how can we avoid having a central place that knows about all of these?

kof: it was this way as well, but I think if this list supposed to grow over time, it should be decentralized, so that each container is responsible for it's registration

immerhin already throws an error if we pass a container to transaction that hasn't been registered with immerhin

apps/designer/app/shared/sync/sync-stores.ts
```diff
@@ -0,0 +1,104 @@
...
+        type: "sendStoreChanges",
+        payload: {
+          target: "designer",
+          changes,
+        },
+      });
+    });
+    return () => {
+      unsubscribeStore();
+    };
```
kof, line 71:
```suggestion
    return unsubscribeStore;
```

apps/designer/app/shared/sync/sync-stores.ts
kof: I think we couod have `registerContainer(namespace, container)`

apps/designer/app/shared/sync/sync-stores.ts
```diff
@@ -0,0 +1,104 @@
...
+  }, []);
+};
+
+export const useDesignerStore = () => {
```
kof, line 75: maybe 
```suggestion
export const useSubscribeDesignerStores = () => {
```

apps/designer/app/shared/sync/sync-stores.ts
TrySound: Maybe later. There is still no immerhin in designer.

apps/designer/app/shared/sync/sync-stores.ts
TrySound: Will add more subscriptions later

apps/designer/app/shared/sync/sync-stores.ts
TrySound: We subscribe here canvas store. The thing is if we want two-way patches this will not be only "subscription".

PR #777, opened 2023-01-13
add figma-design-tokens.ts to .gitignore

rpominov:
## Description

`figma-design-tokens.ts` is generated from `figma-design-tokens.json` which is updated by the tokens plugin in Figma.
If we keep `figma-design-tokens.ts` under GIT control, then after `.json` is updated a random developer will have the `.ts` file updated in their PR.

I didn't add it to .gitignore initially because @TrySound suggested that we should commit it in https://discord.com/channels/955905230107738152/1059756719032516618 , but this isn't working well as described above.


rpominov:
Damn, removing it broke Button in lostpixel 

Screenshot 2023-01-13 at 11 47 59

Trying to figure out how to do this right

PR #782, opened 2023-01-13
Move breakpoints logic to designer

TrySound:
We have stores synchronization and can move the logic from canvas to designer.

Here refactored this way breakpoints, got rid from direct atom changes without store and removed breakpoints pub sub events.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/shared/array-utils.ts
```diff
@@ -1,11 +1,23 @@
-export const filterMutable = <T>(
-  array: T[],
-  predicate: (value: T) => boolean
-) => {
+type Predicate<T> = (value: T) => boolean;
```
kof, line 1: Please don't use abbreviation, I know its what ts documentation uses, but its a know bad idea that ts authors expressed several times

kof: could be Item in this case e.g.

kof (approved):

PR #789, opened 2023-01-16
Fix styles uploads

TrySound:
Forgot to move uploads logic to normalized styles

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project/src/db/tree.ts
+++ b/packages/project/src/db/tree.ts
@@ -1,4 +1,3 @@
-import { formatAsset } from "@webstudio-is/asset-uploader/server";
 import {
   type Tree,
   type ComponentName,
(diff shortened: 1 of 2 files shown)


packages/project/src/db/tree.ts
```diff
@@ -89,42 +26,10 @@ const InstanceDbOut = z.lazy(() =>
     id: z.string(),
     component: z.string(),
     children: z.array(z.union([InstanceDbOut, Text])),
-    cssRules: z.optional(z.array(CssRuleDbOut)),
+    cssRules: z.optional(z.unknown()),
```
kof, line 29: can't be deleted at this point?

kof (approved):

packages/project/src/db/tree.ts
TrySound: Deleted in normalize instances PR

istarkov (approved):

PR #799, opened 2023-01-17
Menu design update

### LINKED ISSUE #798: Implement Menu component from Figma
In Figma we have a Menu and Menu Item components:

https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library

Screenshot 2023-01-17 at 13 26 18

We need to have some components in code that are capable of everything defined in Figma. One to one matching might not be possible in this case as we have two components in code [`DropdownMenu`](https://github.com/webstudio-is/webstudio-designer/blob/main/packages/design-system/src/components/dropdown-menu.tsx) and [`Combobox`](https://github.com/webstudio-is/webstudio-designer/blob/main/packages/design-system/src/components/combobox.tsx) that we use for menus.

rpominov:
## Description

Closes #798.

I've implemented "Menu" and "Menu item" from Figma and updated our components that use menus.

TODO at some later point:

  - Component: "Menu Item Large"
  - Type of "Menu" component: "Dropdown w/large items"
  - Type of "Menu" component: "Context menu"
  
# TODO:
 - [x] `packages/design-system/src/components/icon-button-with-menu.tsx` also needs to be updated (probably use `DropdownMenu` in it) 

## Steps for reproduction

- open main app menu in designer (top left button)
- open menu in any component in the style panel that has a menu
- open project selection menu in dashboard
- https://menu--638245578b47d4399fba191a.chromatic.com/?path=/story/library-menu--menu

## Code Review

  - check demo with the link above (all features from Figma except the ones mentioned in TODO above should be demonstrated in the demo. Please ignore how inputs that trigger the menu look. This is only the demo of menu itself)

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov:
> Found a little mistake, the arrow for submenu looks left-aligned in the design

Good catch, I didn't notice. @taylornowotny what is the right way? Usually the the sub menu arrow is at the right side I think.|

> Also feels like it renders a little bolder or darker than in figma, maybe its just how chrome renders the same font or maybe there is some difference

I cant see this

taylornowotny:
> > Found a little mistake, the arrow for submenu looks left-aligned in the design
> 
> Good catch, I didn't notice. @taylornowotny what is the right way? Usually the the sub menu arrow is at the right side I think

It's supposed to be right-aligned, this was an error in the design that is fixed now.

taylornowotny:
Questions:

1. What is 'radio'? In the design we have 3 variants with differently colored dots. These variants are used to communicate when styles are applied to an element that is represented by a menu item. The dot is not meant to indicate a selection like the check mark is. It should not toggle on/off with any menu action.

2. What is the difference between "Regular" and "Toggleable"? Regular is meant to be toggleable in the first menu. In the second menu none are meant to be toggleable.

3. Did you include keyboard shortcuts? I did not show this in the examples but a keyboard shortcut can be toggled on/off in the component properties.

taylornowotny:
Screenshot 2023-01-19 at 9 25 21 AM

Besides my questions, this is my only design feedback. The horizontal alignment for the submenu is off a bit, but the vertical alignment is good.

rpominov:
> What is 'radio'? In the design we have 3 variants with differently colored dots. These variants are used to communicate when styles are applied to an element that is represented by a menu item. The dot is not meant to indicate a selection like the check mark is. It should not toggle on/off with any menu action.

Got it. Will add an ability to have an arbitrary icon that doesn't toggle on/off.

Initially I thought that design shows behaviour similar to "radio items" here (we use this component under the hood): https://www.radix-ui.com/docs/primitives/components/dropdown-menu 
Screenshot 2023-01-19 at 21 26 37

> What is the difference between "Regular" and "Toggleable"? Regular is meant to be toggleable in the first menu. In the second menu none are meant to be toggleable.

The component in code allows to have items that work like a button (for example "Copy") in the same menu with togglable items (for example "Undock navigator" in our menu):

Screenshot 2023-01-19 at 21 33 35

But if we need all items to be toglgable or all not toglgable the component allows to do that.

> Did you include keyboard shortcuts? I did not show this in the examples but a keyboard shortcut can be toggled on/off in the component properties. (updated examples just now to include shortcuts)

We get some keyboard support from the components that we use under the hood.

> I didn't see titles, submenus, separators, destructive, or disabled items in the select menus. Those are included in select menus, right?

I'll add this to the demo.

taylornowotny:
Thanks @rpominov ! Great work on this one, I know it's a complex component.

Re: radios - We won't need a radio button variant. But it would be nice to have a dot on the submenu item to show that an item inside the submenu is selected. Would that be possible?

rpominov:
> But it would be nice to have a dot on the submenu item to show that an item inside the submenu is selected. Would that be possible?

Once there will be an ability to add arbitrary icon to a menu item that I'm going to add, it will be trivial for developer to add a dot icon when something is selected in submenu in any particular instance of menu.

But it's tricky to make this the default behaviour for all menus because of how component API is structured internally: the Item that opens submenu has no way to know whether the submenu has togglable items and whether they are on or off.

Also, probably not a good idea to make this default for all menus. App menu and context menu are very common UIs and they never do that AFAIK. Here is menu from my browser as example:

Screenshot 2023-01-20 at 10 29 08

taylornowotny:
Thank you for the thoughtful response. We can go without the dot no
problem.

PR #818, opened 2023-01-20
Figma tokens

markxtang:
## Description

Renamed some styles to make it more clear which styles are brand styles as opposed to builder styles. E.g. before the brand color styles were distributed with the other color styles, now they are all under brand. grouping. Also renamed the atomic typography tokens such as font-size so they are in order (smallest value to highest value corresponds to first to last in the array)

Documentation in the Figma Library doc still needs to happen, and I'm not sure whether the steps below apply to this situation. 
## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov (approved):

rpominov:
Thanks

> I'm not sure whether the steps below apply to this situation.

Yes, I think they don't apply

packages/design-system/src/__generated__/figma-design-tokens.json
```diff
@@ -37,7 +37,7 @@
     },
     "lineHeights": {
       "0": {
-        "value": "16",
+        "value": "AUTO",
```
kof, line 40: is that something we want? @rpominov

rpominov: https://discord.com/channels/955905230107738152/1065939291479478343

PR #819, opened 2023-01-20
Migrate InstanceProps into Tree table as props column

TrySound:
InstanceProps is designed to store props per each instance which requires a lot of sql updates on every props change.

Here moved per instance props into array with props for whole tree. This way patches are safer to apply on large trees.

```
type PropsItem = {
  id: string,
  instanceId: string,
  name: string,
  required?: boolean,
  type: 'number' | 'string' | 'boolean' | 'asset',
  value: number | string | boolean
}
```

## Code Review

  - test it locally

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
This also made types/structure easier to read since there is no props.props

packages/prisma-client/prisma/schema.prisma
```diff
@@ -80,6 +80,7 @@ model Tree {
   id           String @id @default(uuid())
   root         String
   instances    String @default("[]")
+  props        String @default("[]")
   presetStyles String @default("[]")
   styles       String @default("[]")
 }
```
kof, line 86: do we still need InstanceProps model?

TrySound: I'd remove it later. Just in case.

PR #820, opened 2023-01-20
Add new label for style panel

TrySound:
Used new local/remote terms instead of set/inherited from figma for color here.

Screenshot 2023-01-20 at 14 06 35

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound:
Oops, set sans-serif globally and broke other diffs. Apply them too?

packages/design-system/.storybook/preview.js
```diff
@@ -7,3 +7,10 @@ export const parameters = {
     },
   },
 };
+
+export const decorators = [
+  (StoryFn) => {
+    document.body.style.fontFamily = "sans-serif";
```
kof, line 13: @rpominov is on this problem right now, you can just remove it from this PR

packages/design-system/src/components/label.tsx
```diff
@@ -0,0 +1,78 @@
+import type { LabelHTMLAttributes, ReactNode, Ref } from "react";
+import { forwardRef } from "react";
+import { styled } from "../stitches.config";
+import { theme } from "../stitches.config";
+
+const StyledLabel = styled("label", {
```
kof, line 6: The thing is the "label" is too generic for this component, because this component is specific to style panel PropertyName.

Maybe this should be renamed in figma, because this seems super misleading.  I started conversation in figma comments about this, lets wait for Taylor to comment

TrySound: Totally agree

packages/design-system/src/components/label.tsx
```diff
@@ -0,0 +1,78 @@
...
+  boxSizing: "border-box",
+  flexShrink: 0,
+
+  // shared style
+  py: 2,
+  border: "1px solid transparent",
+  borderRadius: theme.borderRadius[3],
+  fontSize: theme.fontSize[3],
```
rpominov, line 14: This should come from typography token. In Figma you can see which token to use like this:

Screenshot 2023-01-20 at 18 07 26

In this case it's "Labels - Sentence Case" which will correspond to `typography.labelsSentenceCase` once https://github.com/webstudio-is/webstudio-designer/pull/821 is merged

packages/design-system/src/components/label.tsx
```diff
@@ -0,0 +1,78 @@
+import type { LabelHTMLAttributes, ReactNode, Ref } from "react";
```
rpominov, line 1: Please add a header like this: https://github.com/webstudio-is/webstudio-designer/blob/78af54b0175b7f7be67f1f1c1964b1ea24e10cf6/packages/design-system/src/components/button.tsx#L1-L4

So we would know that this is not a random component but something that should match design in Figma

packages/design-system/src/components/label.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { forwardRef } from "react";
+import { styled } from "../stitches.config";
+import { theme } from "../stitches.config";
+
+const StyledLabel = styled("label", {
+  boxSizing: "border-box",
+  flexShrink: 0,
+
+  // shared style
```
rpominov, line 10: What does this mean?

packages/design-system/src/components/label.tsx
```diff
@@ -0,0 +1,78 @@
...
+  py: 2,
```
rpominov, line 11: Should use `theme.spacing` unless the comment above explains that this is an exception

packages/design-system/src/components/label.tsx
```diff
@@ -0,0 +1,78 @@
...
+  variants: {
+    color: {
+      default: {
+        color: theme.colors.foregroundMain,
+        "&[aria-disabled=true]": {
+          color: theme.colors.foregroundDisabled,
+        },
+      },
+      preset: {
+        px: 4,
```
rpominov, line 29: `theme.spacing` here and in other `px`s below

packages/design-system/src/components/label.tsx
rpominov: Can leave as is with a todo if you're going to merge earlier. Saying just so you know this is coming

TrySound: Will rebase once you merge

packages/design-system/src/components/label.tsx
kof: After talking with Taylor we agreed to keep calling it label because its having general-purpose variants + special ones we use in style panel

This is basically what we do with all design system components to avoid having to create a separate design system for designer

kof (approved):

TrySound:
Looks like rebase broke lost pixel :(

TrySound:
Or @kof broke :)
Screenshot 2023-01-22 at 11 28 16

kof (approved):

PR #822, opened 2023-01-20
Transform styles without zod

TrySound:
Found zod allows to pass anything in transform without type checking. This means we can easily get unexpected result. In this case we could get undefined as asset.

As a solution I defined two versions of types, client side and stored in db. Conversion is implemented on db routes side.

This makes data flow more clear and let us describe only two versions of types instead of tree (in, out, client side).

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project/src/db/styles.ts
+++ b/packages/project/src/db/styles.ts
@@ -39,10 +39,11 @@ export const parseStyles = async (stylesString: string) => {
   const assetIds: string[] = [];
   for (const { value: styleValue } of storedStyles) {
     if (styleValue.type === "image") {
-      for (const item of styleValue.value)
+      for (const item of styleValue.value) {
         if (item.type === "asset") {
           assetIds.push(item.value);
         }
+      }
     }
   }
 
--- a/packages/project/src/db/styles.ts
+++ b/packages/project/src/db/styles.ts
@@ -1,84 +1,72 @@
 import warnOnce from "warn-once";
-import DataLoader from "dataloader";
 import { type Patch, applyPatches } from "immer";
-import { type Asset, prisma } from "@webstudio-is/prisma-client";
+import { prisma } from "@webstudio-is/prisma-client";
+import type { Asset } from "@webstudio-is/asset-uploader";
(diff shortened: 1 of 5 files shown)


kof (approved):

packages/project/src/db/styles.ts
```diff
@@ -34,89 +29,108 @@ const assetsLoader = new DataLoader<string, Asset | undefined>(
...
+                  value: asset,
+                },
+              ];
+            }
+            return [];
+          })
+        )
+      ).flat(),
+    };
+  } else {
```
kof, line 65: nit: that else is not necessary

packages/project/src/db/styles.ts
```diff
@@ -34,89 +29,108 @@ const assetsLoader = new DataLoader<string, Asset | undefined>(
...
-  property: z.string() as z.ZodType<StyleProperty>,
-  value: StyleValueDbOut,
-});
+const loadValue = async (styleValue: StoredStyles[number]["value"]) => {
+  if (styleValue.type === "image") {
+    return {
+      type: "image" as const,
+      value: (
+        await Promise.all(
+          styleValue.value.map(async (item) => {
```
kof, line 50: a little too much nesting, makes it harder to read, would put it as a variable above

packages/project/src/db/styles.ts
```diff
@@ -34,89 +29,108 @@ const assetsLoader = new DataLoader<string, Asset | undefined>(
...
+      type: "image" as const,
+      value: styleValue.value.map((asset) => ({
+        type: asset.type,
+        /**
+         * In the DB we hold only assetId
+         **/
+        value: asset.value.id,
+      })),
+    };
+  } else {
```
kof, line 101: also else not needed

packages/project/src/db/styles.ts
TrySound: Got rid of promises

PR #827, opened 2023-01-21
Add shallowComputed nanostores utility

TrySound:
The new utility allows us to subscribe to filtered array and give the same reference when all items are the same. This prevents unnecessary rerenders in react because getSnapshot considers value unchanged.

This is necessary for styles and we'll need same for props.

The downside is filter and shallow compare will be called on every change for every instance. We will see if this give us performance issues.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/shared/nano-states/nano-states.ts
+++ b/apps/designer/app/shared/nano-states/nano-states.ts
@@ -1,6 +1,6 @@
-import { atom, type WritableAtom } from "nanostores";
+import { atom, computed, type WritableAtom } from "nanostores";
 import { useStore } from "@nanostores/react";
-import type { Instance, PresetStyles, Styles } from "@webstudio-is/react-sdk";
+import { Instance, PresetStyles, Styles } from "@webstudio-is/react-sdk";
 import type {
   DropTargetChangePayload,
   DragStartPayload,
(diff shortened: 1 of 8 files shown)


apps/designer/app/shared/nano-states/nano-states.ts
```diff
@@ -31,6 +31,22 @@ export const useSetPresetStyles = (presetStyles: PresetStyles) => {
...
+    const { instanceId } = stylesItem;
+    let instanceStyles = stylesByInstanceId.get(instanceId);
+    if (instanceStyles === undefined) {
+      instanceStyles = [];
+      stylesByInstanceId.set(instanceId, instanceStyles);
+    }
+    instanceStyles.push(stylesItem);
+  }
+  return {
+    stylesByInstanceId,
```
kof, line 46: could be directly returning styleByIndexId, so that styleByIndexIdStore can be there

kof (approved):
great!

apps/designer/app/shared/nano-states/nano-states.ts
TrySound: We may need other indexes in the future. And it may be cheaper to reuse the same loop

PR #834, opened 2023-01-22
Upgrade to turbo 1.7

TrySound:
We use very old version of turborepo which often fails. Here latest one with persistent tasks for watch mode support.

We also need to enable remote caching to get turbo benefits on CI.

While the size of turbo repo grew. Sice 1.3 a lot of artifacts for unpopular systems were removed so we still get smaller node_modules.

```
du -ckd1 node_modules
-797528
+788132
```

@kof to enable remote caching need to run `pnpm turbo login` and then `pnpm turbo link`.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,6 @@ jobs:
       NODE_ENV: test
       DATABASE_URL: postgres://
       AUTH_SECRET: test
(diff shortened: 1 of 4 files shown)


kof (approved):

TrySound:
And it does not work..

PR #835, opened 2023-01-22
New dashboard has reached feature parity with the old one

kof:
## Description

This is not a complete dashboard implementation, but it can now list, create and open projects, so it can be deployed to production and replace the old version while we work on the remaining features.

In addition implemented: renaming, duplicating, deletion (mark as deleted), keyboard accessibility

## Steps for reproduction

1. click on card to open the project in the designer
1. click on a link at the card bottom if project is published to open published version
1. click on new project and create new project
2. click on rename over menu to rename
3. click on duplicate
4. click on delete
5. Use tab to switch between projects, use arrow keys to switch between focusable elements inside the project card

## Code Review

  - test it on preview
- [ ] hi @istarkov, we need to discuss how we want to build the future APIs with TRPC and probably rewrite what I did here right away

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov:
> test it on preview

Cannot test on Preview because there's a migration: 
Screenshot 2023-01-24 at 10 28 02

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,292 @@
...
+        nextElement?.focus();
+        break;
+      }
+    }
+  };
+
+  const handleDelete = (projectId: string) => {
+    fetcher.submit(
+      { projectId },
+      { method: "delete", action: "/dashboard/projects/delete" }
```
rpominov, line 191: Not a big deal, but a bit strange that we use both `method: "delete"` and `/delete` in path to tell what action to take. Same for `/rename`

If we're trying to follow REST to some extent, we should do something like:

```
/dashboard/projects + PUT = create
/dashboard/projects + POST = edit
/dashboard/projects + DELETE = delete
```

Anyway not a big deal as we probably going to rewrite API at some point.

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,292 @@
...
+          <Text variant="title" truncate>
+            {title}
+          </Text>
+          {isPublished ? (
+            <PublishedLink domain={domain} tabIndex={-1} />
+          ) : (
+            <Text color="hint">Not Published</Text>
+          )}
+        </Flex>
+        {isFeatureEnabled("dashboard2") && (
```
rpominov, line 275: > isFeatureEnabled("dashboard2")

Do we still need this?

packages/project/src/db/project.ts
```diff
@@ -84,35 +84,64 @@ export const create = async ({
...
+  return await clone(project, userId, `${project.title} (copy)`);
+};
+
+export const cloneByDomain = async (domain: string, userId: string) => {
+  const project = await loadByDomain(domain);
+  if (project === null) {
+    throw new Error(`Not found project "${domain}"`);
+  }
+  return await clone(project, userId);
+};
```
rpominov, line 143: We used to clone "prod" build in the old `clone()`, but in `cloneByDomain()` that replaces it we clone "dev" build.

Probably doesn't matter, just wanted to point out in case this is an accidental change we don't want.

apps/designer/app/dashboard/project-card.tsx
rpominov: Ah, I thought dashboard2 was used to hide new dashboard

kof: since I added support for most items, I will just hide the share item behind the feature flag for now

packages/project/src/db/project.ts
kof: oh, def. accidental, forgotten during refactoring

istarkov (approved):

apps/designer/app/routes/dashboard/projects/new.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  const authenticatedUser = await findAuthenticatedUser(request);
+
+  if (authenticatedUser === null) {
+    throw new Error("Not authenticated");
+  }
+
+  const project = await db.project.create({
+    title,
+    userId: authenticatedUser?.id,
```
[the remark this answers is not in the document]
kof: accidental

apps/designer/.env.example
```diff
@@ -22,17 +22,17 @@ S3_ACCESS_KEY_ID=
...
-PUBLISHER_HOST=
+PUBLISHER_HOST=wstd.io
+PUBLISHER_ENDPOINT=https://publisher.webstudio.is
 PUBLISHER_TOKEN=
 
 # Base origin for accessing user-generated build (canvas / preview / etc)
 # Should contatin protocol and host (e.g. https://example.com)
 # Will be detected automatically in development environments like localhost or preview deployment
-BUILD_ORIGIN=
+BUILD_ORIGIN=https://wstd.live
```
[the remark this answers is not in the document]
kof: goal was to show some example of a value when user needs to set up their own production instance ... any suggestions what to show instead? localhost values?

kof:
> Cannot test on Preview because there's a migration:

Since we are not running the migrations on branch, how do we let others test it?

apps/designer/app/dashboard/project-card.tsx
kof: The problem is when there is more than CRUD, e.g. duplicate, you end up needing a separate route for it or alternatively an additional param like "action" or something ...

So there is an option to make all crud operations in /projects and use method  and `ifs` to call all functions from the same action, then have a separate /duplicate and probably also /rename

What would you prefer? also cc @istarkov in relation to trpc what's the way forward?

apps/designer/.env.example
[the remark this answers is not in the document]
rpominov: At localhost we want to leave it empty. 

So for the use-case when developer copies `.env.example` to `.env` when setting up dev environment the first time we want to leave it unset.

Maybe we should do like this:
```
# BUILD_ORIGIN=https://wstd.live
```

apps/designer/app/dashboard/project-card.tsx
rpominov: I have no strong preference between:
 - `/dashboard/projects + PUT = create or duplicate depending on input`
 - `/dashboard/projects + PUT = create; /dashboard/projects/duplicate + POST = duplicate`
 
 But if we're going to migrate to trpc, probably no reason to spend a lot of time trying to make it right now

rpominov:
> Since we are not running the migrations on branch, how do we let others test it?

I don't know how to do this other than setting up separate DB per branch and running the migrations

apps/designer/.env.example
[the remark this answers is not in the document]
kof: good idea

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+const usePublishedLink = ({ domain }: { domain: string }) => {
+  const [url, setUrl] = useState<URL>();
+
+  useEffect(() => {
+    // It uses `location` to detect the default values at localhost.
+    setUrl(new URL(getPublishedUrl(domain)));
+  }, [domain]);
+
+  return { url };
```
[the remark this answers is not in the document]
kof: the problem with getPublishedUrl is that it uses location when there is no env variable defined, like its the case with local env, so when it renders on that backend - browser location doesn't exist and it will mismatch at hydration

basically it boils down to having a localhost:3000 predefined by default serverside and since I haven't found a way to have it without hardcoding and without forcing a user to define it, we detect it on the client

we need a way to get the default host:port without forcing user to set it, remix can be searching for a free port if 3000 is taken, so its not guaranteed, if we hardcode the port we will run into a problem when its taken.

we basically need to get the final port from remix after it has found a free one.

packages/dashboard/src/index.server.ts
```diff
@@ -0,0 +1,2 @@
+export * as db from "./db";
```
TrySound, line 1: Do we need direct access with trpc?

packages/dashboard/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "@webstudio-is/react-sdk": "workspace:^",
+    "bson-objectid": "^2.0.4",
+    "immer": "^9.0.12",
+    "nanoid": "^3.2.0",
+    "react": "^17.0.2",
+    "slugify": "^1.6.5",
+    "uuid": "^9.0.0",
+    "warn-once": "^0.1.1",
+    "zod": "^3.19.1",
+    "@trpc/server": "^10.9.0"
```
TrySound, line 32: Do we need here all these deps?

apps/designer/tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "remix.env.d.ts",
     "**/*.ts",
     "**/*.tsx",
-    "../../@types/**/*.d.ts"
+    "../../@types/**/*.d.ts",
+    "../../packages/dashboard/src/trpc/_router.ts"
```
TrySound, line 9: Is this global type?

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,295 @@
...
+  return {
+    designerLinkRef,
+    handleKeyDown,
+    handleDelete,
+    handleRename,
+    handleDuplicate,
+  };
+};
+
+const proxy =
```
TrySound, line 226: Here proxy, there trpc. Could be more consistent.

apps/designer/app/shared/remix/create-trpc-remix-proxy.ts
```diff
@@ -0,0 +1,40 @@
+import { useFetcher } from "@remix-run/react";
+import type {
+  AnyMutationProcedure,
+  AnyRouter,
+  inferRouterInputs,
+  inferRouterOutputs,
+} from "@trpc/server";
+
+const proxy = (get: (method: string) => void) =>
+  new Proxy({}, { get: (_target, method: string) => get(method) });
```
TrySound, line 10: Why do this scary api with proxy? Could we just add separate useQuery, useMutation which  got trpc something with type info as argument and infer output api? Like relay does for fragments.

rpominov:
Getting the same error as in CI when trying to build locally (wanted to see how TRPC types work)

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,295 @@
...
+  py: theme.spacing[5],
+  px: theme.spacing[7],
+});
+
+const projectNameAvatarStyle = css({
+  fontFamily: theme.fonts.manrope,
+  fontWeight: 200,
+  fontSize: 360,
+  alignSelf: "center",
+  letterSpacing: "-0.05em",
```
rpominov, line 52: The font styles should come from `typography.brandThumbnailLargeDefault` (this token is now in `main`)

Screenshot 2023-01-25 at 12 11 24

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,295 @@
...
+  background: "linear-gradient(0deg, #FBF8FF 25%, #E2E2E2 66%)",
+  WebkitBackgroundClip: "text",
+  backgroundClip: "text",
+  color: "transparent",
+  // @todo figure out the effect from figma in text-shadow
+  //textShadow: "0px 5px 3px rgba(251, 248, 255, 1)",
+  userSelect: "none",
+  outline: "none",
+  "&:hover, &:focus": {
+    fontWeight: 800,
```
rpominov, line 63: This is in `typography.brandThumbnailLargeHover`

apps/designer/app/dashboard/project-card.tsx
```diff
@@ -0,0 +1,295 @@
...
+      case "ArrowLeft": {
+        const nextElement = elements.at(currentIndex - 1) ?? elements[0];
+        nextElement?.focus();
+        break;
+      }
+    }
+  };
+  const { submit: deleteProject } = proxy.delete.useMutation();
+  const { submit: rename } = proxy.rename.useMutation();
+  const { submit: duplicate } = proxy.duplicate.useMutation();
```
rpominov, line 197: Seems like we're ignoring errors from all of these

kof:
@rpominov did you run migrations?

apps/designer/app/shared/remix/create-trpc-remix-proxy.ts
kof: So are we good with this after the discussion or are there serious concerns?

rpominov:
> @rpominov did you run migrations?

I've run migrations now, but still getting the same error when running `pnpm build`.

Not a big deal if this is just me, was just curious to test

apps/designer/app/shared/remix/create-trpc-remix-proxy.ts
```diff
@@ -0,0 +1,40 @@
+import { useFetcher } from "@remix-run/react";
+import type {
+  AnyMutationProcedure,
+  AnyRouter,
+  inferRouterInputs,
+  inferRouterOutputs,
+} from "@trpc/server";
+
+const proxy = (get: (method: string) => void) =>
```
istarkov, line 9: trpc has exported method for this

```
import { createRecursiveProxy } from "@trpc/server/shared";

  createRecursiveProxy(async (opts) => {
    const path = [...opts.path]; // e.g. ["post", "byId", "useMutation"]
    const method = path.pop();
    if (method!=='useMutation') throw ...
```

apps/designer/app/routes/dashboard/projects.$method.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const action = async ({ request, params }: ActionArgs) => {
+  const authenticatedUser = await findAuthenticatedUser(request);
+  if (authenticatedUser === null) {
+    throw new Error("Not authenticated");
+  }
+  const data = Object.fromEntries(await request.formData()) as never;
+  // @todo pass authorization context
+  const caller = projectRouter.createCaller({ userId: authenticatedUser.id });
+  return await caller[params.method as keyof typeof caller](data);
```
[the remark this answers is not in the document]
rpominov: Adding `if (typeof fn === "function")` doesn't solve security issue IMO. We're allowing user to call any function that exists on `caller`, how do we know that this is safe?

I'd just hardcode `const allowedMethods = ["create", ...]; if (allowedMethods.include(params.method)) ...`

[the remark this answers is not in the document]
kof: true, also we need this logic to be some reusable function we can call and it should automatically know which methods  are defined in the router @istarkov

TrySound:
Would be a little easier to review without mixing trpc integration and dashboard features

apps/designer/app/shared/remix/create-trpc-remix-proxy.ts
istarkov: done

apps/designer/app/routes/dashboard/projects.$method.ts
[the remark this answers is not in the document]
istarkov: Added checks on existing procedures.

istarkov (approved):

rpominov (approved):
Looks good other than these unresolved comments:

Screenshot 2023-01-25 at 14 51 09

PR #839, opened 2023-01-24
Use instances index to compute selected instance

TrySound:
Because we tracked the tree and selected instance, traversed and updated the value on every change which usually lead to running same code twice in a row.

Here I added computed index for instances which can be reused in other parts. It is generated by traversing tree. Later will be replaced with for loop.

This allows us to store only selected id and compute instance with `instancesById.get(selectedInstanceId)`.

This store need to be synchronized with designer to use it instead of selected instance data in almost all cases.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/shared/nano-states/nano-states.ts
```diff
@@ -79,6 +101,26 @@ export const useSetDesignTokens = (designTokens: DesignToken[]) => {
   });
 };
 
+export const selectedInstanceIdStore = atom<undefined | Instance["id"]>(
```
istarkov, line 104: May be not export it and use `selectedInstanceStore.get().id`?

istarkov (approved):

apps/designer/app/shared/nano-states/nano-states.ts
```diff
@@ -79,6 +101,26 @@ export const useSetDesignTokens = (designTokens: DesignToken[]) => {
...
+  undefined
+);
+export const selectedInstanceStore = computed(
+  [instancesIndexStore, selectedInstanceIdStore],
+  (instancesIndex, selectedInstanceId) => {
+    if (selectedInstanceId === undefined) {
+      return undefined;
```
kof, line 111:
```suggestion
      return;
```

kof (approved):

Issue #807, opened 2023-01-18
Style Source UI


kof:
Part of Design Tokens User story https://github.com/webstudio-is/webstudio-designer/issues/422

Figma https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=1013%3A31684&t=3rKaN8KoDwdga6FC-0

## Style Source UI 

- [x] Type a new input with autocomplete
- [x] Select an existing source from a list
- [x] Enter a new source
- [x] Hover the source to see the menu
- [x] Menu provides: Remove, Duplicate, Disable, Edit name
- [x] Drag and drop to reorder
- [x] Click to toggle select/unselect
- [x] Double click to edit name
- [x] Local source can only be disabled, nothing else should be possible
- [x] Hit Backspace to delete the last Source item when you are in the input
- [x] Dropdown menu according to design

## Integration in Style Panel

- [x] Render the Style Source UI where local styles are saved on the Local Style Source
- [x] store style sources in db
- [x] select: always only one is selected
- [x] rename
- [x] remove
- [x] duplicate with auto rename
- [x] sorting with dnd
- [x] merge of all styles into single object
- [x] blue property label for selected style source
- [x] regenerate ids on copy/paste
- [x] when creating or adding a token, it should be selected right away
- [x] A min width on style sources of 32px will leave 16px for the target area
- [x] Cursor should be default when hovering any part of the style source. 
- [x] when dragging source item, prevent selection with user-select: none (could be potentially done over a body globally for any drag/drop over our api)
- [ ] #1576
- [x] delete (a destructive one that actually deletes the source from project)
- [x] orange property label (remote) for not selected style source
- [x] convert to token
- [ ] #1577
- [ ] disabled - persistent/multiplayer
- [x] selected - in-memory when instance switched 
   - select the last one selected
   - window reloaded - selected the latest
- [ ] use InputField component instead of DeprecatedTextField

kof:
Added a todo for a delete menu item for deleting a style source entirely from the project, as opposite to removing, which just removes it from the style source selector. We had this planed for style source manager, but we need to add this to the style source menu, because otherwise its gonna be unusable fast

kof:
To make sure this doesn't go under,  I came across the problem that I sometimes forget to create a token and start working on a local source, but then have no way to copy things over and reset the local.

The best idea so far is "Convert to Token": 
1. copies style properties
[...]

PR #850, opened 2023-01-25
Add style sources schema

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/807

We are gonna consolidate design token styles and local styles. On build (project) level there will be list of styles and list of style sources which are basically a group of styles to be referenced.

These style sources can be local with mandatory tree id or token with optional tree id to scope a token on specific page instead of whole project.

Tree will get new list of references to style sources.

Here resulting types

```ts
// Stored as Build.styles
type StylesItem = {
  // same as before but instead of instanceId style is referenced through style source
  styleSourceId: string,
  breakpointId,
  property,
  value
}

type StyleSourcesToken = {
  type: token
  id: string
  treeId?: string
  name: string
}

type StyleSourcesLocal = {
  type: local
  id: string
  treeId: string
  name: string
}

// Stored as Build.styleSources
type StyleSourcesItem = StyleSourcesToken | StyleSourcesLocal

// Stored as Tree.styleRefs
type StyleRefsItem = {
  instanceId: string,
  values: StyleSourcesItem['id']
}
```

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

packages/react-sdk/src/db/style-sources.ts
```diff
@@ -0,0 +1,34 @@
+import { z } from "zod";
```
kof, line 1: Here is what I would call them:

StyleSourcesToken > StyleSourceToken
StyleSourcesLocal > StyleSourceLocal

StyleSourcesItem > StyleSource
StyleRefsItem > StyleSourceSelection

StyleRefs > StyleSourceSelections

kof: lets create project-build package

kof (approved):
great

PR #867, opened 2023-01-26
Share UI

kof:
## Description

Share UI implemented as a story

https://user-images.githubusercontent.com/52824/215069285-0ee7b214-8218-4ead-a468-07a22f65e433.mp4

## Steps for reproduction

1. open stories in designer
2. share project section

## Code Review

  - test it on preview
- [ ] hi @istarkov,  please check if the interface I made fits the data layer you built

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/popover.tsx
```diff
@@ -22,7 +22,7 @@ const StyledContent = styled(PopoverPrimitive.Content, panelStyles, {
   backgroundColor: "white",
   minWidth: 200,
   minHeight: theme.spacing[13],
-  maxWidth: 265,
+  maxWidth: "max-content",
```
rpominov, line 25: What is the idea of `maxWidth: "max-content"`? I see it in many places. What does this do, what the intent as developer?

rpominov (approved):

packages/design-system/src/components/popover.tsx
kof: the container doesn't grow beyond the needs of its content, e.g. a block container would grow otherwise full available width

kof: So in this case content of the tooltip determines it's width

istarkov (approved):

PR #868, opened 2023-01-27
Migrate designer to normalized props

TrySound:

Here added normalized props logic to transactions and replaced useAllUserProps with computed store.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/canvas.tsx
+++ b/apps/designer/app/canvas/canvas.tsx
@@ -27,9 +27,11 @@ import {
   rootInstanceContainer,
   selectedInstanceStore,
   useBreakpoints,
+  useInstanceProps,
   useRootInstance,
   useSetBreakpoints,
   useSetDesignTokens,
+  useSetProps,
   useSetRootInstance,
   useSetStyles,
   useSubscribeScrollState,
@@ -95,13 +97,17 @@ const useAssets = (initialAssets: Array<Asset>) => {
 };
 
 const useCopyPaste = () => {
-  const instance = useStore(selectedInstanceStore);
-  const allUserProps = useAllUserProps();
+  const selectedInstance = useStore(selectedInstanceStore);
+  const instanceProps = useInstanceProps(selectedInstance?.id);
 
-  const selectedInstanceData = useMemo(
-    () => instance && { instance, props: allUserProps[instance.id] },
-    [allUserProps, instance]
-  );
+  const selectedInstanceData = useMemo(() => {
+    if (selectedInstance) {
+      return {
+        instance: selectedInstance,
+        props: instanceProps,
+      };
+    }
+  }, [selectedInstance, instanceProps]);
 
   // We need to initialize this in both canvas and designer,
   // because the events will fire in either one, depending on where the focus is
@@ -150,6 +156,7 @@ export const Canvas = ({ data }: CanvasProps): JSX.Element | null => {
   useSetBreakpoints(data.breakpoints);
   useSetDesignTokens(data.designTokens);
   useAllUserProps(data.props);
+  useSetProps(data.tree.props);
   useSetStyles(data.tree.styles);
   useSetRootInstance(data.tree.root);
   setParams(data.params ?? null);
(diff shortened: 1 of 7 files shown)


apps/designer/app/shared/nano-states/nano-states.ts
```diff
@@ -42,6 +44,40 @@ export const instancesIndexStore = computed(
...
+  const instancePropsStore = useMemo(() => {
+    return shallowComputed([propsIndexStore], (propsIndex) => {
+      if (instanceId === undefined) {
+        return [];
+      }
+      return propsIndex.propsByInstanceId.get(instanceId) ?? [];
+    });
+  }, [instanceId]);
+  const instanceProps = useStore(instancePropsStore);
+  return instanceProps;
```
rpominov, line 78: Does `propsIndexStore` get recomputed only when `propsStore` change? 

Looking at this code it seems like we might be creating separate instances of the computed store per `instanceId`.

Did you test how this works in practice?

apps/designer/app/canvas/canvas.tsx
```diff
@@ -27,9 +27,11 @@ import {
   rootInstanceContainer,
```
rpominov, line 27: (writing here just to make it a thread)

Bug:

- add a "button" to canvas
- set "type" props to "reset"
- select the button on canvas
- copy (Cmd+C)
- click on canvas to make "body" selected
- paste (Cmd+V)
- the "type" props value is not copied

Same bug when selecting in the navigator instead of canvas

apps/designer/app/canvas/shared/instance.ts
```diff
@@ -58,28 +59,32 @@ export const findInsertLocation = (
...
-          dropTarget ?? findInsertLocation(rootInstance, selectedInstanceId)
-        );
-        if (hasInserted) {
-          selectedInstanceIdStore.set(instance.id);
+  useSubscribe(
+    "insertInstance",
+    ({ instance, dropTarget, props: insertedProps }) => {
+      const selectedInstanceId = selectedInstanceIdStore.get();
+      store.createTransaction(
+        [rootInstanceContainer, allUserPropsContainer, propsStore],
```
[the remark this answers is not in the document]
TrySound: Yes, though its data used on server side for now.

apps/designer/app/shared/nano-states/nano-states.ts
TrySound: Yes, propsIndexStore is recomputed every time propsStore is changed. Each instancePropsStore is recomputed when propsIndexStore is changed. Though useStore triggers rerender only when one of props array elements are changed.

Ideally we need to maintain a separate state with index though it's complicated with patches so we recompute shared index for all invocations instead.

apps/designer/app/canvas/canvas.tsx
TrySound: Yeah, we need to generate new id for each inserted prop.

apps/designer/app/shared/sync/sync-stores.ts
```diff
@@ -44,6 +45,7 @@ export const registerContainers = () => {
   store.register("root", rootInstanceContainer);
   store.register("styles", stylesContainer);
   store.register("props", allUserPropsContainer);
+  store.register("props-modern", propsStore);
```
kof, line 48: I would just call the old props `deprecatedProps`

kof (approved):

apps/designer/app/canvas/canvas.tsx
kof: I approved, though if its a new bug please fix it before merging

TrySound: Kinda new. Right now a lot of duplicated prop.id are stored in db. Here the bug just become visible.

TrySound: I'll fix props and styles copy paste in next PR

PR #869, opened 2023-01-27
Fix props with assets loading

TrySound:
Db data with asset ids was parsed as client data with resolved assets. Fixed by reusing props parsing/serializing logic in tree.

Asset inlining complicates code so we need to have two versions of types and transformations.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project/src/db/props.ts
+++ b/packages/project/src/db/props.ts
@@ -1,31 +1,29 @@
 import warnOnce from "warn-once";
-import { type Tree, StoredProps } from "@webstudio-is/project-build";
+import { type Tree, StoredProps, Props } from "@webstudio-is/project-build";
 import { type AllUserProps, type InstanceProps } from "@webstudio-is/react-sdk";
 import { applyPatches, type Patch } from "immer";
 import { prisma } from "@webstudio-is/prisma-client";
 import { formatAsset } from "@webstudio-is/asset-uploader/server";
 
-export const loadByTreeId = async (treeId: Tree["id"]) => {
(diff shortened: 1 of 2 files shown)


kof (approved):

packages/project/src/db/props.ts
```diff
@@ -1,31 +1,29 @@
...
-  });
-
-  if (tree === null) {
-    return [];
+export const parseProps = async (propsString: string) => {
+  let storedProps;
+  try {
+    storedProps = StoredProps.parse(JSON.parse(propsString));
+  } catch (e) {
+    console.log("hello", propsString);
```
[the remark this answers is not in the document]
TrySound: oops, debug

packages/project/src/db/props.ts
```diff
@@ -73,12 +59,60 @@ export const loadByTreeId = async (treeId: Tree["id"]) => {
...
+      };
+      instancePropsMap.set(propsItem.instanceId, instanceProps);
+    }
     instanceProps.props.push(propsItem);
   }
 
   return Array.from(instancePropsMap.values());
 };
 
+export const serializeProps = (props: Props) => {
```
kof, line 91: Do we need to add tests for serialize/parse functions? It feels like the probability of breaking this stuff in the future is high

TrySound: We have almost same code for styles. Maybe we need to change how we handle assets.

PR #873, opened 2023-01-27
Delete design tokens prototype

kof:
## Description

1. With all the new ideas and directions we don't need that initial implementation

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/prisma-client/prisma/migrations/20221218211129_tree_text/schema.prisma
```diff
@@ -91,10 +90,4 @@ model Breakpoints {
   build   Build  @relation(fields: [buildId], references: [id])
   buildId String @id
   values  String @default("[]")
-}
-
-model DesignTokens {
```
TrySound, line 96: I think shouldn't be touched inside migrations

kof: ups

TrySound (approved):

Issue #828, opened 2023-01-21
Delete styles and props along with instances


TrySound:
At the moment we preserve styles and props after deleting their instance to cover edge case in collaboration mode when user delete instance and other user changes style on this instance, the redo apply old value instead of one entered by other user.

We decided to ditch this case and always delete styles and props.

PR #887, opened 2023-01-29
Delete instance styles and props

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/828

Delete instance along with own and descendants styles and props.
This will let us to avoid bloating db with unused data.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/shared/tree-utils.ts
+++ b/apps/designer/app/shared/tree-utils.ts
@@ -1,0 +1,61 @@
+import type { Instance, Props, Styles } from "@webstudio-is/project-build";
+import { removeByMutable } from "./array-utils";
+
+const traverseInstances = (
+  instance: Instance,
+  cb: (child: Instance, parent: Instance) => void
+) => {
+  for (const child of instance.children) {
+    if (child.type === "text") {
+      continue;
+    }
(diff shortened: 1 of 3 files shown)


kof (approved):

apps/designer/app/shared/tree-utils.ts
```diff
@@ -0,0 +1,61 @@
...
+}: {
+  rootInstance: Instance;
+  props: Props;
+  styles: Styles;
+  deletedInstanceId: string;
+}) => {
+  const parentInstances = new Map<Instance["id"], Instance>();
+  const deletedInstances = new Set<Instance["id"]>();
+
+  traverseInstances(rootInstance, (child, instance) => {
```
kof, line 33: even in this case seems like an unnecessary function because all it does is this:

```suggestion
  for (const child of instance.children) {
    if (child.type !== "instance") {
      continue;
    }
};
```

TrySound: Not really, it's recursive function

PR #889, opened 2023-01-29
Drop unused schema

TrySound:
DesignTokens was removed from schema but without migration. Removed everything else unused too.

- Tree.presetStyles
- InstanceProps
- DesignTokens

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
For some reason I thought removing model would have generated drop table when I removed the DesignTokens table

TrySound:
It did, but then you reverted migration along with unrelated changes.

Issue #863, opened 2023-01-26
Offline queue and syncing once online


kof:
As a user I want to know  that my changes will be still synced when I loose the connection and then get it back.

The missing piece of logic during sync is:
- [x] check if the sync was successful, if not - don't remove it from the queue (or put it back) . Right now we are optimistically removing from the queue when syncing
- [x]  after 3 unsuccessful attempts to sync - show the disconnected icon
- [x]  add a tooltip (like in figma) that explains that we will sync as soon as we are online again
- [x] use `addEventListener('offline'` event listener to explain in text of the tooltip that internet is not available, otherwise assume by default server is not available.
- [x] When there are unsaved changes and user tries to close the window - use onbeforeunload to tell them about it and that they will loose the unsynced changes otherwise (until we add offline capability to save those and sync them once online)

- Figma doesn't show offline icon if you don't have unsynced changes and go offline.
- Figma uses this message in the tooltip: "Offline changes will be synced with Figma once you go online."

kof:
For beforeunload there is a hook, just in case https://github.com/streamich/react-use/blob/master/docs/useBeforeUnload.md

PR #893, opened 2023-01-30
Connection/queue status

kof:
Fixes https://github.com/webstudio-is/webstudio-designer/issues/863

## Description

Icon used is temporary, until we have a better one. We can merge this because we only show it when offline, so most people won't see it anyways.

- [x] check if the sync was successful, if not - don't remove it from the queue (or put it back) . Right now we are optimistically removing from the queue when syncing
- [x]  after 3 unsuccessful attempts to sync - show the disconnected icon
- [x]  add a tooltip (like in figma) that explains that we will sync as soon as we are online again
- [x] use `addEventListener('offline'` event listener to explain in text of the tooltip that internet is not available, otherwise assume by default server is not available.
- [x] When there are unsaved changes and user tries to close the window - use onbeforeunload to tell them about it and that they will loose the unsynced changes otherwise (until we add offline capability to save those and sync them once online)

https://user-images.githubusercontent.com/52824/215371856-2d76a364-3918-42b6-a4eb-9693c54d5690.mp4

## Steps for reproduction

1. go to chrome network and set connection to offline or go to the wifi and turn it off (if you are not locally running the server)

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/designer/features/topbar/sync-status.tsx
```diff
@@ -1,50 +1,62 @@
...
+  useEffect(() => {
+    const handle = () => setIsOnline(navigator.onLine);
+    addEventListener("offline", handle);
+    addEventListener("online", handle);
+    return () => {
+      removeEventListener("offline", handle);
+      removeEventListener("online", handle);
+    };
+  }, []);
+  return isOnline;
```
TrySound, line 23:
```
const subscribeOnline = (cb) => {
  addEventListener("offline", handle);
  addEventListener("online", handle);
  return () => {
    removeEventListener("offline", handle);
    removeEventListener("online", handle);
  };
}

const isOnline = () => navigator.onLine

const isOnlineServer = () => true

useSyncExternalStore(subscribeOnline, isOnline, isOnlineServer)
```

apps/designer/app/designer/shared/sync/sync-server.ts
```diff
@@ -0,0 +1,91 @@
+import { sync } from "immerhin";
```
TrySound, line 1: Why you have two sync-server.ts?

kof: because I am great at resolving merge conflicts 🤣

apps/designer/app/designer/features/topbar/sync-status.tsx
kof: it shouldn't be used server-side at all, neat notation but a bit of indirection

TrySound: It's optional. You can not use it if component is client only

apps/designer/app/designer/features/topbar/sync-status.tsx
```diff
@@ -1,50 +1,62 @@
...
-          animation: `${ellipsisKeyframes} steps(4,end) 900ms infinite`,
-          width: 0,
-          overflow: "hidden",
-        }}
-      >
-        <DotsHorizontalIcon width="12" height="12" />
-      </Box>
-    </Flex>
-  );
+const useSetOnline = () => {
```
TrySound, line 15: Why use if it's not hook? atom use useSyncExternalStore underhood too. Why not use it?

TrySound: Won't need to have useEffect

apps/designer/app/designer/shared/sync/queue.ts
```diff
@@ -0,0 +1,67 @@
...
+
+// After this amount of retries without success, we consider connection status error.
+const MAX_RETRY_RECOVERY = 5;
+
+export type SyncStatus = "syncing" | "idle" | "recovering" | "error";
+
+type BaseResult = { ok: boolean };
+
+export const state: {
+  status: WritableAtom<SyncStatus>;
```
TrySound, line 11: status is not a state, it's state container. Maybe split separately since you anyway want to use it as export.

apps/designer/app/designer/features/topbar/sync-status.tsx
kof: I don't want to use useSyncExternalStore in the application logic, it is too low level

kof: changing the name

apps/designer/app/designer/features/topbar/sync-status.tsx
```diff
@@ -1,50 +1,62 @@
...
-          animation: `${ellipsisKeyframes} steps(4,end) 900ms infinite`,
-          width: 0,
-          overflow: "hidden",
-        }}
-      >
-        <DotsHorizontalIcon width="12" height="12" />
-      </Box>
-    </Flex>
-  );
+const setIsOnline = () => {
```
TrySound, line 15: subscribeIsOnline?

apps/designer/app/designer/features/topbar/sync-status.tsx
TrySound: Isn't useEffect low level too?

kof: it is, but at least everyone is intimately familiar with its interface

kof: I don't feel comfortable using syncExternalStore directly

TrySound: They both do exactly same: synchronization. useSyncExternalStore syncrhonize external store, useEffect synchronize dom.

kof: yeah, though external store also runs serverside, but I only need this particular thing to run in frontend only

kof: I just don't feel the need for it here

kof: Don't see the benefit

kof: over atom we got the store anyways, so unless you want to fix some problem here, lets not waste time

TrySound (approved):

Issue #548, opened 2022-11-30
Token based sharing link with permissions


kof:
- [ ] We need to be able to share a project with different users with different permissions, for now 3:
  - view - can use the designer, but changes aren't saved, it's just for trying things out without accidentally modifying project
  - design - changes from the design panel will be saved
  - edit content - content changes will be saved
- [ ] A token has to decide what types of permissions the link gets
- [ ] The same token will be used in the CLI
- [ ] stop using project id like a token, currently it's basically the only secret that gives anyone editing access to any site

Notion has a good example for sharing dialog
<img width="476" alt="Screenshot 2022-11-30 at 13 20 09" src="https://user-images.githubusercontent.com/52824/204807325-94d2ac7f-9072-4823-a4d3-f57576bff6d0.png">

kof:
@taylornowotny what part of it do you think is absolutely necessary for the beta launch and what can be done later?

kof:
Honestly for now, I would just let it be a single mode that allows editing everything and a preview mode that is basically the site, without designer, but the state is the latest, no need to be published.

kof:
Just found this issue we already had before, we need to close it, but lets define what needs to be done now for beta and merge them https://github.com/webstudio-is/webstudio-designer/issues/227

istarkov:
- [x] - Add `trpc-interface` package, depending on ENV variables it uses or the local TRPC server or any remote. The local server is used through `callerLink`, which allows calling local implementation directly using `Router.createCaller`. 

- [x] - Write our permission model using ORY permission language

- [x] - TRPC server created at SAAS. Instead of local, it will communicate with ORY permission system.

- [ ] - Clone project through login step issue #790 

- [ ] - Restrict edit/create operations using `userId` `projectId` for the Project and Build, sync projectId, userId with ORY permission system at SAAS

- [ ] - Add projectId field to Tree ref #742 and restrict edit operations on a tree using `userId` `projectId`

- [ ] - Add permission tokens to allow viewing/previewing projects on Canvas. i.e. `?mode=preview` `?mode=edit` both has DB access

- [ ] - Add read permissions checks for Project data read through `userId`, `projectId` AND `token`, `projectId`

kof:
## Link in with view permission

- [ ] When user comes from a link with 'view' permission  we should show the designer initially in "preview" mode (the eye icon). 
- [ ] User should be able to use everything, but we don't send any patches to the server, we just apply things on the canvas without saving them
[...]

PR #894, opened 2023-01-30
Shared tokens UI + Api

istarkov:
## Description

ref #548

Add shared links with builder/viewer permissions.

`editor` permission is now disabled as we don't know a simple way to allow editing images and text only cc @TrySound would need your opinion on this.

createTrpcRemixProxy now wraps input in `JSON.stringify`  and unwraps before the TRPC call. Otherwise, all params are FormData string entries and may cause errors.

createTrpcRemixProxy useQuery/useMutation now returns Remix useFetcher state parameter.

**`patch` auth call added, we must do any Auth mutate operations in a single call, otherwise if some call failed we are in troubles.**
Example on token relation update.
Remove viewer relation.
Add editor relation. <- If the call failed here user just loses all permissions.
In case of patch both operations are completed in a single transaction - (delete+insert)

## Steps for reproduction

Having migrations works locally only with 
`isFeatureEnabled("share2")` flag.

Click the share button then add/rename/update/copy share links.

The previous Share button is removed.

**TODO In Next PR:**

Different UI for builder and viewer links i.e. [this](https://github.com/webstudio-is/webstudio-designer/issues/548#issuecomment-1406947059).

## Code Review

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


packages/authorization-token/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "@webstudio-is/tsconfig/base.json",
+  "include": ["src", "../../@types/**/*.d.ts"],
+  "compilerOptions": {
+    "module": "ES2020",
+    "target": "ES2020",
+    "baseUrl": ".",
+    "outDir": "./lib",
+    "rootDir": "./src"
```
TrySound, line 9: these tree and include are not needed

packages/trpc-interface/src/authorize/authorization-token.server.ts
```diff
@@ -0,0 +1,106 @@
+import { AppContext } from "../context/context.server";
```
TrySound, line 1: type

packages/trpc-interface/src/authorize/project.server.ts
```diff
@@ -1,14 +1,12 @@
 import { AppContext } from "../context/context.server";
```
TrySound, line 1: type

apps/designer/app/designer/features/topbar/share.tsx
```diff
@@ -1,75 +1,112 @@
...
+
+export const ShareButton = ({ designerUrl }: ShareButtonProps) => {
+  return (
+    <Popover modal>
+      <PopoverTrigger asChild>
+        <Button>Share</Button>
+      </PopoverTrigger>
+
+      <PopoverPortal>
+        <PopoverContent hideArrow>
```
kof, line 106: needs zIndex: zindeces[1] //100 to be on top of other panels like pages

apps/designer/app/designer/features/topbar/share.tsx
```diff
@@ -1,75 +1,112 @@
...
+      projectId: project.id,
+      relation: "viewers",
+      name: "Custom link",
+    });
   };
-  return (
+
+  // @todo loading indicators
+  return state === "submitting" ? (
+    <Flex
```
kof, line 77: we won't need loading ... as discussed

apps/designer/app/routes/designer.tsx
```diff
@@ -14,4 +15,12 @@ export const meta: MetaFunction = () => {
   return { title: "Webstudio" };
 };
 
+export const shouldRevalidate: ShouldRevalidateFunction = ({
```
kof, line 18: I think its not trivial to know why this is needed, please explain in the comment

apps/designer/app/shared/remix/trpc-remix-proxy.ts
```diff
@@ -16,33 +17,46 @@ export const createTrpcRemixProxy = <Router extends AnyRouter>(
...
-
     const method = path.join(".");
 
-    const submit = (input: never) => {
-      return fetcher.submit(input, {
-        method: hook === "useMutation" ? "post" : "get",
-        action: getPath(method),
-      });
-    };
+    const submitOrLoad = useCallback(
```
kof, line 42: I was just discussing with Chance from Remix team that submit is designed to be used for FormData, we should call it send() instead

kof: submitOrLoad is also not very nice, I think  you could just have a separate function for .load()

apps/designer/app/shared/remix/trpc-remix-request.server.ts
```diff
@@ -18,7 +18,24 @@ export const handleTrpcRemixAction = async <Router extends AnyRouter>({
...
   }
 
-  const data = Object.fromEntries(await request.formData()) as never;
+  let input: never;
+
+  if (request.method === "GET") {
+    const url = new URL(request.url);
+    const inputRaw = url.searchParams.get("input");
+    if (typeof inputRaw !== "string") {
+      throw new Error("Missing input");
```
kof, line 27:
```suggestion
      throw new Error(`Bad method name ${inputRaw}`);
```

apps/designer/app/shared/remix/trpc-remix-request.server.ts
```diff
@@ -18,7 +18,24 @@ export const handleTrpcRemixAction = async <Router extends AnyRouter>({
...
+    }
+    input = JSON.parse(inputRaw) as never;
+  } else {
+    const formData = await request.formData();
+    const inputRaw = formData.get("input");
+    if (typeof inputRaw !== "string") {
+      throw new Error("Missing input");
```
kof, line 34: same

apps/designer/app/shared/share-project/share-project2.tsx
```diff
@@ -6,8 +6,6 @@ import {
   Label,
```
kof, line 6: rename this file to share-project.tsx and remove the old one

apps/designer/app/shared/share-project/share-project2.tsx
```diff
@@ -55,29 +59,54 @@ const Permission = ({
 };
 
 type MenuProps = {
-  permission: Permission;
+  relation: Relation;
```
kof, line 62: I don't see how you are importing Relation type, weird magic

apps/designer/package.json
```diff
@@ -37,6 +37,7 @@
     "@trpc/server": "^10.9.0",
     "@vercel/node": "^2.4.1",
     "@webstudio-is/asset-uploader": "workspace:^",
+    "@webstudio-is/authorization-token": "workspace:^",
```
kof, line 40: why not call it authorization? are you afraid unrelated stuff would go in?

packages/authorization-token/src/trpc/authorization-tokens-router.ts
```diff
@@ -0,0 +1,71 @@
...
+      })
+    )
+    .query(async ({ input, ctx }) => {
+      return await db.findMany({ projectId: input.projectId }, ctx);
+    }),
+  create: procedure
+    .input(
+      z.object({
+        projectId: z.string(),
+        relation: z.enum(["editors", "viewers", "builders"]),
```
kof, line 22: is it a convention to call those plural noun? 
what about: editor, viewer, builder
or verb: edit, view, build

packages/authorization-token/src/trpc/authorization-tokens-router.ts
```diff
@@ -0,0 +1,71 @@
...
+        ctx
+      );
+    }),
+  update: procedure
+    .input(
+      z.object({
+        projectId: z.string(),
+        token: z.string(),
+        name: z.string(),
+        relation: z.enum(["editors", "viewers", "builders"]),
```
kof, line 55: seems like I have seen this array quite a few times, maybe worth reusign

packages/prisma-client/prisma/migrations/20230130153140_authorization-tokens-fix/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+DROP TABLE "AuthorizationTokens";
+
+-- DropEnum
+DROP TYPE "AuthorizationPermit";
+
+-- CreateTable
+CREATE TABLE "AuthorizationToken" (
+    "token" TEXT NOT NULL,
+    "projectId" TEXT NOT NULL,
+    "name" TEXT NOT NULL DEFAULT 'Custom Link',
```
kof, line 14: "Custom link" should be a UI concern, this is never useful outside of share dialog UI which can always have this, I guess db should be happy with an empty string or something

kof:
Things we discussed on a call:
- [ ] no loading ... when showing
- [ ] use new popover with title/close
- [ ] use a quick transition (150-200ms) when loading link items on the container height, maybe also fade them in, goal - prevent unpleasant jumping
- [ ] typing in link name should instantly update the link name in the list without any delay or debounce
- [ ] debounce network request
- [ ] fix remix problem with AssetProvider by using SEARCH instead of GET

kof:
- [ ] also make sure story has an open state of the share UI because we need that to be captured by lost-pixel and track the visual regressions

apps/designer/app/shared/share-project/share-project2.tsx
istarkov: It's declared above no need to import. In case we pass unsupported value it would fail. There could be more Relations than this Component will support.

packages/authorization-token/src/trpc/authorization-tokens-router.ts
istarkov: There are 2 type of things which are different.
- Relations i.e `User is "viewer" of Project` or `User is "editor" of Project`
- Permits i.e `User can "view" Project` _Permits are based on relation like User can "view" project if he has Relation "editor"_

From Zanzibar paper we see that Relations often has "editor", "viewer", "parent", "owner" keywords. 
But in [Ory Permission Language](https://www.ory.sh/docs/keto/reference/ory-permission-language) we usually describe that a little bit inverted, instead of `User is "viewer" of Project` there we declare 

```ts
class Project implements Namespace {
  related: {
    viewers: (User | Token)[]
    editors: (User | Token)[]
  }
```

i.e. Project has "viewers" with Users or Tokens

So for Relations we can use any form "viewer" or "viewers" for me 2nd preferred as gives a little bit more information -
"owner", "parent" - means we assume that single owner, single parent exists.
"viewers" - that multiple viewers can exist.

The downside of Zanzibar is that internally there are no unique constraints and in reality no way except name to show that I want single parent, single owner for the Project but many viewers, editors, and builders.

The downside of having `s` is that what if we will want many `parents`, many `owners` etc but the keyword is already an `owner`. I have no strong opinion what's better

kof: We should expect having multiple owners I think

apps/designer/package.json
istarkov: Yep, as of now oss authorization logic is inside trpc-interface

apps/designer/app/shared/remix/trpc-remix-proxy.ts
istarkov: function is same renamed to `remixSubmit` its internal function to call remix submit

apps/designer/app/designer/shared/sync/sync-server.ts
```diff
@@ -1,90 +1,84 @@
...
+      const entries = sync();
+      if (entries.length === 0) {
         return;
       }
 
+      // @todo this entire queueing logic needs to be gone, it's a workaround,
+      // because prisma can't do atomic updates yet with sandbox documents
+      // and backend fetches and updates big objects, so if we send quickly,
+      // we end up overwriting things
       enqueue(() =>
```
TrySound, line 67: Did you reverted to old version accidentally?

apps/designer/app/shared/share-project/share-project.tsx
```diff
@@ -1,28 +1,284 @@
-import { Button, Flex, TextField } from "@webstudio-is/design-system";
+import {
+  Box,
+  Button,
+  css,
+  Flex,
+  Label,
+  DeprecatedPopover,
+  DeprecatedPopoverContent,
+  DeprecatedPopoverTrigger,
```
TrySound, line 9: Why here legacy components?

pnpm-lock.yaml
```diff
@@ -15456,6 +15543,21 @@ packages:
     dependencies:
       map-cache: 0.2.2
 
+  /framer-motion/8.5.5_sfoxds7t5ydpegc3knd667wn6m:
```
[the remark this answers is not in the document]
istarkov: Gonna remove tomorrow with own component in other PR

TrySound (approved):

pnpm-lock.yaml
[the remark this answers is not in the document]
istarkov: Radix has https://www.radix-ui.com/docs/primitives/components/collapsible to replace

apps/designer/app/designer/features/topbar/share.tsx
```diff
@@ -1,75 +1,126 @@
...
+export const ShareButton = ({ designerUrl }: ShareButtonProps) => {
+  const classZIndex = css({ zIndex: theme.zIndices[1] });
+
+  return (
+    <FloatingPanelPopover modal>
+      <FloatingPanelPopoverTrigger asChild>
+        <Button>Share</Button>
+      </FloatingPanelPopoverTrigger>
+
+      <FloatingPanelPopoverContent className={classZIndex()}>
```
kof, line 120:
```suggestion
      <FloatingPanelPopoverContent css={{ zIndex: theme.zIndices[1] }}>
```

apps/designer/app/shared/share-project/share-project.tsx
kof: we just deprecated them while this PR was still not ready, feel free to merge and update separately if you want

kof (approved):

apps/designer/app/designer/features/topbar/share.tsx
istarkov: FloatingPanelPopoverContent now doesn't support css prop. Internals need to be wrapped with styled etc.

Issue #742, opened 2023-01-05
Add "project" relation to Tree


istarkov:
We need this for authorization checks. Also allows faster debug/find needed values.

https://github.com/webstudio-is/webstudio-designer/blob/62d870ad8e3da6ff95d4e6449971d3b843e87a62/packages/prisma-client/prisma/schema.prisma#L79

```
-- prisma
project   Project @relation(fields: [projectId], references: [id])
```

istarkov:
Also, the `projectId` field must be a part of a unique key for the `Tree` table.
Otherwise, it would be possible to pass into the `patch` method other  Project `projectId` with needed permissions for the current user and do any changes.

PR #895, opened 2023-01-30
Add buildId and projectId relations to Tree table

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/742 https://github.com/webstudio-is/webstudio-designer/issues/807

Here I added build and project relations to Tree to have to improve auth security and simplify access without parsing pages.

This is necessary to migrate styles into new style sources format.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/prisma-client/prisma/schema.prisma
```diff
@@ -81,6 +83,10 @@ model Build {
 
 model Tree {
   id              String @id @default(uuid())
```
istarkov, line 85: See #896 I think we need to remove `@id`  attribute here and add `@@id([id, projectId])`, Sorry I totally forgot about this in issues.

TrySound: Will default work with it?

kof (approved):
can't verify the migration work on preview, but changes look good to me overall

PR #900, opened 2023-01-31
Figma tokens

taylornowotny:
added 2-3 new colors

is that enough info for this PR?


rpominov:
> is that enough info for this PR?

Yeah. A lot of info in commit messages anyway.

rpominov (approved):

PR #901, opened 2023-01-31
Dashboard dialogs

kof:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/floating-panel/dialog.tsx
```diff
@@ -80,7 +80,8 @@ const contentStyles = css(floatingPanelStyles, {
   top: "50%",
   left: "50%",
   transform: "translate(-50%, -50%)",
-  width: "90vw",
```
kof, line 83: where is this width coming from?

apps/designer/app/dashboard/new-project.tsx
```diff
@@ -35,15 +46,51 @@ const useNewProject = () => {
...
+              paddingBottom: theme.spacing["10"],
+            }}
+            gap="1"
+          >
+            <Label>Project Title</Label>
+            <form onSubmit={handleCreate}>
+              <TextField placeholder="New Project" onChange={handleChange} />
+            </form>
+          </Flex>
+        </FloatingPanelDialog.Description>
```
rpominov, line 76: Not sure `<Description>` will work well with all this stuff inside. In their own demo Radix developers wrap only a line of text inside it instead of entire dialog content

https://www.radix-ui.com/docs/primitives/components/dialog

Screenshot 2023-01-31 at 09 36 43

And you don't have to always have a descriptions AFAIK.

packages/design-system/src/components/floating-panel/dialog.tsx
rpominov: This is from Radix's example. Not sure what to use instead. Feel free to change

apps/designer/app/dashboard/new-project.tsx
kof: You are right

kof: just read what description is for, its accessibility feature

apps/designer/app/designer/designer.tsx
```diff
@@ -114,6 +114,13 @@ const useCopyPaste = (publish: Publish) => {
   });
 };
 
+const useSetWindowTitle = () => {
+  const [project] = useProject();
+  useEffect(() => {
+    document.title = `${project?.title} | Webstudio`;
+  }, [project?.title]);
+};
```
rpominov, line 122: Why can't we use the standard Remix way to set title?

rpominov (approved):
Looks good otherwise

apps/designer/app/designer/designer.tsx
rpominov: Ah, it's because it can be updated locally after the page is loaded?

kof: Standard way would be using serverside meta function, which doesn't have access to project and I wanted to have project title there because its confusing with multiple tabs open.

We only get project when we render the designer ui

TrySound:
Latest commit failed

PR #909, opened 2023-01-31
Make copy paste data agnostic

TrySound:
Copy paste of instances was implemented closly coupled to our schema. Here I make I implement it with simpler interface without dependency on types. Type and version are instead passed as configuration.

Next step will be decoupling instance copy paste from react and simplifying data management with nanostores.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
> Copy paste of instances was implemented closly coupled to our schema. Here I make I implement it with simpler interface without dependency on types. Type and version are instead passed as configuration.

I miss the information "why" we need it, what does it unblock?

kof:
Do I understand correctly that we are executing copy/paste logic on both, designer and canvas? and you are trying to make it react decoupled?

We could do this much easer by simply forwarding the copy/paste event and having the logic in designer only. Basically always trying to just forward events from canvas and that's it.

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
```diff
@@ -0,0 +1,69 @@
...
+const InstanceCopyData = z.object({
+  instance: Instance,
+  props: Props,
+  styles: Styles,
+});
+
+export type InstanceCopyData = z.infer<typeof InstanceCopyData>;
+
+type InstanceCopyPasteProps = {
+  selectedInstanceData?: InstanceCopyData | undefined;
```
kof, line 17: I don't know why we need undefined when its optional already

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
```diff
@@ -0,0 +1,69 @@
...
+  onCut: (instance: Instance) => void;
+};
+
+// to make it easier to remove React from canvas if we need to
+export const useInstanceCopyPaste = (props: InstanceCopyPasteProps): void => {
+  const { selectedInstanceData, allowAnyTarget, onPaste, onCut } = props;
+  useEffect(() => {
+    selectedInstanceData;
+    onPaste;
+    allowAnyTarget;
```
kof, line 29: wow, what is this doing?

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
```diff
@@ -0,0 +1,69 @@
...
+    return startCopyPaste({
+      version: "@webstudio/instance/v0.1",
```
kof, line 31: we should probably get the version somehow from package

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
```diff
@@ -0,0 +1,69 @@
...
+      type: InstanceCopyData,
+      onCopy: () => {
+        if (selectedInstanceData === undefined) {
+          return;
+        }
+        return selectedInstanceData;
```
kof, line 37: this is pointless, can be removed

apps/designer/app/shared/copy-paste/copy-paste.ts
```diff
@@ -0,0 +1,129 @@
+import { z } from "zod";
+
+const isValidClipboardEvent = (
+  event: ClipboardEvent,
+  options: { allowAnyTarget: boolean }
+) => {
+  const selection = document.getSelection();
+  if (selection !== null && selection.type === "Range") {
```
kof, line 8:
```suggestion
  if (selection?.type === "Range") {
```

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
TrySound: Looks like paste works for non inputs with it

TrySound: Ah, gotcha. Forgot to remove

TrySound:
The main idea is to split all copy/paste setup from data which was spread over few modules.
This will let me easier migrate to style sources.

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
```diff
@@ -0,0 +1,63 @@
...
+          return {
+            ...styleDecl,
+            instanceId: instance.id,
+          };
+        });
+
+        onPaste({ instance, props, styles });
+      },
+    });
+  }, [selectedInstanceData, allowAnyTarget, onPaste, onCut]);
```
rpominov, line 62: A bit weird that we re-start copy-paste every time `selectedInstanceData` changes. 

Probably not a big deal, but I like the previous version with `start/stop/setProps` a bit more even though it's ugly

rpominov (approved):
Couple comments with not strong opinions. Up to you what to do with them :)

apps/designer/app/shared/copy-paste/copy-paste-instance.ts
rpominov: In general, I think `start/stop/setProps` interface is a good default when you want to make a logic from a hook react-agnostic

TrySound: This is the thing. We don't need react to pass data here. We can access it from stores directly.

apps/designer/app/shared/copy-paste/copy-paste.ts
```diff
@@ -0,0 +1,130 @@
+import { z } from "zod";
+
+const isValidClipboardEvent = (
```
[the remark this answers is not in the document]
TrySound: Yeah, let's reconsider this when add new copy paste.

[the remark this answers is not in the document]
kof: future use case can be copy pasting webflow instances, where we would need to differentiate

PR #911, opened 2023-02-01
Fix copy paste styles and props of descendants

TrySound:
Found we do not copy styles and props for descendant instances. Here also removed copy/paste data from react and added a few new tree utilities.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/shared/tree-utils.test.ts
```diff
@@ -33,6 +37,65 @@ const createStyleDecl = (instanceId: string): Styles[number] => {
...
+  const string = expect.any(String) as unknown as string;
+  expect(clonedInstance).toEqual(
+    createInstance(string, [
+      createInstance(string, []),
+      createInstance(string, [createInstance(string, [])]),
+      createInstance(string, []),
+    ])
+  );
+  expect(clonedIds.get(instance.id)).toEqual(clonedInstance.id);
+  expect(clonedIds.get((instance.children[0] as Instance).id)).toEqual(
```
kof, line 94: why is this type assertion needed, its supposed to have the right type over there, no?

apps/designer/app/shared/instance-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+import {
+  rootInstanceContainer,
+  propsStore,
+  stylesStore,
+  selectedInstanceIdStore,
+} from "./nano-states";
+import { findSubtree } from "./tree-utils";
+import { removeByMutable } from "./array-utils";
+
+export const deleteInstance = (targetInstanceId: Instance["id"]) => {
```
kof, line 12: - we have been intentionally decoupling a mutable utility function  from store stuff and calling it with Mutate suffix, so we can test them in isolation

- its weird that you now have both, deleteInstance and deleteInstanceMutable 
- need unit tests for every non-trivial utility we write, this one seems untested

apps/designer/app/shared/tree-utils.test.ts
TrySound: There can be instance | text though we know only instance is there

apps/designer/app/shared/instance-utils.ts
TrySound: It is trivial. Non trivial is traversing of instances. Here we just ditch all matching data with tested utility.

Yeah, deleteInstanceMutable no longer necessary.

kof: I guess trivial is relative, in this case its doing many operations and more will be added, so I would test it, because its a very important one

kof: I would extract the mutable part of this function and write tests using it

kof: could be a snapshot test if its too much data

kof: as for immutable deleteInstance that wrapps the transition, are we using it in many places? if not I would just inline the transaction where its used

kof (approved):

apps/designer/app/shared/instance-utils.ts
TrySound: We gonna use it instead of pubsub.

TrySound: > could be a snapshot test if its too much data

even writing input is a tough call

PR #913, opened 2023-02-01
Add style sources stores

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/807

Here added style sources and style source selections stores, added updates along styles and covered with tests. Patch utilities for rest endpoint are not used to avoid collision with temporary logic in styles.

Next step will be integration styles with style sources on client.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/props-panel/props-panel.stories.tsx
+++ b/apps/designer/app/designer/features/props-panel/props-panel.stories.tsx
@@ -21,6 +21,7 @@ export const NoProps: ComponentStory<typeof PropsPanel> = () => {
   ]);
   return (
     <PropsPanel
+      treeId="treeId"
       selectedInstance={{
         type: "instance",
         id: "1",
@@ -37,6 +38,7 @@ export const RequiredProps: ComponentStory<typeof PropsPanel> = () => {
   propsStore.set([]);
   return (
     <PropsPanel
+      treeId="treeId"
       selectedInstance={{
         type: "instance",
         id: "1",
@@ -53,6 +55,7 @@ export const DefaultProps: ComponentStory<typeof PropsPanel> = () => {
   propsStore.set([]);
   return (
     <PropsPanel
+      treeId="treeId"
       selectedInstance={{
         type: "instance",
         id: "1",
@@ -80,6 +83,7 @@ export const AllProps: ComponentStory<typeof PropsPanel> = () => {
   );
   return (
     <PropsPanel
+      treeId="treeId"
       selectedInstance={{
         type: "instance",
         id: "3",
--- a/apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
+++ b/apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
@@ -1,19 +1,24 @@
 import { useCallback } from "react";
+import ObjectID from "bson-objectid";
 import store from "immerhin";
 import warnOnce from "warn-once";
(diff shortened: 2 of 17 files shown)


apps/designer/app/designer/features/props-panel/props-panel.stories.tsx
```diff
@@ -53,6 +55,7 @@ export const DefaultProps: ComponentStory<typeof PropsPanel> = () => {
   propsStore.set([]);
   return (
     <PropsPanel
+      treeId="treeId"
```
kof, line 58: this looks strange

kof: ah its a story

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
```diff
@@ -70,38 +80,59 @@ export const useStyleData = ({ selectedInstance, publish }: UseStyleData) => {
...
+          const instanceId = selectedInstance.id;
+          const breakpointId = selectedBreakpoint.id;
+
+          // crate style source and its selection on currently selected instance
+          let matchedStyleSourceSelection = styleSourceSelections.find(
+            (styleSourceSelection) =>
+              styleSourceSelection.instanceId === instanceId
+          );
+          if (matchedStyleSourceSelection === undefined) {
+            const styleSourceId = ObjectID.toString();
```
kof, line 95: do we want to keep using ObjectID or switch to uuid?

kof: I think we should also switch uuid to https://twitter.com/_ericelliott/status/1611214704554442752?t=8DUheU9KGq7mtDENA-nlwA&s=19

apps/designer/app/shared/tree-utils.ts
```diff
@@ -46,19 +53,131 @@ export const findSubtree = (
...
+    );
+    clonedStyleSourceSelections.push({
+      values,
+      instanceId,
+    });
+  }
+  return clonedStyleSourceSelections;
+};
+
+// @todo migrate to style source variant
```
kof, line 137: do you have an issue or todo in an issue?

packages/project/src/db/style-sources.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (canEdit === false) {
+    throw new Error("You don't have edit access to this project");
+  }
+
+  const build = await prisma.build.findUnique({
+    where: { id: buildId },
+  });
+  if (build === null) {
+    return null;
```
kof, line 28: If you don't have to use null, I wouldn't use it, since its our level of abstraction, we could avoid it, idk mb this case a special so up to you
```suggestion
    return;
```

kof (approved):
looking good, we are getting close!!!

istarkov (approved):

apps/designer/app/shared/tree-utils.ts
TrySound: This is for the next PR

apps/designer/app/designer/features/style-panel/shared/use-style-data.ts
TrySound: Looks like nanoid have most qualities
https://github.com/paralleldrive/cuid2#the-contenders

PR #919, opened 2023-02-02
A bit more text improvements

rpominov:
## Description

1. `text-decoration`, `text-transform`, and `text-indent` are now taken from Figma tokens as well
2. fixed the code that removes type from token name: `borderWidth0` -> `0`

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -75,6 +78,9 @@ export const typography = {
     fontSize: "12px",
     lineHeight: "16px",
     letterSpacing: "0.005em",
+    textTransform: "capitalize",
```
rpominov, line 81: Not sure this will work well, as `text-transform: capitalize` might be unreliable. But I think we should try and see, before doing something more complicated

TrySound (approved):

PR #925, opened 2023-02-02
Replace bson-objectid with nanoid

TrySound:
- bson-objectid is harder to use with ObjectId().toString()
- nanoid is safer alternative (https://github.com/paralleldrive/cuid2#the-contenders)
- auto import works better

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
I agree nanoid is great, but isn't cuid2 even better?

kof (approved):

istarkov:
IMO for db IDs we need to use uuid compatible format as databases like PostgreSQL hold them as 16byte integer. Should be faster than string but not a lot I think in the case of the index. 

In the past people wrote that better to use `K-Sortable/Sequential/Monotonically Increasing Ids` on DB ids, but having createdAt field it seems like not an issue modern days https://github.com/paralleldrive/cuid2#note-on-k-sortablesequentialmonotonically-increasing-ids

For local JOSN ids like above probability of collision is so low (we don't expect JSON will have like 100000000... ids) IMO most performant and small implementation should be the winner. If nanoid is faster better to use it.

Issue #928, opened 2023-02-02
API Errors throwing and handling.


istarkov:
Based on today's release issues. 

Data loader for saas KV storage https://github.com/webstudio-is/webstudio-designer/blob/97fb27c0ceeb979f22eed14567148aa16daedcd3/apps/designer/app/routes/rest/project.$projectId.tsx#L11

Can return errors as JSON string or as JSON object `{ errors: string[] }` at first it should be unified into a single error format. `Result` type must be exported. (the issue that then we `throw json()` it has Response type and doesn't converted into {errors} object)

Then here https://github.com/webstudio-is/webstudio-designer/blob/97fb27c0ceeb979f22eed14567148aa16daedcd3/packages/http-client/src/index.ts#L22

The whole function is strange. It doesn't take into consideration that `fetch` response can be not a JSON (service not available etc)
Inside catch it not rethrows an error but returns `error.message` and rethrows in a single case if it was project json with non 2xx status. 
Must be fixed into simple implementation. Throw if `response` it not ok. Read `text()` in case of non-2xx errors, read json only if `status` is ok.

Then after that we have code 

```ts
const pages = await loadProject({
  projectId,
  apiUrl: origin,
});
if (pages instanceof Error) {
  throw new Error(pages.message);
}
for (const page of pages) {
```
pages cant be an instance of Error, but it can be a string now and in that case we have `for (const page of pages) {`

What in our case gave error like 

```ts
  throw new Error(
    "Error happened. Please contact us to get help." + page
  );
```

As a fast fix. 

`project.$projectId.tsx` must throw `json` and response status must be non-200 for easier parsing.

i.e. 
```
    throw json(
      {
        errors: error instanceof Error ? error.message : String(error),
        env,
      },
      { status: 500 }
    );
```

http-client packet https://github.com/webstudio-is/webstudio-designer/blob/97fb27c0ceeb979f22eed14567148aa16daedcd3/packages/http-client/src/index.ts#L1

must be able to work with non-json error responses, and must rethrow all errors. The only right type it can return must be `Promise<Array<CanvasData>`

logic at saas should not throw strange errors like `Error happened. Please contact us to get help.`, `new Response("Error"` and and like this `"Error happened. Please contact us to get help."`

The best would be to return in response internal error.  I know that it's considered a bad practice to do so, as some implementation details can leak. But in the absense of normal logging best would be to return real errors and not hide them under `"Error happened. Please contact us to get help."`

[...]

PR #930, opened 2023-02-02
Unify and simplify fetch and errors for endpoints and packages used in SaaS

istarkov:
## Description

Unify and simplify fetch and errors for endpoints and packages used in SaaS
ref #928

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/http-client/src/index.ts
+++ b/packages/http-client/src/index.ts
@@ -22,5 +22,5 @@ export const loadProject = async ({
   // In the case where the status code is not 2xx,
   // we don't know what to do with the response other than just show an error message.
   const message = await projectResponse.text();
-  throw new Error(message.substring(0, 1000));
+  throw new Error(message.slice(0, 1000));
 };
--- a/packages/http-client/src/index.ts
+++ b/packages/http-client/src/index.ts
@@ -19,7 +19,7 @@ export const loadProject = async ({
     return await projectResponse.json();
   }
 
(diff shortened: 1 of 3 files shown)


packages/http-client/src/index.ts
```diff
@@ -7,23 +7,20 @@ export const loadProject = async ({
...
+  const projectResponse = await fetch(projectUrl);
+
+  if (projectResponse.ok) {
+    return await projectResponse.json();
+  }
+
+  // In the case where the status code is not 2xx,
+  // we don't know what to do with the response other than just show an error message.
+  const message = await projectResponse.text();
+  throw new Error(message.substring(0, 1000));
```
TrySound, line 25: slice :)

TrySound (approved):

packages/http-client/src/index.ts
kof, line 25: Are you trying to limit the message because we are posting entire stack trrace?

kof (approved):

packages/http-client/src/index.ts
istarkov: No, often it's HTML content of intermediate proxies, like Cloudflare 522 page or like. The first 1k symbols are enough to get what's happening. Otherwise you usually ends up with huuuge log entries.

PR #943, opened 2023-02-03
Topbar layout and buttons redesign

kof:
## Description

1. Topbar components replaced
2. Toolbar Toggle component added to the ds

https://user-images.githubusercontent.com/52824/216761913-1f801137-af36-400c-ad99-7b01c0e356d8.mp4

## Steps for reproduction

1. open designer

## Code Review

  - test it on preview

- [ ] @taylornowotny I need you to 
  - check for visual correctness 

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/toolbar-toggle.tsx
```diff
@@ -0,0 +1,48 @@
+import { styled, theme } from "../stitches.config";
```
rpominov, line 1: Please add a header similar to https://github.com/webstudio-is/webstudio-designer/blob/9726664bd8dd0f4c62f9605f075bf9564e2abff0/packages/design-system/src/components/button.tsx#L1-L4

packages/design-system/src/components/button.tsx
```diff
@@ -18,6 +18,7 @@ const colors = [
   "positive",
   "neutral",
   "ghost",
+  "dark",
 ] as const;
```
rpominov, line 22: Need to add it here too https://github.com/webstudio-is/webstudio-designer/blob/9726664bd8dd0f4c62f9605f075bf9564e2abff0/packages/design-system/src/components/button.stories.tsx#L17-L23

apps/designer/app/designer/features/breakpoints/trigger-button.tsx
```diff
@@ -41,21 +41,21 @@ export const TriggerButton = forwardRef<
...
   }
-  const variant = willRender(breakpoint, canvasWidth) ? "contrast" : "hint";
+  const variant = willRender(breakpoint, canvasWidth) ? "contrast" : "subtle";
 
   return (
-    <DeprecatedButton
+    <Button
       {...props}
       ref={ref}
       css={{ gap: theme.spacing[3] }}
```
rpominov, line 50: Shouldn't need this gap now

rpominov (approved):
Looks good otherwise

taylornowotny:
Looks great!

Only visual correction is to point out that the breakpoint button does not have an icon in the design. I'm considering whether we even want to represent breakpoints with device icons. Seems problematic for many reasons but this is a separate discussion.

I had written the feedback below but on second thought I'm not sure if it applies since we will be implementing an infinite canvas. You can read it anyways and see what you think. Regardless of the below feedback, I believe our breakpoint button should _not_ be flex-centered becuase it will shift when the left or right containers grow. Would be better to center it absolutely so it stays in the same place.

> I have some feedback on the layout but this can be part of another PR.
> 
> The breakpoint button and the canvas not aligned to each other. I didn't think of this before, but I think they should be, for aesthetic reasons.
> 
> It looks like currently the breakpoint button is centered within the toolbar, but since the containers on the left and right side are not the same size, this means the button will shift when the containers change size. For example if the network connection is lost, the network indicator will show and the right side container will grow wider. This would shift the button to the left.
> 
> Screenshot 2023-02-04 at 10 04 22 AM
> 
> Webflow
> Screenshot 2023-02-04 at 10 17 43 AM
> 
> I think it would look much nicer if the breakpoint button is aligned with the canvas.

taylornowotny (approved):

kof:
@taylornowotny deleted the icon, fixing centering of breakpoint button in a bit

PR #944, opened 2023-02-04
Add size prop to generated icons

TrySound:
We need two sizes for icons: 16px and 22px. New icons can be customized with `size={"small" | "big"}` prop where small is default.

If something else custom is necessary, width and height props still override  size.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Can we also support size as a number? so one can set a custom size instead of width/height

kof:
Actually not only as a number, but as string with unit

packages/icons/svgo-jsx.config.ts
```diff
@@ -33,8 +37,8 @@ export const config: Config = {
   template,
   transformFilename,
   svgProps: {
-    width: "16",
```
kof, line 36: Problem, see for example https://github.com/webstudio-is/webstudio-designer/pull/943/files#diff-feb10459fc59f9de0ebfcab6e8dd0ba63b0c3a7390489ffa967d55dcab1dea0f

play.svg

It is designed at 22px size, when we run the compiler, it sets mixed sizes on it, it ends up partially 16 and partially 22, I think we set width/height but not viewport or something

kof: I am not sure I can just set 16 on that icon manually, it will probably break icons that are designed at 22

TrySound: > It is designed at 22px size

So do we want to have a separate set of icons for topbar and the rest of the app?

TrySound: > I think we set width/height but not viewport or something

and it's fine. If we change viewport we'll need to scale all paths in icon

TrySound: > I am not sure I can just set 16 on that icon manually, it will probably break icons that are designed at 22

they will be scaled

kof: > So do we want to have a separate set of icons for topbar and the rest of the app?

Separate set yes, toolbar only or not not sure, I guess its not just toolbar

kof: taylor informed me yesterday that this is basically a standard to have separate sets for different purposes

TrySound: Let's generate separate set then

kof:
We discussed with @taylornowotny that we actually need a separate set of 22px icons, I would even call them 16 and 22 instead of small/large

kof:
Currently play and offline are the icons that are from 22px set and have manually corrected size in generated code, so when you run generate you see them updated

kof:
Found another one: subscript is using 48px size, something is wrong with this one, needs to be checked with figma

kof:
webstudio.svg is a logo and has width="280" height="210" so I don't even know what to do with this one, this one will require manual size definition anyways in the end

kof:
Hmm, maybe its alright to have `viewBox="0 0 22 22" width="16" height="16"` and doesn't cause any problem at all? viewbox corresponds actual content, width/height can be set to 16 by default and one can scale it anywhere?

Maybe you were on the right path all along, I was just not clearly understanding there is no issue with this?

TrySound:
Yeah, maybe. Though the size should be set anyway.

kof:
Lets try out in this PR using size prop and setting the 22px icons to size={22} using prop in consumer, currently just 2 icons, if that works, we can just keep doing what we doing, we would just make it easier a bit to set the proper size using a size prop in addition to width/height

packages/icons/src/types.ts
```diff
@@ -7,6 +7,7 @@ import type {
 export interface IconProps extends SVGAttributes<SVGElement> {
   children?: never;
   color?: string;
+  size?: "small" | "big";
```
kof, line 10: proposal, instead using small/big, lets literally use 16/22 those are like names

kof: actually wait we don't even need the enum, just use a number and set it for both width and hight

kof: 16 by default

kof (approved):

PR #945, opened 2023-02-04
Refactor closest droppable with instances index

TrySound:

A lot of our utilities can reuse indexed instances instead of retraversing tree. This will also simplify migration to normalized instances.

Here I refactored droppable instance lookup.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/project/src/shared/tree-utils/insert-instance.ts
```diff
@@ -8,8 +8,11 @@ export type InstanceInsertionSpec = {
 export const insertInstanceMutable = (
   rootInstance: Instance,
   instance: Instance,
-  spec: InstanceInsertionSpec
+  spec: undefined | InstanceInsertionSpec
```
kof, line 11: could be just optional

kof (approved):

PR #946, opened 2023-02-04
Fix double delete when focused navigator tree

TrySound:
Looks like calling onDelete in navigator duplicates global delete shortcut.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov (approved):
Didn't realise that. Less code 😅

PR #950, opened 2023-02-06
Add `data-state="open"` support to Button

rpominov:
## Description

When Button is used as a trigger for something that can be open or closed, Radix passes a `data-state` property which we now support:

```js
<Popover>
  <PopoverTrigger asChild>
    {/* Button will have different state depending on Popover state */}
    <Button>Open</Button>
  </PopoverTrigger>
  ...
</Popover>
```

Discussed in https://discord.com/channels/955905230107738152/1071187564788207717

## Code Review

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/button.tsx
```diff
@@ -51,31 +51,31 @@ const perColorStyle = (variant: ButtonColor) => ({
   background: variant === "ghost" ? "transparent" : backgrounds[variant],
   color: foregrounds[variant],
 
-  "&[data-button-state=auto]:hover, &[data-button-state=hover]": {
+  "&[data-state=auto]:hover, &[data-state=hover]": {
```
rpominov, line 54: Now that we handle `data-state` prop explicitly rather than just passing it through it safe to use it here

packages/design-system/src/components/button.tsx
```diff
@@ -125,25 +125,43 @@ type ButtonProps = {
...
       suffix,
       children,
+      "data-state": dataState,
       ...restProps
     }: ButtonProps,
     ref: Ref<HTMLButtonElement>
   ) => {
+    // when button is used as a trigger for something that opens
+    // <SomeTrigger asChild><Button /></SomeTrigger>
+    let finalState = dataState === "open" ? "pressed" : undefined;
```
kof, line 148: you are mapping it to pressed, but is it supposed to be pressed state? was it discussed with Taylor?

kof (approved):

packages/design-system/src/components/button.tsx
rpominov: Yes, discussed and Taylor documented this in Figma

PR #951, opened 2023-02-06
Assets view mode

istarkov:
## Description

In view mode, we need to prevent users from deleting / creating assets in UI. (Server-side protection already done in the previous PR)

The other way can be just to hide buttons from the UI.
Also tooltip duration for disabled actions can be smaller.

## Steps for reproduction

Share the Link with View permissions.
Open that link in incognito mode.
Try to create/delete an asset

image

image

image

image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/shared/fonts-manager/item-menu.tsx
+++ b/apps/designer/app/designer/shared/fonts-manager/item-menu.tsx
@@ -7,10 +7,12 @@ import {
   DeprecatedText2,
   DropdownMenuPortal,
   styled,
+  Tooltip,
 } from "@webstudio-is/design-system";
 import { MenuIcon } from "@webstudio-is/icons";
 import { type FocusEventHandler, useState, useRef, useEffect } from "react";
 import { theme } from "@webstudio-is/design-system";
+import { useAuthPermit } from "~/shared/nano-states";
 
 const MenuButton = styled(DeprecatedIconButton, {
   color: theme.colors.hint,
@@ -42,6 +44,13 @@ const ItemMenu = ({
     };
   }, []);
 
+  const [authPermit] = useAuthPermit();
+
+  const deleteDisabled = authPermit === "view";
+  const tooltipContent = deleteDisabled
+    ? "View mode. You can't delete assets."
+    : undefined;
+
   return (
     <DropdownMenu
       open={isOpen}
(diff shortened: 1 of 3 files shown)


apps/designer/app/designer/shared/fonts-manager/item-menu.tsx
```diff
@@ -42,6 +44,13 @@ const ItemMenu = ({
     };
   }, []);
 
+  const [authPermit] = useAuthPermit();
+
+  const deleteDisabled = authPermit === "view";
+  const tooltipContent = deleteDisabled
+    ? "View mode. You can't delete assets."
```
kof, line 51: probably @taylornowotny

kof: sorry meant to say we should let someone like @taylornowotny with native English check the wording

apps/designer/app/designer/shared/fonts-manager/item-menu.tsx
```diff
@@ -42,6 +44,13 @@ const ItemMenu = ({
     };
   }, []);
 
+  const [authPermit] = useAuthPermit();
+
+  const deleteDisabled = authPermit === "view";
```
kof, line 49: I try to use is/has/can for all booleans for clarity

```suggestion
  const isDeleteDisable = authPermit === "view";
```

apps/designer/app/designer/shared/fonts-manager/item-menu.tsx
```diff
@@ -70,17 +79,20 @@ const ItemMenu = ({
...
+              disabled={deleteDisabled}
+              onClick={(event) => {
+                // Prevent setting the current font to the item.
+                event.stopPropagation();
+              }}
+              onSelect={() => {
+                onDelete();
+              }}
+            >
+              <DeprecatedText2>Delete font</DeprecatedText2>
```
kof, line 93: if its not too much work, please replace deprecated components with new once as you see them, you tested the UI, you know what's supposed to happen, for you it should be couple of minutes.

taylornowotny:
This is good, thank you for thinking about this.

In the case of Assets, I think we should show them in "view" mode rather than hiding them because they may be relevant to some troubleshooting effort, which is one of the use-cases for "view" mode. The upload and delete buttons should be disabled - I agree.

PR #969, opened 2023-02-08
Clone assets

### LINKED ISSUE #872: Assets handling in a cloned project
- [x] Show assets in the cloned project (currently we don't show them)
- [x] Ensure deleting an asset in the original project doesn't delete it from the cloned one
- [x]  Ensure deleting an asset in the cloned project doesn't delete it in the original one

rpominov:
## Description

Closes #872 . Now when you clone a project assets are cloned too.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/src/db/clone.ts
+++ b/packages/asset-uploader/src/db/clone.ts
@@ -1,0 +1,47 @@
+import { prisma, type Prisma, type Project } from "@webstudio-is/prisma-client";
+import {
+  authorizeProject,
+  type AppContext,
+} from "@webstudio-is/trpc-interface/server";
+
+export const cloneAssets = async (
+  {
+    fromProjectId,
+    toProjectId,
+    dontCheckEditPermission = false,
+  }: {
+    fromProjectId: Project["id"];
+    toProjectId: Project["id"];
+    dontCheckEditPermission?: boolean;
+  },
+  context: AppContext,
+  client: Prisma.TransactionClient = prisma
+) => {
+  const canView = await authorizeProject.hasProjectPermit(
+    { projectId: fromProjectId, permit: "view" },
+    context
+  );
+
+  if (canView === false) {
+    throw new Error("You don't have access to this project assets");
+  }
+
+  if (dontCheckEditPermission === false) {
+    const canEdit = await authorizeProject.hasProjectPermit(
+      { projectId: toProjectId, permit: "edit" },
+      context
+    );
+
+    if (canEdit === false) {
+      throw new Error("You don't have access to create this project assets");
+    }
+  }
+
+  const assets = await client.asset.findMany({
+    where: { projectId: fromProjectId, status: "UPLOADED" },
+  });
+
+  await client.asset.createMany({
+    data: assets.map((asset) => ({ ...asset, projectId: toProjectId })),
+  });
+};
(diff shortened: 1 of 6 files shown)


packages/project/src/db/project.ts
```diff
@@ -204,25 +205,40 @@ const clone = async (
...
       context,
       client
     );
 
+    await cloneAssets(
+      {
+        fromProjectId: project.id,
+        toProjectId: newProjectId,
+
+        // Permission check on newProjectId will fail until this transaction is committed.
```
istarkov, line 235: I would add comment like this into function definition too using /** **/ so it will be visible in vscode

istarkov, line 235: Will fail in the open-source version because in open source we take the owner from the Project table, and permission check is executed in other transaction.

istarkov (approved):

packages/project/src/db/project.ts
rpominov: Not sure I understand. 

`dontCheckEditPermission: true` is here specifically to workaround this issue

rpominov: Added

rpominov: Or do you mean that the code comment should say that we have this issue only in open source version?

packages/asset-uploader/src/db/clone.ts
```diff
@@ -0,0 +1,53 @@
...
+    dontCheckEditPermission = false,
+  }: {
+    fromProjectId: Project["id"];
+    toProjectId: Project["id"];
+
+    /**
+     * Don't use unless absolutely have to (e.g. because of transactions)
+     * and unless it's obvious on the call site that permission is checked
+     */
+    dontCheckEditPermission?: boolean;
```
kof, line 20: why would you need this flag instead of just having the right permissions for the action?

packages/project/src/db/project.ts
```diff
@@ -204,25 +205,40 @@ const clone = async (
       ? await db.build.loadByProjectId(project.id, "dev")
       : await db.build.loadByProjectId(project.id, "prod");
 
-  const projectId = uuid();
-  authorizeProject.registerProjectOwner({ projectId }, context);
+  const newProjectId = uuid();
```
kof, line 208: I think we recently switched to using nanoid everywhere, didn't we @TrySound ?

packages/asset-uploader/src/db/clone.ts
rpominov: See further in diff where we call `cloneAssets`: https://github.com/webstudio-is/webstudio-designer/pull/969/files#diff-3130be2fe9639006f0fd0765501c1b1415bfb9952e34ea773bcb0400a02e22efR230

`hasProjectPermit()` returns `false` until project is created in DB

packages/prisma-client/prisma/schema.prisma
```diff
@@ -28,7 +28,7 @@ enum UploadStatus {
 }
 
 model Asset {
-  id          String       @default(uuid())
+  id          String       @default(uuid()) // not unique!
   projectId   String
```
TrySound, line 32: [id, projectId] is unique and project id is mandatory

packages/project/src/db/project.ts
TrySound: Nanoid for json data and uuid for db rows.

kof: why not using one for both? nanoid is apparently even better, no?

kof (approved):

PR #976, opened 2023-02-09
Represet style source selections as map

TrySound:

Patches by indexes are replaced with patches by keys so we could get almost conflict free collaboration.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project-build/src/schema/style-source-selections.ts
+++ b/packages/project-build/src/schema/style-source-selections.ts
@@ -1,0 +1,22 @@
+import { z } from "zod";
+
+const InstanceId = z.string();
+
+const StyleSourceId = z.string();
+
+export const StyleSourceSelection = z.object({
+  instanceId: InstanceId,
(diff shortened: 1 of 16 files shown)


packages/project-build/src/schema/style-source-selections.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from "zod";
+
+const InstanceId = z.string();
```
kof, line 3: Shouldn't we use it from the original place where we have instance id?

kof (approved):

packages/project-build/src/schema/style-source-selections.ts
TrySound: I'd prefer to not export these values separately and Instance.id thing not possible. So let's keep this way. It's more like documentation than any functionality.

PR #977, opened 2023-02-09
Refactor instance reparenting with instances index

TrySound:

Index allows us to pretraverse the tree and provide utils for fast access anything in the tree.

Also removed old implementation from project package.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/shared/tree-utils.ts
```diff
@@ -100,6 +100,48 @@ export const findClosestDroppableTarget = (
...
+  ) {
+    nextPosition -= 1;
+  }
+
+  prevParent.children.splice(prevPosition, 1);
+  if (nextPosition === "end") {
+    nextParent.children.push(instance);
+  } else {
+    nextParent.children.splice(nextPosition, 0, instance);
+  }
```
rpominov, line 142: It seems like this function makes passed `instancesIndex` invalid: modifies the tree, but doesn't update the Maps.

Currently this seem fine, because we create a fresh temporary index every time we apply the function. But I assume we'll have a re-usable index at some point.

rpominov (approved):
Looks good otherwise

apps/designer/app/shared/tree-utils.ts
TrySound: Yes, once we switch to normalized instances we gonna also modify instancesById map.

TrySound: Reusable index is recreated on every store update btw

kof (approved):

PR #983, opened 2023-02-10
Replace findInstanceById with instances index

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/696 https://github.com/webstudio-is/webstudio-designer/issues/651

Got rid of findInstanceById in a few places, fixed dnd to find instances by data attribute instead of element id.

findInstanceById still used in other project utility, will get there later.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/designer/app/designer/shared/tree/index.tsx
```diff
@@ -55,8 +57,22 @@ const instanceRelatedProps = {
...
 
 export const InstanceTree = (
-  props: Omit<TreeProps<Instance>, keyof typeof instanceRelatedProps>
-) => <Tree {...props} {...instanceRelatedProps} />;
+  props: Omit<
+    TreeProps<Instance>,
+    keyof typeof instanceRelatedProps | "findItemById"
+  >
+) => {
+  const { instancesById } = useStore(instancesIndexStore);
```
rpominov, line 65: Not a big deal, but this component is a bit weird now because it takes `rootInstance` as a prop, but takes `instancesById` directly from store.

I'd cleaned this up: either both from props or both from store

apps/designer/app/designer/shared/tree/index.tsx
```diff
@@ -55,8 +57,22 @@ const instanceRelatedProps = {
...
+  const findItemById = useCallback(
+    (_rootInstance: Instance, instanceId: Instance["id"]) => {
```
rpominov, line 67: Looking at `_rootInstance` here makes me realise that `TreeProps<Data>` is a bit wrong. It should be like this:

```js
export type TreeProps<Root, Data extends { id: string }> = {
  root: Root;

  canLeaveParent: (item: Data) => boolean;
  canAcceptChild: (item: Data) => boolean;
  findItemById: (root: Root, id: string) => Data | undefined;
  getItemPath: (root: Root, id: string) => Data[];
  getItemPathWithPositions: (
    root: Root,
    id: string
  ) => Array<{ item: Data; position: number }>;
  getItemChildren: (item: Data) => Data[];
  renderItem: (props: TreeItemRenderProps<Data>) => React.ReactNode;

  selectedItemId?: string;
  onSelect?: (itemId: string) => void;
  onHover?: (item: Data | undefined) => void;
  animate?: boolean;
  onDragEnd: (event: {
    itemId: string;
    dropTarget: { itemId: string; position: number | "end" };
  }) => void;
};
```

Then you could have whatever as root, in this case I'd make `{rootInstance, instancesById}` the root object. And then all these functions like `findItemById` can be defined outside of render again.

Up to you whether to refactor this.

apps/designer/app/canvas/shared/use-drag-drop.ts
```diff
@@ -53,6 +50,16 @@ export type DragEndPayload = {
...
 
+const getInstanceElementById = (id: Instance["id"]) => {
+  return document.querySelector(`[${idAttribute}="${id}"]`);
+};
+
+const getInstanceIdFromElement = (
+  element: Element
+): Instance["id"] | undefined => {
+  return element.getAttribute(idAttribute) ?? undefined;
+};
```
rpominov, line 61: I'd keep `findInstanceByElement, getInstanceElementById, getInstanceIdFromElement` in utils.

Would only change `findInstanceByElement`'s argument from `rootInstance` to `instancesById`

rpominov (approved):

apps/designer/app/canvas/shared/use-drag-drop.ts
TrySound: Do we need them In utils if never reuse?

rpominov: Oleg asked to put in utils originally. But I kinda agree that it makes sense to have these little self contained potentially re-usable pieces of code in utils

apps/designer/app/designer/shared/tree/index.tsx
TrySound: Will refactor this way when figure out all case

Issue #213, opened 2022-08-04
Use context to setup everything global


kof:
We have a number of things that rely on module scope, which makes us need to reset those on every SSR render, but instead we should focus on creating new values/configs and providing them over context instead of resetting it. 

Resetting between renders is a bad pattern that can easily lead to memory leaks as we grow the code base.

- react-sdk/stitches/css.ts - should set everything up and provide over the context
- usePopulateRootInstance - sets the rootInstanceContainer depending on the ref, but instead we should have the container provided over context
- everything that calls `createValueContainer()`
- useAllUserProps()
- useInitializeBreakpoints()

kof:
cc @TrySound I think we still have SSR implemented wrong in all cases that involve atoms that are being initialized once and then reused for each request

kof:
this was done, no? if not would be great to have a checkbox-based todo list

PR #1005, opened 2023-02-11
Resolve props assets on client

TrySound:
Before we checked looked for all asset ids in props, loaded assets and replaced ids with resolved assets. This is all on server. We faced a few problems with it.

- requires to maintain two versions of types: stored in db and resolved for client
- easy to break project by writing to db wrong version

Here I provide to sdk both props and assets and resolve on client with almost same amount of code.

Also replaced store setter with context, see https://github.com/webstudio-is/webstudio-designer/issues/213

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/canvas.tsx
+++ b/apps/designer/app/canvas/canvas.tsx
@@ -9,7 +9,7 @@ import {
   customComponentsMeta,
   setParams,
   type OnChangeChildren,
-  setPropsByInstanceIdStore,
+  ReactSdkProvider,
 } from "@webstudio-is/react-sdk";
 import { publish } from "~/shared/pubsub";
 import { registerContainers, useCanvasStore } from "~/shared/sync";
@@ -21,6 +21,7 @@ import { useTrackSelectedElement } from "./shared/use-track-selected-element";
 import { WrapperComponentDev } from "./features/wrapper-component";
 import {
   propsIndexStore,
+  assetsStore,
   rootInstanceContainer,
   useBreakpoints,
   useRootInstance,
@@ -44,6 +45,11 @@ import { customComponents } from "./custom-components";
 import { useHoveredInstanceConnector } from "./hovered-instance-connector";
 
 registerContainers();
+
+const propsByInstanceIdStore = computed(
+  propsIndexStore,
+  (propsIndex) => propsIndex.propsByInstanceId
+);
 
 const useElementsTree = () => {
   const [rootInstance] = useRootInstance();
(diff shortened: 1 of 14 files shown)


apps/designer/app/canvas/canvas.tsx
```diff
@@ -73,12 +79,21 @@ const useElementsTree = () => {
...
     }
 
-    return createElementsTree({
-      sandbox: true,
-      instance: rootInstance,
-      Component: WrapperComponentDev,
-      onChangeChildren,
-    });
+    return (
+      <ReactSdkProvider
```
kof, line 83: This makes canvas fully coupled to react SDK ... how is this supposed to work some day with a different framework?

TrySound: It is coupled anyway. With other framework we'll use another way to inject stores.

kof: how is it coupled anyway? we are trying to remove every coupling in every step we do ...
I think react provider that has to be provided is not the ideal way of doing it, SDK should provide something agnostic so we don't have to assume on canvas that its already using react.

there is a bunch of options

TrySound: Renderer is full react. It won't be reused in any way. We will reimplement it in other frameworks.

TrySound: I can put provider inside createElementsTree if you want

kof: Anything works for me that doesn't create more coupling with react in SDK. Currently you require SDK to provide this provider

TrySound: done

kof (approved):
I don't like how we do SDK stuff right now in general, still way too much is put in there and creating a separate SDK in a new framework is currently practically impossible 🤷‍♂️

Just food for thoughts

TrySound:
Almost everything in there is renderer and components. Isn't it what supposed to be there?

kof:
Yeah, the problem is more like how do we expect someone writing an SDK for a different framework e.g. vue. There is quite a bit of stuff they have to implement that has to do with the designer. Ideal situation is that they just provide components and are done.

packages/react-sdk/src/props.ts
```diff
@@ -50,20 +42,24 @@ export const useInstanceProps = (instanceId: Instance["id"]) => {
 // this utility is be used for image component in both designer and preview
 // so need to optimize rerenders with computed
 export const usePropAsset = (instanceId: Instance["id"], name: string) => {
+  const { propsByInstanceIdStore, assetsStore } = useContext(ReactSdkContext);
   const assetStore = useMemo(() => {
-    return computed(propsByInstanceIdStore, (propsByInstanceId) => {
-      const instanceProps = propsByInstanceId.get(instanceId);
-      let asset: undefined | Asset = undefined;
-      if (instanceProps) {
+    return computed(
```
istarkov, line 47: I may be wrong, but if we create a computed atom inside the hook, it subscribes to other atoms holding references. 

Any chance that this unmount is called?
https://github.com/nanostores/nanostores/blob/main/computed/index.js#L26

istarkov (approved):
Except I have feeling that creating computed inside hook can cause memory leak, everything else is ok

Issue #1002, opened 2023-02-10
Shortcuts for zoom -+ react when user types them in CssValueInput


kof:
When shortcut uses cmd or ctrl modifiers they don't actually type in inputs, so we don't need to prevent them like in case with undo/redo.

But we also have a shortcuts which don't use modifiers or use "option" modifier which still results in typing in inputs.

Now if you type  =- in inputs we also end up zooming in/out

PR #1006, opened 2023-02-11
Sync zoom and modify with shared shortcuts

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/1002

Synchronizing shortcuts with pubsub is complicated because we need to configure each case. Here instead I synchronized zoom store and run shortcuts on both designer and canvas. This way whole logic tied to state instead of event.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/canvas/shared/use-shortcuts.ts
+++ b/apps/designer/app/canvas/shared/use-shortcuts.ts
@@ -15,7 +15,6 @@ declare module "~/shared/pubsub" {
     cancelCurrentDrag: undefined;
     openBreakpointsMenu: undefined;
     selectBreakpointFromShortcut: number;
-    zoom: "zoomOut" | "zoomIn";
   }
 }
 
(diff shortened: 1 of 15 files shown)


kof (commented):
Works, can we please also add "ctrl +-"? I just realized that this one prevents typing but allows zooming, so one can zoom while being in the focused input, otherwise one has to unfocus first.

cmd+ is already used for browse-level zoom and we don't want to break this complete layout zoom feature

TrySound:
ctrl won't work for windows users, maybe alt?

kof:
alt/option is fine as well, why is control not working? system one like cmd++?

TrySound:
Yes, "windows" keyboards don't have command. ctrl is used instead for same use cases.

TrySound:
Hm, with alt looks like there is an issue too. It types character.

TrySound:
ok, let it be control. We can address this when windows users give feedback.

kof:
> ok, let it be control. We can address this when windows users give feedback.

Actually using control is same problem as using cmd on mac - it overrides native zoom behavior, so people can't actually increase font size of the interface, which is very bad, we had this shortcut before and switched to just +- without modifiers

kof:
I am assuming ctrl++- is the windows zoom shortcut for native browser zoom, isn't it?

TrySound:
Yep. Figma solve this by overriding browser zoom

kof:
Discussed in the chat, both of 2 worlds:
1. keep native shortcut
2. support shift+modifier so that it works in inputs
3. support +- without modifiers in non-inputs

TrySound:
It's done

kof (approved):

Issue #994, opened 2023-02-10
Variable font manrope only has light font weight


kof:
We need to see how a variable font needs to be classified in terms of font weights and other options.

PR #1007, opened 2023-02-11
Support for variable fonts

kof:
Closes https://github.com/webstudio-is/webstudio-designer/issues/994

## Description

Variable fonts were treated like normal fonts, so we were only showing one default font weight as available.
Now we not only detect variable fonts, but we also save all the meta data we get from them, so we can add a variable font settings UI in the future, which is a huge deal for designers, because variable fonts offer very interesting styling options and animations

## Steps for reproduction

1. select instance
2. font family, upload variable versions fonts like Monrope, Inter or Roboto Flex
3. see you can choose any font weight
4. use any styles on the font e.g. italic etc 

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
+++ b/apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
@@ -1,0 +1,48 @@
+import { describe, test } from "@jest/globals";
+import { FontAsset } from "@webstudio-is/asset-uploader";
+import { nanoid } from "nanoid";
+import { isSupportedFontWeight } from "./is-supported-font-weight";
+
+const createServerAsset = (meta: FontAsset["meta"]): FontAsset => ({
+  id: nanoid(),
+  type: "font" as const,
+  name: "test",
+  location: "FS",
+  projectId: "id",
+  size: 2135,
+  format: "ttf",
+  createdAt: "",
+  description: "",
+  meta,
+  path: "",
+});
+
+describe("isSupportedFontWeight", () => {
+  test("static", () => {
+    const meta = {
+      family: "Acumin Pro",
+      style: "italic",
+      weight: 900,
+    } as const;
+    const asset = createServerAsset(meta);
+    expect(isSupportedFontWeight(asset, "400", "Roboto")).toBe(false);
+    expect(isSupportedFontWeight(asset, "400", "Acumin Pro")).toBe(false);
+    expect(isSupportedFontWeight(asset, "900", "Roboto")).toBe(false);
+    expect(isSupportedFontWeight(asset, "900", "Acumin Pro")).toBe(true);
+  });
+
+  test("variable", () => {
+    const meta = {
+      family: "Roboto Flex",
+      variationAxes: {
+        wght: { name: "wght", min: 100, default: 400, max: 1000 },
+      },
+    } as const;
+    const asset = createServerAsset(meta);
+    expect(isSupportedFontWeight(asset, "1", meta.family)).toBe(false);
+    expect(isSupportedFontWeight(asset, "100", meta.family)).toBe(true);
+    expect(isSupportedFontWeight(asset, "500", meta.family)).toBe(true);
+    expect(isSupportedFontWeight(asset, "1000", meta.family)).toBe(true);
+    expect(isSupportedFontWeight(asset, "1001", meta.family)).toBe(false);
+  });
+});
--- a/apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
+++ b/apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
@@ -1,11 +1,10 @@
 import { describe, test } from "@jest/globals";
 import { FontAsset } from "@webstudio-is/asset-uploader";
-import { nanoid } from "nanoid";
 import { isSupportedFontWeight } from "./is-supported-font-weight";
 
 const createServerAsset = (meta: FontAsset["meta"]): FontAsset => ({
-  id: nanoid(),
-  type: "font" as const,
+  id: "111",
+  type: "font",
   name: "test",
   location: "FS",
   projectId: "id",
--- a/apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
+++ b/apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
@@ -1,0 +1,30 @@
+import type { Asset } from "@webstudio-is/asset-uploader";
+
+export const isSupportedFontWeight = (
+  asset: Asset,
+  weight: string,
+  currentFamily: string
+) => {
+  const isCurrentFamily =
+    "meta" in asset &&
+    "family" in asset.meta &&
+    asset.meta.family === currentFamily;
+
+  if (isCurrentFamily === false) {
+    return false;
+  }
+
+  const weightNumber = Number(weight);
+  const { meta } = asset;
+
+  if ("variationAxes" in meta) {
+    const { variationAxes } = meta;
+    return (
+      variationAxes.wght !== undefined &&
+      variationAxes.wght.min <= weightNumber &&
+      variationAxes.wght.max >= weightNumber
+    );
+  }
+
+  return "weight" in meta && meta.weight === weightNumber;
+};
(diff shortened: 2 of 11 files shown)


apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const isSupportedFontWeight = (
+  asset: Asset,
+  weight: string,
+  currentFamily: string
+) => {
+  const isCurrentFamily =
+    "meta" in asset &&
+    "family" in asset.meta &&
+    asset.meta.family === currentFamily;
```
TrySound, line 11: `asset.meta?.family === currentFamily` ?

apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
```diff
@@ -0,0 +1,30 @@
...
+  if ("variationAxes" in meta) {
+    const { variationAxes } = meta;
+    return (
+      variationAxes.wght !== undefined &&
+      variationAxes.wght.min <= weightNumber &&
+      variationAxes.wght.max >= weightNumber
+    );
+  }
+
+  return "weight" in meta && meta.weight === weightNumber;
```
TrySound, line 29: `return meta.weight === weightNumber`

packages/css-engine/src/core/rules.ts
```diff
@@ -142,9 +142,20 @@ export class FontFaceRule {
     this.options = options;
   }
   get cssText() {
-    const { fontFamily, fontStyle, fontWeight, fontDisplay, src } =
-      this.options;
-    return `@font-face {\n  font-family: ${fontFamily}; font-style: ${fontStyle}; font-weight: ${fontWeight}; font-display: ${fontDisplay}; src: ${src};\n}`;
+    const decls = [];
+    let property: keyof FontFaceOptions;
+    for (property in this.options) {
```
TrySound, line 147: Each option can be handled explicitly without loop

packages/fonts/src/font-data.ts
```diff
@@ -51,24 +51,43 @@ export const normalizeFamily = (family: string, subfamily: string) => {
...
+    };
+  }
+
   return {
     format,
     family,
-    ...parsedSubfamily,
+    variationAxes,
+    ...parseSubfamily(subfamily),
   };
```
TrySound, line 92:
```
if (Object.keys(font.variationAxes).length !== 0) {
  return {
    variationAxes: font.variationAxes
    ...
  }
}
return {
  variationAxes: undefined
  ...
}
```

packages/fonts/src/get-font-faces.ts
```diff
@@ -9,13 +9,33 @@ export type PartialFontAsset = {
...
+  fontStyle?: FontMetaStatic["style"];
+  fontWeight?: number | string;
+  fontStretch?: string;
 };
 
-const formatFace = (asset: PartialFontAsset, format: string) => {
+const formatFace = (asset: PartialFontAsset, format: string): FontFace => {
+  if ("variationAxes" in asset.meta) {
+    const { variationAxes } = asset.meta;
+    const wght = "wght" in variationAxes ? variationAxes.wght : undefined;
```
TrySound, line 22: `const {wght} = variationAxes`?

packages/fonts/src/get-font-faces.ts
```diff
@@ -25,8 +45,12 @@ const formatFace = (asset: PartialFontAsset, format: string) => {
   };
 };
 
-const getKey = (asset: PartialFontAsset) =>
-  asset.meta.family + asset.meta.style + asset.meta.weight;
+const getKey = (asset: PartialFontAsset) => {
+  if ("variationAxes" in asset.meta) {
+    return asset.meta.family + Object.values(asset.meta.variationAxes);
```
TrySound, line 50: Let's call .join explicitly

apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
kof: doesn't work)))

apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
kof: same problem, ts doesn't allow

kof: feel free to commit if you want to figure it out

packages/css-engine/src/core/rules.ts
kof: yep, turned out same LOC

apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.ts
kof: the problem there is that meta can be variable font meta object that has no weight and ts doesn't know which one it is for some reason, even though the check above should have taught it at this point

kof: Oh I figured it out, it was confused because asset can also be image asset

apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, test } from "@jest/globals";
+import { FontAsset } from "@webstudio-is/asset-uploader";
+import { nanoid } from "nanoid";
+import { isSupportedFontWeight } from "./is-supported-font-weight";
+
+const createServerAsset = (meta: FontAsset["meta"]): FontAsset => ({
+  id: nanoid(),
+  type: "font" as const,
```
TrySound, line 8: as const not necessary with return type

packages/css-engine/src/core/rules.ts
```diff
@@ -142,9 +142,19 @@ export class FontFaceRule {
     this.options = options;
   }
   get cssText() {
+    const decls = [];
     const { fontFamily, fontStyle, fontWeight, fontDisplay, src } =
       this.options;
-    return `@font-face {\n  font-family: ${fontFamily}; font-style: ${fontStyle}; font-weight: ${fontWeight}; font-display: ${fontDisplay}; src: ${src};\n}`;
+    decls.push(
+      `font-family: ${
+        /\s/.test(String(fontFamily)) ? `"${fontFamily}"` : fontFamily
```
TrySound, line 150: Why cast to string? Type is not sufficient?

packages/fonts/src/get-font-faces.ts
```diff
@@ -9,13 +9,31 @@ export type PartialFontAsset = {
...
-const formatFace = (asset: PartialFontAsset, format: string) => {
+const formatFace = (asset: PartialFontAsset, format: string): FontFace => {
+  if ("variationAxes" in asset.meta) {
+    const { wght, wdth } = asset.meta?.variationAxes ?? {};
+    const fontStretch = wdth
+      ? { fontStretch: `${wdth.min}% ${wdth.max}%` }
+      : undefined;
+    const fontWeight = wght
+      ? { fontWeight: `${wght.min} ${wght.max}` }
+      : undefined;
```
TrySound, line 27: Seems like `in` vs `undefined` a little complicates the logic. Maybe stick to always undefined?

kof: we don't want the property to be in there defined at all, unless we provide a default

packages/css-engine/src/core/rules.ts
kof: true, not necessary after rewrite

packages/fonts/src/get-font-faces.ts
kof: I could set fontWeight: normal by default, but is it worth specifying meaningless default and publishing it to all sites instead of having this?

apps/designer/app/designer/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
kof: correct, was fixed by a type and forgotten

packages/fonts/src/get-font-faces.ts
kof: oh you meant that its only there in the data structure but not rendered in the end, hmm, you are probably right

kof: fixed

TrySound (approved):

Issue #1019, opened 2023-02-13
Preset styles not rendered on a published site


kof:
Repro

1. Create a new site
2. publish
3. go to published site
4. open dev tools, see Body has no margin: 0 for example

On canvas its rendered correctly

TrySound:
All works now

PR #1021, opened 2023-02-13
Add utility to generate css text from data

TrySound:
Ref https://github.com/webstudio-is/webstudio-designer/issues/1019

We do a lot of changes with styles and ofte forget to keep saas up to date. So instead whole logic to generate css will be hidden in shared utility.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/project/src/shared/styles/css.ts
```diff
@@ -0,0 +1,54 @@
...
+import { addGlobalRules } from "./global-rules";
+import { getStyleRules } from "./style-rules";
+
+type Data = {
+  assets: Asset[];
+  build?: Build;
+  tree?: Tree;
+};
+
+export const generateCssText = (data: Data) => {
```
istarkov, line 19: So we can import it into SaaS and be happy

istarkov (approved):

kof (approved):

packages/project/src/shared/styles/css.ts
kof: I thought we already did, this is why we originally created it

Issue #992, opened 2023-02-10
Props panel: Basics


rpominov:
- [x] Understand and clean up the logic behind properties that are always visible. For example, for `Link` `href = ""` [here](https://github.com/webstudio-is/webstudio-designer/blob/main/packages/react-sdk/src/components/link.tsx/#L6) makes it visible by default which is weird and fragile
- [x] Find a way to associate a specific control type with a prop. An `if` like [here](https://github.com/webstudio-is/webstudio-designer/blob/13afe43cc5f3b69fb5bbc84bf96fbe3689e29fba/apps/designer/app/designer/features/props-panel/control.tsx#L218) is not an option, this should come from component's meta somehow
- [x] Ability to define a custom label for a prop in component meta (e.g. "Source", "ID")
- [ ] At the top of the panel show `[Icon] [Name]`

PR #1031, opened 2023-02-14
Props panel: Basics

rpominov:
## Description

#992

1. Cleaned up control types
2. Control type is now chosen by component meta (including Image.src)
3. Added custom labels support
4. Some other changes mentioned in comments in diff

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/components/box.ws.ts
```diff
@@ -14,7 +14,8 @@ const meta: WsComponentMeta = {
   label: "Box",
   Icon: SquareIcon,
   presetStyle,
-  props: props as MetaProps,
+  props: MetaProps.parse(props),
+  initialProps: ["tag"],
```
rpominov, line 18: We used to treat properties that have a default value like initial props, now only:
 - explicitly in `initialProps`
 - `required: true`

packages/generate-arg-types/src/types.ts
```diff
@@ -0,0 +1,89 @@
...
+  Number,
+  Text,
+  MultilineText,
+  Color,
+  Boolean,
+  Radio,
+  Select,
+  Check,
+  ImageUrl,
+]);
```
rpominov, line 87: All supported control types are now defined here and used by both:
- `generate-arg-types`
- props panel

apps/builder/app/builder/features/props-panel/control.tsx
```diff
@@ -21,15 +21,7 @@ import { ImageManager } from "~/builder/shared/image-manager";
...
-const textControlTypes = [
-  "text",
-  "array",
-  "color",
-  "date",
-  "file",
-  "number",
-  "object",
-] as const;
+const textControlTypes = ["text", "color", "number", "multilineText"] as const;
```
kof, line 24: Original names are taken from storybook's controls addon that tries to map data types to control names

https://storybook.js.org/docs/react/essentials/controls#annotation
Should we stick with those and add ours on top?

kof: They just agreed to add multiline-text

https://discord.com/channels/486522875931656193/689023212519948302/1075001187776069682

rpominov: Why do we need any kind of compatibility with what storybook is doing? I feel like this tights our hands and will probably require me an extra day just to understand what kind of compatibility we want

kof: I know we are not using stories right now, but we will need a way to use them some day.

The basic idea was to be able to start webstudio using cli just like storybook and let it pick-up the stories

apps/builder/app/builder/features/props-panel/control.tsx
```diff
@@ -201,34 +188,15 @@ export function Control({
...
-        userProp.name
-      }" component "${component}" is not a string`
-    );
-
-    return <NotImplemented />;
-  }
+  const controlType = argType?.controlType ?? "text";
 
-  if (component === "Image" && userProp.name === "src") {
+  if (controlType === "imageUrl") {
```
kof, line 199: https://storybook.js.org/docs/react/essentials/controls#annotation

See {control: {file, accept}} ...

apps/builder/app/builder/features/props-panel/control.tsx
```diff
@@ -272,12 +240,12 @@ export function Control({
...
           })
         }
-        type={type}
+        type={controlType}
       />
     );
   }
 
-  if (type === "boolean") {
+  if (controlType === "boolean") {
```
kof, line 248: can we stick with "control"?

apps/builder/app/builder/features/props-panel/use-props-logic.test.ts
```diff
@@ -118,126 +117,126 @@ describe("usePropsLogic", () => {
...
-        id: "2",
-        instanceId: "instanceId",
-        name: "arial-label",
-        type: "string",
-        value: "https://example.com",
-        required: false,
-      })
-    ).toBe(false);
-  });
+  // test("should respect initialProps ordering", () => {
```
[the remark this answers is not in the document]
kof: I think the respecting of initialProps ordering should be some simple unit test of a function that respects them, not a hook test

packages/generate-arg-types/src/arg-types.ts
```diff
@@ -34,67 +35,64 @@ export const propsToArgTypes = (
...
 
 const matchers = {
   color: new RegExp("(background|color)", "i"),
   date: /Date$/,
 };
 
+const toPropMeta = (
+  controlType: PropMeta["controlType"],
+  dataType: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kof, line 49: do we really have to use any here? in such cases we need to comment the "why"

kof: maybe something like unknown works?

apps/builder/app/builder/features/props-panel/control.tsx
rpominov: > Original names are taken from storybook's controls addon that tries to map data types to control names

I can't see a list similar to our original names anywhere at the link.

In the table I'm not sure what "Data Type" is. We don't get types in this format from the library that we use in `generate-arg-types`. For example, instead of `array` it would give us something like `string[]`

apps/builder/app/builder/features/props-panel/control.tsx
rpominov: Do you mean I should rename `imageUrl` to `file` and add support of `accept` option?

packages/generate-arg-types/src/arg-types.ts
rpominov: This is existing code, I didn't add this. Can spend some time cleaning up types in this file if you want

apps/builder/app/builder/features/props-panel/control.tsx
kof: yes

packages/generate-arg-types/src/arg-types.ts
kof: try for a few minutes, if it gets weird we can ask a Andarist, it may be a long tail that has resulted in it

apps/builder/app/builder/features/props-panel/control.tsx
rpominov: Will do this when I'll get to implementing the `file` control https://github.com/webstudio-is/webstudio-builder/issues/1011

apps/builder/app/builder/features/props-panel/use-props-logic.test.ts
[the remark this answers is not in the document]
rpominov: Yes. Will sort this in next PR

apps/builder/app/builder/features/props-panel/control.tsx
rpominov: I've added all types of controls from storybook to our system. Some of them we neither generate nor support in props panel. But now it's clear in code.

rpominov: Also created this issue: https://github.com/webstudio-is/webstudio-builder/issues/1036

apps/builder/app/builder/features/props-panel/control.tsx
```diff
@@ -356,17 +323,26 @@ export function Control({
...
         />
       );
     }
 
-    assertUnreachable(type, `Unknown control type ${type}`);
+    if (
+      control === "object" ||
+      control === "date" ||
+      control === "range" ||
+      control === "multi-select"
```
kof, line 335: multi-select can be rendered as check, seems same data, same result, just a different way of selecting multiple items

packages/generate-arg-types/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const Text = z.object({
+  control: z.literal("text"),
+  dataType: z.string(),
+  defaultValue: z.string().nullable(),
+  ...common,
+});
+
+const MultilineText = z.object({
+  control: z.literal("multiline-text"),
```
kof, line 33: add a comment that this has been discussed with storybook team
https://discord.com/channels/@me/1005106717870260275/1075352596027551744
and resulted in https://github.com/storybookjs/storybook/issues/21100

packages/generate-arg-types/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+});
+
+const Boolean = z.object({
+  control: z.literal("boolean"),
+  dataType: z.literal("boolean"),
+  defaultValue: z.boolean().nullable(),
+  ...common,
+});
+
+const Radio = z.object({
```
kof, line 53: I still don't understand why the data structure doesn't match sb's shape straight from their argTypes documentation `{ contact: { control: 'radio', options: ['email', 'phone', 'mail'] }}`
"contact" is a property name, control and options is the shape I am talking about

kof (approved):

packages/generate-arg-types/src/types.ts
kof: same for all of them

rpominov: But don't they match?

Our:

```
control: z.literal("radio"),
options: z.array(z.string()),
```

Theirs:

```
{ control: 'radio', options: ['email', 'phone', 'mail'] }
```

kof: I am talking about the argTypes shape, we sort of make a mixture out of our shape and argTypes shape, but should we mix them all up? It makes it kinda uncear that those are specifically defined by sb, which might be part of the initial confusion about why we even have these names.

It should actually say somewhere that these are argTypes:

```
const argTypes = {
  label: {
    name: 'label',
    type: { name: 'string', required: false },
    defaultValue: 'Hello',
    description: 'demo description',
    table: {
      type: { summary: 'string' },
      defaultValue: { summary: 'Hello' },
    },
    control: {
      type: 'text'
    }
  }
}
```

rpominov: I'd figure this out as part of https://github.com/webstudio-is/webstudio-builder/issues/1036

kof: I disagree, because that issue is about full compatibility with the goal of rendering stories. I am just asking to make sure argTypes shape and format is supported as-is.

kof: or did I misunderstand that issue? its not about rendering stories-based components?

kof: if not its fine, but I thought just changing the shape to argTypes format doesn't require a separate issue and research, its just a shape and semantics

kof: Its just confusing seeing your new shapes you just created to be a mixture of argtypes and webstudio specific types. I think you did it without full understanding and now created an issue to fix what you just created.

rpominov: > or did I misunderstand that issue? its not about rendering stories-based components?

Yes

> if not its fine, but I thought just changing the shape to argTypes format doesn't require a separate issue and research, its just a shape and semantics

It does require a research. 

For example: I know that Storybook supports either an array or an object as `options` of a select. You can do `{options: {"Label 1": "value 1", "Label 2": "value 2"}}`. But this isn't mentioned in the docs.

Basically there's no formal definition of `argTypes` that I know of at the moment.

I want to understand Storybook's interfaces and semantics well before implementing it.

kof: Here are the types storybook defines

https://github.com/ComponentDriven/csf/blob/next/src/story.ts#L63

rpominov: It's a complicated type a lot to understand. And also seem incomplete. 

Also seems incomplete. For example there's a `table` property mentioned here https://storybook.js.org/docs/react/api/argtypes but in code there's no table, probably comes goes under `[key: string]: any` https://github.com/ComponentDriven/csf/blob/e86f8c6b1046ca6e529f9713913d7f6b818a3c11/src/story.ts#L46

kof: that's the basic csf data types though, here are the control types

https://github.com/storybookjs/storybook/blob/next/code/ui/blocks/src/controls/types.ts

apps/builder/app/builder/features/props-panel/control.tsx
```diff
@@ -201,34 +188,15 @@ export function Control({
   onChangePropValue,
   setCssProperty,
 }: ControlProps) {
-  const meta = getComponentMetaProps(component);
+  const meta = getComponentMeta(component)?.props;
```
TrySound, line 191: It was intentionally separately to not bloat runtime with huge jsons.

packages/generate-arg-types/package.json
```diff
@@ -6,6 +6,8 @@
   "homepage": "https://webstudio.is",
   "type": "module",
   "scripts": {
+    "dev": "build-package --watch",
+    "build": "build-package",
```
TrySound, line 10: This is cli only tool. We run it with tsx directly. Why do you want to build it?

TrySound: Got it, you reuse types. We will need to rename it to something like arg-types then. But maybe later.

rpominov: It now exports some code as well

apps/builder/app/builder/features/props-panel/control.tsx
rpominov: Ah, got it. I misunderstood because we still export/import these jsons from the `*.ws.tsx` files and not as a separate export. So they still were in all bundles.

Will:
- switch back to getComponentMetaProps()
- add a comment about intention
- try to export from `*.ws.tsx` separately

packages/generate-arg-types/src/types.ts
rpominov: Refactored to a format similar to `ArgTypes` without trying to get a strict support as discussed in dm

packages/generate-arg-types/src/types.ts
```diff
@@ -0,0 +1,169 @@
...
+
+const controlTypeWithOptions = <T extends string>(name: T) =>
+  z.object({ type: z.literal(name), options: z.array(z.string()) });
+
+const common = {
+  /**
+   * Label for the control (may not match the property name)
+   */
+  name: z.string(),
+  description: z.string().nullable(),
```
kof, line 24: btw why not .optional()?

kof: we actually don'tn want nulls, we use it only when we have to because prismal delivers nulls

kof (approved):

packages/generate-arg-types/src/types.ts
rpominov: yeah, mistake. will fix

rpominov: Made `name`, `description `, and `defaultValue` all optional. This is in line with Storybook and will safe a lot of space in json

istarkov (approved):

apps/builder/app/builder/features/props-panel/control.tsx
rpominov: Done

PR #1037, opened 2023-02-15
Fix clone

istarkov:
## Description

## Steps for reproduction

Now works
http://localhost:3000/rest/project/clone/webstudiois

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/trpc-interface/src/authorize/project.server.ts
+++ b/packages/trpc-interface/src/authorize/project.server.ts
@@ -49,7 +49,9 @@ export const hasProjectPermit = async (
   }
 
   // Allow load webstudiois for clone
-  // @todo rethink permissions to allow everyone has access to various projects
+  // @todo Rethink permissions for this use-case
+  // The plan is to make new permission for projects which are allowed to be publicly clonable by anyone
+  // https://github.com/webstudio-is/webstudio-builder/issues/1038
   if (
     props.permit === "view" &&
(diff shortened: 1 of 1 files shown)


packages/trpc-interface/src/authorize/project.server.ts
```diff
@@ -48,6 +48,15 @@ export const hasProjectPermit = async (
     return true;
   }
 
+  // Allow load webstudiois for clone
+  // @todo rethink permissions to allow everyone has access to various projects
```
kof, line 52: what is meant by this?

istarkov: We have this edge case that everyone can clone webstudiois project
This check on id doesn't look good. If it will be not an edge-case permission system should be changed somehow so we could say that everyone can have access to project X.

istarkov: Probably every such project would have `everyone` user and for every check we would do additional check for `everyone`

kof: The plan is to make new permission for projects which are allowed to be publicly clonable by anyone.

kof: created an issue https://github.com/webstudio-is/webstudio-builder/issues/1038

TrySound (approved):

PR #1048, opened 2023-02-16
Props panel refactoring

### LINKED ISSUE #1013: Props panel: user added properties
- [ ] Always show label (property name) as text rather than input 
- [ ] UI to add a property
- [ ] UI to remove a property
- [ ] Controls should be able to control their layout including position of label and delete button. Example: <br/>Screenshot 2023-02-13 at 11 18 30

----------------------------

User Added Properties

### LINKED ISSUE #1051: Doing ZOD validation on import seems like seriously blocks script startup times
Because of that we can't use react-sdk at cloudflare with Bundled/Unbound model and getting 

```
Error: Failed to publish your Function. Got error: Error: Script startup exceeded CPU time limit.
```

see [Worker startup time](https://developers.cloudflare.com/workers/platform/limits/#worker-startup-time) 

Example https://github.com/webstudio-is/webstudio-builder/blob/main/packages/react-sdk/src/components/body.ws.tsx#L62-L64

Having that it's our tool generates all that data we can use zod on write, not on read.

rpominov:
## Description

Closes #1013

1. Refactored `PropMeta` shape to be flat again instead of trying to be the same as Storybook's `ArgType`
2. Changed `*.props.json` to `*.props.ts`. Fixes #1051
3. Prepared controls files structure, now just need to work on their UI
4. Refactored `usePropsLogic/PropsPanel` to the new logic where we choose the prop before adding and then never change the name
5. A lot of cleanup / small fixes / better TS types

This PR basically completes architecture/logic and now only the work on the controls UI is left. Although complicated controls like URL may still require some work on the logic.

## Steps for reproduction

1. Open props panel
2. Interact with the props

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/project-build/src/schema/props.ts
```diff
@@ -30,6 +30,11 @@ export const Prop = z.union([
     type: z.literal("asset"),
     value: z.string(),
   }),
+  z.object({
+    ...baseProp,
+    type: z.literal("string[]"),
+    value: z.array(z.string()),
+  }),
```
rpominov, line 37: @istarkov @TrySound 

Asking just in case: is it ok to add a new prop type "string[]"? Are there any concerns?

apps/builder/app/builder/features/props-panel/controls/combined.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { ColorControl } from "./color";
+import { NumberControl } from "./number";
+import { CheckControl } from "./check";
+import { RadioControl } from "./radio";
+import { SelectControl } from "./select";
+import { BooleanControl } from "./boolean";
+import { FileImageControl } from "./file-image";
+import type { ControlProps } from "../shared";
+
+export const CombinedControl = ({
```
kof, line 11: does this need to be a component or just a function that renders a component?

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -1,156 +1,240 @@
...
+  // If undefined is returned,
+  // we will not add a prop in storage until we get an onChange from control.
+  //
+  // User may have this experience:
+  //   - they added a prop but didn't touch the control
+  //   - they closed props panel
+  //   - when they open props panel again, the prop is not there
+  //
+  // We want to avoid this if possible, but for some types like "asset" we can't
+  return undefined;
```
kof, line 50: Why would your return an undefined, its the default behavior? Needed someplace for the comment? How about about the function declaration

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -1,156 +1,240 @@
...
+const getStartingProp = (
+  instanceId: Instance["id"],
+  meta: PropMeta,
+  name: string
+): Prop | undefined => {
+  const value = getStartingValue(meta);
+  return value && { id: nanoid(), instanceId, name, ...value };
+};
+
+const getDefaultMetaForType = (type: Prop["type"]): PropMeta => {
```
kof, line 62: seems like this could be just a map from type to meta instead of a function with all the ifs

e.g. 
```
typeToMetaMap = {asset: { type: "string", control: "file-image", required: false }}
```
Also, where is the type "asset" coming from, storybook calls it "file", wouldn't it make sense to call it file too? or is it deeply depending on our internal asset?

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -1,156 +1,240 @@
...
-  };
+type UsePropsLogicOutput = {
+  /**
+   * Initial (not deletable) props
+   */
+  initialProps: PropAndMeta[];
+  /**
+   * Optional props that were added
+   */
+  optionalProps: PropAndMeta[];
```
kof, line 103: are those props added by the user or by the component? not clear from reading the comment

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -1,156 +1,240 @@
...
+  /**
+   * Remaining props that can be added
+   */
+  remainingProps: NameAndLabel[];
```
kof, line 107: remaining sounds like its complete list of the entire remaining properties, if I understand correctly, its the properties that user has manually added? I think there is a better name once I understand the meaning of it

apps/builder/app/builder/features/props-panel/controls/combined.tsx
rpominov: What is the difference? Is there an issue with it being a component?

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -1,156 +1,240 @@
...
-      addedPropsById.set(prop.id, prop);
-      continue;
+
+    return { prop, propName: name, meta: known };
+  });
+
+  // names of the props added by the user during the lifetime of the hook
+  const [newNames, setNewNames] = useState<Prop["name"][]>([]);
+
+  const oldOptional: PropAndMeta[] = [...unprocessedSaved.values()]
```
kof, line 168: isn't values() already returning a new array essentially, why clone it?

kof: ah sorry, its an iterator convertion to array, maybe better Array.from() to be clear

apps/builder/app/builder/features/props-panel/use-props-logic.ts
rpominov: Ah. I'm just not used to use `udefined` and always forget you don't have to return it explicitly.

But yeah, in this case it's also better place for the comment IMO. Can put on the top though, not a strong opinion.

packages/generate-arg-types/src/arg-types.ts
```diff
@@ -44,43 +44,64 @@ const matchers = {
...
+        // @todo: we need some checks here. for example type.name can be "ImageLoader"
+        return makePropMeta("string", "text");
     }
-    case "function":
-    case "symbol":
-      return;
-    default:
-      // @todo: we need some checks here. for example type.name can be "ImageLoader"
-      return PropMeta.parse({ ...common(), control: "text" });
+  } catch (e) {
```
kof, line 102: no shorthands (I know it was there before)
```suggestion
  } catch (error) {
```

kof:
After image selection on Source property

Screenshot 2023-02-22 at 10 48 57

apps/builder/app/builder/features/props-panel/controls/combined.tsx
kof: Component goes through react render circle, its way slower than just a function, do it only if you actually need a component

apps/builder/app/builder/features/props-panel/use-props-logic.ts
rpominov: By user. Fixed the comment

apps/builder/app/builder/features/props-panel/use-props-logic.ts
kof: wrapping into a function is fine, but ifs instead of a simple map looks fishy, because you don't seem to need custom logic there, its just a switch/case type of thing going on here

apps/builder/app/builder/features/props-panel/use-props-logic.ts
rpominov: It's the complete list of the remaining props still available to add

kof: that makes sense then, better to clarify this in the comment, I was not sure

apps/builder/app/builder/features/props-panel/use-props-logic.ts
kof: I think I got confused because optional doesn't imply its controlled/added by the user, maybe better userProps? customProps?

rpominov:
> After image selection on Source property

All UI is temporary in this PR

apps/builder/app/builder/features/props-panel/use-props-logic.ts
kof: addedProps?

kof (approved):
Good work, approving upfront, most issues are semantics

apps/builder/app/builder/features/props-panel/controls/combined.tsx
rpominov: By that logic we should rewrite all our components that don't use hooks as functions. Sounds crazy :)

apps/builder/app/builder/features/props-panel/use-props-logic.ts
rpominov: re-wrote with switch/case

apps/builder/app/builder/features/props-panel/controls/combined.tsx
kof: depends, if you need component in the component tree for debugging purposes or because it needs a state - component is the way to go, but when its just a simple render function that doesn't need to be in the tree and will only make the tree less readable - why making it component

rpominov: https://discord.com/channels/955905230107738152/1077912355087863919

PR #1054, opened 2023-02-16
Simplify text editor store updates

TrySound:
Before we had to preserve instance data to not break styles. Now styles are decoupled and editor can interop whole instances type.

So we no longer need updates with instance-like shape. Text editor can output instances directly preserving only ids.

As a result got rid from children updates in both react-sdk and project packages.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/canvas/features/wrapper-component/wrapper-component.tsx
```diff
@@ -145,7 +145,13 @@ export const WrapperComponentDev = ({
           />
         }
         onChange={(updates) => {
-          onChangeChildren?.({ instanceId: instance.id, updates });
+          store.createTransaction([rootInstanceContainer], (rootInstance) => {
+            const { instancesById } = createInstancesIndex(rootInstance);
```
kof, line 149: I wonder if you should be creating a completely new index every time you want to mutate children as opposed to using immer on the existing index

kof (approved):

apps/builder/app/canvas/features/wrapper-component/wrapper-component.tsx
TrySound: Existing index is not wrapped with draft so will not work.
This will be optimized when we represent instances as map. Still need to migrate stuff.

PR #1056, opened 2023-02-16
Use nanoid for pagse

TrySound:
We already migrated all json ids from bson-objectid to nanoid. Looks like pages can be also migrated from nanoid since not stored directly in db.

Also removed unnecessary zod parses.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project-build/src/db/build.ts
+++ b/packages/project-build/src/db/build.ts
@@ -1,4 +1,4 @@
-import { v4 as uuid } from "uuid";
+import { nanoid } from "nanoid";
 import {
   prisma,
   Build as DbBuild,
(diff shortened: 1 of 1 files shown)


kof (approved):

packages/project-build/src/db/build.ts
```diff
@@ -216,17 +216,18 @@ const createPages = async (
     createNewTreeData({ projectId, buildId }),
     client
   );
-  return Pages.parse({
```
rpominov, line 219: The parse is here because the zod type has additional validation like this https://github.com/webstudio-is/webstudio-builder/blob/f6150c98da78331473254d213533f86f5e64d5ff/packages/project-build/src/schema/pages.ts#L70-L74

Same below. 

Please don't delete `parse` arbitrarily at least when you suspect that I added it. I usually try to add only when needed.

TrySound: Got it, thanks!

TrySound: Though do we need to validate data written in code? Kinda same as with props.

rpominov: I think we should when conditions are:

- Schema contains additional validations not just types (otherwise TS can do the same)
- No big performance concerns like large data or in hot path

We really want to avoid writing invalid data to DB

istarkov (approved):

PR #1059, opened 2023-02-17
Make css helper tool for collapsing prevention on canvas really smart

kof:
Closes https://github.com/webstudio-is/webstudio-builder/issues/604

## Description

-  User can now override helper definitions on canvas
- only boxes that would collapse without helper -  receive helper
- helper is removed when any CSS property is changed on that instance that would prevent collapsing, so that helper can be removed
- Helper doesn't show on the preview or publish
- Helper goes away if an instance inserted as a child
- There is no need to set padding-right or padding-bottom if you just need a small div with a defined or layout-based size, as soon as div is not collapsing, helper is not applied

More: padding will be only added on the side that would collapse otherwise. 
For example when I add a div, it is a block element, it grows automatically full width but has 0 height, in this case spacing heloper with padidng-top: 50px will apply, so that it doesn't collapse.
If user sets height: 100px or does anything that would give it a height - we remove the helper padding right away, so user can actually see the height they set

In other words we prevent elements from collapsing when they have 0 height or width by making them non-zero on canvas, but then we remove those padding universally without shortlists as soon as element doesn't collapse.

## Steps for reproduction

1. add a box
2. set padding-top
3. see it applies on canvas

1. add a box
2. set width
3. see the exact width on canvas, no helper padding

1. add a box
2. set height
3. see the exact height on canvas, no helper padding

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/shared/styles.ts
+++ b/apps/builder/app/canvas/shared/styles.ts
@@ -4,6 +4,7 @@ import { useStore } from "@nanostores/react";
 import { addGlobalRules } from "@webstudio-is/project";
 import {
   assetsStore,
+  isPreviewModeStore,
   selectedInstanceIdStore,
   useBreakpoints,
 } from "~/shared/nano-states";
@@ -31,6 +32,11 @@ import {
 import { useSubscribe } from "~/shared/pubsub";
 
 const cssEngine = createCssEngine({ name: "user-styles" });
+const helpersCssEngine = createCssEngine({ name: "helpers" });
+const fontsAndDefaultsCssEngine = createCssEngine({
+  name: "fonts-and-defaults",
+});
+const presetStylesEngine = createCssEngine({ name: "presetStyles" });
 
 // Helper styles on for canvas in design mode
 const helperStyles = [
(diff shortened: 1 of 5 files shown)


kof:
Hold on, I think I have an idea on how to do all of it much better

TrySound (approved):

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -44,11 +44,16 @@ const helperStyles = [
...
-  `[${idAttribute}]:where(:not(body)[${collapsedAttribute}~=w]) {
+  `[${idAttribute}]:where(:not(body)[${collapsedAttribute}="w"]) {
     padding-right: 50px;
   }`,
   // Has no height, will collapse
-  `[${idAttribute}]:where(:not(body)[${collapsedAttribute}~=h]) {
+  `[${idAttribute}]:where(:not(body)[${collapsedAttribute}="h"]) {
+    padding-top: 50px;
+  }`,
+  // Has no width, will collapse
```
TrySound, line 54: and height

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -54,18 +70,28 @@ const helperStyles = [
   }`,
 ];
 
+const subscribePreviewMode = () => {
+  return isPreviewModeStore.subscribe((isPreviewMode) => {
+    if (isPreviewMode) {
+      helpersCssEngine.clear();
+      helpersCssEngine.render();
```
TrySound, line 77: Maybe instead set some media query as rendering condition?

TrySound: Like will something like "not all" work as condition?

kof: Yeah thought about it too, maybe better? idk, doesn't really matter as long as this is small

TrySound (approved):

PR #1075, opened 2023-02-18
Use remix future dev

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/974

Took an example from template https://github.com/remix-run/remix/tree/83f71db9d4f05c4e682f40b1362cdbaae670c2ec/templates/express

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/server-express.js
```diff
@@ -0,0 +1,65 @@
+require("./env-check");
```
kof, line 1: - why is express server solving anything?

apps/builder/server-express.js
```diff
@@ -0,0 +1,65 @@
...
+app.all("*", (req, res, next) => {
+  purgeRequireCache();
+
+  return createRequestHandler({
+    build: require("./api/index.js"),
+    mode: process.env.NODE_ENV,
+  })(req, res, next);
+});
+
+const port = Number(process.env.PORT) || 3000;
```
kof, line 51: usually people use a port search, otherwise we can't start more than one instance

apps/builder/server-express.js
TrySound: 🤷‍♂️  Remix just decided to throw server management on user.

apps/builder/server-express.js
kof: or does it act as a default and if this doesn't work it would search for a free one?

TrySound: Do we need more than one instance for local dev? Usually not end well because some other part rely on specific port.

TrySound:
Looks like reload still broken
https://github.com/remix-run/remix/issues/5492

apps/builder/server-express.js
kof: remix was never a server, it was app runtime, they had a server for development that afaik we have been using but now you are changing it for express server,so I wonder what happened

apps/builder/server-express.js
kof: port is shared across a lot of things, so usually its a problem sooner or later

kof:
Oh so the latest Remix decided  to recommend this approach instead of what they had before with dev command?

kof:
Reading the release notes here https://github.com/remix-run/remix/releases/tag/remix%401.12.0 now

kof (approved):
well if it solves our problem I am ok with merging it

PR #1078, opened 2023-02-18
Fix flex children sections are showing on wrong elements

### LINKED ISSUE #685: Flex and Flex children sections are showing on wrong elements
When parent is set on display: flex, we show the flex section and flex child section on it, though, flex child should be only shown for it's children and flex section should not be shown on the child unless it's also set to display flex.

istarkov:
## Description

Fix flex children sections are showing on wrong elements
closes #685
most used functions are imported from style-info and all has tests there.

## Steps for reproduction

Create a few Boxes, and see that the Flex Child section is not shown.
Set `display: flex` for some Box. See that Flex Child section is not shown on the element. See that Flex Child section is shown on block children.

## Code Review

- [ ] hi @kof , @TrySound I want you to look namings, too much we have with styles Style, StyleDecl, StyleInfo etc.
I used useInstanceStyleData instead of useInstanceStyle because we already have similar useInstanceStyles

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/shared/style-instance-style-data.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  if (rootInstance === undefined) {
+    return undefined;
+  }
+
+  const ancestorsAndSelf = utils.tree.getInstancePath(rootInstance, instanceId);
+
+  if (ancestorsAndSelf.length < 2) {
+    return undefined;
```
kof, line 104:
```suggestion
    return;
```

apps/builder/app/builder/features/style-panel/shared/style-instance-style-data.ts
```diff
@@ -0,0 +1,108 @@
...
+        styleData[property] = value;
+      }
+    }
+    return styleData;
+  }, [selftAndCascadeInfo, inheritedInfo]);
+
+  return styleData;
+};
+
+export const useParentInstanceId = (instanceId: Instance["id"] | undefined) => {
```
kof, line 90: getParentInstance should be probably a utility in shared/tree-utils with a test

kof: reason is this is a use case that is popular and we can optimize it later using a data structure

apps/builder/app/builder/features/style-panel/shared/style-instance-style-data.ts
```diff
@@ -0,0 +1,108 @@
+import { useStore } from "@nanostores/react";
```
kof, line 1: sounds like file could be just instance-style-data.ts

kof: we already have a file called style-info.ts, seems like this logic is of the same type and should go in there as well

apps/builder/app/builder/features/style-panel/shared/style-instance-style-data.ts
TrySound: @istarkov You can use instances index today. Test free.

```
const { parentInstanceById } = useStore(instancesIndex)
```

apps/builder/app/builder/features/style-panel/shared/style-info.ts
```diff
@@ -282,3 +282,69 @@ export const useStyleInfo = () => {
...
+  const [breakpoints] = useBreakpoints();
+  const selectedBreakpoint = useStore(selectedBreakpointStore);
+  const selectedBreakpointId = selectedBreakpoint?.id;
+
+  const cascadedBreakpointIds = useMemo(
+    () => getCascadedBreakpointIds(breakpoints, selectedBreakpointId),
+    [breakpoints, selectedBreakpointId]
+  );
+
+  const selftAndCascadeInfo = useMemo(() => {
```
kof, line 300: typo

kof (approved):
looking good

PR #1088, opened 2023-02-20
Rename origin to webstudioApi in publish request to the new deployment process

andrasbacsai:
## Description

Just renaming one of the payloads to our publisher API to be consistent.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/routes/rest/publish.ts
+++ b/apps/builder/app/routes/rest/publish.ts
@@ -25,7 +25,7 @@ export const action = async ({ request }: ActionArgs) => {
         method: "PUT",
         headers,
         body: JSON.stringify({
-          origin: url.origin,
+          webstudioApi: url.origin,
           projectId,
           domain,
         }),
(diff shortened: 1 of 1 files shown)


apps/builder/app/routes/rest/publish.ts
```diff
@@ -25,7 +25,7 @@ export const action = async ({ request }: ActionArgs) => {
         method: "PUT",
         headers,
         body: JSON.stringify({
-          origin: url.origin,
+          webstudioApi: url.origin,
```
kof, line 28: is there a better name? builderApiUrl ?

kof (approved):

apps/builder/app/routes/rest/publish.ts
andrasbacsai: Yeah, that could fit better! Thanks!

istarkov (approved):
I like `Origin` postfix instead of Url as based on this 
https://developer.mozilla.org/en-US/docs/Web/API/URL/origin

Looks a little bit strict or narrow for me vs Url. As name explicitly says that no pathname would be specified.


This is:
For http or https URLs, the scheme followed by '://', followed by the domain, followed by ':', followed by the port (the default port, 80 and 443 respectively, if explicitly specified).

PR #1091, opened 2023-02-20
Run eslint on precommit

TrySound:
We often get failed CI after commit because of lint checks. We commit fixes and wait again for CI to complete.

Here I put eslint to nano-staged to run on staged modules. This will let us see errors prematurely and commit less "Fix lint".

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,4 +2,11 @@ module.exports = {
   root: true,
   // This tells ESLint to load the config from the package `eslint-config-custom`
   extends: ["@webstudio-is/eslint-config-custom"],
+  ignorePatterns: [
+    "*.d.ts",
+    "codemod",
+    "packages/*/lib",
+    "packages/prisma-client/prisma/migrations",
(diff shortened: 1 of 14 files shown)


kof:
I have been there, doing anything on recommit that is not done in milliseconds is super annoying, never again :)

Pre-push maybe, but not pre-commit

TrySound:
Pre push won't work with staged files. They already committed. Push will be just deadly slow.

kof:
In that case I would prefer to commit fast, run checks for everything, if needed amend the commit and push, in the worst case you just push again once your local checks run is finished

kof:
fast commits are life changing, i hate everything that gets in the way of fast commits

kof (approved):
lets make it pure --fix thing without failing the commit, then I am ok to try it out and see if its fast enough, but please don't let it prevent commits

kof:
I guess we should close this?

PR #1097, opened 2023-02-21
Fix collapse

### LINKED ISSUE #1087: Collapsing box outline bugs
1. wrong detection after reload

- add box1 > box2
- reload 
- both boxes are outlined, but only box2 one should have outline

2. box collapses on height, but spacing added on width

- add box1 > box2
- set height: 200 on box1
- set display: flex on box1
- we detect horizontal collapse on box 2, but height is 0 because of parent flex

istarkov:
## Description

closes #1087 

- [x] found bug we should not remove all collapsed
- [x] we need to detect edge-case if element explicitly has 0 width / height

## Steps for reproduction

see #1087

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -334,7 +334,7 @@ export const Builder = ({
               css={{
                 height: "100%",
                 width: "100%",
-                minWidth: `${minWidth}px`,
+                minWidth,
               }}
             />
           </Workspace>
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -38,6 +38,7 @@ import { Navigator } from "./features/sidebar-left";
 import { getBuildUrl } from "~/shared/router-utils";
 import { useCopyPaste } from "~/shared/copy-paste";
 import { AssetsProvider } from "./shared/assets";
+import { minWidth } from "./features/breakpoints/width-setting";
 
 registerContainers();
 
@@ -74,7 +75,7 @@ const useNavigatorLayout = () => {
   // We need to render the detached state only once the setting was actually loaded from local storage.
   // Otherwise we may show the detached state because its the default and then hide it immediately.
   const [clientSettings, _, isLoaded] = useClientSettings();
-  return isLoaded ? clientSettings.navigatorLayout : "docked";
+  return isLoaded ? clientSettings.navigatorLayout : "undocked";
 };
 
 const useSubscribeCanvasReady = (publish: Publish) => {
@@ -333,6 +334,7 @@ export const Builder = ({
               css={{
                 height: "100%",
                 width: "100%",
+                minWidth: `${minWidth}px`,
               }}
             />
           </Workspace>
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -143,7 +143,7 @@ export const Canvas = ({
         }
       }
     });
-  }, [data.tree?.root.id]);
+  }, []);
 
   const elements = useElementsTree(getComponent);
 
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from "react";
+import { useMemo, Fragment } from "react";
 import { computed } from "nanostores";
 import type { CanvasData } from "@webstudio-is/project";
 import {
@@ -130,7 +130,8 @@ export const Canvas = ({
     return (
       <>
         <GlobalStyles />
-        {elements}
+        {/* Without fragment elements will be recreated in DesignMode */}
+        <Fragment key="elements">{elements}</Fragment>
       </>
     );
   }
@@ -139,7 +140,7 @@ export const Canvas = ({
     <>
       <GlobalStyles />
       <DesignMode />
-      {elements}
+      <Fragment key="elements">{elements}</Fragment>
     </>
   );
 };
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -34,14 +34,13 @@ import {
   useSetAssets,
   useSetTreeId,
 } from "~/shared/nano-states";
-import { subscribe } from "~/shared/pubsub";
 import { usePublishScrollState } from "./shared/use-publish-scroll-state";
 import { useDragAndDrop } from "./shared/use-drag-drop";
 import { useSubscribeBuilderReady } from "./shared/use-builder-ready";
 import { useCopyPaste } from "~/shared/copy-paste";
 import { customComponents } from "./custom-components";
 import { useHoveredInstanceConnector } from "./hovered-instance-connector";
-import { setDataCollapsed } from "./collapsed";
+import { setDataCollapsed, subscribeCollapsedToPubSub } from "./collapsed";
 
 registerContainers();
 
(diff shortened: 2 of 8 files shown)


apps/builder/app/builder/builder.tsx
```diff
@@ -74,7 +75,7 @@ const useNavigatorLayout = () => {
   // We need to render the detached state only once the setting was actually loaded from local storage.
   // Otherwise we may show the detached state because its the default and then hide it immediately.
   const [clientSettings, _, isLoaded] = useClientSettings();
-  return isLoaded ? clientSettings.navigatorLayout : "docked";
+  return isLoaded ? clientSettings.navigatorLayout : "undocked";
```
kof, line 78: please don't change this behavior by default, we want users to have navigator undocked by default because that's the most common need, in reality people use it all the time

apps/builder/app/builder/builder.tsx
```diff
@@ -333,6 +334,7 @@ export const Builder = ({
               css={{
                 height: "100%",
                 width: "100%",
+                minWidth: `${minWidth}px`,
```
kof, line 337:
```suggestion
                minWidth,
```

apps/builder/app/builder/builder.tsx
istarkov: undocked.
image

docked
image

I see the same picture but in the case of "undocked" there is no layout trash
So I'm not getting why I can't change it

kof: ohh, I thought you changed the other way around, but also, did you try to set it do docked on the client and reload? now it should be loading as undocked with your change and then jumping to docked

kof: probably we should just add a transition for this so that this doesn't feel like a jump but rather as a nice, wanted transition

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
+import { collapsedAttribute } from "@webstudio-is/react-sdk";
```
kof, line 1: could probably go to ./shared/collapse-detection.ts

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
...
+  instanceIdSet.clear();
+  if (instanceIds.length === 0) {
+    return;
+  }
+
+  /**
+   *  Selector to find element itself or common ancestor
+   *  (finding common ancestor is suboptimal solution but it almost never happens)
+   **/
+  const eltSelector =
```
kof, line 17:
```suggestion
  const elementSelector =
```

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const elementsToRecalculate: HTMLElement[] = [];
+  const parentsWithAbsoluteChildren = new Map<HTMLElement, number>();
+
+  for (const element of elements) {
+    if (!(element instanceof HTMLElement)) {
+      continue;
+    }
+
+    if (element.tagName === "_IMG") {
```
kof, line 42: I think we should ignore all void elements https://developer.mozilla.org/en-US/docs/Glossary/Void_element

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
...
+    // Find all Leaf like elements
+    // Leaf like elements are elements that have no children or all children are absolute or fixed
+    if (element.childElementCount === 0) {
+      elementsToRecalculate.push(element);
+    }
+
+    const elementPosution = window.getComputedStyle(element).position;
+
+    if (elementPosution === "absolute" || elementPosution === "fixed") {
+      if (element.parentElement) {
```
kof, line 59: this condition should be around the entire block

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
...
+      elementsToRecalculate.push(element);
+    }
+  }
+
+  // Now combine all operations in batches.
+
+  // 1. Remove all collapsed attributes
+  baseElement.parentElement?.removeAttribute(collapsedAttribute);
+  baseElement.removeAttribute(collapsedAttribute);
+  for (const elt of baseElement.querySelectorAll(`[${collapsedAttribute}]`)) {
```
kof, line 84:
```suggestion
  for (const element of baseElement.querySelectorAll(`[${collapsedAttribute}]`)) {
```

apps/builder/app/canvas/canvas.tsx
```diff
@@ -120,6 +122,29 @@ export const Canvas = ({
...
+            const instanceId = patch.value?.instanceId;
+
+            if (typeof instanceId === "string") {
+              setDataCollapsed(instanceId);
+            }
+          }
+        }
+      }
+    });
+  }, [data.tree?.root.id]);
```
kof, line 146: - lets move this as a custom hook into collapsed
- separate react from non-react - everything inside useEffect is non-react, make a separate function and call it from the hook

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
...
+  elements.push(baseElement);
+
+  const descendants = baseElement.querySelectorAll("[data-ws-id]");
+  elements.push(...descendants);
+
+  const elementsToRecalculate: HTMLElement[] = [];
+  const parentsWithAbsoluteChildren = new Map<HTMLElement, number>();
+
+  for (const element of elements) {
+    if (!(element instanceof HTMLElement)) {
```
kof, line 38: I think you can also check nodeType, seems better than instance checks

kof: nodeType === Node.ELEMENT_NODE

apps/builder/app/canvas/collapsed.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // Find all Leaf like elements
+    // Leaf like elements are elements that have no children or all children are absolute or fixed
+    if (element.childElementCount === 0) {
+      elementsToRecalculate.push(element);
+    }
+
+    const elementPosution = window.getComputedStyle(element).position;
+
+    if (elementPosution === "absolute" || elementPosution === "fixed") {
```
kof, line 58: typo

apps/builder/app/builder/builder.tsx
istarkov: I think we should move docked/undocked into searchParams. Or switch on client rendering. Transitions doesn't look good.

apps/builder/app/canvas/collapsed.ts
istarkov: Do we need this to be used elsewhere?

apps/builder/app/builder/builder.tsx
kof: searchParams won't solve it for permanently setting it, I think we could save it in the db, then ssr knows the setting

kof: but this problem is just one case, we will have more later, so we need a way to allow local settings based on local storage I think

apps/builder/app/canvas/collapsed.ts
kof: nope, in that case features/collapse-detection

kof (approved):

apps/builder/app/canvas/canvas.tsx
TrySound: This is very hacky assumption and will break next week. Let's move it back to hurt everyone's eyes.

istarkov: Just keep in mind that we need to somehow know what instanceId was affected by the style change.

TrySound: The issue is that token style can affect a few instances. Not just one.

PR #1098, opened 2023-02-21
To reduce size of published apps pass getComponent to InstanceRoot, add size limit 

istarkov:
## Description

Now published apps during build can pregenerate imports of used components
https://github.com/webstudio-is/webstudio-builder/blob/89688ee837b45f0877ebe32d3aefb1869a97c0a1/packages/sdk-size-test/app/page-components/page-main.ts#L1-L3

And pas getComponent to InstanceRoot https://github.com/webstudio-is/webstudio-builder/blob/89688ee837b45f0877ebe32d3aefb1869a97c0a1/packages/sdk-size-test/app/routes/$.tsx#L23

Also size-limit is added https://github.com/webstudio-is/webstudio-builder/blob/89688ee837b45f0877ebe32d3aefb1869a97c0a1/packages/sdk-size-test/package.json#L10-L19

### What I did

Create sample remix app with sdk
Build: 256K

Increase size of sup component
Build now 564K

After changes build now 256kB

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/components/subscript.tsx
```diff
@@ -5,6 +5,10 @@ const defaultTag = "sub";
...
...356090547827553099902283601182556875723878812585829342112120643535136234233354548000376373539228441337466447546489972715324870623432473939494074367849041727254265742675895182796020334362260618434065482929109694732775810631500580250568949213398370571061955381036992510061604500623195895685277633841453387092156878758032127460311284924887146975923896612165410078451665187599926607299020458345627963420439715652445650039332697584161527418689052810339622772868014570270031896278770775137289513749285389160134511814790912124554283511450747662061450207874055271983106491319508431939379405139356086244871206328233097256310656806715935871203992140966633225119104450832165354362199377758584328122723097176497270028253352023603346945160822872847275228184687773750722988339131876836902638244934888564346061470214101593355370833759261193543814375325836805068692651602131963859004249450260777932898292974931025747485191547582123684275563739778101515027771884673967343974182564271586530092136733880079123112661604189172284689063826787172246977341470030337709486294246783622917217912573978588957230493800358591236399689631216138583104648370796376626799297615668211984659341559339167444688620035568965184061896502099
```
TrySound, line 10: You lost me after 8th digit

istarkov: Pi number not compressable, so good to use to create huuuuge component ;)

packages/sdk-size-test/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "@remix-run/react": "1.12.0",
+    "@webstudio-is/project": "workspace:^",
+    "@webstudio-is/project-build": "workspace:^",
+    "@webstudio-is/react-sdk": "workspace:^",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
+  },
+  "devDependencies": {
+    "@remix-run/dev": "1.12.0",
+    "@remix-run/eslint-config": "1.12.0",
```
TrySound, line 22: We probably don't need this

istarkov:
Strange locally size-limit gives fully another numbers vs this
image

packages/react-sdk/src/components/components-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+  "Button",
+  "Input",
+  "Form",
+  "Image",
+  "Blockquote",
+  "List",
+  "ListItem",
+] as const;
+
+type Equals<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y
```
kof, line 32: oh nooooo 😭 is there a way to not have it? @Andarist

packages/react-sdk/src/components/components-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Now used only in builder app
+ * @todo Consider using the same approach in the builder app as in the published apps . A dynamic import is needed
+ */
+export const getComponent = (
+  name: string
+): undefined | typeof components[ComponentName] => {
+  return registeredComponents != null && name in registeredComponents
+    ? (registeredComponents[
+        name as ComponentName
+      ] as typeof components[ComponentName])
```
kof, line 66: 😭 @Andarist any better ideas?

packages/react-sdk/src/components/components.ts
```diff
@@ -0,0 +1,48 @@
...
+ * The application imports only the components it uses and uses
+ * getComponent = createGetComponent({ Box, BlaBla })to pass them to RootInstance
+ * see example /packages/sdk-size-test/app/routes/$.tsx
+ */
+import { Body } from "./body";
+import { Box } from "./box";
+import { TextBlock } from "./text-block";
+import { Heading } from "./heading";
+import { Paragraph } from "./paragraph";
+import { Link } from "./link";
```
kof, line 13: why not just export it right away?

packages/react-sdk/src/components/index.ts
```diff
@@ -157,56 +140,12 @@ export const registerComponentPropsMetas = (
...
-export const getComponent = (
-  name: string
-): undefined | typeof defaultComponents[ComponentName] => {
-  return registeredComponents != null && name in registeredComponents
-    ? (registeredComponents[
-        name as ComponentName
-      ] as typeof defaultComponents[ComponentName])
-    : defaultComponents[name as ComponentName];
-};
+export let registeredComponents: RegisteredComponents | null = null;
```
kof, line 148: how about undefined )
```suggestion
export let registeredComponents: RegisteredComponents;
```

packages/react-sdk/src/components/components-utils.ts
Andarist: This type checks type "identity", so for example those won't be treated as the same types:
```ts
type A = { a: number, b: string }
type B = { a: number } & { b: string }
```

I'd say that usually it's not what you actually want, so this can be rewritten like this:
```ts
type Equals<A, B> = [A] extends [B] ? [B] extends [A] ? true : false : false;
```
Or even like this:
```ts
type Equals<A, B> = [A, B] extends [B, A] ? true : false;
```
Although I've seen the latter not working in some quite complex generic cases (so the problem is likely not applicable to you + it's a TS bug)

packages/sdk-size-test/app/routes/$.tsx
```diff
@@ -0,0 +1,25 @@
...
+} from "@remix-run/cloudflare";
+import { useLoaderData } from "@remix-run/react";
+import { InstanceRoot, Root } from "@webstudio-is/react-sdk";
+import { getComponent } from "../page-components/page-main";
+
+export const meta: MetaFunction = () => {
+  return { title: "test" };
+};
+
+export const links: LinksFunction = () => [];
```
kof, line 14: no need for this or meta if that's just for testing

packages/sdk-size-test/app/routes/$.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const loader = async ({ context, request }: LoaderArgs) => {
```
kof, line 16: no need, for that either

packages/sdk-size-test/app/routes/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+import CatchAllContnet, {
+  loader as catchAllloader,
+  meta as catchAllmeta,
+  links as catchAllLinks,
+} from "./$";
+
+export const meta: MetaFunction = (args) => catchAllmeta(args);
+export const loader: LoaderFunction = (args) => catchAllloader(args);
+export const links = catchAllLinks;
+const Content = () => <CatchAllContnet />;
```
kof, line 11: why all that for simple test?

packages/sdk-size-test/app/routes/$.tsx
istarkov: Just tried be as much as near to existing saas app. Also have in mind that probably it will be example app some day

packages/sdk-size-test/app/routes/index.tsx
istarkov: To be as much as near to real app.

kof: maybe you want to use an actual bundler in that case, the same one we use for deployment, I don't really want us to maintain this setup just to test the size, especially if it can become different than our actual deployment

packages/sdk-size-test/server.js
```diff
@@ -0,0 +1,14 @@
...
+// eslint-disable-next-line import/no-internal-modules
+import * as build from "@remix-run/dev/server-build";
+
+const handleRequest = createPagesFunctionHandler({
+  build,
+  mode: process.env.NODE_ENV,
+  getLoadContext: (context) => context.env,
+});
+
+export function onRequest(context) {
```
kof, line 12: export const onRequest

packages/react-sdk/src/components/components.ts
istarkov: Forgot thank you, initially tried do everything in single file

packages/react-sdk/src/components/components-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+  "Superscript",
+  "Subscript",
+  "Button",
+  "Input",
+  "Form",
+  "Image",
+  "Blockquote",
+  "List",
+  "ListItem",
+] as const;
```
Andarist, line 30: Instead of the `Equals` dance you could likely just do this:
```suggestion
] as const satisfies readonly [ComponentName, ...ComponentName[]];
```

istarkov: our prettier don’t like satisfies ;) will update

packages/react-sdk/src/components/components-utils.ts
Andarist: The second cast is not needed as far as I can tell: [TS playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBKC2AHcBTMUByBDeKIwF4YBtAKBhgCIAhEAD0oBpyraATATyZYHoeqAKijpRqAGxDAA1twp8qACRRY2ASzABzWTHmUAClgBOWDccQALbboAy6mczn9KAJVXBzQkbbD3eTgMqIWGBWTrRibKFUAJJQWGJuUZT+AK6IKIYQwIaqiFBJqQBGWTl5SdQpUFDgSdFgiJVJAGIghvC18CYo5RLSAI4pIFDdDjpOttBJE1CxKO3MALowWPigkFCk8jAQWFCqEABmqngwhsps4GIcJADCIEio6Ni4jDAAdB93D2BomDgoxAWCwA3KRSFAOOkYM4UBp9sMzmwvsgfuh8EQDIY9vEADwAbz8MDwCXQAFo1DtCmIUKSfiJSSSUDAAAIQ9IlXJQUnE9RQHiFYKktl4FjEMD-GDqGDIx5-XALABcMDxAF9QSqAHyg0hraBwe4o37o5UsWh0JWq0bsDgWlWkO3gyFMmWouVMojCkAHfXfX7PPBilLwQoZBabfjCZBYmDC6UG2X+wgwKQoDhemNO9MIQ1o7WR1qwamwM5w6AZFBI+OuiBKmGlhEVl1GmAAHxgYBSYjESY7XbzdCjsF1sA0KCgTfQSYAFCxxbgldAcppSABKJUpMBsFBHH5sVsZ9JZqtG4gTt1LAgak0UM5QFKGMCnWHw8uV31omAAQiIve7ADI-3bCUpRLF9ETPCAWAoAB+GAp1AstwOPNEyAoNC0LnJkVjjd83SgtCFhXfClWzWUIDFCVsLPf0QXtYEgA)

You need to cast  `name` to `ComponentName` though as long as `name` is typed as `string` (and I assume that you want it to be typed as `string` because this is a dynamic piece)

packages/react-sdk/src/components/components-utils.ts
Andarist: and actually, at first I tried this without `as const` but thanks to `as const` you can even skip the tuple type complexity and simplify this like this:
```suggestion
] as const satisfies readonly ComponentName[];
```

istarkov: satisfies also doesn't give an expected result https://www.typescriptlang.org/play?#code/MYewdgzgLgBKC2AHcBTMUIwLwwN4CgYiYAhEADwC4YAKASiwD4wBXAGzYBpDiS2BDAEYDq9Jqw7cAvgG58+FOWQAnWFACeiFDADCIJKnQA5fvG04A1inUgAZjA1a7cfcjBoMc-KEiwEbjxMzTBwAbR4iACIyckjOYgSYfABdGH5MH2gYCH4oAEsIWzyUTGUUfgATcDZ1GFC9A3djUxR4gDoOhoDms1Dk5PkAekGYAEkwGHha4HTtRVNENm0hEAA3bXgWLIBzPPW0iZRlZRBleIBBABlLmCt1TE2swW08if9DKCCSwiA

Andarist: Ah, yes - you are right. Can't you just cast `Object.keys` here though? Derive the type + derive the runtime value, sure there is nothing stopping you from casting to a wrong value etc but the whole "unsafe" bit stays contained in a single, very simple, line so it's not that likely that this would end up being a problem.

istarkov: Object.keys immediately increase bundle size in a task I'm solving. All components becomes not tree-shakeable

packages/sdk-size-test/app/routes/$.tsx
kof: we will, probably, but its unlikely going to be this test suit, we need something super easily maintainable for such thing, ideally the real thing we use to bundle, but if not something super minimal

packages/react-sdk/src/components/components-utils.ts
Andarist: Then this is the only thing that comes to my mind right now that would avoid `Equals`:
https://www.typescriptlang.org/play?#code/MYewdgzgLgBKC2AHcBTMUIwLwwN4CgYiYAhEADwC4YAKASiwD4wBXAGzYBpDiS2BDAEYDq9Jqw7cAvgG58+FOWQAnWFACeiFDADCIJKnQA5fvG04A1inUgAZjA1a7cfcjBoMc-KEiwEbjxMzTBwAeUEAKxRgKAA6K3UIGh4iAmJ00gpqAEZuDOIAegLSAWF+HJSYKRgIfigASwhbepRMXBgAbQBpGHqwGATnRxRnf0MMAF0cqvw6GH5MAEFlZX51AB5B+2HR13GIRnkimABJfvh1OAXtRVNENm0hEAA3bXgWaBgAc3rX+f6UCsQMpODBFgAZcEDayYd6fQTaPouAzuYymVqEIA

I'm not super sold if it's that simpler than `Equals` though. It has a nice benefit of the error being more local though and more readable. With the current `Equal` strategy this is all you get:
> Argument of type 'true' is not assignable to parameter of type 'false'.(2345)
https://www.typescriptlang.org/play?#code/MYewdgzgLgBKC2AHcBTMUIwLwwN4CgYiYAhEADwC4YAKASiwD4wBXAGzYBpDiS2BDAEYDq9Jqw7cAvgG58+FOWQAnWFACeiFDADCIJKnQA5fvG04A1inUgAZjA1a7cfcjBoMc-KEiwEbjxMzTBwAbR4iACIyckjOGAiYAHokmGiBYX5I-ABdGH5MH2h5FJgASTAYeHU4Au1FU0Q2bSEQADdteBZoGABzAEsO-MqUZWUQZXiAQQAZGZgrdUwunsFtfsr-QygglAhCfEdtAFEARxZ+NggAHin4kkZsGFC70jzFKDQAE0xQkmm8gB+BzKFjaai2S4QFBeIpqTR7M4XNhPW73Rg0fjUJFQtGkRh0bCPFhgL4oWwbFBfORHCA4tjXPQGdzGUwoeJHZxbFk7NkQUKseBrZQ5DFQUEoOgyIA

The variant with `Object.keys` and a dummy object even comes with autocompletes for the missing keys etc

istarkov: I like your solution, Thank you!

packages/react-sdk/src/components/components-utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  keyof typeof components
+>;
+
+export const getComponentNames = (): ComponentName[] => {
+  const uniqueNames = new Set([
+    ...componentNames,
+    ...Object.keys(registeredComponents || {}),
+  ]);
+
+  return [...uniqueNames.values()] as ComponentName[];
```
kof, line 42: Array.from is nicer to understand the why

packages/react-sdk/src/components/index.ts
```diff
@@ -167,58 +148,12 @@ export const registerComponentPropsMetas = (
...
-export const getComponent = (
-  name: string
-): undefined | typeof defaultComponents[ComponentName] => {
-  return registeredComponents != null && name in registeredComponents
-    ? (registeredComponents[
-        name as ComponentName
-      ] as typeof defaultComponents[ComponentName])
-    : defaultComponents[name as ComponentName];
-};
+export let registeredComponents: RegisteredComponents | undefined = undefined;
```
kof, line 156: lol I thought you can just leave it alone without value

kof (approved):
good job, terrible ts utils are gone, I am not nervous any more

PR #1099, opened 2023-02-21
Separator component

kof:
## Steps for reproduction

1. Add separator 
2. Paste `---` for markdown

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/sidebar-left/panels/components/components.tsx
+++ b/apps/builder/app/builder/features/sidebar-left/panels/components/components.tsx
@@ -110,7 +110,8 @@ export const TabContent = ({ publish, onSetActiveTab }: TabContentProps) => {
       meta?.type === "container" ||
       meta?.type === "control" ||
       meta?.type === "embed" ||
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/components/separator.ws.tsx
```diff
@@ -0,0 +1,31 @@
...
+    value: "gray",
+  },
+  border: {
+    type: "keyword",
+    value: "none",
+  },
+} as const;
+
+export const meta: WsComponentMeta = {
+  type: "embed",
```
kof, line 21: this is a little weird for a separator, I would have called it "void" but image, input etc are also void https://developer.mozilla.org/en-US/docs/Glossary/Void_element

kof: we condidered that name before, but it was weird to call image or input void from general perspective, but maybe we should

PR #1107, opened 2023-02-22
Drop turbo

TrySound:
I tried to setup turbo cache without luck recently. Without cache there is no much use for turbo in our project. It can be removed in favor of pnpm which is even faster alone.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

kof:
I think we are not doing it since the issues with turbo will be resolved very soon

PR #1115, opened 2023-02-23
Radio and Checkbox components update

### LINKED ISSUE #963: New component: Radio
This is the “Radio Group Item” within the Radio Group Radix component. Radios are similar to checkboxes except that radio groups can only have one option selected while checkboxes can have multiple.

Screenshot 2023-02-07 at 1 27 29 PM

rpominov:
## Description

Closes #963

## Steps for reproduction

1. https://radio-chekbox--638245578b47d4399fba191a.chromatic.com/?path=/story/library-radio--demo
2. https://radio-chekbox--638245578b47d4399fba191a.chromatic.com/?path=/story/library-checkbox--demo

## Code Review

  - check the demos by the links above, specifically make sure I got all the dimensions right as it was tricky with sub-pixels
  
## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/checkbox.tsx
```diff
@@ -1,86 +1,130 @@
...
+import { forwardRef, type ComponentProps, type Ref } from "react";
+import * as Primitive from "@radix-ui/react-checkbox";
+import { type CSS, css, theme, styled } from "../stitches.config";
+
+const checkedIconStyles = css({
+  display: "none",
+  "[data-state=checked] > &": { display: "block" },
+});
+
+const CheckedIcon = () => (
```
kof, line 15: could potentially be just css, no?

packages/design-system/src/components/checkbox.tsx
```diff
@@ -1,86 +1,130 @@
...
-  "&::after": {
-    boxSizing: "border-box",
-  },
+import { forwardRef, type ComponentProps, type Ref } from "react";
+import * as Primitive from "@radix-ui/react-checkbox";
+import { type CSS, css, theme, styled } from "../stitches.config";
+
+const checkedIconStyles = css({
+  display: "none",
+  "[data-state=checked] > &": { display: "block" },
```
kof, line 12: instead of nesting, better is css custom variable

packages/design-system/src/components/checkbox.tsx
```diff
@@ -1,86 +1,130 @@
...
-  display: "flex",
-  height: "100%",
-  justifyContent: "center",
-  width: "100%",
+const indicatorStyles = css({
+  display: "block",
+  background: theme.colors.backgroundControls,
+  borderRadius: theme.borderRadius[2],
+  color: theme.colors.backgroundPrimary,
+  "&[data-disabled]": {
```
kof, line 97: shouldn't this be data-state=disabled?

kof: nope

kof (approved):

packages/design-system/src/components/checkbox.tsx
rpominov: I don't like the boilerplate of variables when encapsulation is not a concern

packages/design-system/src/components/checkbox.tsx
kof: second

taylornowotny:
@rpominov I used icons for the radio and checkbox - sorry if that made it harder for you becuase of the sub-pixels. I can re-design them with divs - just let me know if that's needed and I'll do it.

Feedback:

1. It looks like both the radios and checkboxes are not centered in their containers. 
Screenshot 2023-02-23 at 12 01 25 PM
Screenshot 2023-02-23 at 12 02 48 PM

2. The weight of the lines is 1.5px. 1px is too thin.

3. Focus indicator doesn't show for unchecked checkboxes or radios.

4. Pressing ENTER when focusing does not toggle between checked & unchecked

rpominov:
> It looks like both the radios and checkboxes are not centered in their containers.
> The weight of the lines is 1.5px. 1px is too thin.

These are most likely artefacts because of sub-pixel values, so yeah, let's re-design

> I don't see the focus indicator on unchecked checkboxes or radios.
> Pressing ENTER when focusing does not toggle between checked & unchecked

Focus behaviour is implemented by Radix, it will be tricky to change it and I don't think we should. Radix team probably put more thought into it.

https://www.radix-ui.com/docs/primitives/components/radio-group#keyboard-interactions
https://www.radix-ui.com/docs/primitives/components/checkbox#keyboard-interactions

Also, I made the focus visible only when keyboard is used, this might be the reason you didn't see focus sometimes, try to use keyboard.

But if we want focus to be visible even on mouse interactions, this is easy to change.

kof:
@istarkov I think you have experience with subpixel values, are there problems? I thought it works

kof:
Quick googling gave me some pointers that there are apparently differences in how browsers may render subpixel and that it also can dependon usage of transforms and if its border or something else https://stackoverflow.com/questions/34676263/sub-pixels-calculated-and-rendered-differently-among-browsers

taylornowotny:
The design uses icons for all the visuals. There's no need to mess with subpixels, just use the icons as they are used in the design.

If this is not possible then I will re-design them to align to the pixel grid and not use icons.

taylornowotny:
@rpominov understood about accessibility. It seems to be working fine.

rpominov:
> The design uses icons for all the visuals. Just a 16px icon inside a 16px container. The icon or the color of the icon changes between variants and that's all there is to it. It's really simple in figma so I was hoping you could build it the same way. No need to worry about sub pixels.
> 
> If this is not possible then I will re-design them to align to the pixel grid and not use icons.

Got it. Will refactor with icons.

rpominov:
@taylornowotny I've switched to icons. Please check again

taylornowotny:
Looking much better, thanks @rpominov 

I noticed the checked and indeterminate icons were smaller than the unchecked box and that was not intentional so I fixed the icon. Please replace the old icon with the new one.

I also noticed that the icon art was not centered inside the 16px container. If you export the whole 16px icon (not just the art layer) then you shouldn't have to worry about this. CSS can't center something with subpixel size but figma can.

rpominov:
> I noticed the checked and indeterminate icons were smaller than the unchecked box and that was not intentional so I fixed them. Please replace the old icons with the new ones.

Ok

> I also noticed that the icon art was not centered inside the 16px container. If you export the whole 16px icon (not just the art layer) then you shouldn't have to worry about this becuase it's already centered. CSS can't center something with subpixel size but figma can.

I import icons with the container as 16x16 SVGs. This is just how browser renders SVGs:

Screenshot 2023-02-24 at 20 41 28

taylornowotny:
Ok, understood and thank you!

PR #1117, opened 2023-02-23
Add target and prefetch to a link

kof:
## Description

Adding target and prefetch to initial properties of a Link component along with the defaults

## Steps for reproduction

1. add a link
2. try target prop
3. try prefetch options

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/custom-components/link.tsx
+++ b/apps/builder/app/canvas/custom-components/link.tsx
@@ -46,10 +46,11 @@ export const preserveBuildParams = (href: string, sourceSearch: string) => {
   return `${url.pathname}${search === "" ? "" : `?${search}`}`;
 };
 
+// @todo this copy-paste has to go away, along with this wrapper component
 type Props = Omit<ComponentProps<"a">, "href" | "target"> & {
   href?: string;
   target?: "self" | "blank" | "parent" | "top";
-  preload?: "none" | "prefetch" | "prerender";
+  prefetch?: "none" | "intent" | "render";
 };
 
 type Ref = ElementRef<"a">;
--- a/apps/builder/app/canvas/custom-components/link.tsx
+++ b/apps/builder/app/canvas/custom-components/link.tsx
@@ -11,10 +11,10 @@ import { preserveSearchBuildParams } from "~/shared/router-utils";
 const isAbsoluteUrl = (href: string) => {
   try {
     new URL(href);
-    return true;
-  } catch (e) {
+  } catch {
     return false;
   }
+  return true;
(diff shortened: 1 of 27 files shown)


apps/builder/app/canvas/custom-components/link.tsx
```diff
@@ -46,7 +46,11 @@ export const preserveBuildParams = (href: string, sourceSearch: string) => {
   return `${url.pathname}${search === "" ? "" : `?${search}`}`;
 };
 
-type Props = Omit<ComponentProps<"a">, "href"> & { href?: string };
+type Props = Omit<ComponentProps<"a">, "href" | "target"> & {
+  href?: string;
+  target?: "self" | "blank" | "parent" | "top";
+  preload?: "none" | "prefetch" | "prerender";
```
rpominov, line 52: What is this `preload` prop? 

Remix has [prefetch](https://remix.run/docs/en/v1/components/link#prefetch) which you already have below. 

And `<a>` doesn't have a `preload` attribute https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

packages/react-sdk/src/components/link.tsx
```diff
@@ -1,9 +1,15 @@
 import { forwardRef, type ElementRef, type ComponentProps } from "react";
 
-type Props = Omit<ComponentProps<"a">, "href"> & { href?: string };
+type Props = Omit<ComponentProps<"a">, "href" | "target"> & {
+  href?: string;
+  target?: "self" | "blank" | "parent" | "top";
+  prefetch?: "none" | "intent" | "render";
```
rpominov, line 6: It's a bit wrong to have preload here since this is not an attribute of regular <a> link. Strictly speaking this prop should be defined in remix-link.tsx, and we should generate separate props for it etc.

But maybe we can be loose here for simplicity, I'd just add a code comment at least.

apps/builder/app/canvas/custom-components/link.tsx
kof: that's me having committed intermediate change and forgotten.
The question is how we ended up not having any errors there? Why are all the link props type copy pasted across the places?

packages/react-sdk/src/components/link.tsx
kof: yes and the way we have it here right now doesn't allow generating the props from the actual component being used, system is broken

rpominov (approved):

Issue #1100, opened 2023-02-22
Update component: Title


taylornowotny:
The Title component was recently updated. Section titles now have states: **open/hover**, **closed**, and **focus**. Clicking a section title should expand/collapse the section. The chevron is gone!

The Title component also includes a "panel with tabs" variant that needs to be built. It contains the new **Panel Tab Trigger** component which has a separate issue #966

https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=0%3A1&t=HLuCPsO7LxFFuq11-1
<img width="1085" alt="Screenshot 2023-02-21 at 4 01 00 PM" src="https://user-images.githubusercontent.com/86495896/220486277-61bb016b-6292-40fa-8975-f5d397869da3.png">

Please let me know if you have questions.

kof:
@rpominov please look into it after properties panel, if its quick to make, lets do it, because its going to clean up the look of the entire style panel.

Additionally I would:
- [ ] persist the open/closed state to local storage (we already have setSetting/setSettings for this)
- [ ] make it open by default

taylornowotny:
To clarify - some section like Backgrounds, will be closed by default, until the user clicks the "+" button in the title to add a background. Then the section will open to show the background that they added. The section design will specify this.

rpominov:
- So sometimes we'll have a "+" button, in which case panel is open when it has items?
- And sometimes we'll have an arrow button, in which case panel is closed by default and can be just opened closed normally?  (Example: "Size" section)

<img width="270" alt="Screenshot 2023-02-22 at 12 43 54" src="https://user-images.githubusercontent.com/825702/220568091-a44ca2d3-628b-49e0-9b11-4f1abefac810.png">

This is kinda confusing. Maybe we should always have an arrow and additionally a plus button sometimes?

kof:
> To clarify - some section like Backgrounds, will be closed by default, until the user clicks the "+" button in the title to add a background. Then the section will open to show the background that they added. The section design will specify this.

@taylornowotny you meant to say they are not closed, but have no items and therefore you only see basically the title container until you add a first item

kof:
@rpominov no, + button is for adding items - for removing items, consistently. The thing about panels like backgrounds and borders is that its a multi-value property and until you have at least one, there is nothing to show.

We have been struggling to find a better option for collapsing ux for entire time and at the end arrived at the same approach Figma has taken.

[...]

Issue #1010, opened 2023-02-13
Props panel: basic control types (Switch, Radios, Checkboxes, Text, MultilineText, Select, Number, Color)


rpominov:
- [ ] Switch
- [ ] Radios
- [ ] Checkboxes
- [ ] Text (example: ID)
- [ ] MultilineText
- [ ] Select
- [ ] Number
- [ ] Color
- [ ] Make sure to call onChange at `blur` when it makes sense

<img width="263" alt="Screenshot 2023-02-13 at 11 36 52" src="https://user-images.githubusercontent.com/825702/218398031-f9c35163-bb5b-4576-8fcb-c9456832a38f.png">

Design for `Number` is not specified explicitly (I forgot to ask for it), but it's shown here:

<img width="263" alt="Screenshot 2023-02-13 at 11 58 15" src="https://user-images.githubusercontent.com/825702/218401905-69eac9ed-3c73-49cb-abd6-e7596fda3add.png">

rpominov:
Mostly done, but will keep open until I get back from vacation and fix or create issues for remaining things like:

- align `TextArea` with Figma
- align `Select` with Figma
- update / use "Color Picker" for color

kof:
Lets make separate issues and close this one

rpominov:
- #1153
- #1155
- #1154

PR #1134, opened 2023-02-27
Props panel UI

### LINKED ISSUE #992: Props panel: Basics

rpominov:
## Description

- Added `[Icon] Label` at the top. Closes #992
- Implemented Controls from #1010 . Underlying components like TextArea or Select are still not aligned with Figma and will need to be updated later
- Cleaned up props panel layout as much as possible without the new Title (#1100)

## Steps for reproduction

1. Open props panel
2. Or: https://props-panel-controls--6382151c8b47d4399fb9fc69.chromatic.com/?path=/story/builder-features-props-panel--story

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/primitives/is-truncated.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const useIsTruncated = (ref: RefObject<HTMLElement>, text: string) => {
+  const [isTruncated, setIsTruncated] = useState(false);
+
+  useEffect(() => {
+    if (ref.current) {
+      const { offsetWidth, scrollWidth } = ref.current;
+      // https://stackoverflow.com/a/10017343/478603
+      setIsTruncated(offsetWidth < scrollWidth);
+    }
+  }, [ref, text]);
```
kof, line 15: ref doesn't change

kof: usually if you need a ref in the deps its a smell

kof: you could use ref function instead, so you can know if element has changed and then set the truncated

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -207,7 +204,7 @@ export const usePropsLogic = ({
   const handleChange = ({ prop, propName }: PropOrName, value: PropValue) => {
     updateProp(
       prop === undefined
-        ? { id: nanoid(), instanceId: instance.id, name: propName, ...value }
+        ? { id: nanoid(), instanceId: instanceId, name: propName, ...value }
```
kof, line 207:
```suggestion
        ? { id: nanoid(), instanceId, name: propName, ...value }
```

packages/design-system/src/components/primitives/is-truncated.tsx
rpominov: Yes. It doesn't change. Having it in deps makes no difference, but if I remove eslint will complain.

apps/builder/app/builder/features/props-panel/use-props-logic.ts
rpominov: Fixed

packages/design-system/src/components/primitives/is-truncated.tsx
rpominov: Added a comment

apps/builder/app/builder/features/props-panel/props-panel.tsx
```diff
@@ -216,16 +255,57 @@ export const PropsPanel = ({
...
+    },
+    deleteProp: (id) => {
+      store.createTransaction([propsStore], (props) => {
+        props.delete(id);
+      });
+    },
+  });
+
+  return (
+    <PropsPanelUI
```
kof, line 304: Lets call it PropsPanel and instead call  this component PropsPanelContainer or something?

Also this component contains only hooks, does it still make sense to have this component at all if it could be all combined in one hook?

apps/builder/app/builder/features/props-panel/props-panel.tsx
```diff
@@ -101,34 +135,31 @@ const Property = ({
...
+}) =>
+  renderControl({
+    meta,
+    prop,
+    propName,
+    onDelete,
+    onChange: (propValue, asset) => {
+      onChange(propValue);
+
+      // @todo: better way to do this?
```
kof, line 147: looks like we could make a better API for type of thing, I bet there is a name for the pattern, it almost looks like a transformer/map but instead it sets the property

kof: not for right now, but we most likely will have more of these in the future

apps/builder/app/builder/features/props-panel/props-panel.tsx
rpominov: > Also this component contains only hooks, does it still make sense to have this component at all if it could be all combined in one hook?

We need `PropsPanelUI` for storybook. If PropsPanelUI would call a hook that contains all logic that would defeat the purpose of separating.

> Lets call it PropsPanel and instead call this component PropsPanelContainer or something?

`PropsPanelContainer` is weird as the name of main exported component that is used in the app.

apps/builder/app/builder/features/props-panel/props-panel.tsx
```diff
@@ -12,24 +11,55 @@ import {
...
   item ? getLabel(item, item.name) : "";
 
+const Row = ({ children, css }: { children: React.ReactNode; css?: CSS }) => (
+  <Flex css={{ px: theme.spacing[9], ...css }} gap="2" direction="column">
+    {children}
+  </Flex>
+);
+
+const InstanceInfo = ({
+  componentMeta,
```
kof, line 47: will have a small conflict with my pr https://github.com/webstudio-is/webstudio-builder/pull/1146

apps/builder/app/builder/features/props-panel/props-panel.tsx
rpominov: Yeah, will need to think how to solve this later. Don't want to do this now. This is not new code added in this PR.

apps/builder/app/builder/features/props-panel/props-panel.tsx
```diff
@@ -12,24 +11,55 @@ import {
...
+import {
+  getComponentPropsMeta,
+  getComponentMeta,
+  type WsComponentMeta,
+} from "@webstudio-is/react-sdk";
 
 const itemToString = (item: NameAndLabel | null) =>
   item ? getLabel(item, item.name) : "";
 
+const Row = ({ children, css }: { children: React.ReactNode; css?: CSS }) => (
```
kof, line 40: is there a benefit for using ReactNode instead of JSX.Element? also you probably want to import the specific type {ReactNode}

apps/builder/app/builder/features/props-panel/controls/color.tsx
```diff
@@ -1,33 +1,46 @@
...
+        value={localValue.value}
+        onChange={(event) => localValue.set(event.target.value)}
+        onBlur={localValue.save}
+        onKeyDown={(event) => {
+          if (event.key === "Enter") {
+            localValue.save();
+          }
         }}
-        css={{ flexGrow: 1 }}
+        css={{ width: 120 }}
```
kof, line 42: same

apps/builder/app/builder/features/props-panel/props-panel.tsx
rpominov: Not sure I understand the problem. `ReactNode` is kinda default type for `children`. So I use it all the time for `children`. For example ` ComponentProps<"div">["children"] === ReactNode`

apps/builder/app/builder/features/props-panel/controls/number.tsx
```diff
@@ -1,34 +1,51 @@
...
+          );
+        }}
+        onBlur={localValue.save}
+        onKeyDown={(event) => {
+          if (event.key === "Enter") {
+            localValue.save();
+          }
         }}
-        css={{ flexGrow: 1 }}
+        css={{ width: 98 }}
```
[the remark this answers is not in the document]
rpominov: From Figma. I think it's ok to deviate sometimes from spacing values we decided to use.

@taylornowotny if you want to fix widths of controls in Figma (the one that are 98 and 120), I'll update in code later.

apps/builder/app/builder/features/props-panel/controls/color.tsx
rpominov: Commented above

apps/builder/app/builder/features/props-panel/props-panel.tsx
rpominov: Fixed the import

apps/builder/app/builder/features/props-panel/controls/number.tsx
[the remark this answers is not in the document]
kof: Taylor is OOO right now, lets create an issue and a @todo with a link

apps/builder/app/builder/features/props-panel/props-panel.tsx
kof: we need a standard convention for this type of components again

apps/builder/app/builder/features/props-panel/controls/number.tsx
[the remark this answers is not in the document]
rpominov: https://github.com/webstudio-is/webstudio-builder/issues/1150

kof (approved):

Issue #877, opened 2023-01-27
Backgrounds section redesign


kof:
https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=2440%3A48041

What's left:

- [x] - Use + in the title instead of "Add background" button
- [x] - Focus should work inside layer items using arrows (left/right)
- [x] - Between layers using (top/down)
- [x] - Label colors should be moved somewhere outside of Layer items
- [x] - Fix theme.spacing, do not use px values

istarkov:
# Steps to implement.


## Implement background panel 

- [ ] - Done

<img width="126" alt="image" src="https://user-images.githubusercontent.com/5077042/220930435-6d50c86a-533f-4530-97f6-95adeb4b4f8e.png">

Initially, with "Add background" button, there is no drag/reorder action and hide icon.

Add a floating panel and put inside it the existing background panel section.

## Implement  Floating panel

- [ ] - Done

<img width="121" alt="image" src="https://user-images.githubusercontent.com/5077042/220933797-36875e65-1679-4507-bb43-1567319b7ff5.png">

Omitting position control.

<img width="53" alt="image" src="https://user-images.githubusercontent.com/5077042/220934031-56ab6d87-432f-4421-9952-23e4ef747bda.png">

Code control doesn't parse or do anything with inserted gradient and uses it as is.

## Parse code control at linear gradient section.

- [ ] - Done

Light parsing just extracts gradients without full css `background` parsing.


## Last steps

- [ ] - Add position control.
<img width="53" alt="image" src="https://user-images.githubusercontent.com/5077042/220934031-56ab6d87-432f-4421-9952-23e4ef747bda.png">

- [ ] - Use `+` in the title instead of "Add background" button
- [ ] - Add drag/reordering for the background panel
- [ ] - Add hide icon.
- [ ] - Probably make better parsing of gradient code section i.e. full css background parse.

taylornowotny:
Docs finished! A lot has changed since yesterday so please inspect everything carefully.

The background-color property is now a permanent control, rather than a list item, in response to your feedback.

The position control will be a separate issue. This is a new component and requires some explanation.

Added backgrounds are **CSS Value List Item** component #1118. These will be used again in other sections.

Please let me know if I can assist in any way.

taylornowotny:
[...]

PR #1135, opened 2023-02-27
Css value list item control

### LINKED ISSUE #1118: new CSS Value List Item component
This component is needed for the Backgrounds section #877 and other future sections. 

The thumbnail is a nested **Thumbnail** component. It's nothing special, I just organized it this way so the thumbnail could be changed on the list item without breaking the instance in figma. Let me know if I need to make a separate issue for the thumbnail.

https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=0%3A1&t=GFXJS5fMcAiXXhiQ-1
Screenshot 2023-02-23 at 4 19 05 PM

istarkov:
## Description

closes #1118 ref #877

Added:

[SmallToggleButton](https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=4%3A3236&t=Nn8L5VPCruHNv1no-0) and [CssValueListItem](https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=2002%3A6155&t=Nn8L5VPCruHNv1no-0) components

Changed 
SmallIconButton - state now open instead of hover, reused base component SmallButton.
Label - had wrong height 22px instead of 20, changed.

Pictures.

image

image

image

## Steps for reproduction

See in storybook above components

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/css-value-list-item.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+const Thumbnail = styled("div", {
+  width: theme.spacing[10],
+  height: theme.spacing[10],
+  backgroundImage: "linear-gradient(yellow, red)",
+});
+
+const ListItem = (props: {
+  hidden: boolean;
+  labelColor: "default" | "preset" | "local" | "remote";
```
kof, line 40: this is also the same with style source UI, but we put so far style source UI into builder, probably need to move it over to design system as well cc @TrySound

packages/design-system/src/components/css-value-list-item.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+  state: undefined | "open";
```
kof, line 41: just optional works too
```suggestion
  state?: "open";
```

packages/design-system/src/components/css-value-list-item.tsx
```diff
@@ -0,0 +1,106 @@
...
+  },
+});
+
+type Props = ComponentProps<typeof Item> & {
+  hidden?: boolean;
+  label: React.ReactElement;
+  thumbnail: React.ReactElement;
+  buttons?: React.ReactElement;
+  // to support Radix trigger childAs
+  ["data-state"]?: "open";
```
kof, line 53: This didn't work?
```suggestion
  "data-state"?: "open";
```

packages/design-system/src/components/css-value-list-item.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Flex gap={2}>
+        <ThumbHolder>{thumbnail}</ThumbHolder>
+        {label}
+      </Flex>
+
+      <Flex grow={true} />
+
+      <Flex
+        gap={2}
+        onClick={(e) => {
```
kof, line 91: please add a linter, I am tired of being a linter for this

```suggestion
        onClick={(event) => {
```

packages/design-system/src/components/css-value-list-item.stories.tsx
istarkov: It's a storybook and I want to explicitly pass undefined to be sure I checked most variants.
So its intented
image

packages/design-system/src/components/small-icon-button.stories.tsx
```diff
@@ -13,6 +13,8 @@ const iconsMap = {
   "<TrashIcon>": <TrashIcon />,
 } as const;
 
+const states = [undefined, ...smallIconButtonStates];
```
kof, line 16: that undefined looks weird

kof (approved):

packages/design-system/src/components/small-icon-button.stories.tsx
istarkov: I don't want support useless states in types. Having that they used for storybook only

rpominov:
> Label - had wrong height 22px instead of 20, changed.

Fixed this too. In main

packages/design-system/src/components/css-value-list-item.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+import * as React from "react";
+import { SmallIconButton } from "./small-icon-button";
+import {
+  FloatingPanelPopover,
+  FloatingPanelPopoverContent,
+  FloatingPanelPopoverTrigger,
+} from "./floating-panel-popover";
+import { StorySection, StoryGrid } from "./storybook";
+
+export default {
```
rpominov, line 19: If this is a demo of a component from Figma, please put under "libaray" (add `title` here same as in other files)

packages/design-system/src/components/primitives/small-button.tsx
```diff
@@ -0,0 +1,103 @@
+/**
+ * Implementation of the "Small Icon Button" component from:
+ * https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=4%3A3171
+ */
```
rpominov, line 4: This should be in `small-icon-button.tsx`, but not here

packages/design-system/src/components/small-icon-button.tsx
```diff
@@ -9,104 +9,43 @@ import {
...
-  variant?: typeof smallIconButtonVariants[number];
-  state?: typeof smallIconButtonStates[number];
+  variant?: (typeof smallIconButtonVariants)[number];
+  state?: (typeof smallIconButtonStates)[number];
   focused?: boolean;
   css?: CSS;
-
   // might be set when <SmallIconButton> is asChild
-  "data-state"?: string;
+  "data-state"?: (typeof smallIconButtonStates)[number];
```
rpominov, line 32: This type is wrong. It should be just `string` and then a check in code. We don't control what radix may put in `data-state`

rpominov (approved):
Good otherwise

packages/design-system/src/components/small-toggle-button.tsx
```diff
@@ -0,0 +1,57 @@
...
+    ref: Ref<HTMLButtonElement>
+  ) => {
+    return (
+      <TogglePrimitive.Root
+        asChild
+        defaultPressed={defaultPressed}
+        pressed={pressed}
+        onPressedChange={onPressedChange}
+        disabled={disabled}
+      >
```
rpominov, line 49: We usually do a spread here to support everything that the underlying Radix component support instead of picking individual props

packages/icons/icons/drag-handle.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <rect y="3" x="5" width="6" height="1" rx="0.5" />
+  <rect y="6" x="5" width="6" height="1" rx="0.5" />
+  <rect y="9" x="5" width="6" height="1" rx="0.5" />
+  <rect y="12" x="5" width="6" height="1" rx="0.5" />
+</svg>
```
rpominov, line 6: Not sure we should put drag handle in icons. Because it's not an icon from Figma point of view. I would probably inline svg in the component code.

And if it's an 16x16 icon you're probably have trouble with layout later as it should be smaller

Screenshot 2023-02-27 at 19 31 00

packages/design-system/src/components/small-icon-button.tsx
istarkov: But we may control what we support exactly. I explicitly set only supported state here.

rpominov: The problem is we won't get a TS error when Remix passes something we don't support

PR #1140, opened 2023-02-27
Base breakpoint

kof:
## Description

- rename default mobile to base
- set initial canvas width to the max size within the lowest breakpoint boundary

## Steps for reproduction

1. create new project
2. see that by default you are on base breakpoint
3. see that the initial width is at max width in the before the first next breakpoint begins (767px)

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


taylornowotny (approved):

istarkov:
Now if my screen is not set to a full width I always get 767px width of the screen 
image

What hides left and right part of the canvas
image

So the minimal width we support now is 1330px.

Moreover now impossible to do mobile development, for example, lets see common viewport widths here https://experienceleague.adobe.com/docs/target/using/experiences/vec/mobile-viewports.html

This makes impossible to develop sites for 50+% of the web devices

apps/builder/app/builder/features/breakpoints/use-update-canvas-width.ts
```diff
@@ -3,21 +3,27 @@ import { useCallback, useEffect } from "react";
...
 
   // Ensure the size is within currently selected breakpoint when returning to design mode out of preview mode,
   // because preview mode enables resizing without constraining to the selected breakpoint.
   useEffect(() => {
     if (isPreviewMode || selectedBreakpoint === undefined) {
       return;
     }
-    setCanvasWidth(Math.max(selectedBreakpoint.minWidth, minWidth));
-  }, [isPreviewMode, selectedBreakpoint, setCanvasWidth]);
+    const width = Math.max(
```
istarkov, line 21: At this point, we can measure the width of iframe parent and use it here to not set 767px when awailable width is less

istarkov (approved):

apps/builder/app/builder/features/breakpoints/use-update-canvas-width.ts
kof: done

TrySound (approved):

PR #1146, opened 2023-02-28
Instance name setting

kof:
## Description

Closes https://github.com/webstudio-is/webstudio-builder/issues/1143

Allows to select an instance and change name via settings > name

Used in navigator, outlines, props panel (design untouched, will be updated soon)

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/shared/tree/index.tsx
```diff
@@ -98,3 +104,10 @@ export const InstanceTreeNode = (
     renderItem={instanceRelatedProps.renderItem}
   />
 );
+
+export const getInstanceLabel = (
+  instance: { label?: string },
+  meta: WsComponentMeta
+) => {
+  return instance.label ? humanizeString(instance.label) : meta.label;
```
TrySound, line 112: Do we need to "humanize" user input? Can be confusing.

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
```diff
@@ -0,0 +1,44 @@
...
+type SettingUpdate = { setting: "label"; value: string };
+
+export const useSettingsLogic = ({
+  selectedInstance,
+}: {
+  selectedInstance: Instance;
+}) => {
+  const changes = useRef<Map<SettingUpdate["setting"], SettingUpdate["value"]>>(
+    new Map()
+  );
```
TrySound, line 15: Types won't work if one setting change value type. Let it be just an object. Map doesn't optimize here anything.

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
```diff
@@ -0,0 +1,44 @@
...
+  };
+
+  const update = () => {
+    store.createTransaction([instancesStore], (instances) => {
+      const instance = instances.get(selectedInstance.id);
+      if (instance !== undefined) {
+        instance.label = changes.current.get("label");
+      }
+    });
+  };
```
TrySound, line 32: updateLabel

apps/builder/app/builder/shared/tree/index.tsx
kof: Idk, let me check with figma and others, would be really nice to have consistent naming without working hard to do so

kof: removed

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
```diff
@@ -0,0 +1,44 @@
...
+    new Map()
+  );
+
+  const set = ({ setting, value }: SettingUpdate) => {
+    switch (setting) {
+      case "label": {
+        changes.current.set(setting, value);
+      }
+    }
+  };
```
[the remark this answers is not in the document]
kof: wdym?  instead of "label" use "setLabel" ? setting is meant as a name of the setting, noun

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
kof: this function is meant for all the things, but its also internal and not exported/returned, and we don't know what we are going to need in the future, maybe we will indeed need separate functions for each update

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const SettingsPanel = ({ selectedInstance }: SettingsPanelProps) => {
+  const { set, handlers } = useSettingsLogic({ selectedInstance });
+  const label = getComponentMeta(selectedInstance.component)?.label;
+  return (
+    <Flex css={{ px: theme.spacing[9] }}>
+      <Flex gap="1" direction="column" grow>
+        <Label>Instance Name</Label>
+        <TextField
+          {...handlers}
```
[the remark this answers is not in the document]
kof: pls remind me what am I missing here with this spread? seems to infer everything correctly?  @istarkov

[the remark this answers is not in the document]
TrySound: Easy to pass not existing fields. Prisma for example breaks when spread null and gives any as result.

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
```diff
@@ -20,9 +20,7 @@ export const SettingsPanel = ({ selectedInstance }: SettingsPanelProps) => {
           defaultValue={selectedInstance.label}
           onChange={(event) => {
             const { value } = event.target;
-            if (value.trim()) {
-              set({ setting: "label", value });
-            }
+            value.trim() ? set("label", value) : set("label", undefined);
```
TrySound, line 23: Why not if/else? Ternary is harder to understand when operands effectful

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
```diff
@@ -0,0 +1,45 @@
...
+export const useSettingsLogic = ({
+  selectedInstance,
+}: {
+  selectedInstance: Instance;
+}) => {
+  const changes = useRef<SettingUpdate>({});
+
+  const set = (
+    setting: keyof SettingUpdate,
+    value: SettingUpdate[keyof SettingUpdate]
```
TrySound, line 17: Now value is not collocated with key.

I would not prematurely complicate this code. It would even more clear to have all logic inside settings panel component.

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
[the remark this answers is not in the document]
istarkov: Yes, remember a bug a few weeks ago about scrub is no longer works, containerRef baseRef renaming and we had no warning because of the spread

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
kof: I hate else's haha, but I can make the ternary effectless...

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
kof: Yeah, its a bit premature, but I think its not terrible and we will need to split it for sure

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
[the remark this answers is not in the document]
kof: right, that was it, making explicit

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
kof: > Now value is not collocated with key.

still getting typechecked in the sense you can set a wrong value with a wrong key

kof: oh nooo, since its optional an object, it now allows to set unexisting value

kof: ah no, all good

TrySound: I mean if the other key will have number value will be `string | number` on all keys. So you loose type safety here. You can use generic or just use named functions like setLabel for simplicity. This is not generic logic to keep it abstracted.

kof: true, didn't think further

apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const useSettingsLogic = ({
+  selectedInstance,
+}: {
+  selectedInstance: Instance;
+}) => {
+  const changes = useRef<SettingUpdate>({});
+
+  const setLabel = (value: SettingUpdate[keyof SettingUpdate]) => {
+    changes.current.label = value || undefined;
```
TrySound, line 16: add comment why || so nobody automatically refactor to ??

PR #1148, opened 2023-03-01
Make Image value not an Array

istarkov:
## Description

ref #877

As a part of above I'm changing 
```
export const ImageValue = z.object({
  type: z.literal("image"),
  value: z.array(z.object({ type: z.literal("asset"), value: ImageAsset })),
});
```
to 
```
export const ImageValue = z.object({
  type: z.literal("image"),
  value: z.object({ type: z.literal("asset"), value: ImageAsset }),
});
```
In favour of using in next PRs 

```
const ArrayValue = z.object({
  type: z.literal("array"),
  value: z.array(z.union([UnitValue, KeywordValue, UnparsedValue, ImageValue])),
});
```

For now, to not break compatibility with existing code I'm not converting to an array immediately. 
Just style types simplification.

'array' type is not now supported on db level, parse/serialize will not work. Will be done in following PRs

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound:
Maybe use "List" instead of "Array"? How is it called in css?

istarkov:
I think to rename as layers

TrySound (approved):

PR #1149, opened 2023-03-01
Background section redesign part 1.1

istarkov:
## Description

ref #877
As in this comment part 1 https://github.com/webstudio-is/webstudio-builder/issues/877#issuecomment-1441897685

image

You can add/remove layers. View/Hide not working. No drag.

## Steps for reproduction

Create new project. (existing could not work well because of migration)
Create background layers. Add images. Set various background layers params. 

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.ts
```diff
@@ -0,0 +1,252 @@
...
+        newPropertyStyle = structuredClone(propertyStyle);
+      } else if (styleValue?.cascaded?.value.type === "layers") {
+        propertyChanged = true;
+        newPropertyStyle = structuredClone(styleValue?.cascaded?.value);
+      } else {
+        propertyChanged = true;
+        newPropertyStyle = { type: "layers", value: [] };
+      }
+
+      // All layers must have the same number of layers
```
istarkov, line 168: // All properties must have ...

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-thumbnail.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+export const LayerThumbnail = (props: { layerStyle: StyleInfo }) => {
+  const backgroundImageStyle = props.layerStyle.backgroundImage?.value;
+
+  if (backgroundImageStyle?.type === "image") {
+    const asset = backgroundImageStyle.value.value;
+    const remoteLocation = asset.location === "REMOTE";
+
+    // Avoid image flickering on switching from preview to asset (during upload)
```
istarkov, line 66: remove

apps/builder/app/builder/features/style-panel/sections/backgrounds/backgrounds.tsx
```diff
@@ -1,19 +1,149 @@
...
+            createBatchUpdate
+          )}
+          deleteProperty={(property, options) => {
+            deleteLayerProperty(
+              layerNum,
+              currentStyle,
+              deleteProperty,
+              createBatchUpdate
+            )(property, options);
+          }}
```
TrySound, line 117:
```
          deleteProperty={deleteLayerProperty(
               layerNum,
               currentStyle,
               deleteProperty,
               createBatchUpdate
             )}
```

apps/builder/app/builder/features/style-panel/sections/backgrounds/backgrounds.tsx
```diff
@@ -1,19 +1,149 @@
...
-    <>
-      {styleConfigsByCategory.map((entry) => renderProperty(entry))}
-      <ShowMore
-        styleConfigs={moreStyleConfigsByCategory.map((entry) =>
-          renderProperty(entry)
-        )}
-      />
-    </>
+    <Flex gap={1} direction="column">
+      {Array.from(Array(layersCount).keys()).map((layerNum) => (
```
TrySound, line 100:
```
Array.from(Array(layersCount), (_, layerNum) => (
```

TrySound (approved):

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
```diff
@@ -0,0 +1,201 @@
...
+  currentStyle: StyleInfo;
+  setProperty: SetBackgroundProperty;
+  deleteProperty: DeleteBackgroundProperty;
+};
+
+const safeDeleteProperty = (deleteProperty: DeleteBackgroundProperty) => {
+  const result: DeleteProperty = (property, options) => {
+    // isBackgroundLayeredProperty is typeguard and ts don't understand === false
+    if (!isBackgroundLayeredProperty(property)) {
+      throw new Error("");
```
kof, line 36: error message?

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
```diff
@@ -0,0 +1,201 @@
...
+type BackgroundContentProps = {
+  currentStyle: StyleInfo;
+  setProperty: SetBackgroundProperty;
+  deleteProperty: DeleteBackgroundProperty;
+};
+
+const safeDeleteProperty = (deleteProperty: DeleteBackgroundProperty) => {
+  const result: DeleteProperty = (property, options) => {
+    // isBackgroundLayeredProperty is typeguard and ts don't understand === false
+    if (!isBackgroundLayeredProperty(property)) {
```
kof, line 35:
```suggestion
    if (isBackgroundLayeredProperty(property) === false) {
```

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
```diff
@@ -0,0 +1,201 @@
...
+  };
+  return result;
+};
+
+const safeSetProperty = (setBackgroundProperty: SetBackgroundProperty) => {
+  const result: SetProperty = (property) => {
+    // isBackgroundLayeredProperty is typeguard and ts don't understand === false
+    if (!isBackgroundLayeredProperty(property)) {
+      throw new Error("");
+    }
```
kof, line 48: same

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
kof: uhh didn't see the comment, wtf

kof: what if you assign to a constant first?

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
```diff
@@ -0,0 +1,201 @@
...
+Not stackable:
+
+background-color
+*/
+
+export const BackgroundContent = (props: BackgroundContentProps) => {
+  const setProperty = safeSetProperty(props.setProperty);
+  const deleteProperty = safeDeleteProperty(props.deleteProperty);
+
+  const eltRef = useRef<HTMLDivElement>(null);
```
kof, line 96: elementRef

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.test.ts
```diff
@@ -0,0 +1,177 @@
...
+import { describe, test } from "@jest/globals";
+import type { StyleInfo } from "../../shared/style-info";
+import type { CreateBatchUpdate } from "../../shared/use-style-data";
+import type {
+  LayersValue,
+  StyleProperty,
+  StyleValue,
+} from "@webstudio-is/css-data";
+
+// @todo remove at node18
```
kof, line 16: can we add it from some central place?

kof: why do you even need a structured clone for tests?

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.test.ts
```diff
@@ -0,0 +1,177 @@
...
+globalThis.structuredClone = (v) => JSON.parse(JSON.stringify(v));
```
kof, line 17: value, not v

kof: please reread everything and remove abbreviations, a lot of them, not commenting each

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.ts
```diff
@@ -0,0 +1,252 @@
...
+    const styleValue = style[property];
+
+    const valueStyle = styleValue?.value;
+    const localStyle = styleValue?.local;
+    const cascadedStyle = styleValue?.cascaded;
+
+    if (valueStyle?.type === "layers") {
+      const styleValue = valueStyle.value[layerNum];
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      result[property]!["value"] = styleValue;
```
kof, line 98: you could have avoided it, I bet it doesn't know what type you are operating on, so it can't tell if there is a value property

kof: all of these non-null assertions look fishy to me, I bet you can do better than this

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-thumbnail.tsx
```diff
@@ -0,0 +1,89 @@
...
+  width: theme.spacing[10],
+  height: theme.spacing[10],
+  backgroundImage: "linear-gradient(yellow, red)",
+});
+
+const NoneThumbnail = styled("div", {
+  width: theme.spacing[10],
+  height: theme.spacing[10],
+  background:
+    "repeating-conic-gradient(rgba(0,0,0,0.22) 0% 25%, transparent 0% 50%) 0% 33.33% / 40% 40%",
```
kof, line 17: do we have that gradient in the tokens eventually?

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-thumbnail.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  if (backgroundImageStyle?.type === "image") {
+    const asset = backgroundImageStyle.value.value;
+    const remoteLocation = asset.location === "REMOTE";
+
+    const loader = remoteLocation
+      ? loaders.cloudflareImageLoader({
+          resizeOrigin: env.RESIZE_ORIGIN,
+        })
+      : loaders.localImageLoader();
```
kof, line 70: this logic probably already exists in the image package, aka getLoaderByAssetLocation()

packages/project-build/src/db/styles.ts
```diff
@@ -97,7 +145,9 @@ export const parseStyles = async (
 /**
  * prepare value to store in db
  */
-const serializeValue = (styleValue: StyleDecl["value"]) => {
+const serializeValue = (
```
kof, line 148: same for serializeStyles, needs tests

kof (changes requested):
lets put in a little more work, but if you really need to merge, then lets merge and do follow-up

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
istarkov: Nothing except the above works

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.test.ts
istarkov: I wasn't able to add it to jest via globals as usual. We need it for the test because it's used in the client code I'm testing.
In any case, we need to switch on the modern node and this will gone.

kof: needs an issue

istarkov: done https://github.com/webstudio-is/webstudio-builder/issues/1191

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-thumbnail.tsx
istarkov: This would need for the image package knowledge which is better to avoid 
like what is the asset, what is asset remote etc

And I think that during this https://github.com/webstudio-is/webstudio-builder/issues/1179 we would try to avoid any code like this at all, as it becoming hard

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-thumbnail.tsx
istarkov: This is used in Image Control, its Chessboard for transparent background. I used it here that instead of webflow we have `none` default layer. And it's mostly the same as transparent.

packages/project-build/src/db/styles.ts
```diff
@@ -19,7 +19,44 @@ import {
 const parseValue = (
```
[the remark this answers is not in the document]
istarkov: From what I know Bogdan wants to move this logic on the client, out of here. 
It looks like a time waste to write tests for this logic.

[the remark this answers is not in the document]
TrySound: Yes, to make collaboration work this should be moved client side. Need to figure out how to integrate assets into css engines.

packages/project-build/src/db/styles.ts
TrySound: Need to kill this code

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
kof: here is how I would write it otherwise

```ts
const safeDeleteProperty = (deleteProperty: DeleteBackgroundProperty): DeleteProperty => {
  return (property, options) => {
    const isLayered = isBackgroundLayeredProperty(property);
    if (isLayered) {
      return deleteProperty(property, options);
    }
    throw new Error(`Property ${property} should be background style property`);
  };
};
```

kof: cc @Andarist another thing where typescript is definitely being crazy stupid, maybe you already know the issue for that?

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.ts
```diff
@@ -0,0 +1,254 @@
...
+export const setLayerProperty =
+  (layerNum: number, style: StyleInfo, createBatchUpdate: CreateBatchUpdate) =>
+  (propertyName: LayeredBackgroundProperty) =>
+  (newValue: BackgroundStyleValue, options?: StyleUpdateOptions) => {
+    const batch = createBatchUpdate();
+
+    const layerCount = Math.max(getLayerCount(style), layerNum + 1);
+
+    for (const property of layeredBackgroundProps) {
+      let propertyChanged = false;
```
kof, line 151: this style with a flag that changes depending on conditions in a long block is generally problematic: hard to read, easy to break,
much better would be an early return for each case

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-layers.ts
```diff
@@ -0,0 +1,254 @@
...
+      if (newPropertyStyle.value.length < layerCount) {
+        propertyChanged = true;
+        newPropertyStyle.value = newPropertyStyle.value.concat(
+          new Array(layerCount - newPropertyStyle.value.length).fill(
+            layeredBackgroundPropsDefaults[property]
+          )
+        );
+      }
+
+      if (newPropertyStyle.value.length > layerCount) {
```
kof, line 180: this condition is the same as the one in line 171

kof: ah sorry, didn't realize the arrow pointing the other direction

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
Andarist: It's interesting that `=== false` doesn't work. I guess that it could, it's just probably `!` is more "idiomatic"/common in JS and that has been implemented in TS. 

The last suggestion by @kof works though (with both aliased `isLayered` and without that "alias"). So the "key" here is to flip the structure of the code and handle the error case later or to use `!`. 

Note that, in a similar vein, `=== true` doesn't work either (just in the flipped situation).

Andarist: Do u have a coding style that prescribed `=== false/true`?

packages/project-build/src/schema/styles.ts
```diff
@@ -3,19 +3,37 @@ import {
...
+      StoredImageValue,
+      InvalidValue,
+    ])
+  ),
+});
+
 export const StoredStyleDecl = z.object({
   styleSourceId: z.string(),
   breakpointId: z.string(),
   // @todo can't figure out how to make property to be enum
```
kof, line 34: this todo still valid?

kof (approved):

packages/project-build/src/db/styles.ts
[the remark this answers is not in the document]
kof: we can still have tests and move that logic together with tests?

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
kof: I think this is a terrible ts implementation that should be treated like a bug, it literally disrespects javascript's language semantics

kof: how do I create a minimal reproducible case for this?

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAlhAgsKcBuBTGBeGAzAQwBsIMAoUSWNYgVy1wAoBKHAPhgG8yZ48ZGCZKkytuvXgCcMUWpLA8YAXxiLps+VygBPAA4YAXDACMSsmbJx+gpCnQNsuQiQxjFNIvRYA6HfoDc5mSW1gCEQnaiXO50GD5+GIFKQA

Andarist: Minimal repro for the said issue:
https://www.typescriptlang.org/play?ts=5.1.0-dev.20230308#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXywGcAVGZDACwE8AKHAIwCsAueNAa1RwHdUBKVg0YFC8AN4BfAFBTQkWAjB5CGeENZikOHKxUwsqAOYBueBPgAfeKmQQIMrIng0ipctTpM+fcVPhqmADpEbSlpKUdnVzJKWiEfAF4k+AwyEB8xPwDGYNCJIA

PR #1162, opened 2023-03-01
Scroll navigator into view when selecting instances on canvas/navigator

kof:
Closes https://github.com/webstudio-is/webstudio-builder/issues/1157

## Description

When selecting in navigator - scroll into view on canvas and vice-versa

## Steps for reproduction

1. insert lots of elements to make navigator and canvas scroll
2. click in navigator last and first - see canvas scrolling
3. same as 2 but for canvas

## Code Review

  - test it on preview
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/tree/tree-node.tsx
+++ b/packages/design-system/src/components/tree/tree-node.tsx
@@ -282,6 +282,16 @@ export const TreeItemBody = <Data extends { id: string }>({
   const isDragging = dropTargetItemId !== undefined;
   const isDropTarget = dropTargetItemId === itemData.id;
 
+  // Scroll the selected button into view when selected from canvas.
+  useEffect(() => {
+    if (isSelected && isDragging === false && isDropTarget === false) {
+      itemButtonRef.current?.scrollIntoView({
+        behavior: "smooth",
+        block: "nearest",
+      });
+    }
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -282,6 +282,16 @@ export const TreeItemBody = <Data extends { id: string }>({
...
 
+  // Scroll the selected button into view when selected from canvas.
+  useEffect(() => {
+    if (isSelected && isDragging === false && isDropTarget === false) {
+      itemButtonRef.current?.scrollIntoView({
+        behavior: "smooth",
+        block: "nearest",
+      });
+    }
+  }, [isSelected, isDragging, isDropTarget]);
```
rpominov, line 293: Seems like this will cause scroll not only when it becomes selected, but also when `isDragging` or `isDropTarget` becomes `false`.

Not sure this is a real scenario, but I would use refs to make `isSelected` the only dependency just in case

rpominov (approved):
Otherwise looks good

packages/design-system/src/components/tree/tree-node.tsx
kof: yep, done

PR #1166, opened 2023-03-02
Fix infinite dependencies loop

kof:
## Description

Naming instance over settings has infinite update loop

## Steps for reproduction

1. select instance
2. go to settings
3. change name
4. hit enter
5. should not land in the loop

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
+++ b/apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
@@ -16,6 +16,8 @@ export const useSettingsLogic = ({
   selectedInstance: Instance;
 }) => {
   const changes = useRef<SettingUpdate>({});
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/settings-panel/use-settings-logic.ts
```diff
@@ -24,12 +26,12 @@ export const useSettingsLogic = ({
 
   const updateLabel = useCallback(() => {
     store.createTransaction([instancesStore], (instances) => {
```
TrySound, line 28: const selectedInstance = selectedInstanceStore.get()

TrySound (approved):

PR #1168, opened 2023-03-02
Markdown bugs

kof:
## Description

1. Don't handle global paste even when inside content-editable (before that when you paste inside texteditable, it would insert an instance additionally outside)
2. Insert links in paragraphs as RichTextLink
3. Make sure rich text link has the same props as link

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/copy-paste/init-copy-paste.ts
+++ b/apps/builder/app/shared/copy-paste/init-copy-paste.ts
@@ -115,7 +115,7 @@ export const initCopyPaste = (options: Options) => {
       document.removeEventListener("cut", handleCut);
     }
     if (onPaste) {
-      document.removeEventListener("paste", handlePaste);
+      document.removeEventListener("paste", handlePaste, { capture: true });
     }
   };
 };
--- a/apps/builder/app/shared/copy-paste/init-copy-paste.ts
+++ b/apps/builder/app/shared/copy-paste/init-copy-paste.ts
@@ -1,5 +1,3 @@
-import { textEditingInstanceIdStore } from "../nano-states";
-
 const isValidClipboardEvent = (event: ClipboardEvent) => {
   const selection = document.getSelection();
   if (selection?.type === "Range") {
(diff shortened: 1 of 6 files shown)


apps/builder/app/shared/copy-paste/init-copy-paste.ts
```diff
@@ -77,6 +79,9 @@ export const initCopyPaste = (options: Options) => {
   };
 
   const handlePaste = (event: ClipboardEvent) => {
+    if (textEditingInstanceIdStore.get()) {
+      return;
+    }
```
TrySound, line 84: Found more solid solution. Here we already check contenteditable as invalid element. Though the problem is that lexical catches it earlier.

We can fixed it by capturing event. Add `{ capture: true}` for paste event.

kof: I tried, but the problem is when I get the event, the target will be already removed from the dom and replaced by lexical, so I can't properly detect contenteditable.

kof: let me try some more tricks

TrySound: Yes, I get this too with current event. Though capture worked before lexical removed element from dom.

kof: you were right, we even have that check, I must have somehow removed the capture or something

apps/builder/app/shared/copy-paste/init-copy-paste.ts
```diff
@@ -107,7 +102,9 @@ export const initCopyPaste = (options: Options) => {
     document.addEventListener("cut", handleCut);
   }
   if (onPaste) {
-    document.addEventListener("paste", handlePaste);
+    // Capture is required so we get the element before content-editable removes it
+    // This way we can detect when we are inside content-editable and ignore the event
+    document.addEventListener("paste", handlePaste, { capture: true });
```
TrySound, line 107: Probably need for removeEventListener too

kof: wow first time seeing that capture is required for removal as well

TrySound (approved):

PR #1174, opened 2023-03-03
Link block sdk component

### LINKED ISSUE #1173: Link Block component
To have a link that can contain instances instead of text we need a non-editable link block component.

kof:
Closes #1173 

## Description

- link block component that allows adding child instances
- also replaced box icon

## Steps for reproduction

1. add link block
2. insert image component and text component
3. select it
4. add regular link, see if it still works correctly
5. add text-editable link, see if it still works correctly

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/components/link-block.ws.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { props } from "./__generated__/link-block.props";
+
+const presetStyle = {
+  boxSizing: {
+    type: "keyword",
+    value: "border-box",
+  },
+  display: {
+    type: "keyword",
+    value: "inline-block",
```
TrySound, line 12: Should it be inline block?

kof: yeah, its more likely use case, also wf uses it

TrySound (approved):

rpominov:
Maybe I'm missing something, but it's strange that adding this didn't require to create additional two wrappers as we do with other links. Probably this new link doesn't support Remix etc.

kof:
True, the block links now don't have the ability to link to without full page reload

kof:
Fixing it now

PR #1178, opened 2023-03-05
Right types for the RootInstance

istarkov:
## Description

RootInstance dont need all the data we pass

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/tree/root.ts
+++ b/packages/react-sdk/src/tree/root.ts
@@ -23,8 +23,12 @@ export type Data = {
   params?: Params;
 };
 
+export type RootPropsData = Omit<Data, "build"> & {
+  build: Pick<Data["build"], "instances" | "props">;
+};
+
 type RootProps = {
-  data: Data;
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/tree/root.ts
```diff
@@ -23,8 +23,12 @@ export type Data = {
   params?: Params;
 };
 
+export type RootPropsData = Omit<Data, "build"> & {
+  build: Pick<Data["build"], "instances" | "props">;
+};
```
TrySound, line 28: Could it be `Omit<Data, "build"> & { instances: Build['instances'], props: Build['props'] }`?

istarkov: In the future definitely yes, for now it just same type as used at saas

TrySound (approved):

PR #1180, opened 2023-03-06
Object Fit & object position

kof:
## Description

This implements object fit and inside a position control to be reused with background position and others.

https://user-images.githubusercontent.com/52824/223282840-92b54da6-686d-45fe-8fe8-4422beb7572e.mp4

## Steps for reproduction

1. sizes section
2. object position

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-engine/src/core/to-value.ts
```diff
@@ -74,6 +74,10 @@ export const toValue = (
     return value.value.map((v) => toValue(v, options)).join(",");
   }
 
+  if (value.type === "tuple") {
+    return `${toValue(value.value[0])} ${toValue(value.value[1])}`;
```
istarkov, line 78: value.value is array, why not `map(toValue).join(' ')`

kof: done

apps/builder/app/builder/features/style-panel/controls/position/position.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+const toPosition = (value: TupleValue) => {
+  return {
+    left: value.value[0].value,
+    top: value.value[1].value,
+  };
+};
+
+// @todo SetValue has legacy string value support, that needs to be removed
+const toTuple = (
```
istarkov, line 21: The comment is confusing,  `string` now can come because of computed style  not because of setValue
(setProperty uses toStyleValue inside, so no strings in our styles)

packages/design-system/src/components/icon-button.tsx
```diff
@@ -73,7 +77,7 @@ export const IconButton = styled("button", {
       },
       active: {
         backgroundColor: theme.colors.blue10,
-        color: "White",
+        color: "#fff",
```
istarkov, line 80: mb add to the theme?

istarkov (approved):

packages/css-data/src/schema.ts
```diff
@@ -97,6 +97,15 @@ const ArrayValue = z.object({
...
+export const TupleValueItem = z.union([UnitValue, KeywordValue, UnparsedValue]);
+export type TupleValueItem = z.infer<typeof TupleValueItem>;
+
+export const TupleValue = z.object({
+  type: z.literal("tuple"),
+  value: z.array(TupleValueItem),
+});
+export type TupleValue = z.infer<typeof TupleValue>;
+
 export const validStaticValueTypes = [
```
istarkov, line 109: We need to document what means `validStaticValueTypes` every time I see this I'm using the search

apps/builder/app/builder/features/style-panel/controls/position/position.tsx
kof: we also have cases where we pass a string from control, either way the important thing here to not pass a string ever again, no matter why

packages/css-data/src/schema.ts
kof: commented

taylornowotny:
Functionality is perfect! Great work. Two minor corrections below:

- [ ] Layout: The spacing is not consistent with the design. Please see the Anatomy section and update the spacing accordingly. Please make sure that the update is also applied to the background position controls #877 

- [ ] Units: sorry I forgot to specify which units to include in the inputs. Webflow has: PX, %, VH, & VW. I am not aware of any limitations on which units to include here, but they should be consistent between object-position and background-position. Backgrounds has a unitless option which doesn't make sense to me - I think that should be removed.

PR #1192, opened 2023-03-07
Fix collapse on size change

### LINKED ISSUE #1183: Collapse detector fails on initial load in this particular case
Screenshot 2023-03-06 at 23 49 54

Link to page in DM

Steps

1. open debug page
2. on initial load collapse detector sets both vertical and horizontal paddings
3. if you change any CSS property, it detects correctly and padding goes away
4. after reload it will be back

istarkov:
## Description

closes #1183 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/dom-hooks/use-window-resize.ts
+++ b/apps/builder/app/shared/dom-hooks/use-window-resize.ts
@@ -65,3 +65,16 @@ export const useWindowResize = (onResize: () => void) => {
     };
   }, [onResize]);
 };
+
+export const useWindowResizeDebounced = (onResize: () => void) => {
+  useEffect(() => {
+    // Call on leading
+    emitter.on("resizeStart", onResize);
+    // and trailing edge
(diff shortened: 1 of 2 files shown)


apps/builder/app/shared/dom-hooks/use-window-resize.ts
```diff
@@ -65,3 +65,16 @@ export const useWindowResize = (onResize: () => void) => {
     };
   }, [onResize]);
 };
+
+export const useWindowResizeDebounced = (onResize: () => void) => {
+  useEffect(() => {
+    // Call on leading
+    emitter.on("resizeStart", onResize);
```
istarkov, line 72: Can be called twice but above we have todo - add throttling, but better to make debounce with leading and trailing edges.

TrySound (approved):

PR #1196, opened 2023-03-08
Use pnpm to run monorepo dev

TrySound:
Turbo is just an additional layer which starts pnpm which starts actual dev scripts.

Here I removed turbo in favour of pnpm run --recursive --parallel Which does the same and probably faster (at least when I start it feels less laggy).

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

package.json
```diff
@@ -8,10 +8,10 @@
     "build-packages": "pnpm run --recursive --parallel --filter='!./apps/*' --filter='!./packages/sdk-size-test' build",
     "build": "pnpm build-packages && pnpm --filter=@webstudio-is/sdk-size-test build && pnpm --filter=@webstudio-is/builder build",
     "checks": "turbo run checks",
-    "dev": "turbo run dev --parallel",
+    "dev": "pnpm run --parallel --recursive dev",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "build-storybook": "build-storybook",
-    "storybook": "turbo run storybook:run",
+    "storybook": "pnpm run --parallel --recursive storybook:run",
```
kof, line 14: don't you loose turbo cache?

TrySound: It's dev mode. There is no cache here. `persistent: true` before was called `cache: false`

PR #1200, opened 2023-03-08
Instance name bugs

kof:
## Description

1. Don't set the label when id didn't change
2. workaround stale defaultValue in TextField

## Steps for reproduction

1. change instance name
2. select a different instance (see previous instance name is still in the text field)
3. select an different instance or different tab - instance name from previous instance went into a the currently selected instance

4. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/settings-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/settings-panel.tsx
@@ -17,7 +17,7 @@ export const SettingsPanel = () => {
         <Label>Instance Name</Label>
         <TextField
           /* Key is required, otherwise when label is undefined, previous value stayed */
-          key={selectedInstance.label}
+          key={selectedInstance.id}
           onBlur={handleBlur}
           onKeyDown={handleKeyDown}
           placeholder={label}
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/features/settings-panel/settings-panel.tsx
```diff
@@ -16,6 +16,8 @@ export const SettingsPanel = () => {
       <Flex gap="1" direction="column" grow>
         <Label>Instance Name</Label>
         <TextField
+          /* Key is required, otherwise when label is undefined, previous value stayed */
+          key={selectedInstance.label}
```
istarkov, line 20: Why not id for key?

kof: yep, changed to id locally

istarkov (approved):

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
TrySound: Why only locally?

kof: damn forgot to commit

PR #1203, opened 2023-03-09
Allow button children and inner text over properties

kof:
Fixes https://github.com/webstudio-is/webstudio-builder/issues/1198

## Description

- Button has a standard behavior with space key pressed that can't be turned off, so we can't type a space
- Even if we work around the space behavior somehow, we still  don't have an ability for rich-text to accept any instance and that makes a use case for icon button and button with icon on left/right impossible

The solution I went for is to make it container for accepting child instances and edit text using InnerText property. This may be not as intuitive as direct rich-text but its more capable.

Webflow does the same thing for a form button, though it uses an input element.

Screenshot 2023-03-09 at 03 10 05

## Steps for reproduction

1. add button
1. change inner text
1. add image or text components

## Code Review

  - test it on preview
- [ ] @taylornowotny - please test 

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/components/__generated__/button.props.ts
+++ b/packages/react-sdk/src/components/__generated__/button.props.ts
@@ -452,4 +452,10 @@ export const props: Record<string, PropMeta> = {
     control: "text",
     type: "string",
   },
+  innerText: {
+    required: false,
+    control: "text",
+    type: "string",
+    defaultValue: "Edit Inner Text in Properties",
+  },
 };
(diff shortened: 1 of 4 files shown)


TrySound (approved):

taylornowotny:
Two issues with the component:

1. The "open" state of the section should have black icons (foreground/icon/main)

2. What is the "open, but empty" variant for? When closed it shows a blue dot, which would indicate that it is not empty.

PR #1204, opened 2023-03-09
Fix text highlighing

kof:
## Description

Closes https://github.com/webstudio-is/webstudio-builder/issues/1193

When user is selecting text inside content editable and mouse goes up on a different instance - we don't want to select a different instance because that would cancel the text selection.

## Steps for reproduction

1. enter text editing
2. select text and let mouse go up on a different instance outside of the current
3. see text now stays highlighted, before it would select a different instance

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/shared/use-track-selected-element.ts
+++ b/apps/builder/app/canvas/shared/use-track-selected-element.ts
@@ -20,6 +20,8 @@ const eventOptions = {
   passive: true,
 };
 
+const isHighlighting = () => getSelection()?.type === "Range";
+
 export const useTrackSelectedElement = () => {
   const selectedInstanceId = useStore(selectedInstanceIdStore);
   const [editingInstanceId, setEditingInstanceId] = useTextEditingInstanceId();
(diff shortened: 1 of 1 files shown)


apps/builder/app/canvas/shared/use-track-selected-element.ts
```diff
@@ -38,6 +40,12 @@ export const useTrackSelectedElement = () => {
 
   useEffect(() => {
     const handleClick = (event: MouseEvent) => {
+      // When user is selecting text inside content editable and mouse goes up
+      // on a different instance - we don't want to select a different instance
+      // because that would cancel the text selection.
+      if (isHighlighting()) {
```
TrySound, line 46: Why click on other instance after selecting text in editor?

kof: see video: https://github.com/webstudio-is/webstudio-builder/issues/1193

you start selecting in our instance, then mouse-up happens over another instance

TrySound (approved):

PR #1211, opened 2023-03-09
Reverting the decision to make nicer looking link targets

kof:
User should know the underlying platform

this has also resulted in the need to have target wrapped in all links components

Screenshot 2023-03-09 at 14 19 50

## Steps for reproduction

1. create link
2. select target prop

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

rpominov (approved):
Probably not in scope of this PR, but something that came to mind:

Target can also point to an iframe instead of using a keyword like `_blank`. So maybe we should leave it as `string` instead of enum.

kof:
@rpominov how do we express that? do we have a combobox type?

rpominov:
> do we have a combobox type?

No, only option is just a text input currently

kof:
> No, the only option is a Text input currently

This is not an option, because it requires user to know the standard targets, we need a combobox for this, will keep a select for now, using a custom target is something I only know about but never used in practice

packages/prisma-client/prisma/migrations/20230309142820_link-target/migration.ts
```diff
@@ -0,0 +1,73 @@
...
+        for (const build of builds) {
+          const props = JSON.parse(build.props);
+          let hasConversion = false;
+
+          // prop.value === 'self' > prop.value === '_self'
+          for (const prop of props) {
+            if (
+              prop.name === "target" &&
+              typeof prop.value === "string" &&
+              prop.value[0] !== "_"
```
TrySound, line 48: startsWith('_') a little cleaner

TrySound (approved):

packages/prisma-client/prisma/migrations/20230309142820_link-target/migration.ts
kof: agree, don't care for migration though

PR #1225, opened 2023-03-10
Switch pages on client side

TrySound:
Switching between pages takes very long time though all data already loaded. Here I added validation to not reload builder data when page is changed and render canvas with static id which is not really used for anything so iframe no longer reloads.

Switching pages now instant.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/nano-states/nano-states.ts
+++ b/apps/builder/app/shared/nano-states/nano-states.ts
@@ -12,6 +12,7 @@ import type {
   Instances,
   InstancesItem,
   Page,
+  Pages,
   Prop,
   Props,
   StyleDecl,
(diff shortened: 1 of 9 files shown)


apps/builder/app/shared/nano-states/nano-states.ts
```diff
@@ -40,7 +41,32 @@ const useValue = <T>(atom: WritableAtom<T>) => {
...
+};
+
+export const selectedPageIdStore = atom<undefined | Page["id"]>(undefined);
+export const useSetSelectedPageId = (pageId: Page["id"]) => {
+  useSyncInitializeOnce(() => {
+    selectedPageIdStore.set(pageId);
+  });
+};
+
+export const selectedPageStore = computed(
```
kof, line 58: it would be better to have a  page by id map, so we can pick whatever page we need by id any time

kof (approved):

apps/builder/app/shared/nano-states/nano-states.ts
kof: also think about future multi-canvas situation where you can have the different pages on different canvases visible at the same time

TrySound: > it would be better to have a page by id map, so we can pick whatever page we need by id any time

We need to move migrate away from `{ homePage, pages }` structure in db in favor of flat list. But maybe later.

> think about future multi-canvas situation

yep, state will be a little more complex

istarkov (approved):

PR #1228, opened 2023-03-12
Address selected instance with tree path

TrySound:
With slots there can be same ids in one tree so we need new way to address instances. Here I propose the list from target instance to the root.

Now instead of computing parents by normalized tree (one instance of slot can have many parents) we just access already computed list from rendered tree.

This address can be computed when user click in dom which can be traversed or it can be passed top-down with props.

The main con of this approach is collaboration. Reparanted instances on by one user makes "selected address" stale. Though we have similar situation with single id when instance is deleted.

Isolated mode for slots/components will make this problem less noticable.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/footer/breadcrumbs.tsx
+++ b/apps/builder/app/builder/features/footer/breadcrumbs.tsx
@@ -1,4 +1,3 @@
-import { useMemo } from "react";
 import { useStore } from "@nanostores/react";
 import { ChevronRightIcon } from "@webstudio-is/icons";
 import {
@@ -8,10 +7,10 @@ import {
   DeprecatedText2,
 } from "@webstudio-is/design-system";
 import {
-  instancesIndexStore,
-  selectedInstanceIdStore,
+  instancesStore,
+  selectedInstanceAddressStore,
 } from "~/shared/nano-states";
-import { getInstanceAncestorsAndSelf } from "~/shared/tree-utils";
+import { getAncestorInstanceAddress } from "~/shared/tree-utils";
 
 type BreadcrumbProps = {
   children: JSX.Element | string;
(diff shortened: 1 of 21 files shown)


apps/builder/app/builder/features/footer/breadcrumbs.tsx
```diff
@@ -8,10 +7,10 @@ import {
   DeprecatedText2,
 } from "@webstudio-is/design-system";
 import {
-  instancesIndexStore,
-  selectedInstanceIdStore,
+  instancesStore,
+  selectedInstanceAddressStore,
 } from "~/shared/nano-states";
-import { getInstanceAncestorsAndSelf } from "~/shared/tree-utils";
+import { getAncestorInstanceAddress } from "~/shared/tree-utils";
```
kof, line 13: address vs path, if we go with actual path inside then Path word is more accurate

apps/builder/app/shared/copy-paste/plugin-markdown.ts
```diff
@@ -207,13 +208,19 @@ export const parse = (clipboardData: string, options?: Options) => {
...
-
+  // paste to the root if nothing is selected
+  const instanceAddress = selectedInstanceAddressStore.get() ?? [
+    selectedPage.rootInstanceId,
+  ];
+  const [targetInstanceId] = instanceAddress;
   const dropTarget = findClosestDroppableTarget(
     instancesIndexStore.get(),
-    selectedInstanceIdStore.get()
+    // @todo accept instance address
```
kof, line 222: what will happen when pasting into slot rn?

apps/builder/app/shared/tree-utils.ts
```diff
@@ -15,6 +15,49 @@ import {
...
+
+// this utility is temporary solution to compute instance addresses
+// before all logic is migrated to get it from rendered context
+// @todo should be deleted before adding slots
+export const getInstanceAddress = (
+  instances: Instances,
+  instanceId: Instance["id"]
+) => {
+  const parentInstancesById = new Map<Instance["id"], Instance["id"]>();
+  for (const instance of instances.values()) {
```
kof, line 45: is this iterating over ALL instances of the project?

kof (approved):

apps/builder/app/shared/copy-paste/plugin-markdown.ts
TrySound: Nothing really, though will be easier to use already traversed list of ancestors

apps/builder/app/shared/tree-utils.ts
TrySound: Yep, will get rid of it in upcoming PRs

apps/builder/app/builder/features/footer/breadcrumbs.tsx
TrySound: We use path a lot of routing and fs in node. Afraid to confuse with naming

PR #1230, opened 2023-03-13
Add lostpixel to builder

andrasbacsai:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


package.json
```diff
@@ -14,6 +14,7 @@
     "storybook": "pnpm run --parallel --recursive storybook:run",
     "storybook:build": "npm run build && turbo run storybook:build",
     "storybook:ds:build": "turbo run storybook:build --filter=@webstudio-is/design-system",
+    "storybook:builder:build": "turbo run storybook:build --filter=@webstudio-is/builder",
```
kof, line 17: we need something better than a separate command per package, because you don't want this to become a list of 10-20 packages

apps/builder/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "**/*.tsx",
     "../../@types/**/*.d.ts"
   ],
+  "exclude": ["node_modules", "lostpixel.config.ts"],
```
TrySound, line 10: Maybe fix errors instead of suppressing them?

andrasbacsai: The error is that `lostpixel.config.ts` should be under `rootDir`.
image

Should I just move it there? Or what is the suggested fix for this?

package.json
andrasbacsai: Fixed this with a script.

apps/builder/tsconfig.json
andrasbacsai: I only excluded `lostpixel.config.ts`. Locally `pnpm checks` ends with a LOT more errors if I do not specify `node_modules` as well.

Is it good now?

.github/workflows/vis-reg-tests.yml
```diff
@@ -2,20 +2,33 @@ name: Visual Regression Tests
...
-
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/ci-setup
 
       - name: Build Storybook for Design System
-        run: pnpm storybook:ds:build
+        run: bash scripts/build-storybook.sh design-system
```
kof, line 13: would this not work?
`turbo run storybook:build --filter=@webstudio-is/design-system` 

If I understand correctly you created a custom bash script to just pass the filter argument?

apps/builder/lostpixel.config.ts
```diff
@@ -0,0 +1,9 @@
+import * as process from "process";
+
+export const config = {
```
kof, line 3: the idea to use typescript here is to actually typecheck the config, but we are not importing lost-pixel types here to use them, I think we should use the types from lost pixel like in the docs https://docs.lost-pixel.com/user-docs/setup/project-configuration#typescript

or just use json/js config, because what's the point of using ts here otherwise?

kof: I think we had lost pixel in deps before and it was a big one, so kinda only makes sense if we can install lost pixel types separately from the library, otherwise I would just use json for now 

cc @d-ivashchuk

kof:
I also just noticed that we haven't deleted .lostpixel from the root, so old setup baselines are still in the repo

apps/builder/lostpixel.config.ts
andrasbacsai: For example `apiKey` does not exist in any of the types excluded in `lost-pixel`. I will keep it as a `.js` file for now.

.github/workflows/vis-reg-tests.yml
andrasbacsai: Yeah. Could use that way as well.

kof (approved):

TrySound (approved):

PR #1232, opened 2023-03-13
Create page on client

TrySound:
Newly created instance does not propagate to client because we no longer reload on navigating to new page.

Here moved page and root instance creation to client with patches.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/sidebar-left/panels/pages/settings.tsx
+++ b/apps/builder/app/builder/features/sidebar-left/panels/pages/settings.tsx
@@ -1,4 +1,5 @@
 import { z } from "zod";
+import store from "immerhin";
 import { useState, useCallback, ComponentProps } from "react";
 import { useStore } from "@nanostores/react";
 import { useDebouncedCallback } from "use-debounce";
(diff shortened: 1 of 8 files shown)


kof (approved):
Tested

apps/builder/app/builder/features/sidebar-left/panels/pages/settings.tsx
```diff
@@ -255,20 +235,43 @@ export const NewPageSettings = ({
...
-      fetcher.submit(values, {
-        method: "put",
-        action: restPagesPath({ projectId }),
-      });
+      const pageId = nanoid();
+      store.createTransaction(
+        [pagesStore, instancesStore],
+        (pages, instances) => {
+          if (pages === undefined) {
+            return;
```
kof, line 243: when this happens, can we still call onSuccess()?

TrySound: Never. Just type guard. It will be removed we replace current pages format to map like all other data.

PR #1233, opened 2023-03-13
Improve instance selection

TrySound:
"Click" based implemention had aa few important bugs. Selecting other instance was not possible when some text was selected in editing mode. This is consequence of a fix when other instance is selected user select text and move cursor outside of editor while still selecting.

Here's how all solved. We track both mousedown and mouseup. When their targets are different we do nothing because this is most likely text selection or drag'n'drop.

Editing state a little simplified by just looing for closest contenteditable instead of editing id comparison.

Instance address now computed by traversing dom.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -37,6 +37,7 @@ import { customComponents } from "./custom-components";
 import { useHoveredInstanceConnector } from "./hovered-instance-connector";
 import { setDataCollapsed, subscribeCollapsedToPubSub } from "./collapsed";
 import { useWindowResizeDebounced } from "~/shared/dom-hooks";
+import { subscribeInstanceSelection } from "./instance-selection";
 
 registerContainers();
 
@@ -82,6 +83,13 @@ const DesignMode = () => {
   // in both places
   useCopyPaste();
   useHoveredInstanceConnector();
+
+  useEffect(() => {
+    const unsubscribeInstanceSelection = subscribeInstanceSelection();
+    return () => {
+      unsubscribeInstanceSelection();
+    };
+  }, []);
 
   return null;
 };
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -84,12 +84,7 @@ const DesignMode = () => {
   useCopyPaste();
   useHoveredInstanceConnector();
 
-  useEffect(() => {
-    const unsubscribeInstanceSelection = subscribeInstanceSelection();
-    return () => {
-      unsubscribeInstanceSelection();
-    };
-  }, []);
+  useEffect(subscribeInstanceSelection, []);
 
   return null;
 };
--- a/apps/builder/app/canvas/instance-selection.ts
+++ b/apps/builder/app/canvas/instance-selection.ts
@@ -1,0 +1,112 @@
+import { getComponentMeta, idAttribute } from "@webstudio-is/react-sdk";
+import {
+  instancesStore,
+  selectedInstanceAddressStore,
+  textEditingInstanceIdStore,
+} from "~/shared/nano-states";
+import { publish } from "~/shared/pubsub";
+import {
+  type InstanceAddress,
+  getAncestorInstanceAddress,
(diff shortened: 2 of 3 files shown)


apps/builder/app/canvas/canvas.tsx
```diff
@@ -83,6 +84,13 @@ const DesignMode = () => {
   useCopyPaste();
   useHoveredInstanceConnector();
 
+  useEffect(() => {
+    const unsubscribeInstanceSelection = subscribeInstanceSelection();
+    return () => {
+      unsubscribeInstanceSelection();
+    };
+  }, []);
```
kof, line 92:
```suggestion
  useEffect(subscribeInstanceSelection, []);
```

apps/builder/app/canvas/instance-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+    element.closest(`[${idAttribute}]`) ?? undefined;
+  while (matched) {
+    const instanceId = matched.getAttribute(idAttribute) ?? undefined;
+    if (instanceId !== undefined) {
+      instanceAddress.push(instanceId);
+    }
+    matched = matched.parentElement?.closest(`[${idAttribute}]`) ?? undefined;
+  }
+  if (instanceAddress.length === 0) {
+    return undefined;
```
kof, line 32:
```suggestion
    return;
```

kof: also wouldn't it be better to just return empty array?

apps/builder/app/canvas/instance-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+  for (const instanceId of instanceAddress) {
+    const instance = instances.get(instanceId);
+    if (
+      instance !== undefined &&
+      getComponentMeta(instance.component)?.type === "rich-text"
+    ) {
+      return getAncestorInstanceAddress(instanceAddress, instanceId);
+    }
+  }
+  return undefined;
```
kof, line 50: no need for this, its the default

apps/builder/app/canvas/instance-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+    // notify whole app about click on document
+    // e.g. to hide the side panel
+    publish({ type: "clickCanvas" });
+
+    // prevent selecting instances inside text editor while editing text
+    if (element.closest("[contenteditable=true]")) {
+      return;
+    }
+
+    const instanceAddress = getInstanceAddressFromElement(element);
```
kof, line 80: I think its elementAddress not instance address, instance has an id, it doesn't need an address, but the specific thing we selected on the canvas does, which is element

kof: in this case getInstanceAddressFromElement > getElementAddress

apps/builder/app/canvas/instance-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (instanceAddress === undefined) {
+      return;
+    }
+
+    // the first click in double click or the only one in regular click
+    if (event.detail === 1) {
+      selectedInstanceAddressStore.set(instanceAddress);
```
kof, line 87: that would be selectedInstanceAddressStore > selectedElementAddressStore

apps/builder/app/canvas/instance-selection.ts
TrySound: Element is a concept from dom and not every element is instance.

apps/builder/app/canvas/instance-selection.ts
kof: the fact that one item always exists is something you would have to assume, in the downstream logic, its not guaranteed by types or checks

apps/builder/app/canvas/instance-selection.ts
kof: yes technically its an instance element, but do you really need that detail in the semantics?

kof (approved):

PR #1236, opened 2023-03-14
Improve instance hovering

TrySound:
Label again started flickering because hovered instance and hovered outline store changes have delay in between. Solution as before is to provide all info with outline. Here I just passed both label and component to get the shape for getInstanceLabel.

Removed react hook from instance hovering and refactored with instance selectors.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/tree-utils.ts
+++ b/apps/builder/app/shared/tree-utils.ts
@@ -34,6 +34,16 @@ export const getAncestorInstanceSelector = (
   return instanceSelector.slice(ancestorIndex);
 };
 
+export const areInstanceSelectorsEqual = (
+  left?: InstanceSelector,
+  right?: InstanceSelector
+) => {
+  if (left === undefined || right === undefined) {
+    return false;
+  }
+  return left.join(",") === right.join(",");
+};
+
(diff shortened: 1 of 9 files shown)


apps/builder/app/shared/tree-utils.ts
```diff
@@ -34,6 +34,16 @@ export const getAncestorInstanceSelector = (
...
 };
 
+export const areInstanceSelectorsEqual = (
+  left?: InstanceSelector,
+  right?: InstanceSelector
+) => {
+  if (left === undefined || right === undefined) {
+    return false;
+  }
+  return left.join(",") === right.join(",");
```
kof, line 44: fun way to do the same `String(left) === String(right)` but also wondering if this is going to be a very frequently called function and actual walking over array would be faster

kof (approved):

apps/builder/app/shared/tree-utils.ts
TrySound: I think join is used in .toString() of arrays. So kinda same, join is just more explicit.

Issue #1120, opened 2023-02-24
Refactor Link component


rpominov:
We have 6 link components: (rich-text + regular) * (`<a>` + `<RemixLink>` + "the one that does `preserveBuildParams`")

What should be done:
 - remove "the one that does `preserveBuildParams`" and solve this on `WebstudioComponentDev` level or on data level
 - generate separate props for `<a>` and `<RemixLink>`, so that we can freely change Remix's link props

rpominov:
@kof Probably better to do this after https://github.com/webstudio-is/webstudio-builder/issues/1129 ? Wdyt?

kof:
@rpominov lets hold off on this, #1129 is an entire user story, it has 3 parts, I am not sure which part would be blocking.

This issue also has 2 parts, I think the second part is already defined as part of #1129, bu the first part can be done completely independently

rpominov:
I can do the first part

rpominov:
> generate separate props for `<a>` and `<RemixLink>`, so that we can freely change Remix's link props

@istarkov Could you take this remaining todo from here as part of #1129

PR #1237, opened 2023-03-14
Refactor link behaviour in preview

### LINKED ISSUE #1231: When in preview, external links should open in new window
Screenshot 2023-03-13 at 14 39 40

It the preview mode, which you get in by clicking "play" button above, there should be special behaviour for links. We should:

1. Either switch page: when link corresponds to a page, no matter what "target=" says
2. Or open link in a new window (tab): when it's an absolute link

rpominov:
## Description

- Removed Link wrapper that used to apply `preserveBuildParams` (part 1 of #1120)
- Removed `preserveBuildParams` completely, because we no longer need it as we request page switch instead of letting link to navigate
- Removed `mode` parameter from BuildParams, because it's always "edit" now
- Fixes #1231

## Steps for reproduction

1. Open preview mode
2. Click on some links

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/canvas/features/webstudio-component/link.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const useHandleLinkClick = () => {
+  const pages = useStore(pagesStore);
+  const [isPreviewMode] = useIsPreviewMode();
+
+  return (event: MouseEvent) => {
+    event.preventDefault();
+
+    if (isPreviewMode === false || pages === undefined) {
```
TrySound, line 29: Inside event handlers you can access previewModeStore and pagesStore directly to not trigger rerender when any of them are changed.

apps/builder/app/canvas/features/webstudio-component/link.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (href === null) {
+      return;
+    }
+
+    const page = findPageByIdOrPath(pages, href);
+
+    if (page) {
+      publish({
+        type: "switchPage",
+        payload: { pageId: page.id === pages.homePage.id ? "home" : page.id },
```
TrySound, line 44: homePage have actual id why use "home"?

apps/builder/app/canvas/features/webstudio-component/link.ts
```diff
@@ -0,0 +1,53 @@
...
+      });
+      return;
+    }
+
+    if (isAbsoluteUrl(href)) {
+      window.open(href, "_blank");
```
TrySound, line 50: Does it work from inside iframe? Could be more restrictions in other browsers. Could you check so we could have future proof implementation. Can maybe will be more sense to put href in switchPage event.

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -117,8 +120,8 @@ export const WebstudioComponentDev = ({
     [componentAttribute]: instance.component,
     [idAttribute]: instance.id,
     onClick: (event: MouseEvent) => {
-      if ((event.target as HTMLElement).closest("a")) {
-        event.preventDefault();
+      if (event.currentTarget instanceof HTMLAnchorElement) {
+        handleLinkClick(event);
```
TrySound, line 124: Does it still work with bold inside link?

apps/builder/app/canvas/features/webstudio-component/link.ts
rpominov: Tested in Chrome, Brave, Safari, Firefox: all good. Couldn't find anything about it not working via Google.

I think doing it in switchPage will be more risky because then browser may think it's not caused by click.

apps/builder/app/shared/pages/use-switch-page.ts
```diff
@@ -0,0 +1,30 @@
...
+import { builderPath } from "~/shared/router-utils";
+import { useAuthToken } from "~/shared/nano-states";
+import { useSubscribe } from "~/shared/pubsub";
+
+export const useSwitchPage = () => {
+  const [project] = useProject();
+  const [authToken] = useAuthToken();
+  const navigate = useNavigate();
+
+  return (pageId: "home" | Page["id"]) => {
```
TrySound, line 13: Since we can access stores from anywhere it could look like this. A little less abstraction around remix.
```
const navigate = useNavigate();
navigate(getPageUrl(pageId))
```

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
rpominov: Yeah. 

The reason I've switched from `.closest("a")` is because when you click on "bold" in `Body > Link > Bold` we get 3 events: 

1. `{currentTarger: Body, target: Bold}`
2. `{currentTarger: Link, target: Bold}` — this one we want to preventDefault
3.  `{currentTarger: Bold, target: Bold}`

And all 3 satisfy `event.target.closes('a')`

apps/builder/app/shared/pages/use-switch-page.ts
rpominov: How can I access store without a hook?

apps/builder/app/canvas/features/webstudio-component/link.ts
rpominov: This is just because we want to set remove `?pageId` from URL when it's home. So we need access to `pages.homePage.id` to check whether this is the home page. I decided to check here since I already have `pages.homePage` here.

apps/builder/app/canvas/features/webstudio-component/link.ts
TrySound: Btw can be just a function without hook closure

apps/builder/app/canvas/features/webstudio-component/link.ts
TrySound: It's just confusing that pageId can be not actual page id.

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
rpominov: Makes more sense to me to check in general onClick and then call specific handler. Maybe we'll have `handleButtonClick` later or something.

If I move check to `handleLinkClick` then and call it always, then it will be weird that it's called `handleLinkClick` and not just `handleClick`, and if I rename to `handleClick` then semantics/intent will be confusing.

apps/builder/app/canvas/features/webstudio-component/link.ts
rpominov: Removed "home"

TrySound (approved):

kof (approved):

apps/builder/app/canvas/features/webstudio-component/link.ts
```diff
@@ -0,0 +1,46 @@
...
+const isAbsoluteUrl = (href: string) => {
+  try {
+    new URL(href);
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+export const handleLinkClick = (event: MouseEvent) => {
```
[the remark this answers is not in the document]
TrySound: We gonna have anchor api to replace libraries like floating-ui/popperjs. And all is fine with using existing name.

Link is widely used everywhere as `<a>` wrappers. So better stick with it.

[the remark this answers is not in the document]
rpominov: I've no strong opinion, except that renaming something like that is aways pain. Will require db migration among other things for example.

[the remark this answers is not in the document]
kof: ignore it, @TrySound dm me about it since I don't understand what you mean

PR #1239, opened 2023-03-14
Replace tree preview with navigator tree

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/248

Requires a lot of work to support with all data changes

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
its kinda an important feature, just removing it is not the solution. Also, I don't see why it should be any harder to have the tree preview than having the navigator, its virtually the same thing

kof:
I also just noticed that navigator doesn't live-update during dragging right now, so to remove the live-preview on the right when navigator is undocked, we need to also make sure navigator live-updates

kof (changes requested):
discussed in dm

kof (approved):

PR #1242, opened 2023-03-14
Blocking alerts

### LINKED ISSUE #198: Prevent users from using unsupported browser
It can be frustrating if user thinks we support the browser just because it looks ok on the surface. Since we are not testing anything  but chrome for the Beta release, we should have an overlay with explanation in all other browsers to prevent frustration.

kof:
Fixes #198

## Description

1. disables builder in non-chromium browsers
2. disables builder when window is smaller than 900px
3. allows testing in unsupported browsers via feature flag "unsupportedBrowsers" or "*"

## Steps for reproduction

1. open builder in safari
2. resize the window to < 900 and then return back to 900+

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/blocking-alerts/use-alerts.tsx
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+    setMessage(
+      "Webstudio currently supports Google Chrome and Microsoft Edge. We plan to support more browsers in the near future."
+    );
+  }, []);
+  return message;
+};
+
+export const useAlerts = () => {
```
TrySound, line 35: Why hook and not component?

apps/builder/app/builder/features/blocking-alerts/alert.tsx
```diff
@@ -0,0 +1,35 @@
+import { css, Flex, Text, theme } from "@webstudio-is/design-system";
+import { AlertIcon } from "@webstudio-is/icons";
+
+const containerStyle = css({
+  position: "absolute",
+  top: 0,
+  left: 0,
+  right: 0,
+  bottom: 0,
```
TrySound, line 9: inset has good support
https://developer.mozilla.org/en-US/docs/Web/CSS/inset

kof: wow completely missed this property cc @taylornowotny I think this is awesome, I think this could be great for style panel as well

TrySound: It's kinda as "margin". In ui better to linked changes like in space ui.

TrySound (approved):

PR #1249, opened 2023-03-15
Add SectionTitle; Add utility for focusing using arrow keys

### LINKED ISSUE #1100: Update component: Title

### LINKED ISSUE #1271: Labels in Section Titles
Some sections may represent a single CSS property. In these cases, we want to have all the functionality of the Label component (colors, hovers, tooltips) within the Section Title.

The **Section Title** component is now using a nested **Label** component for the title text AND the **Label** component now has "Section Title" variants. The new Label variants are only different in that they are using text style: "Title".

On most sections this label will remain in the default color but _when the section title represents a single CSS property,_ the label may get a preset/local/remote color. For example, clicking the "+" in the Background Section Title adds a background AND changes Label color to "local" to indicate that a background has been set. See the Background section docs for more.

New docs on the **Section Title** component:
Screenshot 2023-03-20 at 7 48 25 PM

New variants of the Label component:
Screenshot 2023-03-20 at 8 01 59 PM

Let me know if I can assist in any way!

rpominov:
## Description

Closes #1100
Closes #1271

1. Added `SectionTitle` https://section-title--638245578b47d4399fba191a.chromatic.com/?path=/story/library-section-title--demo
2. Added basic `ArrowFocus` utility
3. Used the new `SectionTitle` in `CollapsibleSection`, so props and style panels now use the new `SectionTitle`
4. Made `CollapsibleSection` open by default as specified in #1100

## Steps for reproduction

1. Open style panel
2. Open props panel
3. Check storybook demos by the links above

## Code Review

- [ ] hi @istarkov, I need you to
  - review the `arrow-focus` utility to see whether it'll work for you in backgrounds 
  - check whether `SectionTitle` component looks and works as expected by the link above

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
- [ ] Input With Nested Buttons

this one is def. broken because arrows now don't allow to move caret in the input

packages/design-system/src/components/primitives/arrow-focus.stories.tsx
```diff
@@ -0,0 +1,108 @@
+import { StorySection } from "../storybook";
+import {
+  handleArrowFocus,
+  setArrowFocusRow,
+  setArrowFocusColumn,
+} from "./arrow-focus";
+
+const main = {
```
kof, line 8: I used to use this style initially but then realize if we don't wrap it into css() and don't use CSS type , you don't get the validation by ts

packages/design-system/src/components/primitives/arrow-focus.tsx
```diff
@@ -0,0 +1,162 @@
+import type { KeyboardEvent } from "react";
+
+const focusableSelector = (extra: string) =>
```
kof, line 3: I am surprised you ended up not using react-aria, I thought we decided to use it?

packages/design-system/src/components/primitives/arrow-focus.tsx
```diff
@@ -0,0 +1,162 @@
...
+  string,
+  ["vertical" | "horizontal", "+" | "-"] | undefined
+> = {
+  ArrowUp: ["vertical", "-"],
+  ArrowDown: ["vertical", "+"],
+  ArrowLeft: ["horizontal", "-"],
+  ArrowRight: ["horizontal", "+"],
+};
+
+const ROW_ATTRIBUTE = "data-arrow-focus-row";
```
kof, line 33: why isit not using role="row" attribute like react aria does?

packages/design-system/src/components/section-title.tsx
```diff
@@ -0,0 +1,174 @@
...
+const containerStyle = css({
+  position: "relative",
+  height: theme.spacing[15],
+  [addIconColor]: theme.colors.foregroundSubtle,
+  "&:hover, &:has(:focus-visible), &[data-state=open]": {
+    [addIconColor]: theme.colors.foregroundMain,
+  },
+});
+
+const buttonStyle = css({
```
kof, line 30: hmm? isn't this supposed to be our button component?

packages/design-system/src/components/primitives/arrow-focus.stories.tsx
rpominov: This is stories so I didn't think to much about it. Does it matter in stories?

packages/design-system/src/components/primitives/arrow-focus.tsx
rpominov: It doesn't do what we need: https://discord.com/channels/955905230107738152/973300187567095858/1085820876060106812

packages/design-system/src/components/primitives/arrow-focus.stories.tsx
kof: yeah, idk

packages/design-system/src/components/section-title.tsx
rpominov: This is not our button. This is just styles for `<button>` to make it take entire space of the title. This is what you click when you click on the title.

packages/design-system/src/components/primitives/arrow-focus.tsx
rpominov: Talked in DM: I misunderstood what FocusScope is, need to have another look. Maybe we can use it.

packages/design-system/src/components/section-title.tsx
kof: damn, ok lets call it titleButtonStyle, because I totally thought its the + button there

packages/design-system/src/components/section-title.tsx
```diff
@@ -0,0 +1,174 @@
...
+    position: "absolute",
+    top: 0,
+    bottom: 0,
+    left: theme.spacing[2],
+    right: theme.spacing[2],
+    borderRadius: theme.borderRadius[4],
+    border: `2px solid ${theme.colors.borderFocus}`,
+  },
+  variants: {
+    hasAddIcon: { true: { paddingRight: theme.spacing[16] } },
```
kof, line 56: hasAddButton?

packages/design-system/src/components/primitives/arrow-focus.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </ul>
+      </StorySection>
+
+      <StorySection title="Grid">
+        <ul onKeyDown={handleArrowFocus}>
+          {["A", "B", "C"].map((row) => (
+            <li key={row}>
+              {["1", "2", "3"].map((column) => (
+                <button
+                  {...setArrowFocusRow(row)}
```
istarkov, line 91: setArrowFocusRow looks like a not usual naming for things like this.
For example react-aria names are `focusWithinProps`, `hoverProps`
Downshift has namings like: `getLabelProps`, `getMenuProps` etc. 

For me would be good to have a name like `arrowFocusRowProps` or `getArrowFocusRowProps`

packages/design-system/src/components/primitives/arrow-focus.tsx
```diff
@@ -0,0 +1,162 @@
...
+    "summary",
+    "iframe",
+    "object",
+    "embed",
+    "audio[controls]",
+    "video[controls]",
+    "[contenteditable]",
+    "[tabindex]",
+  ]
+    .map((selector) => `${selector}:not([disabled]):not([hidden])${extra}`)
```
istarkov, line 20: Never seen `hidden` attribute in a wild.

Also instead of `:not[disabled])` you can just use  
```
  'button:enabled',
  'select:enabled',
  'textarea:enabled',
  'input:enabled',
```
in your list

packages/design-system/src/components/primitives/arrow-focus.tsx
```diff
@@ -0,0 +1,162 @@
...
+  | "all"
+  | "within-row"
+  | "within-column"
+  | "first-per-row"
+  | "first-per-column";
+
+const getFocusable = (
+  filter: Filter,
+  root: Element,
+  current: Element
```
istarkov, line 64: The assumption that current is the currently focused element?

```
root = event.currentTarget, 
current = event.target
```

packages/design-system/src/components/primitives/arrow-focus.tsx
```diff
@@ -0,0 +1,162 @@
...
+  if (
+    filter === "all" ||
+    filter === "within-row" ||
+    filter === "within-column"
+  ) {
+    let extraSelector = "";
+    if (filter !== "all") {
+      const attribute =
+        filter === "within-row" ? ROW_ATTRIBUTE : COLUMN_ATTRIBUTE;
+      extraSelector = `[${attribute}="${current.getAttribute(attribute)}"]`;
```
istarkov, line 75: Need to be commented on why we want the attribute to have the same value as current[attribute]?

rpominov:
@istarkov I'll most likely remove the custom utility and replace it with react-aria's FocusScope after all

packages/design-system/src/components/section-title.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+}) => {
+  const [isOpen, setOpen] = useState(initialIsOpen);
+  return (
+    <div style={{ display: "flex", alignItems: "center", gap: 8 }}>
+      <div style={{ width: 240, border: "dashed 3px #e3e3e3" }}>
+        <SectionTitle
+          isOpen={isOpen}
+          onOpenChange={setOpen}
+          hasItems={["local", "remote"]}
+          {...props}
```
istarkov, line 24: please don't use spreads especially in examples 
`onAdd` is it come from props or how?

packages/design-system/src/components/section-title.tsx
```diff
@@ -0,0 +1,174 @@
...
+          <div className={buttonTextStyle()}>{children}</div>
+          {dots.length > 0 && (
+            <div className={dotsSlotStyle()}>
+              {dots.map((color) => (
+                <div key={color} className={dotStyle({ color })} />
+              ))}
+            </div>
+          )}
+        </button>
+        {onAdd && (
```
istarkov, line 154: I always don't like logic like this, if we have a handler then we draw UI element.
I would prefer to have here only placeholder for the Add button like its done https://github.com/webstudio-is/webstudio-builder/blob/43617f1a59aa2eb4f9a385ca08b778f69b5f8f24/packages/design-system/src/components/css-value-list-item.tsx#L112

istarkov: The issue is opening logic. 

Also its interesting how to use that `onAdd` in our UI having that we render Section Collapsible elements and style Section in different places. 
How to call onAdd inside Background section if it called on some parent

istarkov (approved):
Super!

packages/design-system/src/components/primitives/arrow-focus.tsx
TrySound: hidden is basically html version of display: none. Allows to not rebuild dom.

TrySound: Sick, never knew there is :enabled

taylornowotny:
Two issues with the component:

   1.  The "open" state of the section should have black icons (token: foreground/icon/main)

  2.  What is the "open, but empty" variant for? When closed it shows a blue dot, which would indicate that it is not empty.

rpominov:
> What is the "open, but empty" variant for? 

This is how the following logic is implemented in code:

> Sections that do not have a an Add (+) button are open by default. Sections that do have the button are closed by default. The + button adds items to the section and each item will have a - button to remove them.

In code:
 - **All** sections are open by default
 - If sections has items that can be added/removed and currently empty, then it looks like closed even if it's open

> When closed it shows a blue dot, which would indicate that it is not empty.

In the demo, when you click on it an item gets added, so it's no longer empty after that

packages/design-system/src/components/section-title.stories.tsx
rpominov: I don't see anything wrong with spread here. 

Intent here: `Wrapped` wraps `SectionTitle` with some code that otherwise would have been repeated a lot in demos, it controls some of the props, and passes through the rest. 

Spread is perfect to express: "passes through the rest".

packages/design-system/src/components/section-title.tsx
rpominov: > The issue is opening logic.

Yeah, main reason I've put button inside the component instead of having a `suffix` prop or something is because I wanted to pass the `onClick` callback from the component. And there're also some minor reasons.

> Also its interesting how to use that onAdd in our UI having that we render Section Collapsible elements and style Section in different places.
How to call onAdd inside Background section if it called on some parent
Or we can make SectionTitle components as like as we do custom Sections.

Yeah, we'll have to refactor sections a bit somehow. One options is to make each section render the entire collapsible instead of just its content.

rpominov:
> The "open" state of the section should have black icons (token: foreground/icon/main)

The + icon is black in the open state currently. Maybe you've looked at "open, but empty"

packages/design-system/src/components/section-title.stories.tsx
rpominov: Talked in DM: I didn't know something like this is not a TS error:

```
const props = {invalidProp: "something"}
<Something foo="bar" {...props} />
```

I agree it's a good idea to avoid props spread when possible because of this.

rpominov:
- [x] Fixed minor review comments such as renaming or props spread
- [x] Replaced custom focus manager with FocusScope (removed rows/columns support for now, don't want to do it as part of this PR). This resolves all comments on the custom utility that is removed now
- [x] Added Label as specified in: https://github.com/webstudio-is/webstudio-builder/issues/1271
- [ ] There's still an open question regarding how click/focus should be handled on the Label https://github.com/webstudio-is/webstudio-builder/issues/1271#issuecomment-1477611492

packages/design-system/src/components/primitives/arrow-focus.tsx
```diff
@@ -0,0 +1,31 @@
+import { FocusScope, useFocusManager } from "react-aria";
```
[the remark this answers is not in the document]
rpominov: Actually they seem to allow to install individual packages https://react-spectrum.adobe.com/react-aria/getting-started.html#installation

Will install it

TrySound (approved):

kof (approved):

PR #1250, opened 2023-03-15
Pass selected instance selector to tree

TrySound:
Selector is a list of ids from root to target to allow identify instances even with duplicated ids.

Here is the first step to support selectors in navigator. Passing selected instance selector already helped to get rid of one getItemPath usage.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/sidebar-left/navigator/navigator.tsx
+++ b/apps/builder/app/builder/features/sidebar-left/navigator/navigator.tsx
@@ -20,7 +20,6 @@ type NavigatorProps = {
 
 export const Navigator = ({ isClosable, onClose }: NavigatorProps) => {
   const selectedInstanceSelector = useStore(selectedInstanceSelectorStore);
-  const selectedInstanceId = selectedInstanceSelector?.[0];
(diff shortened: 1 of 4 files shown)


kof (approved):

rpominov:
> instances even with duplicated ids

Just curious, how this can happen? Just in case two clients create the same id at random? Or if someone maliciously use existing ids and send them directly via API?

rpominov:
Thinking a bit more, using selectors instead of ids doesn't help with duplicates at all. Because selectors can match too. 

Example:

```
  Body (123)
    Box (456)
    Box (456)
```

Not arguing agains using selectors if they simplify code somehow. But they don't help with duplicating ids IMO.

kof:
@rpominov this is specifically for slots and components, because these can contain identical instances on the same page if you have the slot or component used twice on the same page. Slot/Component instance itself will have a unique id so the paths will be different based on that. We don't assume there will be just plain instances with same ids.

rpominov:
Got it 👍

PR #1252, opened 2023-03-15
changed word "inherited" to "remote"

taylornowotny:
and word "set" to "local" for parity with code
also changed background.canvas to background.workspace


rpominov:
@taylornowotny 

> also changed background.canvas to background.workspace

I don't see this change in the code. It's still "canvas". Maybe you didn't commit it via plugin?

Will merge the PR nevertheless, we can make this change in another PR.

rpominov (approved):

PR #1253, opened 2023-03-15
Refactor tree drag item and drop target with selectors

TrySound:
Before supporting selector in callbacks need to upport selectors in data. Here internal drag item and drop target supported as selectors.

In the future these states will be exposed so for drop target added custom representation without element to allow sending it via pubsub between realms.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/tree/horizontal-shift.ts
+++ b/packages/design-system/src/components/tree/horizontal-shift.ts
@@ -1,5 +1,6 @@
 import { useMemo, useState } from "react";
-import type { DropTarget, Placement } from "../primitives/dnd";
+import type { Placement } from "../primitives/dnd";
+import type { ItemDropTarget, ItemSelector } from "./item-utils";
 import { getPlacementIndicatorAlignment } from "./tree-node";
 
 export type ShiftedDropTarget<Data> = {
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/tree/horizontal-shift.ts
```diff
@@ -66,7 +69,11 @@ export const useHorizontalShift = <Data extends { id: string }>({
       placement: shiftPlacement(currentDepth),
     } as const;
 
-    const isDragItem = (item: Data | undefined) => item?.id === dragItem.id;
+    const [dragItemId] = dragItemSelector;
+    const isDragItem = (item: Data | undefined) => item?.id === dragItemId;
+
+    const dropTargetPath = getItemPath(root, data.id);
+    dropTargetPath.reverse();
```
kof, line 76: maybe it was more ergonomic to keep the order after all? :)

kof (approved):

packages/design-system/src/components/tree/horizontal-shift.ts
TrySound: It's still legacy path. We convert here to target -> root order. Reverse won't be necessary here soon.

PR #1256, opened 2023-03-15
Make canvas iframe receive no pointer events for the time of dragging

kof:
Fixes https://github.com/webstudio-is/webstudio-builder/issues/1213

## Description

when user drags color picker thumb, they can move mouse over canvas, with CORS canvas will hide the events and we loose control

## Steps for reproduction

1. see video in the issue
2. also affected dragging components onto canvas

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -14,7 +14,7 @@ import { useSharedShortcuts } from "~/shared/shortcuts";
 import { SidebarLeft } from "./features/sidebar-left";
 import { Inspector } from "./features/inspector";
 import {
-  useEnabledCanvasPointerEvents,
+  isCanvasPointerEventsEnabledStore,
   useProject,
 } from "./shared/nano-states";
 import { Topbar } from "./features/topbar";
@@ -53,6 +53,7 @@ import type { Asset } from "@webstudio-is/asset-uploader";
 import { useSearchParams } from "@remix-run/react";
 import { useSyncInitializeOnce } from "~/shared/hook-utils";
 import { BlockingAlerts } from "./features/blocking-alerts";
+import { useStore } from "@nanostores/react";
 
 registerContainers();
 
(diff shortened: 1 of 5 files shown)


TrySound:
Can be solved a little easier. Need to add event.prventDefault() to mouse move.

TrySound:
Can be solved a little easier if add event.preventDefault() to mouse move.

TrySound:
react-colorful does it
https://github.com/omgovich/react-colorful/blob/master/src/components/common/Interactive.tsx#L76

kof:
@TrySound it won't, we have a special case with iframe and CORS, browser behaves differently here, it kills all events as you hover over it

apps/builder/app/builder/features/style-panel/shared/color-picker.tsx
```diff
@@ -234,47 +264,49 @@ export const ColorPicker = ({
...
-        if (
-          styleValue?.type === "rgb" ||
-          styleValue?.type === "keyword" ||
-          styleValue?.type === "intermediate" ||
-          styleValue === undefined
-        ) {
-          onChange(styleValue);
-          return;
-        }
+    <>
```
TrySound, line 267: Why fragment here?

apps/builder/app/builder/builder.tsx
```diff
@@ -319,12 +323,7 @@ export const Builder = ({
             <CanvasIframe
               ref={iframeRefCallback}
               src={canvasUrl}
-              pointerEvents={
-                dragAndDropState.isDragging &&
-                dragAndDropState.origin === "panel"
-                  ? "none"
-                  : "all"
-              }
+              pointerEvents={isCanvasPointerEventsEnabled ? "all" : "none"}
```
TrySound, line 326: Man says "all" is svg only. Better use "auto"
https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

apps/builder/app/builder/features/sidebar-left/panels/components/components.tsx
```diff
@@ -134,6 +138,7 @@ export const TabContent = ({ publish, onSetActiveTab }: TabContentProps) => {
           dragItem: utils.tree.createInstance({ component: componentName }),
         },
       });
+      toggleCanvasPointerEvents(false);
```
TrySound, line 141: Here every change of store will trigger additional rerender because you invoke hook with subscription to store. You can just use store.set() directly to avoid this.

TrySound (approved):

PR #1269, opened 2023-03-20
Switch builder to separate placement indicator computing

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/pull/1268

Now all builder UI rely on separate placement computing

- for background layers had to add default getValidChildren to exclude placement indicator child
- fixed text selecting in safari while dragging with preventDefault on move

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

packages/design-system/src/components/primitives/dnd/use-drop.ts
```diff
@@ -72,9 +68,6 @@ type UseDropProps<Data> = {
   // |___________|
   emulatePointerAlwaysInRootBounds?: boolean;
 
-  // Distance from an edge when placement is put next to an element edge
-  placementPadding?: number;
```
kof, line 76: I am missing it, how did you replace this setting?

TrySound: It is moved to computeIndicatorPlacement. Only there is needed.

kof (approved):

PR #1275, opened 2023-03-22
Component card

kof:
## Description

https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=3217-86010&t=Muy6FruCcEJ5xYIj-4

Card component for adding components to the canvas

Screenshot 2023-03-22 at 18 44 41

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/component-card.tsx
```diff
@@ -0,0 +1,91 @@
...
+        {icon}
+        <div className={textStyle()}>
+          {label
+            .split(/\s/)
+            .slice(0, 2)
+            .map((word, index) => (
+              <span className={wordStyle()} key={index}>
+                {word}
+              </span>
+            ))}
```
rpominov, line 84: We show only first two words? For example, "Very Important Button" will become "Very Important"?

packages/design-system/src/components/component-card.tsx
```diff
@@ -0,0 +1,91 @@
...
+const wordStyle = css(textVariants.small, {
+  display: "flex",
+  flexGrow: 0,
+  textAlign: "center",
+  justifyContent: "center",
+  whiteSpace: "nowrap",
+  textOverflow: "ellipsis",
+  overflow: "hidden",
+  width: "100%",
+});
```
rpominov, line 58: We could use https://caniuse.com/css-line-clamp for multiline text overflow

rpominov: (Just in case you didn't consider it)

packages/design-system/src/components/component-card.tsx
TrySound: Maybe better line-clamp or text-ellipsis?
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

packages/design-system/src/components/component-card.tsx
kof: I was looking for something like this, didn't know it exists, but we also have another problem: single-word name is supposed to be centered vertically while the 2-words name would be split in 2 lines, where each word could be truncated

If we go this route with line-clamp, it would def. not allow ellipsis for each word, but maybe that's not super important after all, idk, I am sure @taylornowotny didn't know this is possible either

packages/design-system/src/components/component-card.tsx
kof: basically going with line-clamp is shortening on the total length, where with manual split its shortening each word while making sure we get at least 2 words in if there are 2+, with automatic approach, first word could be long enough so we never see the second word at all

kof: Screenshot 2023-03-22 at 15 09 19

vs. 


Screenshot 2023-03-22 at 15 19 57

rpominov (approved):
Approving, as current trimming solution seem like a viable option

packages/design-system/src/components/component-card.tsx
kof: The case of 2 long words truncated each is kinda weird and unnatural, not sure if @taylornowotny meant it this way

kof: I am trying to get vertically centered and single time truncated work

kof: alright got this done better with line clamp, still tradeoffs but seems like better once

rpominov (approved):

PR #1290, opened 2023-03-25
Locally-generated Prisma client

Andarist:


packages/scripts/build-package.ts
```diff
@@ -55,6 +59,15 @@ for (const rel of assets) {
...
 }
 
+if (
+  await access("./src/_generated")
+    .then(() => true)
+    .catch(() => false)
+) {
+  await cp("./src/_generated", "./lib/_generated", { recursive: true });
+  await cp("./src/_generated", "./lib/cjs/_generated", { recursive: true });
+}
```
Andarist, line 69: a downside of this solution is that you end up with the generated client *twice* - one in the ESM and one in the CJS directory and that includes `libquery_engine-darwin-arm64.dylib.node` that weighs 14MB~.

I think it's a fair price to pay for this script to be decoupled from what actually happens with Prisma. This script, at the moment, doesn't know what it copies - it's just using a convention to copy over any `src/_generated` directory to dist files.

packages/prisma-client/.gitignore
```diff
@@ -0,0 +1 @@
+src/_generated
```
TrySound, line 1: We use `__generated__` everywhere 8n the project

Andarist: roger that, I adjusted the name of this directory

packages/scripts/build-package.ts
```diff
@@ -1,19 +1,24 @@
...
 
-import { rm, cp } from "node:fs/promises";
+import { rm, cp, access } from "node:fs/promises";
 import { join } from "node:path";
 import { totalist } from "totalist";
 import { build, context } from "esbuild";
 
 const args = process.argv.slice(2);
 const watch = args.includes("--watch");
+const noGeneratedAsEntries = args.includes("--no-generated-as-entries");
```
Andarist, line 10: This is a little bit quirky... but we have 2 use cases here:
- `esbuild` should ignore the generated directory in `@webstudio-is/prisma-client`, it contains more than just `.ts` files and `esbuild` trips over the shape of `.d.ts` within it. In this case, we only want to copy-over the generated directory into the dist dirs
- in other packages though... the generated files are used as entrypoints~ so they should be handled by `esbuild` and not ignored

TrySound: Can't we just ignore .d.ts here? Shouldn't break anything

Andarist: That was something I wanted to do at first but then I realized that there are still other files there that we need to copy over to the dist directory. So it's just easier to copy over the whole thing and ignore it for esbuild since it literally doesn't have to touch it at all.

packages/scripts/build-package.ts
```diff
@@ -55,6 +60,18 @@ for (const rel of assets) {
...
+  noGeneratedAsEntries &&
+  (await access("./src/__generated__")
+    .then(() => true)
+    .catch(() => false))
+) {
+  await cp("./src/__generated__", "./lib/__generated__", { recursive: true });
+  await cp("./src/__generated__", "./lib/cjs/__generated__", {
+    recursive: true,
+  });
+}
```
TrySound, line 73: await inside if look always too hard to read. Maybe like this?

```
if (noGeneratedAsEntries) {
  try {
    await access("./src/__generated__")
    await cp(...)
  } catch {}
}
```

TrySound (approved):

Andarist:
@TrySound did you have a chance to test this out on your machine?

TrySound:
See the error when run dev

```
Error: ENOENT: no such file or directory, open '.../apps/builder/api/schema.prisma'
```

Andarist:
@TrySound could you take another look?

pnpm-lock.yaml
```diff
@@ -7292,6 +7310,43 @@ packages:
       webpack: 4.46.0
     dev: true
 
+  /@storybook/core-client@6.5.16(react-dom@17.0.2)(react@17.0.2)(typescript@5.0.3)(webpack@5.75.0):
```
TrySound, line 7313: These changes look like from old diff. Could you revert lock file and install again?

TrySound (approved):
All works for me. Great job!

PR #1294, opened 2023-03-26
Implement slots

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1116

Slot is copy pastable part of the tree which is always in sync. Kinda like primitive components.

The idea is the following. Slot is copy pasted as any other instance though its children are preserved as is without cloning with new ids.

This lead to the problem with multiple slot children. While copy paste they are preserved as is though when add new child to slot, the other instance of it will not get new child.

To workaround this we add another Fragment component which groups all children into single one and new children are always reflected in all slot copies.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Nice, works except the outline of the slot itself

apps/builder/app/builder/shared/navigator-tree.tsx
```diff
@@ -3,26 +3,34 @@ import { useStore } from "@nanostores/react";
...
 
   const dragItemSelector =
     state.dragPayload?.type === "reparent"
       ? state.dragPayload.dragInstanceSelector
       : undefined;
 
-  const isItemHidden = useCallback((_instanceId: Instance["id"]) => false, []);
+  const isItemHidden = useCallback(
+    (instanceId: Instance["id"]) =>
+      // fragments is internal tool to group other instances
```
kof, line 29: tool > component

packages/react-sdk/src/components/slot.tsx
```diff
@@ -0,0 +1,17 @@
...
+type Props = {
+  children?: ReactNode;
+};
+
+export const Slot = forwardRef<ElementRef<"div">, Props>((props, ref) => {
+  return (
+    <div
+      {...props}
+      ref={ref}
+      style={{ display: props.children ? "contents" : "block" }}
```
kof, line 12: maybe we could set contents on fragment, so we can always have slot selectable? its def. inconsistent to not be able to select the slot, especially once we want to add props to it

kof: oh I forgot why we needed display: contents - it was because we didn't want the slot to interfere with the layout

I guess we need some other way to  get selection working then

packages/react-sdk/src/components/fragment.ws.ts
```diff
@@ -0,0 +1,11 @@
+import type { WsComponentMeta, WsComponentPropsMeta } from "./component-meta";
+
+export const meta: WsComponentMeta = {
+  type: "container",
+  label: "Fragment",
+  Icon: () => null,
+};
+
+export const propsMeta: WsComponentPropsMeta = {
+  props: {},
```
kof, line 10: a little weird we need these

packages/react-sdk/src/components/slot.tsx
kof: ok here is what we can do:
1. try getting getClientRects()
2. if length is 0 but element has children, 
3. walk children calculate their group height/width
4. in case of fragment its also going to have length 0, so we would need to do this in a loop to avoid having specific logic for slots and fragments

I think we should do it in a way that allows us outlining of elements even when they have display: contents in a generic way, without using the logic of fragment/slot component.

This way we can outline even when user component uses this and we don't know anything about it

kof (changes requested):
lets get the outline working and releaseeee 🔥

rpominov (commented):
Getting "Maximum call stack size exceeded" when pasting slot inside itself.

Otherwise looks good.

PR #1296, opened 2023-03-27
Compute bounding box by children when display: contents

TrySound:
Slots have display: contents to not affect layout. So we need a way to show correct hover and selected state. Here's added utility to traverse all descendants and get their rects and then compute the sum of those rects.

## Code Review

  - test it on preview

Added ability to set display: contents from UI under features flag.

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
Looking great, why do you need a feature flag for this?

PR #1305, opened 2023-03-27
Support minWidth and maxWith in media queries, both are optional

kof:
## Description

This is in preparation for supporting webflow breakpoints which use a combination of min and max width
## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project/src/shared/breakpoints/sort.ts
+++ b/packages/project/src/shared/breakpoints/sort.ts
@@ -6,6 +6,6 @@ import type { Breakpoints } from "@webstudio-is/project-build";
  */
 export const sort = (breakpoints: Breakpoints) => {
   return Array.from(breakpoints.values()).sort((breakpointA, breakpointB) => {
-    return breakpointA.minWidth - breakpointB.minWidth;
+    return (breakpointA.minWidth ?? 0) - (breakpointB.minWidth ?? 0);
   });
 };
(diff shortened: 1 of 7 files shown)


packages/project/src/shared/breakpoints/sort.ts
```diff
@@ -6,6 +6,6 @@ import type { Breakpoints } from "@webstudio-is/project-build";
  */
 export const sort = (breakpoints: Breakpoints) => {
   return Array.from(breakpoints.values()).sort((breakpointA, breakpointB) => {
-    return breakpointA.minWidth - breakpointB.minWidth;
+    return (breakpointA.minWidth ?? 0) - (breakpointB.minWidth ?? 0);
```
TrySound, line 9: Still does not consider maxWidth

kof: yes, I was thinking about it if a query contains both min and max with, can we really sort when there are intersections?

I think we should not have automated sorting at all and instead allow sorting in the UI like with style sources

kof: I don't see a good way to meaningfully sort them automatically when there are combinations of both min and max, I think we should just keep current sorting and later remove it entirely when we have manual sorting ability

PR #1306, opened 2023-03-28
Fix possible cyclic slot and hide its styles and props panels

TrySound:
One cyclic case I missed. When paste slot into itself as target then check miss unresolved fragment inside. Here added additional check before copying and covered with tests.

For slots show only settings panel.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/inspector/inspector.tsx
+++ b/apps/builder/app/builder/features/inspector/inspector.tsx
@@ -1,4 +1,4 @@
-import { useRef } from "react";
+import { useEffect, useRef, useState } from "react";
 import { useStore } from "@nanostores/react";
 import type { Publish } from "~/shared/pubsub";
 import {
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/inspector/inspector.tsx
```diff
@@ -68,6 +68,29 @@ export const Inspector = ({ publish, navigatorLayout }: InspectorProps) => {
     );
   }
 
+  if (selectedInstance.component === "Slot") {
```
kof, line 71: can we please just not show the tab triggers where they are right now instead of copy pasting entire block?
we may have other conditions in the future, so showStylePanel and showPropsPanel would be nice

TrySound: Default is also different. Though most of the time is not correct because tab remembers style.

kof: wait we don't want to remove it, just disable it

kof: there is even disabled prop on trigger from radix

PR #1315, opened 2023-03-28
Remove scroll state from builder as not useful

TrySound:
Scroll state in builder used only to hide select, hover and text toolbar while scrolling. Though they all already track scrolling to do the same and more importantly restore state after scroll. So there is no need to have it builder.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound:
They all track scrolling in canvas frame. Nothing tracks scroll in builder.

PR #1316, opened 2023-03-28
Add gradient color to Button Component

lokeswaran-aj:
## Description

This PR contains the changes for the Gradient color for the button component

Here is an example of share button with `gradient` color:

https://user-images.githubusercontent.com/74011196/228329202-54f1b143-cc99-47d5-bd17-decf170fbe82.mov

## Steps for reproduction

1. Create a Button component
2. set `color="gradient"` in Button Component

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] updated [test cases](https://github.com/webstudio-is/webstudio-builder/blob/main/apps/builder/docs/test-cases.md) document
- [x] added tests
- [x] if any new env variables are added, added them to `.env.example` and the `builder/env-check.js` if mandatory


packages/design-system/src/components/button.tsx
```diff
@@ -41,11 +43,16 @@ const foregrounds: Record<ButtonColor, string> = {
...
+  gradient: theme.colors.foregroundContrastMain,
 };
 
 // CSS supports multiple gradients as backgrounds but not multiple colors
-const backgroundColors = (base: string, overlay: string) =>
-  `linear-gradient(${overlay}, ${overlay}), linear-gradient(${base}, ${base})`;
+const backgroundColors = (base: string, overlay: string) => {
+  if (base === "$colors$backgroundGradientPrimary") {
+    return `linear-gradient(${overlay}, ${overlay}), ${base}`;
+  }
```
kof, line 53: not sure why this was needed

kof: oh because base is a gradient and before it was a color

kof: @rpominov 
with this conditional checking that particular base the function became quite bad, it would probably make sense to rather inline that gradient and remove the function

rpominov: This solution seems ok to me, except we should replace `base === "$colors$backgroundGradientPrimary"` with `base === theme.colors.backgroundGradientPrimary`.

Logic with this solution:
 - `backgroundColors` applies two colors as a background, and does it using a gradient hack
 - but if one of the colors is a gradient itself, it should put it as is

-----------------------

Alternative solution: in `perColorStyle` do something like:

```
background: variant === "gradient" ? `linear-gradient(${overlay}, ${overlay}), ${base}` : backgroundColors(
      backgrounds[variant],
      theme.colors.backgroundButtonHover
    )
```

Not sure which one is better, both seem ok to me.

rpominov:
Please also add `gradient` here https://github.com/webstudio-is/webstudio-builder/blob/1163d2c8634cb588b0cb9393b7b2c48e031a42c9/packages/design-system/src/components/button.stories.tsx#L24-L31

kof:
Problem is that if is with that specific color, kinda not very useful
function, wouldn't use it at least in case of the gradient base

rpominov:
> Problem is that if is with that specific color, kinda not very useful
function, wouldn't use it at least in case of the gradient base

Yeah, let's do `variant === "gradient" ?` then

packages/design-system/src/components/button.tsx
kof: or just do what you suggested both is kinda ok

lokeswaran-aj: Hii @kof We are going with the perColorStyle solution, right?

kof: yeah, lets do this

lokeswaran-aj: Updated and pushed the changes

kof:
merged, thank you!

PR #1318, opened 2023-03-29
Replace lodash string utilities with title-case

TrySound:
Change case is a cool collection of string utilities. See https://github.com/blakeembrey/change-case

Here replaced lodash snakecase and capitlize used for humanizeString with single title-case from that collection.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov:
Is the output different like this?

```
titleCase("background-color")  === "Background-Color"
humanizeString("background-color") === "Background Color"
```

If so, maybe we have to keep our own utility with this behaviour

TrySound:
Yep, found this too. Yes, let's keep humanizeString. Just changed underlying utilities.

rpominov (approved):

PR #1326, opened 2023-03-30
Support cross-project instance copy paste

TrySound:
Here a little refactored copy paste instance plugin.

Now we copy all data including slots and tokens though paste write them as is or do not write at all if already exist.

This let us to not do unnecessary patches for copy paste within project while still have whole UI pasted in different project.

There is still a problem with pasting styles which always have different breakpoint ids in each project. We will need to implement merging in other PR.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
slot copy/paste works, still missing assets and breakpoints

PR #1332, opened 2023-03-31
Support cross-project breakpoints copy paste

TrySound:
Here added utility to merge pasted breakpoints into existing ones or add new one if breakpoint does not match any.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/copy-paste/plugin-instance.ts
+++ b/apps/builder/app/shared/copy-paste/plugin-instance.ts
@@ -97,12 +97,14 @@ const getTreeData = (targetInstanceSelector: InstanceSelector) => {
     treeStyleSourceIds
   );
 
-  const treeBreapointIds = new Set<Breakpoint["id"]>();
-  const treeStyles = getMapValuesBy(stylesStore.get(), (styleDecl) => {
-    treeBreapointIds.add(styleDecl.breakpointId);
-    return treeStyleSourceIds.has(styleDecl.styleSourceId);
-  });
+  const treeStyles = getMapValuesBy(stylesStore.get(), (styleDecl) =>
+    treeStyleSourceIds.has(styleDecl.styleSourceId)
+  );
 
+  const treeBreapointIds = new Set<Breakpoint["id"]>();
+  for (const styleDecl of treeStyles) {
+    treeBreapointIds.add(styleDecl.breakpointId);
+  }
   const treeBreapoints = getMapValuesByKeysSet(
     breakpointsContainer.get(),
     treeBreapointIds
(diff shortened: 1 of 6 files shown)


apps/builder/app/shared/copy-paste/plugin-instance.ts
```diff
@@ -93,11 +97,19 @@ const getTreeData = (targetInstanceSelector: InstanceSelector) => {
     treeStyleSourceIds
   );
 
-  const treeStyles = getMapValuesBy(stylesStore.get(), (styleDecl) =>
-    treeStyleSourceIds.has(styleDecl.styleSourceId)
+  const treeBreapointIds = new Set<Breakpoint["id"]>();
+  const treeStyles = getMapValuesBy(stylesStore.get(), (styleDecl) => {
+    treeBreapointIds.add(styleDecl.breakpointId);
+    return treeStyleSourceIds.has(styleDecl.styleSourceId);
+  });
```
rpominov, line 104: Seem like we're putting into `treeStyles` only styles that satisfy `treeStyleSourceIds.has()`, but we put into `treeBreapointIds` **all** breakpoints. Is this a bug?

TrySound: It's filter function which iterates whole map. Just didn't want to add the second iteration.

rpominov: Yeah, but does this mean we're copying and merging all breakpoints?

kof: we should def. only copy breakpoints that are used on the copied instances

apps/builder/app/shared/tree-utils.ts
```diff
@@ -493,24 +495,58 @@ export const insertStyleSourceSelectionsCopyMutable = (
...
     const styleDeclCopy = {
       ...styleDecl,
       styleSourceId: newStyleSourceId,
+      breakpointId: newBreakpointId,
     };
     styles.set(getStyleDeclKey(styleDeclCopy), styleDeclCopy);
   }
 };
+
+const areBreakpointsSimilar = (left: Breakpoint, right: Breakpoint) => {
```
kof, line 529: this could be part of the css-engine I think, similar to matchMedia, maybe call it equalMedia()?

kof (approved):

apps/builder/app/shared/copy-paste/plugin-instance.ts
TrySound: You are right. I get what you mean now. Will add test

PR #1340, opened 2023-03-31
Webflow breakpoints

kof:
## Description

To enable copy-pasting instances from webflow without creating new breakpoints, we decided to switch by default to the breakpoints webflow uses.

User can still switch to mobile-first breakpoints like we had before once we have a new breakpoints editor, but then when copy/pasting from webflow they will get the missing breakpoints added if the copied styles are using those.

I had to temporarily hide the breakpoints editor behind a feature flag because we need the new editor to allow editing min or max width, so instead of adjusting this current UI, we need a new one.

I think its not a big deal right now to not have the breakpoints editor, webflow never had one.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/breakpoints/breakpoints-editor.tsx
```diff
@@ -73,7 +73,13 @@ const BreakpointEditorItem = ({
...
           css={{ textAlign: "right", width: 50 }}
           variant="ghost"
-          defaultValue={breakpoint.minWidth}
+          defaultValue={
+            "minWidth" in breakpoint
+              ? breakpoint.minWidth
+              : "maxWidth" in breakpoint
+              ? breakpoint.maxWidth
+              : 0
+          }
```
TrySound, line 82: `breakpoint.minWidth ?? breakpoint.maxWidth ?? 0`

apps/builder/app/builder/features/breakpoints/breakpoints.tsx
```diff
@@ -30,6 +29,7 @@ import {
   selectedBreakpointIdStore,
   selectedBreakpointStore,
 } from "~/shared/nano-states/breakpoints";
+import { sortMedia } from "@webstudio-is/css-engine";
```
TrySound, line 32: Nice

apps/builder/app/builder/features/breakpoints/breakpoints.tsx
```diff
@@ -133,28 +133,35 @@ export const Breakpoints = () => {
...
-                  >
-                    {breakpoint.label}
-                    <DropdownMenuItemRightSlot>
-                      {breakpoint.minWidth}
-                    </DropdownMenuItemRightSlot>
-                  </DropdownMenuCheckboxItem>
-                );
-              })}
+              {Array.from(breakpoints.values())
+                .sort(sortMedia)
```
TrySound, line 137: Functions passed to sort usually called "compare". So better "compareMedia"

apps/builder/app/builder/features/breakpoints/breakpoints.tsx
```diff
@@ -133,28 +133,35 @@ export const Breakpoints = () => {
...
+                .reverse()
```
TrySound, line 138: What' wrong with default order direction?

apps/builder/app/builder/features/breakpoints/use-update-canvas-width.ts
```diff
@@ -27,7 +27,10 @@ export const useUpdateCanvasWidth = () => {
     if (isPreviewMode || selectedBreakpoint === undefined) {
       return;
     }
-    const maxWidthBelowNextBreakpoint = (nextBreakpoint?.minWidth ?? 0) - 1;
+    const maxWidthBelowNextBreakpoint =
+      (nextBreakpoint && "minWidth" in nextBreakpoint
+        ? nextBreakpoint.minWidth ?? 1
+        : 1) - 1;
```
TrySound, line 33: Split the logic inside parentheses into separate variable and add some comments. The logic I getting less clear now.

apps/builder/app/builder/features/breakpoints/use-update-canvas-width.ts
```diff
@@ -36,7 +39,7 @@ export const useUpdateCanvasWidth = () => {
       : maxWidthBelowNextBreakpoint;
 
     const width = Math.max(
-      selectedBreakpoint.minWidth ?? 0,
+      "minWidth" in selectedBreakpoint ? selectedBreakpoint.minWidth ?? 0 : 0,
```
TrySound, line 42: Maybe no need to check for "in" operation. `selectedBreakpoint.minWidth ?? 0` look equal

apps/builder/app/builder/features/breakpoints/width-setting.tsx
```diff
@@ -27,13 +27,19 @@ export const WidthSetting = () => {
   // We want to enable unconstrained resizing in a preview mode
   const min = isPreviewMode
     ? minWidth
-    : Math.max(minWidth, selectedBreakpoint.minWidth ?? 0);
+    : Math.max(
+        minWidth,
+        "minWidth" in selectedBreakpoint ? selectedBreakpoint.minWidth ?? 0 : 0
```
TrySound, line 32: And here, same as before

apps/builder/app/builder/features/breakpoints/breakpoints-editor.tsx
kof: right I had a different type before, it hat do check like I did before, but now it can be simplified

apps/builder/app/builder/features/breakpoints/use-update-canvas-width.ts
kof: now yes, before I had a union type, a property was not available potentially so needed these checks

apps/builder/app/builder/features/breakpoints/breakpoints.tsx
kof: its just the order that makes visually sense and is aligned with webflow, while the sortMedia function sorts so that these are rendered in a correct order in the cssom

TrySound: So like from big to small for mobile first approach?

apps/builder/app/builder/features/breakpoints/breakpoints.tsx
```diff
@@ -133,45 +134,56 @@ export const Breakpoints = () => {
...
+                      onMouseLeave={() => {
+                        setBreakpointPreview(selectedBreakpoint);
+                      }}
+                      onSelect={() => {
+                        selectedBreakpointIdStore.set(breakpoint.id);
+                      }}
+                    >
+                      {breakpoint.label}
+                      <DropdownMenuItemRightSlot>
+                        {"minWidth" in breakpoint
```
TrySound, line 157: Nullish coalescing

packages/css-engine/src/core/sort-media.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, test, expect } from "@jest/globals";
+import { compareMedia } from "./sort-media";
```
TrySound, line 2: R3name .odule?

TrySound (approved):

PR #1344, opened 2023-04-01
Remove isEditable option from style source

TrySound:

isEditable is inferrable from the local "source" of style source. And no need to be passed into every item.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
What about future use cases like tag styles? I think we may have more sources that are not editable potentially

TrySound:
We will add `|| source === 'tag'`

TrySound:
Like we already have all the info necessary in input. We already have check by source.

kof (approved):

PR #1345, opened 2023-04-01
Upgrade to TypeScript 5

Andarist:


kof (approved):

packages/design-system/src/components/avatar.tsx
```diff
@@ -293,7 +293,15 @@ export const Avatar = React.forwardRef<
               mb: `-${theme.spacing[2]}`,
             }}
           >
-            <Status size={size && size > 2 ? "2" : "1"} variant={status} />
+            <Status
+              size={
+                (typeof size === "number" && size > 2) ||
+                (typeof size === "string" && parseInt(size, 10) > 2)
```
TrySound, line 299: typeof hides real type and Number.parseInt.
Ideally just define correct type in props.

PR #1349, opened 2023-04-02
Add style source states UI

TrySound:

Here added support for states in style source input under feature flag. Styles editing and rendering still doesn't work. Here only UI.

image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
@TrySound design has a bunch of differences to what I see on the screenshot

https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=2229-53886&t=LAbB95NsERSvIyX8-4

kof:
@taylornowotny lets discuss the list of pseudo states that we always show by default, figma includes: hover, focus, pressed only

kof:
- [ ] Truncation seems to be broken in stories
Screenshot 2023-04-02 at 20 29 32

kof:
- [ ] add pseudo states to stories

apps/builder/app/builder/features/style-panel/style-source/style-source-input.tsx
```diff
@@ -221,10 +235,20 @@ const markAddedValues = <Item extends IntermediateItem>(
   return items.map((item) => ({ ...item, isAdded: valueIds.has(item.id) }));
 };
 
+const userActionStates = [
+  ":hover",
+  ":active",
+  ":focus",
+  ":focus-visible",
+  ":focus-within",
+];
```
kof, line 244: list needs discussion

TrySound: Took the list here
https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes#user_action_pseudo-classes

kof: Discussed this list with taylor, I think we are fine, the initial idea was to stick with wf naming and initial list, but they have weird names, e.g. they call :active "pressed" or they call :focus-visible "focus keyboard" its clear that they were trying to be better understood by non-developers, but this type of renaming is harmful long-term, we will have an explanation under the  dropdown for each


tldr your list is fine

TrySound:
>  state menu items seem to be missing those other 2 states 

Very depends on logic so did not implement yet. Will do in separate PR

PR #1351, opened 2023-04-02
Generate declaration files before publishing packages

Andarist:


packages/asset-uploader/package.json
```diff
@@ -11,6 +11,7 @@
     "checks": "pnpm typecheck && pnpm lint && pnpm test",
     "dev": "build-package --watch",
     "build": "build-package",
+    "dts": "tsc --emitDeclarationOnly --declaration --declarationDir lib/types",
```
TrySound, line 14: Does it still require "include": ["src"] in tsconfig? Because of it .ts files in the root of packages are not checked.

Andarist: I wasn't exactly sure why you were using `include: ['src']` in the first place so I left it alone. If you don't need it then we could likely remove it from `tsconfig.json` **and** pass it to the args list here (we want `./lib/types` output to reflect `./src` and not `./`, so we need it for the dts emit itself)

Andarist: I would move this particular work to another PR though to keep the changes incremental and smaller

apps/builder/app/builder/builder.tsx
```diff
@@ -3,7 +3,7 @@ import { type Publish, usePublish, useSubscribe } from "~/shared/pubsub";
 import type { Build } from "@webstudio-is/project-build";
 import type { Project } from "@webstudio-is/project";
 import { theme, Box, type CSS, Flex, Grid } from "@webstudio-is/design-system";
-import type { AuthPermit } from "@webstudio-is/trpc-interface";
+import type { AuthPermit } from "@webstudio-is/trpc-interface/server";
```
TrySound, line 6: I think type should be available in non server route. Otherwise auto import can give us many issues.

patches/mitt@3.0.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/package.json b/package.json
+index ad38c7bd3a6eef608058259f8b2523b3a12b935d..ae9dc6f91a216aaa793294036443dbd839e3375d 100644
+--- a/package.json
++++ b/package.json
+@@ -9,6 +9,7 @@
+   "source": "src/index.ts",
+   "typings": "index.d.ts",
+   "exports": {
++    "types": "./index.d.ts",
```
TrySound, line 9: Aren't we staying with root "types" field?

apps/builder/app/builder/builder.tsx
Andarist: Right now there is no `@webstudio-is/trpc-interface` according to `exports` (it was like that before my change). It was working accidentally because the root fields were desynchronized from `exports`. So with the current state of this PR the autocomplete should work fine since the lang server doesn't recognize `@webstudio-is/trpc-interface` as a valid import source

patches/mitt@3.0.0.patch
Andarist: Root `types` will be used in the `-sass` repository. In this one, we've switched to `moduleResolution: bundler` and with that when `exports` are present in any package they become the only source of information, so packages in `node_modules` have to have valid `exports`

patches/totalist@3.0.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/package.json b/package.json
+index 6b4a8743c0d9a927e31e89fb5eeeafed3cb6b6de..b3b3f97818054feba7890c9f70f9d6248c6b1b3d 100644
+--- a/package.json
++++ b/package.json
+@@ -8,6 +8,7 @@
+   "types": "index.d.ts",
+   "exports": {
+     ".": {
++      "types": "./index.d.ts",
```
TrySound, line 9: Luke released you PR
https://github.com/lukeed/totalist/releases/tag/v3.0.1

TrySound (approved):
Looks great!

PR #1361, opened 2023-04-03
Add assets store

TrySound:

Here I replaced assetContainersStore with assetsStore. Now deleting and uploading states are maintained locally in assets provider. To preserve the order of uploading assets assetsStore map has value as undefined. This way assets.get(id) provides the same value. Though when iterating we see missing asset.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/assets/use-assets.tsx
+++ b/apps/builder/app/builder/shared/assets/use-assets.tsx
@@ -28,6 +28,9 @@ import { normalizeErrors, toastUnknownFieldErrors } from "~/shared/form-utils";
 import { assetsStore } from "~/shared/nano-states";
 import { useStore } from "@nanostores/react";
 
+// stubbed asset is necessary to preserve position of asset
+// while uploading and after it is uploaded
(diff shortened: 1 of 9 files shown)


apps/builder/app/builder/shared/assets/use-assets.tsx
```diff
@@ -1,26 +1,52 @@
...
+  PreviewAsset,
   UploadingAssetContainer,
 } from "./types";
 import { usePersistentFetcher } from "~/shared/fetcher";
 import type { ActionData } from "~/builder/shared/assets";
 import { normalizeErrors, toastUnknownFieldErrors } from "~/shared/form-utils";
+import { assetsStore } from "~/shared/nano-states";
+import { useStore } from "@nanostores/react";
+
+const stubAssets = (ids: Asset["id"][]) => {
```
istarkov, line 31: Comment would be good why we need stub asset

istarkov (approved):

PR #1366, opened 2023-04-04
Breakpoints redesign stage 2

kof:
## Description

- Canvas behavior should be exactly as in Webflow now - responsive canvas.
- No constraints for resizing, you can switch to different breakpoints when you change the canvas width just like in preview

## Not done in this PR

- [ ] breakpoints dropdown menu redesign
- [ ] on-canvas resizable handles
- [ ] flash-animate breakpoints list when breakpoint is switched through resize
- [ ] flash-animate left and right borders when breakpoint is switched
- [ ] gravity of resize handle dragging when reaching next breakpoint
- [ ] vertical labels with @media example that matches current breakpoint
- [ ] migrate min-width 0 breakpoints to base?
- [ ] hide outline while changing size
- [ ] ensure there is never both min and max in one media

## Todo

- [x] draggable positioning calc
- [x] shortcuts for zoom
- [x] sync width and selected breakpoint
- [x] remove scale
- [x] fix mobile-first breakpoints
- [x] tabbable toolbar items
- [x] icon for base
- [x] fix flicker
- [x] After breakpoint delete/add CascadeIndicator doesn't refresh
- [x] fix when there is no base, min-width 0 should have still a full width

## Steps for reproduction

1. click button
3. expect xyz

## Code Review

- [ ] hi @TrySound, @istarkov  , I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-engine/src/core/find-matching-media.ts
```diff
@@ -0,0 +1,20 @@
+import { compareMedia } from "./compare-media";
+import { matchMedia } from "./match-media";
+import type { MediaRuleOptions } from "./rules";
+
+// Find media rule that matches the given width when rendered.
+export const findMatchingMedia = <Media extends MediaRuleOptions>(
+  media: Array<Media>,
+  width: number
```
TrySound, line 8: Isn't it the same as match media?

taylornowotny:
When opening any breakpoint the canvas goes to max size and there's a message saying:

 "Please increase the canvas width.
"Mobile portrait" breakpoint maximum width is 479px."

I can resize with the menu but it's jumpy and flashes at max size at each interval.

The intended behavior from the docs is:
> "Switching breakpoints with the toggles will change the canvas width to the smallest available size for that breakpoint
> - if bp has a min-width, the min-width is the smallest size
> - if bp has a max-width, then the smallest size for that bp is the max-width of the next smallest bp +1 pixel
> - This behavior is same as Webflow so you can use them as a reference
> - Reasoning: A best practice for the user is to apply styles to the smallest width of a given bp, bc if content is designed for the smaller end then it will display properly on the larger end, but not necessarily the other way around."

taylornowotny:
Screenshot 2023-04-07 at 9 29 48 AM

+ can we give this button a min-width of 121px so it doesn't resize when the canvas width value goes from 4 digits to 3 digits & back? The movement is a bit distracting.

kof:
> can we give this button a min-width of 121px so it doesn't resize when the canvas width value goes from 4 digits to 3 digits & back? The movement is a bit distracting.

I was also thinking to hide the scale value when its 100%, since most of the time its not useful, webflow hides it too, but then I saw the breakpoints selector toolbar jumping around and removed this logic, with min-width we could potentially work this around  and make it stay

taylornowotny:
canvas setting button looks great now when resizing!

istarkov:
By default project opens with 0 width canvas
https://webstudio-builder-59fpf7vb1-webstudio-is.vercel.app/builder/71f88d3e-9231-489e-8ba0-8adc10215335
So nothing seen at all

istarkov:
If I set width for example 354px
image
Then switch on 768 clicking here
image
Then click 0
Canvas width is 0 again

istarkov:
If I set canvas width 
image
Then resize window it resets to defaults, 
image

IMO precents should be changed but not pixel size, if user (like me) has changed size explicitly

apps/builder/app/builder/features/breakpoints/cascade-indicator.tsx
```diff
@@ -0,0 +1,126 @@
...
+  return (
+    <>
+      <div
+        className={cascadeIndicatorStyle()}
+        data-direction="left"
+        style={indicatorStyle.left}
+      />
+      <div
+        className={cascadeIndicatorStyle()}
+        data-direction="right"
```
istarkov, line 121: In UI I see single indicator, what the reason to have left and right?
Any comments why we need both would be good.

istarkov:
After breakpoint delete/add CascadeIndicator doesn't refresh
image

apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
```diff
@@ -0,0 +1,56 @@
...
+              <ToolbarToggleItem
+                variant="subtle"
+                ref={breakpoint.id === selectedBreakpoint.id ? ref : undefined}
+                value={breakpoint.id}
+                key={breakpoint.id}
+              >
+                {breakpoint.minWidth ??
+                  breakpoint.maxWidth ??
+                  (breakpoint.id === selectedBreakpoint.id ? (
+                    <BpStarOnIcon size={22} />
```
istarkov, line 44: Probably I'm doing something wrong but how to see Star icon on the indicator? What I need to do?

apps/builder/app/builder/features/breakpoints/group-breakpoints.ts
```diff
@@ -0,0 +1,17 @@
+import { compareMedia } from "@webstudio-is/css-engine";
+import { isBaseBreakpoint } from "~/builder/shared/breakpoints";
+//import type { Breakpoint } from "@webstudio-is/project-build";
```
istarkov, line 3: remove

apps/builder/app/builder/features/breakpoints/group-breakpoints.test.ts
```diff
@@ -0,0 +1,26 @@
...
+describe("Group breakpoints", () => {
+  test("for ui", () => {
+    const initial = [
+      { minWidth: 1920 },
+      { minWidth: 1440 },
+      { minWidth: 1280 },
+      {},
+      { maxWidth: 991 },
+      { maxWidth: 767 },
+      { maxWidth: 479 },
```
istarkov, line 13: Are we going to mix breakpoints?

apps/builder/app/builder/features/breakpoints/group-breakpoints.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    const grouped = [
+      { minWidth: 1920 },
+      { minWidth: 1440 },
+      { minWidth: 1280 },
+      {},
+      { maxWidth: 991 },
+      { maxWidth: 767 },
+      { maxWidth: 479 },
+    ];
+    expect(groupBreakpoints(initial)).toStrictEqual(grouped);
```
istarkov, line 24: I don't get this test at all, as it looks like you test that initial array is equal to grouped. Having that initial === grouped it makes zero sense.

apps/builder/app/builder/features/breakpoints/group-breakpoints.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const groupBreakpoints = <
```
istarkov, line 5: Doesn't look like a group, but reorder/order as you don't have any groups here.

apps/builder/app/builder/features/workspace/workspace.tsx
```diff
@@ -45,6 +39,37 @@ const useSetWorkspaceRect = () => {
...
+) => {
+  let canvasHeight;
+  let canvasLeft;
+
+  if (workspaceRect?.height) {
+    canvasHeight = workspaceRect.height / (scale / 100);
+  }
+
+  if (workspaceRect?.width && canvasWidth) {
+    canvasLeft = Math.max((workspaceRect.width - canvasWidth) / 2, 0);
```
istarkov, line 55: I have a feeling that canvasLeft>0 only if transform scale === 1, 
that allows to not calculate left position but use for canvas positioning ordinary Flex box with justify-content: center. 
Less calculus is always better

packages/css-engine/src/core/find-applicable-media.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const media = [
+  {},
+  { maxWidth: 991 },
+  { maxWidth: 767 },
+  { maxWidth: 479 },
+  { minWidth: 1280 },
+  { minWidth: 1440 },
+  { minWidth: 1920 },
+];
```
istarkov, line 12: What if maxWidth and minWidth would have intersections? How grouping etc would work too

packages/css-engine/src/core/find-applicable-media.ts
```diff
@@ -0,0 +1,20 @@
+import { compareMedia } from "./compare-media";
+import { matchMedia } from "./match-media";
+import type { MediaRuleOptions } from "./rules";
+
+// Find media rule that matches the given width when rendered.
+export const findApplicableMedia = <Media extends MediaRuleOptions>(
+  media: Array<Media>,
+  width: number
+) => {
+  const sortedMedia = media
```
istarkov, line 10: Looks like you need `groupBreakpoints` here too?

kof:
@istarkov 

> By default project opens with 0 width canvas
> https://webstudio-builder-59fpf7vb1-webstudio-is.vercel.app/builder/71f88d3e-9231-489e-8ba0-8adc10215335
> So nothing seen at all

I am getting unexpected server error on this one, how do I see it?

kof:
> I am getting unexpected server error on this one, how do I see it?

Found an example of 0 canvas width, fixing

istarkov (approved):

apps/builder/docs/test-cases.md
```diff
@@ -65,7 +65,7 @@
 
    - Clicking on different breakpoints resizes the canvas
    - Canvas width slider lets you resize canvas only within its bounds
-   - Zoom slider works
+   - Scal slider works
```
TrySound, line 68: scal

apps/builder/app/shared/shortcuts/shortcuts.ts
```diff
@@ -93,4 +54,6 @@ export const useSharedShortcuts = () => {
     { enableOnFormTags: true, enableOnContentEditable: true },
     []
   );
+
+  useHotkeys("esc", escapeSelection, { enableOnContentEditable: true }, []);
```
TrySound, line 58: Ivan removed it from here. Is it rebase?

apps/builder/app/builder/shared/nano-states/index.ts
```diff
@@ -24,3 +24,19 @@ export const projectContainer = atom<Project | undefined>();
...
+  [canvasWidthContainer, workspaceRectStore],
+  (canvasWidth, workspaceRect) => {
+    if (
+      canvasWidth === undefined ||
+      workspaceRect === undefined ||
+      canvasWidth <= workspaceRect.width
+    ) {
+      return 100;
+    }
+    return parseFloat(((workspaceRect.width / canvasWidth) * 100).toFixed(2));
```
TrySound, line 40: Number.parseFloat

apps/builder/app/builder/shared/breakpoints/is-base-breakpoint.ts
```diff
@@ -0,0 +1,4 @@
+export const isBaseBreakpoint = (breakpoint: {
```
TrySound, line 1: shared/breakpoints.ts?

TrySound (approved):

apps/builder/app/builder/features/style-panel/style-panel.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from "@webstudio-is/design-system";
 import type { Instance } from "@webstudio-is/project-build";
 import type { Publish } from "~/shared/pubsub";
-import { selectedBreakpointStore } from "~/shared/nano-states/breakpoints";
+import { selectedBreakpointStore } from "~/builder/shared/breakpoints/stores";
```
TrySound, line 10: Why reexport from index if import from stores?

PR #1370, opened 2023-04-04
Drop chromatic from dependencies

TrySound:
We probably don't need to have chromatic in dependencies.
Especially since the latest version is getting 20MB bigger

https://packagephobia.com/result?p=chromatic

Also removed builds from chromatic workflows, "source" export condition allows to avoid it.

## Code Review

- [ ] hi @Andarist, I need you to do

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -80,8 +80,6 @@ jobs:
 
       - uses: ./.github/actions/ci-setup
 
-      - run: DATABASE_URL="-" AUTH_SECRET="-" pnpm build
-
       - name: Chromatic
         id: chromatic
         uses: chromaui/action@v1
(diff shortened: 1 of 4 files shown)


kof (approved):
sure if we are still able to run storybook as before?

.github/workflows/chromatic.yml
```diff
@@ -52,9 +52,6 @@ jobs:
 
       - uses: ./.github/actions/ci-setup
 
-      # only packages have to be built as they might be imported by Storybooks
-      - run: pnpm build-packages
```
Andarist, line 56: is this not needed as Storybooks rely on the source condition?

Andarist (approved):

.github/workflows/chromatic.yml
TrySound: yes

PR #1372, opened 2023-04-05
Simplify assets uploading

TrySound:

- with persistent fetcher there is no need in assets provider
- replaced async FileReader api with sync URL.createObjectURL for preview while uploading

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/assets/asset-upload.tsx
+++ b/apps/builder/app/builder/shared/assets/asset-upload.tsx
@@ -1,13 +1,33 @@
 import { type ChangeEvent, useRef } from "react";
-import { Button, Flex, Tooltip } from "@webstudio-is/design-system";
+import { Button, Flex, Tooltip, toast } from "@webstudio-is/design-system";
 import { UploadIcon } from "@webstudio-is/icons";
-import type { AssetType } from "@webstudio-is/asset-uploader";
+import {
+  type AssetType,
+  MAX_UPLOAD_SIZE,
+  toBytes,
+} from "@webstudio-is/asset-uploader";
 import { FONT_MIME_TYPES } from "@webstudio-is/fonts";
-import { useAssets } from "./use-assets";
+import { useUploadAsset } from "./use-assets";
 import { useAuthPermit } from "~/shared/nano-states";
 
+const maxSize = toBytes(MAX_UPLOAD_SIZE);
+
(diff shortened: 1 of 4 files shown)


apps/builder/app/builder/shared/assets/use-assets.tsx
```diff
@@ -54,100 +43,81 @@ const setAsset = (asset: Asset) => {
...
-  files: File[]
-): Promise<[UploadingAssetContainer, FormData][]> => {
-  const assets: Array<Promise<[UploadingAssetContainer, FormData]>> = [];
+const getFilesData = (type: AssetType, files: File[]): FileData[] => {
+  return files.map((file) => {
+    return {
+      id: crypto.randomUUID(),
+      type,
+      file,
+      objectURL: URL.createObjectURL(file),
```
istarkov, line 59: In order to remove mem leaks from what I know you should call `revokeObjectURL()` after upload completes

istarkov: But that can cause the image blink if the uploaded Image is not loaded, not sure. Probably can be called for all uploaded images after few minutes delay.

apps/builder/app/builder/shared/assets/asset-upload.tsx
```diff
@@ -1,13 +1,33 @@
 import { type ChangeEvent, useRef } from "react";
-import { Button, Flex, Tooltip } from "@webstudio-is/design-system";
+import { Button, Flex, Tooltip, toast } from "@webstudio-is/design-system";
 import { UploadIcon } from "@webstudio-is/icons";
-import type { AssetType } from "@webstudio-is/asset-uploader";
+import {
+  type AssetType,
+  MAX_UPLOAD_SIZE,
```
istarkov, line 6: Not current PR issue, but MAX_UPLOAD_SIZE should be builder env variable and not a const inside asset-uploader

apps/builder/app/builder/shared/assets/use-assets.tsx
istarkov: Anyway checked uploading ton of huge files, don't see any leak.

istarkov: Seems revokeObjectURL exists ;-)

istarkov (approved):

PR #1376, opened 2023-04-05
Upload files without creating assets

TrySound:

Now we use assets endpoint only to upload asset and them add its data with store patch. This means assets are now fully handled with patches.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/src/db/create.ts
+++ b/packages/asset-uploader/src/db/create.ts
@@ -1,18 +1,11 @@
 import type { FontMeta } from "@webstudio-is/fonts";
-import {
-  prisma,
-  type Location,
-  type Project,
-  type Asset,
-} from "@webstudio-is/prisma-client";
-import { Env, type ImageMeta } from "../schema";
+import type { Location, Project, Asset } from "@webstudio-is/prisma-client";
+import type { ImageMeta } from "../schema";
 import { formatAsset } from "../utils/format-asset";
 import {
   authorizeProject,
   type AppContext,
 } from "@webstudio-is/trpc-interface/server";
-
-const env = Env.parse(process.env);
 
 type BaseOptions = {
   id: string;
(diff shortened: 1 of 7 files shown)


packages/asset-uploader/src/db/create.ts
```diff
@@ -46,91 +37,21 @@ export const createAssetWithLimit = async (
...
-          {
-            projectId,
-            status: "UPLOADING",
-            createdAt: { gt: new Date(Date.now() - UPLOADING_STALE_TIMEOUT) },
-          },
-        ],
-      },
-    });
-
-    if (count >= env.MAX_ASSETS_PER_PROJECT) {
```
rpominov, line 70: Do we still check limit at the server side somewhere? Need to make sure we check as client-side check is easy to bypass.

TrySound: We can do truncation in patches, wdyt?

packages/asset-uploader/src/patch.ts
```diff
@@ -38,4 +42,28 @@ export const patchAssets = async (
...
+  }
+  await prisma.asset.createMany({
+    data: addedAssets.map((asset) => ({
+      id: asset.id,
+      location: asset.location,
+      name: asset.name,
+      size: asset.size,
+      format: asset.format,
+      projectId: asset.projectId,
+      meta: JSON.stringify(asset.meta),
```
rpominov, line 65: I worry that we trust all this data from client

TrySound: Our whole data management trust client data. Though at least we validate it with zod.

rpominov: Yeah, and I worry about that :)
Maybe it's ok for now because we want to move fast, but this can cause security issues or just break things. Potential issues I see here:

1. If I know `name` of someone's asset that I don't have access to, I can set `name` of my asset to that name, and get the file
2. I can cause backend to read / delete local files if I set `location` to `LOCAL`
3. I can set wrong `size/format/meta` and cause bugs. This is probably relatively ok, same as submitting somehow broken tree etc., although in case of size, this may allow to bypass size limits if we have them.

packages/asset-uploader/src/db/create.ts
rpominov: Sounds good

PR #1390, opened 2023-04-07
Fixed missing Prisma-generated type reexport

Andarist:


packages/prisma-client/src/types.ts
```diff
@@ -6,5 +6,6 @@ export type {
   Project,
   Asset,
   DashboardProject,
+  AuthorizationRelation,
```
TrySound, line 9: what???

TrySound:
Could you also fix turbo in publish.yaml?

packages/prisma-client/src/types.ts
Andarist: the problem is that TS has to insert smth like `import('@webstudio-is/prisma-client').AuthorizationRelation` in the generated `.d.ts` for `@webstudio-is/authorization-token` but to do that the type has to be exported

TrySound (approved):

PR #1403, opened 2023-04-08
Provide asset path getter to css engine

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1179

Styles currently get urls like this

https://assets.webstudio.is/e55bd4Lu_400x400_8X-Q89JB6QQIxu8912gCK.jpg

While image component get optimized one like this

https://image-transform.wstd.io/cdn-cgi/image/width=384, quality=80, format=auto/https://assets.webstudio.is/e55bd4Lu_400x400_8X-Q89JB6QQIxu8912gCK.jpg

We need a way to provide different url than hardcoded in asset. Here I added getAssetPath support for css engine to inject right path by asset id.

This will also unlock us from decoupling styles and assets on server side.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-engine/src/core/to-value.ts
+++ b/packages/css-engine/src/core/to-value.ts
@@ -1,12 +1,12 @@
+import type { Asset } from "@webstudio-is/asset-uploader";
 import type { StyleValue } from "@webstudio-is/css-data";
 import { DEFAULT_FONT_FALLBACK, SYSTEM_FONTS } from "@webstudio-is/fonts";
 
-type ToCssOptions = {
-  withFallback: boolean;
-};
+export type GetAssetPath = (assetId: Asset["id"]) => undefined | string;
 
-const defaultOptions = {
-  withFallback: true,
+type ToCssOptions = {
+  withFallback?: boolean;
+  getAssetPath?: GetAssetPath;
 };
 
 // exhaustive check, should never happen in runtime as ts would give error
@@ -16,8 +16,9 @@ const assertUnreachable = (_arg: never, errorMessage: string) => {
 
 export const toValue = (
   value?: StyleValue,
-  options: ToCssOptions = defaultOptions
+  options: ToCssOptions = {}
 ): string => {
+  const { withFallback = true, getAssetPath } = options;
   if (value === undefined) {
     return "";
   }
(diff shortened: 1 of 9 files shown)


packages/css-engine/src/core/to-value.ts
```diff
@@ -16,16 +16,17 @@ const assertUnreachable = (_arg: never, errorMessage: string) => {
 
 export const toValue = (
   value?: StyleValue,
-  options: ToCssOptions = defaultOptions
+  options: ToCssOptions = {}
 ): string => {
+  const { withFallback = true, getAssetPath } = options;
```
kof, line 21: hmm I wasn't planing on adding application-level logic to css-engine, is there a different way to do this outside the engine?

TrySound: Then we need to map style object. Though we probably already do this. And types with asset will be css engine specific.

kof: maybe we could pass some object with toString method and cssEngine would just stringify it without knowing any internals?

TrySound: And then we need to map asset to apply loader. So maybe even change the type from full asset to {id,path}

kof: is that bad?

TrySound: Let's see

TrySound: And we also gonna have 2 toValue because style panel heavily use one from css-engine but it will not be compatible with data from db.

istarkov: I like Bogdan solution its just some path converter, alternatives from my POV makes code less readable abd cause creation of client only types for styles, or splitting toValue logic

TrySound: Yep, just tried to add separate types. And it's huge amount of conversions and complexities.

kof: I am open to any suggestions that allow us to decouple css-engine from application

kof: you could add some sort of plugin/transformer in a generic way that would do this mapping

TrySound: Maybe more like this? No reliance on "id".
```
toValue(value, { resolveImagePath(imageValue): string }?
```

PR #1404, opened 2023-04-09
Allow to map individual style values when stringify to css

TrySound:
Related to https://github.com/webstudio-is/webstudio-builder/pull/1403

Here I removed app specific case from css-engine toValue by providing transformValue function which is called on every individual style value inside nested structure (like var or layers).

With this function we can resolve assets and remove complex resolving logic from server.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-engine/src/core/to-value.ts
+++ b/packages/css-engine/src/core/to-value.ts
@@ -1,44 +1,50 @@
 import type { StyleValue } from "@webstudio-is/css-data";
 import { DEFAULT_FONT_FALLBACK, SYSTEM_FONTS } from "@webstudio-is/fonts";
 
-type ToCssOptions = {
-  withFallback: boolean;
-};
-
-const defaultOptions = {
-  withFallback: true,
-};
+export type TransformValue = (styleValue: StyleValue) => undefined | StyleValue;
 
 // exhaustive check, should never happen in runtime as ts would give error
 const assertUnreachable = (_arg: never, errorMessage: string) => {
   throw new Error(errorMessage);
 };
 
+const fallbackTransform: TransformValue = (styleValue) => {
+  if (styleValue.type === "fontFamily") {
+    const firstFontFamily = styleValue.value[0];
+    const fallbacks = SYSTEM_FONTS.get(firstFontFamily);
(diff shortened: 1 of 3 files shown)


packages/css-engine/src/core/to-value.ts
```diff
@@ -1,44 +1,50 @@
...
-  withFallback: true,
-};
+export type TransformValue = (styleValue: StyleValue) => undefined | StyleValue;
 
 // exhaustive check, should never happen in runtime as ts would give error
 const assertUnreachable = (_arg: never, errorMessage: string) => {
   throw new Error(errorMessage);
 };
 
+const fallbackTransform: TransformValue = (styleValue) => {
```
kof, line 11: do you want to move the transform to the application in a separate step?

kof: because if you moved it right away, instead of providing a transform for withFallback: false, you would be able to turn this around and provide transform always and not provide it for withFallback: false

TrySound: This is kinda default stuff, not builder specific. Do we need to move it into application?

TrySound: > because if you moved it right away, instead of providing a transform for withFallback: false, you would be able to turn this around and provide transform always and not provide it for withFallback: false

I will need to put it into every addStyleRule in builder and in sdk. And somehow share it caz the logic is quite complicated

kof (approved):

packages/css-engine/src/core/to-value.ts
kof: > This is kinda default stuff, not builder specific. Do we need to move it into application?

yes because this transform contains application level logic and dependency

TrySound: Alright, someday then 😄

kof: we can move it later as well if its too much work

TrySound: I mean yes, better later

PR #1412, opened 2023-04-11
UX improvement for Inputs that have both keywords and units

### LINKED ISSUE #956: UX improvement for Inputs that have both keywords and units
The current implementation shows a select menu button on the input when a user starts typing letters, but a unit menu button when the user types numbers (sometimes - it doesn't seem to show up consistently). This is not a good pattern becuase the unit menu should be visible at all times so the user knows that this input accepts numbers and not just key words.

>  My suggestion is similar to the current implementation except that we never show a keyword menu button and always show units (when both are allowed). Keywords can be typed + an autocomplete menu shows, but no keyword select menu.

In the future we could have both menus showing, but the above suggestion would work pretty well and hopefully not require a lot of work.

istarkov:
## Description

closes #956

+ In case if user not edited the value but immediately clicked to unit select we are trying to convert em px etc values

## Steps for reproduction

1. Edit `width` value, enter 10
go to unit select set `em` - see 10em as final result

2. Click on unit select of the width, change em to px see 10em converted to 140px or like (depends on em size)

3. Enter `Auto` in width, click Enter. 
Click on UnitSelect select EM, see browser style is used to set em width.

4. Enter Auto in width, switch on Unit select, set px, see browser style is used.

5. See that now unit selector is visible always. With some exclusions like aspectRatio as it has keywords and numeric values.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/css-value-input/convert-units.ts
+++ b/apps/builder/app/builder/features/style-panel/shared/css-value-input/convert-units.ts
@@ -1,5 +1,4 @@
 import type { Unit } from "@webstudio-is/css-data";
-// import { selectedInstanceUnitSizesStore } from "~/shared/nano-states";
 
 const convertibleUnits = ["px", "ch", "vw", "vh", "em", "rem"] as const;
 
--- a/apps/builder/app/builder/features/style-panel/shared/css-value-input/convert-units.ts
+++ b/apps/builder/app/builder/features/style-panel/shared/css-value-input/convert-units.ts
@@ -1,0 +1,29 @@
+import type { Unit } from "@webstudio-is/css-data";
+// import { selectedInstanceUnitSizesStore } from "~/shared/nano-states";
+
+const convertibleUnits = ["px", "ch", "vw", "vh", "em", "rem"] as const;
+
+type ConvertibleUnit = (typeof convertibleUnits)[number];
+
+export type UnitSizes = Record<ConvertibleUnit, number>;
+
+const isConvertibleUnit = (unit: Unit): unit is ConvertibleUnit =>
+  convertibleUnits.includes(unit as ConvertibleUnit);
+
+export const convertUnits =
+  (unitSizes: UnitSizes) =>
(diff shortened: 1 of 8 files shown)


apps/builder/app/builder/features/style-panel/shared/css-value-input/convert-units.ts
```diff
@@ -0,0 +1,29 @@
+import type { Unit } from "@webstudio-is/css-data";
+// import { selectedInstanceUnitSizesStore } from "~/shared/nano-states";
```
kof, line 2: commented

apps/builder/app/builder/features/style-panel/shared/css-value-input/convert-units.ts
```diff
@@ -0,0 +1,29 @@
...
+  convertibleUnits.includes(unit as ConvertibleUnit);
+
+export const convertUnits =
+  (unitSizes: UnitSizes) =>
+  (value: number, from: Unit, to: Unit): number => {
+    if (from === to) {
+      return value;
+    }
+
+    if (!isConvertibleUnit(from)) {
```
kof, line 20: !i

packages/design-system/src/components/input-field.tsx
```diff
@@ -175,24 +177,38 @@ export const InputField = forwardRef(
...
-      ref={mergeRefs(ref, containerRef ?? null)}
-    >
-      <Input {...props} ref={inputRef} />
-    </Container>
-  )
+  ) => {
+    // Our input field can contain multiple focused elements,
+    // so we need to use useFocusWithin to track focus within the container.
+    const { focusWithinProps } = useFocusWithin({
+      onFocusWithin: (e) => onFocus?.(e),
```
kof, line 196: why do you need this wrapping instead of this?
```suggestion
      onFocusWithin: onFocus,
```

kof (approved):

PR #1413, opened 2023-04-11
Breakpoints stage 3

kof:
## Description

1. Input instead of a slider

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

- [ ] hi @istarkov  , I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov:
Adding breakpoints is confusing and has bug.

Bug is that indicator length doesn't change on adding new breakpoint, to reproduce add 1600 breakpoint, then see 768 has no indicator under

image

The other issue that until done is clicked 
image
Edited breakpoint is shown at top, that's confusing as it changes it position, value etc

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+          >
+            PX
+          </Text>
+        }
+      />
+    </Flex>
+  );
+};
+/*
+const useScrub = ({
```
istarkov, line 71: delete

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  const onChange = (value: number) => {
+    setCanvasWidth(value);
+    const applicableBreakpoint = findApplicableMedia(
+      Array.from(breakpoints.values()),
+      value
+    );
+    if (applicableBreakpoint) {
+      selectedBreakpointIdStore.set(applicableBreakpoint.id);
```
istarkov, line 42: Mb doesn't do that during the edit, also from what I know stores are dumb, and better to not update if the value is already the same, to not force rerendering

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+    <InputField
+      id={id}
+      value={intermediateValue ?? value}
+      ref={scrubRef}
+      inputRef={inputRef}
+      type="number"
+      suffix={suffix}
+      min={min}
+      onChange={(event) => {
+        setIntermediateValue(event.target.valueAsNumber);
```
istarkov, line 154: wow, I never used `valueAsNumber`

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+  css: CSS;
+  tabIndex?: number;
+  min?: number;
+}) => {
+  const [intermediateValue, setIntermediateValue] = useState<number>();
+  const handleChangeComplete = (value: number) => {
+    onChange(Math.max(value, min ?? 0));
+    setIntermediateValue(undefined);
+  };
+  const handleChangeCompleteRef = useRef(handleChangeComplete);
```
istarkov, line 127: Never updated so would hold outdated value of `handleChangeComplete`

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const { scrubRef, inputRef } = useScrub({
+    value,
+    onChange: handleChangeCompleteRef.current,
+  });
+
+  useEffect(
+    () => () => {
```
istarkov, line 135: Really hard to read that it's unsubscribe logic, 
better to use curly braces and explicit return + 
add a comment that we need this because of the Dialog blabla

packages/design-system/src/components/primitives/use-scrub.ts
```diff
@@ -0,0 +1,83 @@
...
+    const scrub = numericScrubControl(scrubRefCurrent, {
+      // @todo: after this https://github.com/webstudio-is/webstudio-builder/issues/564
+      // we can switch back on using just initial value
+      //
+      // For now we are reusing controls for different selectedInstanceData,
+      // and the best here is to call useEffect every time selectedInstanceData changes
+      // and recreate numericScrubControl.
+      // Until we have decision do we use key properties for this or not,
+      // on of the solution to get value inside scrub is to use ref and lazy getter.
+      // Getter to avoid recreating scrub on every value change
```
istarkov, line 52: We decided to not use keys, and seems like 
// Getter to avoid recreating scrub on every value change
is enough

packages/design-system/src/components/primitives/use-scrub.ts
```diff
@@ -0,0 +1,83 @@
+import { useEffect, useRef } from "react";
+import {
+  numericScrubControl,
+  type NumericScrubValue,
+} from "./numeric-gesture-control";
+import { unstable_batchedUpdates as unstableBatchedUpdates } from "react-dom";
+
+export const useScrub = ({
```
istarkov, line 8: I would add comments what `scrubRef` means, and why its needed at all

packages/design-system/src/components/primitives/use-scrub.ts
```diff
@@ -0,0 +1,83 @@
...
+  onChange,
+  onChangeComplete,
+  shouldHandleEvent,
+}: {
+  value: NumericScrubValue;
+  onChange: (value: NumericScrubValue) => void;
+  onChangeComplete?: (value: NumericScrubValue) => void;
+  shouldHandleEvent?: (node: EventTarget) => boolean;
+}) => {
+  const scrubRef = useRef<HTMLDivElement | null>(null);
```
istarkov, line 19: It looks to me that it's not needed at all, and we can use `inputRef.parent.appendChild(<div tabindex=-1)` whenever needed. The only thing it used is to preserve focus inside input container

istarkov (approved):

packages/design-system/src/components/primitives/use-scrub.ts
kof: so removing the comment then

apps/builder/app/builder/features/breakpoints/width-input.tsx
kof: we have to do it as we scroll, breakpoints need to always reflect the size of the canvas

istarkov: I mean u need to do this 
```
if (selectedBreakpointIdStore.get() !== applicableBreakpoint.id) {
  selectedBreakpointIdStore.set(applicableBreakpoint.id);
}
```
otherwise it forces all listeners to render

packages/design-system/src/components/primitives/use-scrub.ts
kof: its also used for focus inside css value input ` scrubRef.current?.focus();`

istarkov: For same reason, to preserve focus inside container

TrySound (approved):

PR #1428, opened 2023-04-12
Fix cascade indicators on scrub move

istarkov:
## Description

Map used based on
https://react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
+++ b/apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
@@ -4,7 +4,7 @@ import {
   ToolbarToggleGroup,
   ToolbarToggleItem,
 } from "@webstudio-is/design-system";
-import { useRef } from "react";
+import { useCallback, useRef } from "react";
 import { CascadeIndicator } from "./cascade-indicator";
 import { BpStarOffIcon, BpStarOnIcon } from "@webstudio-is/icons";
 import { useSetInitialCanvasWidth } from ".";
@@ -18,11 +18,14 @@ import { groupBreakpoints } from "~/shared/breakpoints";
 export const BreakpointsSelector = () => {
   const breakpoints = useStore(breakpointsStore);
   const selectedBreakpoint = useStore(selectedBreakpointStore);
-  const ref = useRef(null);
+  const refs = useRef(new Map<string, HTMLButtonElement>());
   const setInitialCanvasWidth = useSetInitialCanvasWidth();
+  const getButtonById = useCallback((id: string) => refs.current.get(id), []);
+
   if (selectedBreakpoint === undefined) {
     return null;
   }
+
   return (
     <Toolbar>
       <ToolbarToggleGroup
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
```diff
@@ -42,7 +45,13 @@ export const BreakpointsSelector = () => {
...
               <ToolbarToggleItem
                 variant="subtle"
-                ref={breakpoint.id === selectedBreakpoint.id ? ref : undefined}
+                ref={(node) => {
+                  if (node) {
+                    refs.current.set(breakpoint.id, node);
+                    return;
+                  }
+                  refs.current.delete(breakpoint.id);
+                }}
```
TrySound, line 54: I think ref should be stable otherwise remounts on every render. But I'm not sure

istarkov: I copypasted from react https://react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback and here are elements not refs
So map is not stable but map values are

apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
```diff
@@ -18,11 +18,14 @@ import { groupBreakpoints } from "~/shared/breakpoints";
 export const BreakpointsSelector = () => {
   const breakpoints = useStore(breakpointsStore);
   const selectedBreakpoint = useStore(selectedBreakpointStore);
-  const ref = useRef(null);
+  const refs = useRef(new Map<string, HTMLButtonElement>());
```
kof, line 21: why do you need a map, seems like it could be a normal ref

istarkov: https://react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback

kof: oh you want many nodes

apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
kof: I don't think ref needs to be stable

apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
istarkov: It can't be a single ref, as the issue was that the ref was outdated. And not set

kof (approved):

apps/builder/app/builder/features/breakpoints/breakpoints-selector.tsx
TrySound: Ah got it. It does not remount, just recall again

> React will also call your ref callback whenever you pass a different ref callback. In the above example, (node) => { ... } is a different function on every render. When your component re-renders, the previous function will be called with null as the argument, and the next function will be called with the DOM node.

PR #1452, opened 2023-04-14
Upgrade to react 18

TrySound:
Ref https://react.dev/blog/2022/03/29/react-v18 https://react.dev/blog/2022/03/08/react-18-upgrade-guide

- bumped all deps to use react 18
- storybook still somehow enforces react 17 (but not react-dom) so forces with overrides until bump of storybook
- both storybook and remix still use legacy api so concurrent features are not yet enabled. Will do in separate PRs
- fixed weird react-hot-toast error which expose function as renderable type
- fixed a few untyped useCallback which failed with new react types
- sdk size became a little bigger

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):
!!!!

istarkov:
Something with local development, canvas loaded like 1ce per 10 reloads. Investigating

kof:
even before switching I had frequently the case that builder loaded but canvas was empty and only after another reload it loaded properly, I am sure we have some problem with ready state of the canvas

kof (approved):

TrySound:
Merged with sync mode. So all works like in react 17 for now

PR #1453, opened 2023-04-14
Add image proxy route

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1179

Here added image proxy - simple remix route to interop image requests locally and in preview. We have now only single image loader which accepts image base url.

Here support only local deployment. Preview and saas preview to RESIZE_ORiGIN env.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/env/env.public.server.ts
```diff
@@ -28,9 +28,10 @@ const env = {
   PUBLISHER_HOST: process.env.PUBLISHER_HOST || null,
   BUILD_REQUIRE_SUBDOMAIN: process.env.BUILD_REQUIRE_SUBDOMAIN === "true",
 
-  // Must be set for Vercel deployments
-  RESIZE_ORIGIN: process.env.RESIZE_ORIGIN,
-
+  IMAGE_BASE_URL:
```
istarkov, line 31: Having proxy this doesn't look right. The idea that from the client we always have image/assets `src`
like 
```jsx
<img src={`/cdn-cgi/image/${resizeoptions}/${name}` />
```

istarkov: So in env.public this should be a `_PATH`

apps/builder/app/routes/cgi.image.$options.$name.ts
```diff
@@ -0,0 +1,17 @@
+import { createReadStream } from "node:fs";
```
istarkov, line 1: Looks like localhost only proxy, but should be for vercel too

kof (approved):
I don't have really understanding what this is oding

apps/builder/app/routes/$.tsx
```diff
@@ -15,13 +15,10 @@ export const loader = async ({ request }: LoaderArgs) => {
   }
 
   const params: Params = {
+    imageBaseUrl: env.IMAGE_BASE_URL,
```
istarkov, line 18: Lets add comment above about possible values it can be. Also how it different from assetBaseUrl
As it can be just basePath and can be baseUrl.
Same comments for the IMAGE_BASE_URL
Also lets hardcode default for `IMAGE_BASE_URL` at code if not already

apps/builder/app/routes/asset.image.$name.ts
```diff
@@ -0,0 +1,35 @@
...
+  const quality = url.searchParams.get("quality");
+  const format = url.searchParams.get("format");
+  if (width === null || quality === null || format === null) {
+    throw Error("Options are invalid");
+  }
+  const options = `width=${width},quality=${quality},format=${format}`;
+
+  if (env.RESIZE_ORIGIN !== undefined) {
+    const assetUrl = `${env.ASSET_BASE_URL}${name}`;
+    return fetch(`${env.RESIZE_ORIGIN}/cdn-cgi/image/${options}/${assetUrl}`);
```
istarkov, line 26: Add todo and issue please in SaaS, 
to add here a secret key and use our own worker implementation or ruleset for transforms. 
In case of no SECRET we should just give an original Image.
This is to avoid exploiting our server.

apps/builder/app/routes/asset.image.$name.ts
```diff
@@ -0,0 +1,35 @@
...
+  const width = url.searchParams.get("width");
+  const quality = url.searchParams.get("quality");
+  const format = url.searchParams.get("format");
+  if (width === null || quality === null || format === null) {
+    throw Error("Options are invalid");
+  }
+  const options = `width=${width},quality=${quality},format=${format}`;
+
+  if (env.RESIZE_ORIGIN !== undefined) {
+    const assetUrl = `${env.ASSET_BASE_URL}${name}`;
```
istarkov, line 25: `ASSET_BASE_URL` plz add comment with possible defaults in env ts file 
/**
* Possible values are `https://assets-dev....`
*//

istarkov (approved):
Good step forward removal of our domains from image urls

apps/builder/app/routes/$.tsx
TrySound: Hardcoded for builder but would keep required for saas which use own worker implementation.

PR #1454, opened 2023-04-14
Ability to link by #ID; Fixed a bug with links in RTE

### LINKED ISSUE #991: Props panel: always show ID
Screenshot 2023-02-10 at 16 26 29

- [ ] Make sure all our components can take an `id` prop and pass it to `<element id=`
- [ ] Add `id` prop to "always visible" in all components

Not essential:

- [ ] Show a error if `id` is not unique across all components on the page
- [ ] Always show `id` first and add a separator after it
- [ ] Custom label `ID` instead of `Id`

### LINKED ISSUE #993: Props panel: control type URL
Screenshot 2023-02-10 at 17 08 49

- [x] Add control type URL and use it for `Link.href`
- [x] Ability to enter arbitrary url
- [x] Ability to choose a page
- [x] Make sure to call onChange at `blur` when user enters URL

Not essential:

- [ ] Section link
- [x] Email link
- [x] Phone link
- [x] Attachment link

Link Properties

### LINKED ISSUE #1380: RTE link styling regression
When editing RTE, link suddenly gets different styles than after exiting editing mode

https://user-images.githubusercontent.com/52824/230070052-df32dec5-b3a1-460d-a41d-dde94eb08e70.mp4

### LINKED ISSUE #1475: Ability to link to a section on a specific page
Context: https://github.com/webstudio-is/webstudio-builder/pull/1454#issuecomment-1511164203

We should consider an ability to create a link to a specific section on a specific page. Perhaps have a section picker next to the page picker here:

Screenshot 2023-04-17 at 14 54 21

rpominov:
## Description

1. Made `href` always present (# is used when not set, similar to Webflow). Fixes #1380
2. Added "Section" option in URL Control. Closes #993 
3. Added "ID" prop that is visible in all components. Closes #991 
4. Closes #1475

## Steps for reproduction

1. Add ID prop to an instance
4. Add a link that references that instance by ID

## Code Review

- [ ] hi @TrySound, I need you to
  - have a look at how I deal with stores to get list of available IDs

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/props-panel/controls/url.tsx
+++ b/apps/builder/app/builder/features/props-panel/controls/url.tsx
@@ -1,4 +1,5 @@
 import { useStore } from "@nanostores/react";
+import { computed } from "nanostores";
 import {
   instancesStore,
   pagesStore,
@@ -33,7 +34,6 @@ import {
 } from "../shared";
 import type { Instance, Page } from "@webstudio-is/project-build";
 import { SelectAsset } from "./select-asset";
-import { computed, type ReadableAtom } from "nanostores";
 
 type UrlControlProps = ControlProps<
   "url",
@@ -222,6 +222,7 @@ const BaseAttachment = ({ prop, onChange, onSoftDelete }: BaseControlProps) => (
   </Row>
 );
 
+// store that contains IDs off all instances in the selected page
 const pageInstancesStore = computed(
   [selectedPageStore, instancesStore],
   (page, instances) => {
@@ -249,7 +250,7 @@ const pageInstancesStore = computed(
   }
 );
 
-let sectionsStore: ReadableAtom<Map<string, string>> = computed(
+let sectionsStore = computed(
   [pageInstancesStore, propsStore],
   (pageInstances, props) => {
     const sections = new Map<Instance["id"], string>();
--- a/apps/builder/app/builder/features/props-panel/controls/url.tsx
+++ b/apps/builder/app/builder/features/props-panel/controls/url.tsx
@@ -259,6 +259,7 @@ let sectionsStore: ReadableAtom<Map<string, string>> = computed(
       if (
         prop.type === "string" &&
         prop.name === "id" &&
+        prop.value.trim() !== "" &&
         pageInstances.has(prop.instanceId)
       ) {
         sections.set(prop.instanceId, prop.value);
--- a/apps/builder/app/builder/features/props-panel/controls/url.tsx
+++ b/apps/builder/app/builder/features/props-panel/controls/url.tsx
@@ -23,9 +23,8 @@ import {
   LinkIcon,
   PageIcon,
   PhoneIcon,
-  SectionLinkIcon,
 } from "@webstudio-is/icons";
-import { type ReactNode, useState } from "react";
+import { type ReactNode, useState, useMemo } from "react";
 import {
   type ControlProps,
   getLabel,
@@ -39,10 +38,7 @@ import {
 } from "@webstudio-is/project-build";
 import { SelectAsset } from "./select-asset";
 
-type UrlControlProps = ControlProps<
-  "url",
-  "string" | "page" | "asset" | "instance"
->;
+type UrlControlProps = ControlProps<"url", "string" | "page" | "asset">;
 
 type BaseControlProps = {
   id: string;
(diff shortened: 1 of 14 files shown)


apps/builder/app/builder/features/props-panel/controls/url.tsx
```diff
@@ -211,9 +222,94 @@ const BaseAttachment = ({ prop, onChange, onSoftDelete }: BaseControlProps) => (
   </Row>
 );
 
+// store that contains IDs off all instances in the selected page
+const pageInstancesStore = computed(
```
kof, line 226: this could be useful globally I guess, no? I have a feeling we must be potentially even doing this already somewhere @TrySound

apps/builder/app/builder/features/props-panel/controls/url.tsx
```diff
@@ -211,9 +222,94 @@ const BaseAttachment = ({ prop, onChange, onSoftDelete }: BaseControlProps) => (
...
+      }
+    }
+
+    return sections;
+  }
+);
+
+// too hard to mock all the stores that sectionsStore is derived from
+// so we have this for Storybook
+export const setMockUrlSectionsStore = (store: typeof sectionsStore) => {
```
kof, line 275: another option is to pass the sections to a component <Components sections={sectionsStore.get()} /> 
So this way when you need to mock it, you pass it directly to the component, I just ended up doing this in BreakpointsSelector

I created a BreakpointsSelectorContainer that holds the stores, so I can have a pure BreakpointsSelector I can render in stories

apps/builder/app/builder/features/props-panel/props-panel.stories.tsx
```diff
@@ -40,6 +42,13 @@ pagesStore.set({
...
 
+const urlSections = new Map([
+  [unique(), "links"],
+  [unique(), "about"],
+]);
+
+setMockUrlSectionsStore(atom(urlSections));
+
 const imageAsset = (name = "cat", format = "jpg"): Asset => ({
   id: unique(),
```
kof, line 53: wouldn't it be better to just use nanoid everywhere instead of unieque?

kof:
I just realized that it's impossible to link using a section on a different page. Same problem at webflow. How did we not see that upfront?
It seems that the right place for this action is under the page tab, where you can either choose a page, or section or both

cc @taylornowotny

kof (approved):

apps/builder/app/builder/features/props-panel/controls/url.tsx
rpominov: Yeah, wasn't sure about this too. @TrySound What do you think?

apps/builder/app/builder/features/props-panel/controls/url.tsx
rpominov: In this case we would have to pass this prop to `<PropsPanel>`, which feels front from the separations of concerns point of view.

kof: yeah, not great, but we should have a general internal standard for mocking stores, so that we don't end up having hundreds of mock functions

kof: maybe @TrySound has an opinion on this

rpominov:
> I just realized that it's impossible to link using a section on a different page

Created an issue: #1475

apps/builder/app/builder/features/props-panel/props-panel.stories.tsx
rpominov: Ah, just realised it will not work, because IDs are visible in lostpixel screenshot and have to be stable

apps/builder/app/builder/features/props-panel/controls/url.tsx
TrySound: See findTreeInstanceIds in project-build

apps/builder/app/builder/features/props-panel/props-panel.stories.tsx
```diff
@@ -40,6 +42,13 @@ pagesStore.set({
   ],
 });
 
+const urlSections = new Map([
+  [unique(), "links"],
+  [unique(), "about"],
+]);
+
+setMockUrlSectionsStore(atom(urlSections));
```
TrySound, line 50: I see no problem in mocking writable atoms. We would have testing element here.

I usually add createInstancePair and createPropPair utilities to make formatting better and not duplicate id

```ts
pagesStore.set({
  homePage: { id: 'home', rootInstanceId: 'root' }
  pages: []
})
instances.set(new Map([
  ['root', {
    id: 'root',
    component: 'Box',
    children: []
  }]
]))
props.set(new Map([
  ['prop1', {
    type: "string",
    name: "id",
    value: "links",
    instanceId: 'root'
  }],
  ['prop3', {
    type: "string",
    name: "id",
    value: "about",
    instanceId: 'root'
  }],
]))
```

TrySound (approved):

rpominov:
@kof @taylornowotny Please test how page / section select works in the video or in preview:

https://user-images.githubusercontent.com/825702/232760359-2628cd49-f725-4425-864e-0d6ad68db03e.mov

taylornowotny:
@rpominov This is perfect - great job! 🔥 I'm impressed that you figured out the ideal UX without those details being provided in the design.

kof:
nice

apps/builder/app/builder/features/props-panel/controls/url.tsx
```diff
@@ -175,29 +186,104 @@ const BaseEmail = ({ prop, onChange, id }: BaseControlProps) => {
   );
 };
 
+const useSections = (page: { rootInstanceId: string }) => {
+  const store = useMemo(
+    () =>
+      computed([instancesStore, propsStore], (instances, props) => {
```
kof, line 192: hmm, not sure why you need computed, seems like you could just call .get() on those stores

apps/builder/app/builder/features/props-panel/use-props-logic.ts
```diff
@@ -113,9 +108,20 @@ const getAndDelete = <Value>(map: Map<string, Value>, key: string) => {
...
- */
+const systemPropsMeta: { name: string; meta: PropMeta }[] = [
+  {
+    name: "id",
+    meta: {
+      required: false,
+      control: "text",
+      type: "string",
+      rows: 0,
+      label: "ID",
```
kof, line 119: Id?

apps/builder/app/builder/features/props-panel/controls/url.tsx
rpominov: I call `useStore()` on the `store` created by `computed`. `computed` inside `useMemo` is a pattern we use everywhere

packages/react-sdk/src/components/link.tsx
```diff
@@ -19,10 +19,14 @@ export const Link = forwardRef<HTMLAnchorElement, Props>((props, ref) => {
 
   const { assetBaseUrl } = getParams();
 
-  let url: string | undefined;
+  let url = "#";
+
   switch (href?.type) {
     case "page":
       url = href.page.path;
+      if (href.idPropValue !== undefined) {
```
kof, line 27: why not href.hash // "#asdfasdf"

like .hash in new URL()

rpominov:
> One point of feedback - the select button that says 'Current Page' is showing placeholder text. Current Page is a pre-selected option from the menu, not placeholder text, so for this we would use the "set value" state rather than the "placeholder" state (or whatever equivalent you have on the code component).

We show it as a placeholder currently, because when it says "Current Page" prop is not yet set to link to the current page.

The text "Current Page" is here only to let user know that the sections list is from the current page if they don't select a specific page.

We could instead actually pre-select current page at the moment user switches to the "Page" tab, but then logic of entire URL control changes: currently switching between tabs doesn't change the prop value. This allows to explore tabs w/o loosing value set in the different tab.

kof (approved):
really don't like the idPropValue name, otherwise great

taylornowotny:
@rpominov Good ideas.

Is this pattern reasonable?

Change it to "Choose page" (placeholder state)
Section menu lists all sections that have an ID on all pages
Choosing a section without choosing a page first will select the page that the section is on

This seems like the ideal UX but if that won't work then we can leave it as is.

rpominov:
> Section menu lists all sections that have an ID on all pages
Choosing a section without choosing a page first will select the page that the section is on

Ah, great idea! Shouldn't be a problem, will do.

PR #1461, opened 2023-04-15
name and color update to the token handling the scroll thumb color

taylornowotny:


rpominov (approved):

kof:
where is a follow up issue for actually using this token?

PR #1469, opened 2023-04-16
Refactor canvas resize to use scrub

kof:
## Description

Now user can keep resizing outside of the window boundary

Also:
- fixes https://github.com/webstudio-is/webstudio-builder/issues/1465

## Steps for reproduction

1. start resizing, keep going outside of window

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/workspace/canvas-tools/resize-handles.tsx
```diff
@@ -78,59 +77,63 @@ const handleIcon = (
...
     }
-  };
+    const scrub = numericScrubControl(ref.current, {
+      getInitialValue() {
+        return canvasWidthStore.get();
+      },
+      getValue(state, movement) {
+        const value =
+          side === "left"
+            ? state.value - movement * 2
```
istarkov, line 104: What is 2? Speed of movement? Probably better make it const with some good name and explain why 2 is selected. Also does this mean that all values are odd or even depending on initial value?

apps/builder/app/builder/features/workspace/canvas-tools/resize-handles.tsx
```diff
@@ -78,59 +77,63 @@ const handleIcon = (
...
+            : state.value + movement * 2;
+        return Math.max(value, minCanvasWidth);
+      },
+      onStatusChange(status) {
+        if (status === "scrubbing") {
+          isCanvasPointerEventsEnabledStore.set(false);
```
istarkov, line 110: mb disable events if isResizingCanvasStore is true instead of adding additional atom?

istarkov:
Sometimes during the scrub all the text etc are selected
image

istarkov (approved):

apps/builder/app/builder/features/workspace/canvas-tools/resize-handles.tsx
kof: will add a comment, yes, it's compensating because we center the canvas, so that cursor stays at the edge

apps/builder/app/builder/features/workspace/canvas-tools/resize-handles.tsx
kof: isCanvasPointerEventsEnabledStore is not only for resizing, but also for drag&drop and potentially more interactions, I could automatically set isCanvasPointerEventsEnabledStore when isResizingCanvasStore is true, but I am not sure this would be better, maybe less obvious even

packages/design-system/src/components/primitives/numeric-gesture-control.stories.tsx
```diff
@@ -20,7 +20,7 @@ const useNumericScrubControl = ({
     }
     ref.current.value = String(value);
     const { disconnectedCallback } = numericScrubControl(ref.current, {
```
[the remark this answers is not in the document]
kof: I agree, originally was called after web components, but it has nothing to do with them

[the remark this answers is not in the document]
kof: will rename now

kof:
> But can't find short example how to reproduce, just after playing with scrub got this

will set globally user-select: none on the body during scrub

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -127,17 +157,15 @@ export const numericScrubControl = (
 
         onStatusChange?.("scrubbing");
         targetNode.addEventListener("pointermove", handleEvent);
+        targetNode.ownerDocument.documentElement.style.userSelect = "none";
```
istarkov, line 160: Not this situation issue as Im 100% that documentElement will never have custom userSelect style,
But styles usually updated, saving previous value/state and restoring it.

istarkov (approved):

PR #1470, opened 2023-04-16
Make the base `tsconfig.json` to assume declaration emit

Andarist:
## Description

This PR "swaps" the guiding principle behind what options are in the base `tsconfig.json`.

Previously TS was configured to *not* emit declarations and we were forcing for them to be emitted with CLI flags. This had an unfortunate result - errors about "not portable type definitions" were not displayed in the IDE because to its best knowledge the type definitions were never supposed to be emitted. So it wasn't worthwhile to surface this error.

With the adjusted options those errors will be raised in the IDE which should improve the feedback loop for those errors.

## Code Review

- [ ] hi @TrySound, @istarkov, @kof - you need to recheck if you have any doubts about this approach, if you see anything wrong with it in the context of this project

## Before merging

- [ ] test IDE setup
- [ ] test build script
- [ ] test typechecking and dts emit


apps/builder/tsconfig.json
```diff
@@ -10,6 +10,7 @@
   "compilerOptions": {
     "paths": { "~/*": ["app/*"] },
     "baseUrl": ".",
-    "rootDir": "./app"
+    "rootDir": "./app",
+    "declaration": false
```
TrySound, line 14: Does this mean we stuck with rootDir option and cannot run checks on configurations in the root?

Andarist: No, I don't think so. I didn't change anything here - I simply want to disable declarations here. An app won't have declarations emitted for it so we don't care if types are portable here or not.

Andarist, line 14:
```suggestion
    "declaration": false,
    "declarationDir": null
```

TrySound (approved):

PR #1481, opened 2023-04-18
Define component states in meta

TrySound:

Here added ability to extend states in style source menu list with a list defined in component meta.

As an example added `[aria-current=page]` support to links. For now can be statically set with props panel.

Note: review with hidden whitespaces

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/components/link.ws.tsx
+++ b/packages/react-sdk/src/components/link.ws.tsx
@@ -26,7 +26,7 @@ export const meta: WsComponentMeta = {
   label: "Link Text",
   Icon: LinkIcon,
   presetStyle,
-  states: ["[aria-current=page]"],
+  states: [{ value: "[aria-current=page]", label: "Current page" }],
   children: ["Link text you can edit"],
 };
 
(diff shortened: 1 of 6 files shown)


packages/react-sdk/src/components/link.ws.tsx
```diff
@@ -26,6 +26,7 @@ export const meta: WsComponentMeta = {
   label: "Link Text",
   Icon: LinkIcon,
   presetStyle,
+  states: ["[aria-current=page]"],
```
kof, line 29: I think we should be able to add a label there that would be visible in UI

TrySound: Default humanizer is not good enough?

packages/react-sdk/src/components/link.ws.tsx
```diff
@@ -26,6 +26,7 @@ export const meta: WsComponentMeta = {
   label: "Link Text",
   Icon: LinkIcon,
   presetStyle,
+  states: [{ value: "[aria-current=page]", label: "Current page" }],
```
kof, line 29: lets rename value to selector?

kof (approved):
looking great, one comment left

PR #1495, opened 2023-04-19
Update css-engine transformValue on every update

TrySound:
We have a bug that asset cannot be applied in background style section. This is because image transformer was set only once on rule creation. We need to reset it with actual assets on every styles updates.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/shared/styles.ts
+++ b/apps/builder/app/canvas/shared/styles.ts
@@ -191,19 +191,16 @@ const getOrCreateRule = ({
   const key = `${instanceId}:${breakpointId}:${state}`;
   let rule = wrappedRulesMap.get(key);
   if (rule === undefined) {
-    const params = getParams();
-    rule = cssEngine.addStyleRule(
-      `[${idAttribute}="${instanceId}"]${state}`,
-      {
-        breakpoint: breakpointId,
-        style: {},
-      },
-      createImageValueTransformer(assets, {
-        assetBaseUrl: params.assetBaseUrl,
-      })
-    );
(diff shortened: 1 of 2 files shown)


istarkov (approved):

packages/css-engine/src/core/rules.ts
```diff
@@ -11,6 +11,9 @@ class StylePropertyMap {
   constructor(transformValue?: TransformValue) {
     this.#transformValue = transformValue;
   }
+  setTransformer(transformValue: TransformValue) {
```
kof, line 14: do we have tests for transformers in css-engine? if not please add a few, its important to keep this core piece well-tested

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -191,19 +191,16 @@ const getOrCreateRule = ({
...
+    rule = cssEngine.addStyleRule(`[${idAttribute}="${instanceId}"]${state}`, {
+      breakpoint: breakpointId,
+      style: {},
+    });
     wrappedRulesMap.set(key, rule);
   }
+  const params = getParams();
+  rule.styleMap.setTransformer(
+    createImageValueTransformer(assets, { assetBaseUrl: params.assetBaseUrl })
+  );
```
kof, line 203: we need to start adding tests to these functions in styles.ts, they are super mission critical and completely untested

kof (approved):

packages/css-engine/src/core/rules.ts
TrySound: Yes, we have

PR #1499, opened 2023-04-20
Remove handleRequest from react-sdk

TrySound:
We should not have any server specific functions in react-sdk. Any small change can break client environment.

And moving to /server entry point still problematic because this entry point is node specific while we want to support as many environments as possible.

Recently remix added defaults for different environments so user do not need to add any entry.client.tsx or entry.server.ts at all.

Defaults do not work for us in builder though doing great in size test setup.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-size-test/package.json
```diff
@@ -20,16 +20,18 @@
...
   "dependencies": {
-    "@remix-run/cloudflare-pages": "1.15.0",
-    "@remix-run/react": "1.15.0",
+    "@remix-run/cloudflare": "^1.15.0",
+    "@remix-run/cloudflare-pages": "^1.15.0",
+    "@remix-run/react": "^1.15.0",
     "@webstudio-is/project": "workspace:^",
     "@webstudio-is/project-build": "workspace:^",
     "@webstudio-is/react-sdk": "workspace:^",
+    "isbot": "^3.6.10",
```
TrySound, line 29: auto installed dependency

kof (commented):
We need to find a way to decouple builder from whatever server/framework is used on the canvas. Adding canvas-specific Remix logic on the canvas seems like a step in the opposite direction.

The plan was to split react-sdk in 2 packages: app and components. So Remix things would be part of app package.

TrySound:
Added comment to remove canvas handler at all once canvas is split into separate app.

kof (approved):

PR #1504, opened 2023-04-20
Fix switching to base breakpoint on a small screen

kof:
when base breakpoint selected, always use a width between max and min breakpoints at the lowest value

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/breakpoints/find-initial-width.test.ts
+++ b/apps/builder/app/builder/features/breakpoints/find-initial-width.test.ts
@@ -13,11 +13,31 @@ const breakpoints = [
 ];
 
 describe("Find initial width", () => {
-  test("base", () => {
+  test("base default", () => {
+    expect(
+      findInitialWidth(breakpoints, breakpoints[0], workspaceWidth)
+    ).toStrictEqual(992);
+  });
+
+  test("base while workspace is small", () => {
+    expect(findInitialWidth(breakpoints, breakpoints[0], 500)).toStrictEqual(
+      992
+    );
+  });
+
+  test("base min-width only", () => {
+    const workspaceWidth = 1000;
+
+    const breakpoints = [
+      { id: "0", label: "Base" },
+      { id: "4", label: "Large", minWidth: 1280 },
+      { id: "5", label: "Extra Large", minWidth: 1440 },
+    ];
     expect(
       findInitialWidth(breakpoints, breakpoints[0], workspaceWidth)
     ).toStrictEqual(workspaceWidth);
   });
+
   test("tablet", () => {
     expect(
       findInitialWidth(breakpoints, breakpoints[1], workspaceWidth)
--- a/apps/builder/app/builder/features/breakpoints/find-initial-width.test.ts
+++ b/apps/builder/app/builder/features/breakpoints/find-initial-width.test.ts
@@ -25,7 +25,7 @@ describe("Find initial width", () => {
     );
   });
 
-  test("base min-width only", () => {
+  test("base and min-width only", () => {
     const workspaceWidth = 1000;
 
     const breakpoints = [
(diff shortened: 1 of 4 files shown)


apps/builder/app/builder/features/breakpoints/find-initial-width.ts
```diff
@@ -11,12 +12,24 @@ export const findInitialWidth = (
...
 ) => {
-  // When on a base breakpoint, we want to show maximum width canvas
-  if (
-    selectedBreakpoint.minWidth === undefined &&
-    selectedBreakpoint.maxWidth === undefined
-  ) {
-    return workspaceWidth;
+  if (isBaseBreakpoint(selectedBreakpoint)) {
+    // Finding the canvas width when user selects base breakpoint is a bit more complicated.
+    // We want to find the lowest possible size that is bigger than all max breakpoints and smaller than all min breakpoints.
```
istarkov, line 17: Wrong comment, the algorithm below is right only if groupedBreakpoints returns non-intersected Set of min max breakpoints that's not true. Example.
image
 
So here it returns just `return Math.max(next.maxWidth + 1, defaultWidth);`
`groupBreakpoints` have no logic that resulting groups must not intersect

apps/builder/app/builder/features/breakpoints/find-initial-width.test.ts
```diff
@@ -13,11 +13,31 @@ const breakpoints = [
 ];
 
 describe("Find initial width", () => {
-  test("base", () => {
+  test("base default", () => {
```
istarkov, line 16: Tests should contain no min|max breakpoints,
no breakpoints except Base at all.
Otherwise most of your logic is not covered with test at all

apps/builder/app/builder/features/breakpoints/find-initial-width.ts
```diff
@@ -11,12 +12,24 @@ export const findInitialWidth = (
   selectedBreakpoint: Breakpoint,
   workspaceWidth: number
 ) => {
-  // When on a base breakpoint, we want to show maximum width canvas
-  if (
-    selectedBreakpoint.minWidth === undefined &&
-    selectedBreakpoint.maxWidth === undefined
-  ) {
-    return workspaceWidth;
+  if (isBaseBreakpoint(selectedBreakpoint)) {
```
istarkov, line 15: If the only breakpoint is the base breakpoint it should return workspaceWidth no returns 320

istarkov (approved):
Plz fix tests and only base breakpoint issue. Comments are not describing edgecase situations

apps/builder/app/builder/features/breakpoints/find-initial-width.ts
kof: yes, we need to decide what to do about intersecting min and max breakpoints, what happens if we have a 300px canvas and there is min and max breakpoint that applies at that width? We would have 2 lines one is on the right, one is on the left.

maybe that's alright, but user shouldn't want to have this in practice because there is no point in defining a breakpoint that has an intersection, since you are trying to address unique resolutions

kof: added a comment regarding intersecting breakpoints for now, not sure we should prevent user from ever having an intersection, because reasons for having an intersection could be complex e.g. copy/paste and user should decide how they define them

apps/builder/app/builder/features/breakpoints/find-initial-width.test.ts
kof: I think we should always have the base breakpoint

kof:
Any more use cases you see uncovered with tests?

istarkov (approved):

PR #1505, opened 2023-04-20
NavLink with current page state

kof:
## Description

This is using NavLink instead of Link by default to allow styling Current Page state so we can use it for navigations.

## Steps for reproduction

Can't be tested because works only on a published site and we need to fix publishing on staging and locally

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

kof:
I guess we have to deploy to prod and see if this works

PR #1507, opened 2023-04-21
Enforce server constraints for packages

TrySound:
Ref https://remix.run/docs/en/1.15.0/guides/constraints#server-code-pruning

Remix use NAME.server.ts extension to enforce omitting module from client bundles. Though we heavily used `/server` in packages which had no affect and blocked us in using a few node apis which were not polifilled with empty modules while bundling for client.

Here I replaced `/server` with `/index.server` and removed all legacy server entry points which are not longer necessary with new typescript resolution.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -3,7 +3,7 @@ import { type Publish, usePublish } from "~/shared/pubsub";
 import type { Build } from "@webstudio-is/project-build";
 import type { Project } from "@webstudio-is/project";
 import { theme, Box, type CSS, Flex, Grid } from "@webstudio-is/design-system";
-import type { AuthPermit } from "@webstudio-is/trpc-interface/server";
+import type { AuthPermit } from "@webstudio-is/trpc-interface/index.server";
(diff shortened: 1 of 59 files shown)


istarkov:
How to check? Import into any client modele will cause errors?

istarkov (approved):

TrySound:
createWriteStream is not be easily "polyfilled" so cannot be bundled in client code

Andarist:
Hm, as per the comment in Remix's code this convention should only be applied to the `app` directory:
https://github.com/remix-run/remix/blob/cb7416beb87f2f008f05bafb0c25d263bf59851f/packages/remix-dev/compiler/plugins/emptyModulesPlugin.ts#L18-L26

Since you use it outside of `app` too (in your packages in the monorepo), I wonder if this won't break for you at some time in the future. In other words, I wouldn't exactly rely on this being applied to packages

TrySound:
Honestly no idea. This whole remix is super flaky and uncertain. And the most importantly blocks me from using some node api in packages. We can resolve this if any issue appear in the future.

PR #1508, opened 2023-04-21
Fix data not synchronized if you reopen the projects from the dashboard

istarkov:
## Description

Fix data not synchronized if you reopen the projects from the dashboard

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/sync/sync-stores.ts
+++ b/apps/builder/app/shared/sync/sync-stores.ts
@@ -303,4 +303,10 @@ export const useBuilderStore = (publish: Publish) => {
       };
     });
   }, [publish]);
+
+  useEffect(() => {
+    return () => {
+      handshakenStore.set(false);
+    };
+  }, []);
 };


apps/builder/app/shared/sync/sync-stores.ts
```diff
@@ -303,4 +303,10 @@ export const useBuilderStore = (publish: Publish) => {
       };
     });
   }, [publish]);
+
+  useEffect(() => {
+    return () => {
+      handshakenStore.set(false);
```
TrySound, line 309: I thinking maybe reset handshake in the beginning instead?

istarkov: will be dependent on order of useEffects

istarkov: Also some atoms initiated before, and here can be all project dependent cleanup if we will not find any context solution

TrySound (approved):

PR #1515, opened 2023-04-22
Use s3 rest api instead of library

TrySound:


kof:
Does it work with CF R2? We aren't actually using S3 from aws

TrySound:
Yep, r2 gives compatible endpoints. This is why js clients worked. Though they kinda add a lot of code to do same fetch but with super puper nice api based on commands.

kof (commented):
1. I don't know if the client was handling some kind of edge cases you don't handle any more
2. I don't know which additional apis or options we will need in the future

Client was conceptually a safer choice that would safe us from nasty bugs and potentially help us migrate to newer versions, so I am against shaving a few kb and in return creating potential issues now or in the future.

Client is something they have to maintain, now this thing is what WE need to maintain

TrySound:
While client can actually change api and input data format, rest api is set in stone because become so popular and was adopted by many platforms and tools. We have minio, r2, rclone and other clients. All use same api. And if anything really changes on r2 side it probably won't be reflected by aws client.

The most complex thing here is authorization with signature v4. This is what I don't even try to solve here. In the future it will be simpler and accept less data (saw in code comments).

So what do we have in those 83 packages? Whole aws api we may never use.

This amount of packages doing nothing abuses node_modules resolution algorithm and makes cold start longer.

kof:
Screenshot 2023-05-13 at 12 06 26

not great

TrySound:
@kof Wrong PR?

istarkov (approved):

PR #1516, opened 2023-04-22
Radix Tabs

kof:
## Description

1. This is the first attempt to use advanced/complex components on cavnas

## Todo

- [x] Tree Template - a data structure we can place into `children` in meta that we can use to create real data structure and render. Default template should have by default 2 tabs and 2 contents that work out of the box.
- [x] Remove category from Trigger, Content, Tabs List metas, so they don't show up standalone in components list, but they still can be copy/pasted from tab. 
- [x] Recover from misplaced component error or any error on canvas after dragging component in the right position or removing it
      Screenshot 2023-04-22 at 14 10 00
- [ ]  generate-arg-types doesn't seem to work with interfaces (not blocking right now, since we can retype the types, but will be blocking for the next stage with user components) 

## Steps for reproduction

1. click button
3. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/package.json
+++ b/packages/react-sdk/package.json
@@ -43,6 +43,7 @@
   "dependencies": {
     "@nanostores/react": "^0.4.1",
     "@react-aria/utils": "^3.13.3",
+    "@radix-ui/react-tabs": "^1.0.0",
     "@webstudio-is/asset-uploader": "workspace:^",
     "@webstudio-is/css-data": "workspace:^",
     "@webstudio-is/css-engine": "workspace:^",
--- a/packages/react-sdk/src/components/index.ts
+++ b/packages/react-sdk/src/components/index.ts
@@ -33,6 +33,10 @@ import { meta as RadioButtonFieldMeta } from "./radio-button-field.ws";
 import { meta as RadioButtonMeta } from "./radio-button.ws";
 import { meta as CheckboxFieldMeta } from "./checkbox-field.ws";
 import { meta as CheckboxMeta } from "./checkbox.ws";
+import { meta as TabsRootMeta } from "./tabs.ws";
+import { meta as TabsList } from "./tabs-list.ws";
+import { meta as TabsTrigger } from "./tabs-trigger.ws";
+import { meta as TabsContent } from "./tabs-content.ws";
 
 // these are huge JSON objects that we want to be tree-shaken when not used!
 import { propsMeta as SlotMetaPropsMeta } from "./slot.ws";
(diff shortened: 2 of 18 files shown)


packages/react-sdk/package.json
```diff
@@ -42,6 +42,7 @@
   },
   "dependencies": {
     "@nanostores/react": "^0.4.1",
+    "@radix-ui/react-tabs": "^1.0.0",
```
TrySound, line 45: Maybe we should have something like radix-sdk for such stuff?

kof: Yes, I thought about this too, even more so I would basically want to let user search and add all available component packages aka marketplace ... just not sure when to start with this, since we need to think the UX through for how this will work, including versioning.

Additionally we need to change how we generate props jsons from types to support external webstudio components, they should come directly prebuilt with everything, ready to get installed

packages/react-sdk/src/components/index.ts
```diff
@@ -157,7 +169,11 @@ export const defaultPropsMetas: Record<string, WsComponentPropsMeta> = {
   RadioButton: RadioButtonPropsMeta,
   CheckboxField: CheckboxFieldPropsMeta,
   Checkbox: CheckboxPropsMeta,
-};
+  Tabs: TabsRootPropsMeta,
```
TrySound, line 172: Prefix with radix? Like RadixTabs

packages/react-sdk/src/components/index.ts
```diff
@@ -68,6 +72,10 @@ import { propsMeta as RadioButtonFieldPropsMeta } from "./radio-button-field.ws"
 import { propsMeta as RadioButtonPropsMeta } from "./radio-button.ws";
 import { propsMeta as CheckboxFieldPropsMeta } from "./checkbox-field.ws";
 import { propsMeta as CheckboxPropsMeta } from "./checkbox.ws";
+import { propsMeta as TabsRootPropsMeta } from "./tabs.ws";
```
TrySound, line 75: Maybe avoid Root in this naming? We don't use it in filenames

packages/react-sdk/src/components/index.ts
kof: Was thinking about it as well, especially if we go with "Radix UI" as a category, need to discuss this with Taylor. Ideally this would be part of a UI that lets user install those components but we in the end still need to decide how we represent the section of installed components or even packages. So the closest thing I can think of right now is to just put all Radix UI components into a section/category

TrySound: Label can still be without prefix. I mean would be good to prefix the identifier so we could avoid fatal conflicts if someone adds another library with same component name.

kof:
https://github.com/webstudio-is/webstudio-builder/pull/2022

PR #1517, opened 2023-04-22
Allow to specify tree template in component meta

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/pull/1516

Here added new utility "tree template" to describe children in component meta and generate instances from it while inserting.

Added simple template to form component as example.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/components/form.ws.tsx
+++ b/packages/react-sdk/src/components/form.ws.tsx
@@ -22,6 +22,14 @@ export const meta: WsComponentMeta = {
   label: "Form",
   Icon: FormIcon,
   presetStyle,
+  children: [
+    { type: "instance", component: "Input", children: [] },
+    {
+      type: "instance",
+      component: "Button",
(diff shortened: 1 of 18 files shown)


packages/react-sdk/src/components/form.ws.tsx
```diff
@@ -22,6 +22,14 @@ export const meta: WsComponentMeta = {
   label: "Form",
   Icon: FormIcon,
   presetStyle,
+  children: [
```
kof, line 25: Thiis was a test, right?

kof (commented):
we need it to be able to have props and styles, so we can fully assemble complex component

kof (approved):

packages/react-sdk/src/components/form.ws.tsx
TrySound: Is there practical usage in forms now?

PR #1521, opened 2023-04-24
Custom domains

andrasbacsai:
## Description

- Added custom domains to the prisma schema
- Added basic UI for testing (should not use in production)
- Related PR: https://github.com/webstudio-is/webstudio-saas/pull/144

### DIFF
--- a/apps/builder/app/builder/features/topbar/publish.tsx
+++ b/apps/builder/app/builder/features/topbar/publish.tsx
@@ -15,6 +15,7 @@ import {
   FloatingPanelPopoverTrigger,
   FloatingPanelPopoverContent,
   FloatingPanelPopoverTitle,
+  ScrollArea,
 } from "@webstudio-is/design-system";
 import { useIsPublishDialogOpen } from "../../shared/nano-states";
 import type { Project } from "@webstudio-is/project";
@@ -110,7 +111,9 @@ export const PublishButton = ({ project }: PublishButtonProps) => {
       </FloatingPanelAnchor>
 
       <FloatingPanelPopoverContent css={{ zIndex: theme.zIndices[1] }}>
-        <Content project={project} />
+        <ScrollArea css={{ display: "grid" }}>
+          <Content project={project} />
+        </ScrollArea>
         <FloatingPanelPopoverTitle>Publish</FloatingPanelPopoverTitle>
       </FloatingPanelPopoverContent>
     </FloatingPanelPopover>
--- a/apps/builder/app/builder/features/topbar/publish.tsx
+++ b/apps/builder/app/builder/features/topbar/publish.tsx
@@ -21,6 +21,8 @@ import type { Project } from "@webstudio-is/project";
 import { getPublishedUrl, restPublishPath } from "~/shared/router-utils";
 import { theme } from "@webstudio-is/design-system";
 import { useAuthPermit } from "~/shared/nano-states";
+import { isFeatureEnabled } from "@webstudio-is/feature-flags";
+import { Domains } from "./domains";
 type PublishButtonProps = { project: Project };
 
 const Content = ({ project }: PublishButtonProps) => {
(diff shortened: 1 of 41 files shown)


istarkov:
plz rebase and fix errors

istarkov:
We are trying to not use rest as above, we are using trpc 

Examples are: 

https://github.com/webstudio-is/webstudio-builder/blob/e1fa2075a535d11a42bc8b99ef36a06d499ab6a7/apps/builder/app/routes/rest.authorization-token.$method.ts#L15

https://github.com/webstudio-is/webstudio-builder/blob/e1fa2075a535d11a42bc8b99ef36a06d499ab6a7/apps/builder/app/routes/dashboard.projects.$method.ts#L15

And then we use endpoints from the UI

https://github.com/webstudio-is/webstudio-builder/blob/9c316b04992a56dd3e8c15d1e1f90a702fd07d18/apps/builder/app/shared/share-project/share-project-container.tsx#L8-L9

apps/builder/app/builder/features/topbar/publish.tsx
```diff
@@ -18,16 +18,38 @@ import {
...
+  restAddCustomDomainPath,
+  restRemoveCustomDomainPath,
+  restPublishPath,
+} from "~/shared/router-utils";
 import { theme } from "@webstudio-is/design-system";
 import { useAuthPermit } from "~/shared/nano-states";
 type PublishButtonProps = { project: Project };
 
+async function removeDomain(domain: string, userId: string) {
+  return await fetch(restRemoveCustomDomainPath(), {
```
istarkov, line 32: See comment about trpc

apps/builder/app/routes/rest.domain.ts
```diff
@@ -0,0 +1,78 @@
...
+      errors: "Domain already exists! Contact us for support.",
+    };
+  }
+
+  const foundCname = await trpcClient.customDomain.verifyCNAME.mutate({
+    customCname,
+    customDomain,
+  });
+
+  if (!foundCname.success) {
```
istarkov, line 52: Just error is enough in a type to define is it error or not. No need in `success`

apps/builder/app/routes/rest.domain.ts
```diff
@@ -0,0 +1,78 @@
...
+      userId,
+    },
+  });
+}
+export const action = async ({ request }: ActionArgs) => {
+  const { customDomain, customCname, projectId, userId } = schema.parse(
+    await request.formData()
+  );
+
+  const customDomainFound = await prisma.customDomain.findFirst({
```
istarkov, line 35: It should be not a problem if a custom domain is found in our db.  Few users can enter the same domain, final routing depends on cname.

apps/builder/app/routes/rest.domain.ts
```diff
@@ -0,0 +1,78 @@
...
+      projectId,
+      userId,
+    });
+    return {
+      success: true,
+    };
+  }
+  // Should handle this error better
+  // CF duplicate domain error or other unknown error
+  return {
```
istarkov, line 74: Should be not an issue if it is already created. From my understanding, all custom domain has the same fallback origin value. So it can already be created by the same user or other user

apps/builder/app/routes/rest.domain.ts
istarkov, line 35: Before doing anything we should check that user has owner permission on the project

packages/prisma-client/prisma/schema.prisma
```diff
@@ -124,3 +128,16 @@ view DashboardProject {
...
 }
+
+model CustomDomain {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+
+  domain String @unique
+
+  user   User?   @relation(fields: [userId], references: [id])
+  userId String?
```
istarkov, line 139: No sure we need userId reference as it is avalibale through the Project

packages/project/src/shared/schema.ts
```diff
@@ -8,14 +8,19 @@ export const Title = z
     (val) => val.length >= MIN_TITLE_LENGTH,
     `Minimum ${MIN_TITLE_LENGTH} characters required`
   );
-
+export const CustomDomains = z.object({
+  domain: z.string(),
+  projectId: z.string(),
+  userId: z.string(),
```
istarkov, line 14: types are inconsistent with db types. In db they are optional here are not

packages/prisma-client/prisma/schema.prisma
```diff
@@ -124,3 +128,16 @@ view DashboardProject {
...
+
+  Project   Project? @relation(fields: [projectId], references: [id])
+  projectId String?
```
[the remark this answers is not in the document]
istarkov: The only reason is clean up, right? Delete all custom domains without references from cloudflare. But this can be done 
Select all custom domains - find all not in our db - delete

apps/builder/app/routes/rest.domain.remove.ts
```diff
@@ -0,0 +1,32 @@
...
+    where: {
+      userId,
+      domain,
+    },
+  });
+  if (domainFound && domainFound.id) {
+    await trpcClient.customDomain.removeCloudflareCustomHostname.mutate({
+      domain,
+    });
+    await prisma.customDomain.delete({ where: { id: domainFound.id } });
```
istarkov, line 24: Must be oss implementation at least with throw new Error, but with same types as SaaS

PR #1523, opened 2023-04-24
Run dts as additional checks in actions

TrySound:
This fixes the bug introduced in my previous PRs and reproducable only by running dts generation. Here added it to checks so we could avoid broken publishes while releasing.

## Code Review

- [ ] hi @Andarist, I need you to do

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,4 +25,4 @@ jobs:
 
       - run: pnpm prettier --cache --check "**/*.{js,md,ts,tsx}"
 
-      - run: pnpm turbo run size-test build checks
+      - run: pnpm turbo run size-test build checks dts
--- a/packages/react-sdk/src/embed-template.ts
+++ b/packages/react-sdk/src/embed-template.ts
@@ -47,7 +47,7 @@ export const EmbedTemplateStyleDecl = z.object({
 
 export type EmbedTemplateStyleDecl = z.infer<typeof EmbedTemplateStyleDecl>;
 
-type EmbedTemplateInstance = {
+export type EmbedTemplateInstance = {
   type: "instance";
   component: string;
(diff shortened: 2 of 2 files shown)


packages/react-sdk/src/embed-template.ts
```diff
@@ -38,22 +38,25 @@ const EmbedTemplateProp = z.union([
...
+      type: z.literal("instance"),
+      component: z.string(),
+      props: z.optional(z.array(EmbedTemplateProp)),
+      children: WsEmbedTemplate,
+    })
+);
 
-export const WsEmbedTemplate = z.array(
-  z.union([z.lazy(() => EmbedTemplateInstance), EmbedTemplateText])
+export const WsEmbedTemplate = z.lazy(() =>
```
TrySound, line 58: Wrapping whole type with lazy seems like generate smaller type in .d.ts

.github/workflows/main.yml
```diff
@@ -25,4 +25,4 @@ jobs:
 
       - run: pnpm prettier --cache --check "**/*.{js,md,ts,tsx}"
 
-      - run: pnpm turbo run size-test build checks
+      - run: pnpm turbo run size-test build checks dts
```
Andarist, line 28: small caveat: `checks` include `typecheck`, so essentially now you might be type-checking twice here. Once through `checks` and once through `dts`. I **think** that there is no way to *just* emit declarations without typechecking using `tsc` but I could be wrong here. I think that some other tools might be able to do this since they use TS programmatic API and IIRC there is away to avoid requesting semantic diagnostics there.

TrySound: Sure, I would not fix it now. But some time later.

Andarist (approved):

PR #1524, opened 2023-04-24
Units menu in css value input

### LINKED ISSUE #1464: Units menu in css value input
Issues we know of with the current menu:

- user can't discover some basic keywords until they remove input or start typing words
- we want to allow using any unit css supports without blowing up the menu length
- we want to make the choice for units in the menu useful for 99% of use cases, but still short
- we want to support unitless values only when they are actually unitless in CSS (right now margin/padding allow unitless, but they are not)

We have been discussing it for a while, here is the version we came up with:

- [ ] rename "UnitMenu" to just "Menu" because we are going to bring in some keywords
- [x] put in the menu the most popular units/keywords to keep it short and usable in this order:
    "-",
    "px",
    "%",
    "em",
    "rem",
    "dvw",
    "dvh",
    "auto",
    "normal",
    "none"
- [x] add "auto" , "normal" , "none" to properties that support these keywords into the menu
- [x] use "-" when value is unit-less (detect unitless value from mdn-data)
- [x] use "-" when keyword value is selected
- [x] allow any unit to be typed even if its not in a menu, when its typed and user hits enter - we show it as a unit, just not part of the menu 
- [x] adjust margin for keywords menu to 2px from the bottom edge of the input

istarkov:
## Description

closes #1464

- [ ] rename "UnitMenu" to just "Menu" because we are going to bring in some keywords
- [ ] put in the menu the most popular units/keywords to keep it short and usable in this order:
    "-",
    "px",
    "%",
    "em",
    "rem",
    "dvw",
    "dvh",
    "auto",
    "normal",
    "none"
- [ ] add "auto" , "normal" , "none" to properties that support these keywords into the menu
- [ ] use "-" when value is unit-less (detect unitless value from mdn-data)
- [ ] use "-" when keyword value is selected
- [ ] allow any unit to be typed even if its not in a menu, when its typed and user hits enter - we show it as a unit, just not part of the menu 
- [ ] adjust margin for keywords menu to 2px from the bottom edge of the input 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/css-value-input/unit-select.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/css-value-input/unit-select.tsx
@@ -1,5 +1,5 @@
 import { useState, useMemo } from "react";
-import type { Unit } from "@webstudio-is/css-data";
+import { keywordValues, type Unit } from "@webstudio-is/css-data";
 import { properties, units } from "@webstudio-is/css-data";
 import * as SelectPrimitive from "@radix-ui/react-select";
 import {
@@ -12,18 +12,34 @@ import {
   nestedSelectButtonUnitless,
 } from "@webstudio-is/design-system";
 import { ChevronDownIcon, ChevronUpIcon } from "@webstudio-is/icons";
+import type { CssValueInputValue } from "./css-value-input";
+import { isValid } from "../parse-css-value";
+import { toPascalCase } from "../keyword-utils";
 
 type UnitOption = {
-  id: Unit;
+  id: string;
   label: string;
 };
 
-const visibleLengthUnits = ["px", "em", "rem", "ch", "vw", "vh"] as const;
+// To make sorting stable
+const unitPreferedSorting = [
+  ...units.length,
+  ...units.percentage,
+  "number",
+  ...units.angle,
+  ...units.decibel,
+  ...units.flex,
+  ...units.frequency,
+  ...units.resolution,
+  ...units.semitones,
+  ...units.time,
+];
+
+const visibleLengthUnits = ["px", "em", "rem", "dvw", "dvh"] as const;
 
 type UseUnitSelectType = {
   property: string;
-  value?: Unit;
-  showUnitless: boolean;
+  value: CssValueInputValue;
   onChange: (value: Unit) => void;
   onCloseAutoFocus: (event: Event) => void;
 };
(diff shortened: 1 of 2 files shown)


kof:
- [ ] design was updated with added separator between units and kewords like auto

Screenshot 2023-04-25 at 22 55 42


https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=881-39217&t=xB0otzl0K9V87Btw-4

apps/builder/app/builder/features/style-panel/shared/css-value-input/unit-select.tsx
```diff
@@ -12,71 +16,187 @@ import {
...
+          type: "unit",
+          label: unit.toLocaleUpperCase(),
+        });
       }
     }
 
-    if (showUnitless && !options.some((o) => o.id === "number")) {
-      options.push({ id: "number", label: nestedSelectButtonUnitless });
+    // Special case for 0, which is often used as a unitless value
+    const showUnitless =
```
kof, line 100: seems like majority of the logic in this hook could be a normal function which we could then test quite easily, wdyt?

kof: a little worried about all the use cases this hook handles inside of all those if statements, not sure what's tested here

kof: could be buildOptions()

apps/builder/app/builder/features/style-panel/shared/parse-css-value.ts
```diff
@@ -114,10 +114,18 @@ export const parseCssValue = (
       const values = keywordValues[
         property as keyof typeof keywordValues
       ] as ReadonlyArray<string>;
-      if (values?.includes(input)) {
+      const regExOnlyLetters = /[^a-zA-Z]+/g;
```
kof, line 117: 1. lettersRegex
2. why not `\w`

kof (approved):
works as expected from my side, we can let Taylor test later when he is back, but I am good with the results 👍

- [x] There are minor naming nits.
- [x] minor design fix with the separator line
- [x] concern regarding the complexity of that ad-hoc logic inside the hook that is probably testable in a standalone function

PR #1525, opened 2023-04-24
Forms

kof:
https://github.com/webstudio-is/webstudio-builder/issues/84#issuecomment-1522834976

## Description

1. Added new components:
   - textarea
   - radio-button and radio-button-field
   - checkbox and checkbox-field
   - success-message
   - error-message
 2. Added `states` like `:disabled`, `:valid`, etc. to `input`
 3. Added `custom-components/form` that renders success-message and error-message conditionally
 4. Added `success` and `error` style states to `custom-components/form`
 5. Style panel now shows "Select an instance on the canvas" if the selected component is not rendered on canvas (e.g. error-message may not be rendered)
 
Not in this PR:

- The mailer package
- Backend handler for forms (will be in a PR in SAAS)
- Logic in Form that actually submits to backend and shows success/error messages

## Steps for reproduction

1. Add a "Form" to canvas
2. Add "Checkbox Field", "Radio Button Field", "Text Area" to the form
3. Try changing `initialState` property of the form

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
- [ ] Should we put some margin between checkbox and text?
- [ ] Label can't contain div (text)

rpominov:
> Label can't contain div (text)

Do you mean this is not allowed by the standard like `<p><div/></p>` or what is the problem?

kof:
@rpominov the radio and checkbox on canvas are rendered as label, `<label><div></label>`

kof:
Screenshot 2023-05-01 at 22 53 32

I just realized what @rpominov was talking about: there is Checkbox label in add component panel and there  is navigator, which includes the container and checkbox itself

It seems like what we are adding is not a checkbox, it's Checkbox Field or Checkbox Container because checkbox itself is inside, this would be semantically more correct @taylornowotny but webflow actually calls the thing in add component panel Checkbox but then in navigator it calls the same thing a "Checkbox"

I think we kinda need to think again about this, because we can't have atm a different label in add component panel than in the navigator and if we call container "Checkbox", then we need to call the <input type="checkbox"> something else.

I see 2 options:

- Checkbox Field > Checkbox
- Checkbox > Checkbox Input

taylornowotny:
@kof 

We have a "Input Label" component in the design for this, but we need a solution for the avg user who will not know to manually assign the Label to the correct Input.
Screenshot 2023-05-01 at 2 08 52 PM

Each input should have a label, right?

Suggestion:
Adding any input to the canvas will also add an Input Label component attached to that input. eg. User adds a Text Field to the canvas and they get an Input Label to go along with it.

Two questions:
1. Do we name the Label after the Input to help the user stay organized?
ie. Add a Radio, get an Input Label named "Radio Label"
2. Do we group the Label with the Input inside a div like Webflow does? I don't see why they have done this with some of the input types (radio) but not others.

kof:
@taylornowotny might be an overkill for text fields/text area, but yeah agree for radio and checkbox

Oh I think I understand now how we can have different labels depending on the add components panel or tree: add components is also actual label from meta, but inside the tree, it can be the Instance Name that we set upfront.

So we already have everything we need to make them distinct. In Add component panel I would use short names, in navigator more precise names

kof:
> Do we group the Label with the Input inside a div like Webflow does? I don't see why they have done this with some of the input types (radio) but not others.

I guess an input doesn't necessary need a label, can have a placeholder or could be only thing at all, but radio/checkbox always needs a label

rpominov:
> Should we put some space between checkbox/radio and text?

Will add a padding to preset styles of "Chekbox"

> Label can't contain div (text)

Yes, according to this https://stackoverflow.com/a/18609649/478603 `div` inside `label` is not allowed by spec. 

Doesn't seem like it causes any issues in practice, but just to be safe I'll fix this by adding a `tag` property to "Text Block".

> move "Button" component to "General" section

Ok

> need a way to order the components in the sections - "Form" component should be the first item in the Forms section. Order should be same as the design.

Created a separate issue #1550 

> Change name of "Input" to "Text Field" to match design

Ok

> Add Forms components from the design like "Text Area" and "Radio" and "Checkbox"

Already added in this PR

> Oh I think I understand now how we can have different labels depending on the add components panel or tree: add components is also actual label from meta, but inside the tree, it can be the Instance Name that we set upfront.
> So we already have everything we need to make them distinct. In Add component panel I would use short names, in navigator more precise names

We don't quite have everything we need. We would need something like `defaultInstanceName` in meta. Will add it, and have "Checkbox" text in the panel and "Checkbox Field" in tree.

rpominov (approved):

taylornowotny:
There are a few issues:

1. "Field Label" and "Checkbox" icons are not consistent with the design.
Screenshot 2023-05-02 at 1 42 17 PM

2. We can do a lot better on the inputs naming. Let's eliminate the word "field":

current: Radio Button Field -> Radio Button -> Text Block
better: Radio -> Radio Input -> Radio Label

current: Checkbox Field -> Checkbox -> Text Block
better: Checkbox -> Checkbox Input -> Checkbox Label

current: Text Field
better: Text Input

3. Adding a Text Input or Text Area should add an Input Label automatically. If user doesn't want it they can delete it, but accessibility best practices dictate that all inputs and fields should have labels so we want to make that easy for the users.

rpominov:
> "Field Label" and "Checkbox" icons are not consistent with the design.

Seems like already fixed in https://github.com/webstudio-is/webstudio-builder/pull/1558

> We can do a lot better on the inputs naming. Let's eliminate the word "field":

Will fix

> Adding a Text Input or Text Area should add an Input Label automatically. If user doesn't want it they can delete it, but accessibility best practices dictate that all inputs and fields should have labels so we want to make that easy for the users.

This is not trivial because currently we can only insert extra instances as children of the main component that's being added. So if we want to add label along with "Text Input" we would have to do similar to Checkbox/Radio: separate Text Input into two components a wrapper and the input itself etc.

taylornowotny:
> > Adding a Text Input or Text Area should add an Input Label automatically. If user doesn't want it they can delete it, but accessibility best practices dictate that all inputs and fields should have labels so we want to make that easy for the users.
> 
> This is not trivial because currently we can only insert extra instances as children of the main component that's being added. So if we want to add label along with "Text Input" we would have to do similar to Checkbox/Radio: separate Text Input into two components a wrapper and the input itself etc.

Ok, it's fine if new Text Input/Text Areas don't come with a label since the Form component comes with two Text Inputs and two Labels anyways.

Question - how are the Labels connected to the Inputs? Is that the "For" property? If so, how does that work?

Once I understand how this works I will try to make it clearer to the user, through tooltips or something.

rpominov:
Yeah, labels are connected via the "For" property that has the same value as the input's "ID".

Example:

```html
<label for="username">Username</label>
<input id="username" />
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attributes

kof:
Webflow automatically sets the id and for attribute when you drop a form on the canvas, but doesn't actually show the for attribute on the label element in the settings, so you as a user never learn to do that.

When you add a label yourself, you as a user obviously need to set it manually. When it comes to input type text, this acts mainly as screen readers accessibility feature because label explain what the input is for. When it is used with checkbox/radio - then its also accessibility for any pointer device, because this makes it easier to click the radio by clicking label.

In case of radio-checkbox we are wrapping the checkbox/radio with a label and this solves the pointer accessibility without for attribute.

Ideally when we add form component on the canvas, we have for/id for input text just like webflow + the for and id is a property that is visible. I think we already have the EmbedTemplate ability to do that, its just a matter of setting those.

rpominov:
> Ideally when we add form component on the canvas, we have for/id for input text just like webflow + the for and id is a property that is visible. I think we already have the EmbedTemplate ability to do that, its just a matter of setting those.

EmbedTemplate only works for children of the component that user adds. We can't add siblings using it. And some more problems: 
 - would need ability to generate dynamic unique IDs, which EmbedTemplate doesn't allow to do
 - there's to predefine props on the instance that's being added, only on the child instances defined in EmbedTemplate

kof:
@rpominov I agree, while adding a props definition to the meta is not hard, I am not sure how we could have unique id's generated out of the template, lets talk to @TrySound and create an issue for both, can be discussed but postponed for after beta launch

PR #1527, opened 2023-04-25
GPT integration

gsppe:
This branch implements an early and WIP integration of GPT into Webstudio: it mainly implements a rudimentary end-to-end integration so that I can do prompt engineering work and test results visually.

Eventually this feature will be decoupled from Webstudio builder and it'll be a separate plugin/package.

Tracking progress here https://github.com/webstudio-is/webstudio-builder/issues/1450#issuecomment-1523612309


apps/builder/app/builder/features/sidebar-left/panels/ai/ai.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { CloseButton, Header } from "../../header";
+import type { TabName } from "../../types";
+
+type TabContentProps = {
+  onSetActiveTab: (tabName: TabName) => void;
+  publish: Publish;
+};
+
+export const TabContent = ({ publish, onSetActiveTab }: TabContentProps) => {
+  const [isGenerating, setIsGenerating] = useState(false);
```
kof, line 21: you will want to have an enum with more than 2 states I bet

apps/builder/app/env/env.server.ts
```diff
@@ -46,6 +46,10 @@ const env = {
   S3_BUCKET: process.env.S3_BUCKET,
   S3_ACL: process.env.S3_ACL,
   ASSET_CDN_URL: process.env.ASSET_CDN_URL,
+
+  // AI Features
+  OPENAI_KEY: process.env.OPENAI_KEY,
```
kof, line 51: when env is for plugins, we need to think of a way to be clear about that, maybe prefixing with plugin name, maybe passing it as a json string with plugin name as an env variable name?

PR #1531, opened 2023-04-26
Migrate preset styles to array of declarations

TrySound:
Before preset styles used `{ tag: { property: value } }` format. Now we are switching to `{ tag: [{ state,property,value }] }` to support specifying states in preset styles.

Same format will be used for embed template styles.

Here switched only format. States in preset styles not supported in builder yet.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
looking great

PR #1533, opened 2023-04-26
Show initial and html style in style panel

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1067

We need to move away from getComputedStyle for not local styles in style panel.

Here added default html style from browser sources. Now we can render it and properties initial styles instead of computed styles.

Also fixed in properties data generation currentcolor -> currentColor and removed same from border presets.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/style-info.ts
+++ b/apps/builder/app/builder/features/style-panel/shared/style-info.ts
@@ -1,6 +1,11 @@
 import { useMemo } from "react";
 import { useStore } from "@nanostores/react";
-import type { Style, StyleProperty, StyleValue } from "@webstudio-is/css-data";
+import {
+  html,
+  type Style,
+  type StyleProperty,
+  type StyleValue,
+} from "@webstudio-is/css-data";
 import { properties } from "@webstudio-is/css-data";
 import {
   StyleSourceSelections,
@@ -428,11 +433,34 @@ export const useStyleInfo = () => {
     return getPresetStyleRule(component, tagName);
   }, [instances, selectedInstanceSelector, selectedInstanceIntanceToTag]);
 
+  const htmlStyle = useMemo(() => {
+    const instanceId = selectedInstanceSelector?.[0];
+    if (instanceId === undefined) {
+      return;
+    }
+    const tagName = selectedInstanceIntanceToTag?.get(instanceId);
+    if (tagName === undefined) {
+      return;
+    }
+    const styles = html[tagName];
+    if (styles === undefined) {
+      return;
+    }
+    const style: Style = {};
+    for (const styleDecl of styles) {
+      style[styleDecl.property] = styleDecl.value;
+    }
+    return style;
+  }, [selectedInstanceSelector, selectedInstanceIntanceToTag]);
+
   const styleInfoData = useMemo(() => {
     const styleInfoData: StyleInfo = {};
     for (const property of styleProperties) {
       // temporary solution until we start computing all styles from data
       const computed = browserStyle?.[property];
+      const defaultValue =
+        htmlStyle?.[property] ??
+        properties[property as keyof typeof properties].initial;
       const preset = presetStyle?.[property];
       const inherited = inheritedInfo[property];
       const cascaded = cascadedInfo[property];
@@ -444,7 +472,7 @@ export const useStyleInfo = () => {
         cascaded?.value ??
         inherited?.value ??
         preset ??
-        computed;
+        defaultValue;
       if (value) {
         if (property === "color") {
           styleInfoData[property] = {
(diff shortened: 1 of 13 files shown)


kof:
Screenshot 2023-04-26 at 20 38 12

something is off with typography, font-weight and color are not supposed to be "thin" and currentColor preview box doesn't render it as black, even though it is showing as black

TrySound:
This is not currentColor. It's inherit. I thought we already support it.

TrySound:
As for bold out font weight control cannot interpret "normal" and "bold". Always expect a number.

apps/builder/app/builder/features/style-panel/shared/style-info.ts
```diff
@@ -428,11 +433,34 @@ export const useStyleInfo = () => {
     return getPresetStyleRule(component, tagName);
   }, [instances, selectedInstanceSelector, selectedInstanceIntanceToTag]);
 
+  const htmlStyle = useMemo(() => {
+    const instanceId = selectedInstanceSelector?.[0];
+    if (instanceId === undefined) {
+      return;
+    }
+    const tagName = selectedInstanceIntanceToTag?.get(instanceId);
```
rpominov, line 441: `tagName` still comes from the rendered DOM node :( 

Basically we can't rely on anything that comes from `SelectedInstanceConnector` if we want to allow to not render instances on canvas.

Not sure solving this is in scope of this PR, just an observation.

packages/css-data/bin/mdn-data.ts
```diff
@@ -49,6 +49,18 @@ const normalizedValues = {
...
 
+const beautifyKeyword = (property: string, keyword: string) => {
+  if (keyword === "currentcolor") {
+    return "currentColor";
+  }
+  // builder style panel cannot interpret "normal" and "bold"
+  // always expected numeric value
+  if (property === "font-weight" && keyword === "normal") {
+    return "400";
+  }
```
rpominov, line 60: Strange that comment says `"normal" and "bold"`, but in code we handle only `normal`

rpominov (approved):

packages/css-data/src/__generated__/keyword-values.ts
```diff
@@ -3836,7 +3836,7 @@ export const keywordValues = {
     "unset",
   ],
   fontWeight: [
-    "normal",
+    "400",
     "bold",
     "bolder",
     "lighter",
```
rpominov, line 3842: There's also bolder/lighter that we may need to handle above

apps/builder/app/builder/features/style-panel/shared/style-info.ts
TrySound: It's not possible to solve unless data always provides currently used tag name

packages/css-data/bin/mdn-data.ts
TrySound: You can find bold in html.ts too

packages/css-data/src/__generated__/keyword-values.ts
TrySound: Ah, yeah. didn't notice. Need to revert this one

TrySound: Fixed bold. Though bolder and lighter are relative. Will leave as is. We anyway do not use this list. Possible weights are extracted from font.

kof:
@TrySound 
>This is not currentColor. It's inherit. I thought we already support it. The problem there is no color on body. So I guess it should fallback.

currentColor is inheritance, black is being defined somewhere above but we seem to not walk the tree or something? not sure, maybe we just need a defined color on body tag?

I tested on body, is it the same on other tags or maybe it was just broken on body

kof (approved):

packages/css-data/bin/mdn-data.ts
```diff
@@ -49,6 +49,23 @@ const normalizedValues = {
   "text-size-adjust": autoValue,
 } as const;
 
+const beautifyKeyword = (property: string, keyword: string) => {
+  if (keyword === "currentcolor") {
+    return "currentColor";
+  }
+  // builder style panel cannot interpret "normal" and "bold"
+  // always expected numeric value
+  if (property === "font-weight") {
```
kof, line 58: this is suspicious for bugs
- there is more keywords
-shouldn't the color input be able to interpret those keywords correctly?

TrySound: font-weight control. Yes it should, but we can fix it later. Its logic is not trivial.

kof: got it, lets create an issue, add everything you know that needs to happen, add the link here in a todo

PR #1534, opened 2023-04-27
Implement breakpoints popover design

kof:
Part of https://github.com/webstudio-is/webstudio-builder/issues/1407

## Description

Implements design changes from https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=3719-47947&t=IyN7o4sC7xAxaPNG-4

- new List and ListItem primitive components just for keyboard accessibility
- deprecated List and ListItem we had before because they are only useful in asset manager and we already learned how we want to treat keyboard differently

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov (approved):
Wrote in DM, duplicating here:

One of the ways to implement breakpoints is to use DropdownMenu for entire popover, and add extra controls at top and at the bottom.

This way keyboard interactions of DropdownMenu will be active. Although not sure these keyboard interactions will work well when the extra controls are added.

Current way is ok too, but you can't navigate menu using arrows.

rpominov (approved):

PR #1547, opened 2023-05-02
Upload files by url

TrySound:
Here splitted asset upload into 2 phases

1. initial upload creates asset in db with uploading state and generates unique name (POST /rest/assets endpoint)

2. use unique name to upload file, extract meta and mark asset as uploaded (POST /rest/assets/$name endpoint)

This give us more straightforward data flow without creating temporary asssets with fake id which are changed later.

And also unlock splitting files into separate table to allow securely copy paste assets between projects (still wip).

Fixed persistent fetcher to support sequential submits.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/routes/rest.assets.tsx
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    if (request.method === "POST") {
+      const formData = await request.formData();
+      const projectId = formData.get("projectId") as string;
+      const assetId = formData.get("assetId") as string;
+      const filename = formData.get("filename") as string;
+      if (projectId === null || assetId === null || filename === null) {
+        throw Error("Project id, asset id or filename are missing");
+      }
+      const name = await initiateUpload(
```
istarkov, line 36: I dislike the name as its not upload but name getter, 
something similar to `presigned url`, as only this name allows to upload file later.
`createPresignedFileName` or something better

packages/asset-uploader/src/upload.ts
```diff
@@ -1,30 +1,125 @@
...
+  { name, request }: { name: string; request: Request },
   client: AssetClient
 ) => {
-  const asset = await createAssetWithLimit(
-    projectId,
-    () => {
-      return client.uploadFile(request);
+  const asset = await prisma.asset.findFirst({
+    where: {
+      name: { equals: name },
```
istarkov, line 91: IMO you must add @unique constraint to the Asset table on `name`

TrySound: Not really. Name can be reused for cloned projects. Though I will make name an id in File table.

apps/builder/app/routes/rest.assets.$name.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  if (params.name === undefined) {
+    throw new Error("Name is undefined");
+  }
+
+  try {
+    if (request.method === "POST") {
+      const asset = await completeUpload(
+        {
+          name: params.name,
```
istarkov, line 21: What if everything in URL is considered as not secure, so bad actor can intercept and upload other photo with knowledge of `name`
Not an issue at all anyway as I have no idea why anyone would do that, does POST urls can leak as intents on mobile. And s3 presigned urls have no other security too

PR #1556, opened 2023-05-02
Fix preset styles

TrySound:
There was too early return

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
what is it fixing though?
any test possible?

style-info is probably now a quite complex thing we need tests for

TrySound:
Preset styles were skipped because of early return. Yes, though hard to write test for it. Will be easier with new implementation.

Issue #84, opened 2022-06-24
Forms UX


taylornowotny:
Users need to be able to control forms through an intuitive UX. Currently we can add a from to a page in Webstudio but users also need to be able to:

- add form field labels
- add & style placeholder text
- control form submission data
- style success, waiting, and error states
- style a field focus state
- customize inputs (multiple fields, radio & checkbox input, dropdown select)
- add recaptcha (after beta)

and on top of that, the html of the form needs to follow [accessibility best practices](https://www.w3.org/WAI/tutorials/forms/).

Webflow handles this with a 'form block' element in which you can add multiple types of fields:
<img width="248" alt="Screen Shot 2022-06-24 at 10 31 09 AM" src="https://user-images.githubusercontent.com/86495896/175612854-2b7cd2a0-d313-4de6-923a-25d60cb505cd.png">

And additionally this menu appears in the settings tab:
<img width="241" alt="Screen Shot 2022-06-24 at 10 33 55 AM" src="https://user-images.githubusercontent.com/86495896/175612974-64b790f4-0b1d-4ad5-89cc-55afa32eb11d.png">
Notice the toggle to change the state from Normal to Success to Error so those states can be styled and customized.

rpominov:
Details discussed on a call regarding the first iteration of forms:

When form is submitted, a email will be send to the owner of the site. Switch between normal/success/error states while in builder will be accomplished with a prop.

- [ ] Add the remaining components: checkbox, radio, ...
- [ ] Check that meta has all the useful states for each form component, including form itself 
- [ ] We move Form component to custom components and rename to RemixForm
- [ ] We add action for all forms and we identify which form is being submitted automatically
- [ ] We handle success and error responses and show SuccessMessage and ErrorMessage components
- [ ] Consider having one component FormMessage instead of ErrorMessage and SuccessMessage, then use status="success"|"error"
- [ ] Get the user email (signup email) from data directly
- [ ] Switch between form states happens over props (already implemented)
- [ ] Create mailer package, use mailchannels api, format the email

rpominov:
Another to-do (copied from https://github.com/webstudio-is/webstudio-builder/pull/1525#issuecomment-1532138860):

- [ ] We can do a lot better on the inputs naming. Let's eliminate the word "field":

current: Radio Button Field -> Radio Button -> Text Block
better: Radio -> Radio Input -> Radio Label

current: Checkbox Field -> Checkbox -> Text Block
[...]

PR #1568, opened 2023-05-04
Changes to enable forms submission in saas

rpominov:
Part of #84 
Should be deployed together with https://github.com/webstudio-is/webstudio-saas/pull/148

## Description

1. Exposed `User.email` in the data we send to saas
2. Used Remix's form in the Form component
3. Refactoring in `form-handlers`

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/routes/rest.project.$projectId.tsx
+++ b/apps/builder/app/routes/rest.project.$projectId.tsx
@@ -9,7 +9,9 @@ import { getUserById, type User } from "~/shared/db/user.server";
 export const loader = async ({
   params,
   request,
-}: LoaderArgs): Promise<Data & { user?: { email: User["email"] } }> => {
+}: LoaderArgs): Promise<
+  Data & { user: { email: User["email"] } | undefined }
+> => {
   try {
     const projectId = params.projectId ?? undefined;
 
--- a/apps/builder/app/routes/rest.project.$projectId.tsx
+++ b/apps/builder/app/routes/rest.project.$projectId.tsx
@@ -1,13 +1,15 @@
 import { json, type LoaderArgs } from "@remix-run/node";
 import type { Data } from "@webstudio-is/react-sdk";
+import { db as projectDb } from "@webstudio-is/project/index.server";
 import { sentryException } from "~/shared/sentry";
 import { loadProductionCanvasData } from "~/shared/db";
 import { createContext } from "~/shared/context.server";
+import { getUserById, type User } from "~/shared/db/user.server";
 
 export const loader = async ({
   params,
   request,
-}: LoaderArgs): Promise<Data> => {
+}: LoaderArgs): Promise<Data & { user?: { email: User["email"] } }> => {
   try {
     const projectId = params.projectId ?? undefined;
 
@@ -22,7 +24,20 @@ export const loader = async ({
       context
     );
 
-    return pagesCanvasData;
+    const project = await projectDb.project.loadByParams(
+      { projectId },
+      context
+    );
+
+    const user =
+      project === null || project.userId === null
+        ? undefined
+        : await getUserById(project.userId);
+
+    return {
+      ...pagesCanvasData,
+      user: user ? { email: user.email } : undefined,
+    };
   } catch (error) {
     // If a Response is thrown, we're rethrowing it for Remix to handle.
     // https://remix.run/docs/en/v1/api/conventions#throwing-responses-in-loaders
--- a/packages/form-handlers-mailchannels-test/src/index.ts
+++ b/packages/form-handlers-mailchannels-test/src/index.ts
@@ -18,14 +18,14 @@ export default {
     formData.append("time", new Date().toLocaleString("en-US"));
 
     const result = await mailchannelsHandler({
-      formInto: {
+      formInfo: {
         formData,
         projectDomain: "example.com",
         projectId: "test-project",
         pageUrl: "https://example.com/some-page",
         email: to,
+        senderDomain: "webstudio.is",
       },
-      senderDomain: "webstudio.is",
     });
 
     if (result.success) {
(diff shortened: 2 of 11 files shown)


apps/builder/app/routes/rest.project.$projectId.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { sentryException } from "~/shared/sentry";
 import { loadProductionCanvasData } from "~/shared/db";
 import { createContext } from "~/shared/context.server";
+import { getUserById, type User } from "~/shared/db/user.server";
 
 export const loader = async ({
   params,
   request,
-}: LoaderArgs): Promise<Data> => {
+}: LoaderArgs): Promise<Data & { user?: { email: User["email"] } }> => {
```
istarkov, line 12: If we allow more strict typings then your type having how you use it below is not optional but 
`user: { email: User["email"] } | undefined`

apps/builder/app/routes/rest.project.$projectId.tsx
```diff
@@ -22,7 +24,20 @@ export const loader = async ({
...
+    );
+
+    const user =
+      project === null || project.userId === null
+        ? undefined
+        : await getUserById(project.userId);
+
+    return {
+      ...pagesCanvasData,
+      user: user ? { email: user.email } : undefined,
```
istarkov, line 39: Not a bug, just for us to keep in mind 
In the only non protected endpoint we are now leaking user email

packages/form-handlers-mailchannels-test/src/index.ts
```diff
@@ -13,13 +13,19 @@ export default {
     }
 
     const formData = new FormData();
+    formData.append("ws--form-id", "test-form");
```
istarkov, line 16: plz move `ws--form-id` as shared const identifier, examples https://github.com/webstudio-is/webstudio-builder/blob/d61c39f22db07e0171fbc232eddb2d8b650b0dfd/packages/react-sdk/src/tree/webstudio-component.tsx#L61-L63C1

packages/form-handlers-mailchannels-test/src/index.ts
```diff
@@ -13,13 +13,19 @@ export default {
...
     formData.append("name", "Bob");
```
istarkov, line 17: Where is Alice?

packages/form-handlers/src/shared.ts
```diff
@@ -1,34 +1,93 @@
...
+  let txt = `There has been a new submission of your form at ${pageUrl}:\n\n`;
+
+  html += "<table><tbody>";
+
+  for (const [key, value] of getFormEntries(formData)) {
+    html += `<tr><td><strong>${key}:</strong></td><td>${value}</td></tr>`;
+    txt += `${key}: ${value}\n`;
+  }
+
+  html += "</tbody></table>";
```
istarkov, line 50: I would never recommend to generate email html with hands, https://github.com/mjmlio/mjml
Example of title, table
https://mjml.io/try-it-live/tq4Ea7NQaP

packages/form-handlers/src/shared.ts
```diff
@@ -1,34 +1,93 @@
...
 
   return {
-    subject,
-    plainText: `${prefix}\n\n${plainTextRows}`,
-    html: `<!DOCTYPE html><html><body><p>${prefix}</p><table><tbody>${htmlRows}</tbody></table></body></html>`,
+    from: `${projectDomain}@${senderDomain}`,
```
istarkov, line 53: what if self hosted domain, what we want to see here, why not pass from as an argument instead of generating it here

packages/form-handlers/src/shared.ts
```diff
@@ -1,34 +1,93 @@
...
+    const json = JSON.parse(text);
+    return typeof json === "object" && json !== null
+      ? { json, text }
+      : { text };
+  } catch {
+    return { text: text === "" ? response.statusText : text };
+  }
+};
+
+export const getErrors = (
```
istarkov, line 75: The whole body funciton looks like zod object with transform, like

`z.union(z.object({error: z.string()}).blabla( v => [v.errors]), ...)` not sure this work

packages/form-handlers/src/shared.ts
rpominov: > what if self hosted domain, what we want to see here

Don't know. Question for @kof .

> why not pass from as an argument instead of generating it here

In this case this logic will be in worker, but I'm trying to move as much logic as possible into the code that will be shared between saas and oss

packages/form-handlers/src/shared.ts
rpominov: Zod feels like overkill. I want to keep this package light

packages/form-handlers/src/shared.ts
rpominov: With custom domain I think we want: `example.com@webstudio.mail`.

Why not `form@example.com` or something? In this case deliverability of emails may suffer:
- the service may even refuse to send the email depending on the domain
- it will be less trustworthy for spam filters

packages/form-handlers/src/shared.ts
rpominov: BTW, dependencies like `mjml` will go into the worker bundle if we add them here

apps/builder/app/routes/rest.project.$projectId.tsx
```diff
@@ -22,7 +24,20 @@ export const loader = async ({
...
     );
 
-    return pagesCanvasData;
+    const project = await projectDb.project.loadByParams(
+      { projectId },
+      context
+    );
+
+    const user =
+      project === null || project.userId === null
```
kof, line 33: userId can be null? I wonder how is that possible

packages/form-handlers/src/shared.ts
kof: from can be easily passed, no strong reason to build it here

PR #1573, opened 2023-05-05
Move assets data into File table

TrySound:

We want to be able to copy paste assets between projects. Though assets are tied to specific project and we need to duplicate its data client side which can be a security hole.

Here I added File table with all file data from assets grouped by name. So now client side will still duplicate File data within session but this data will no persist in db/storage.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


rpominov (approved):

packages/prisma-client/prisma/migrations/20230501151941_asset-to-file/migration.ts
```diff
@@ -0,0 +1,43 @@
...
+        SELECT
+          "name",
+          "format",
+          "size",
+          "description",
+          "createdAt",
+          "meta",
+          "status"
+        FROM "Asset"
+        GROUP BY
```
istarkov, line 30: why? do we have duplicates?

TrySound: Yes, project cloning reuse same files in storage and duplicate only entries in db.

PR #1582, opened 2023-05-06
Reduce published site size

TrySound:
Image prop were added to the bundle because of object rest and spread. Here I moved it to function and marked its execution as pure to cut 15kB.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/image/src/index.ts
+++ b/packages/image/src/index.ts
@@ -4,15 +4,19 @@ export type { ImageLoader } from "./image-optimize";
 export * as loaders from "./image-loaders";
 import { props } from "./__generated__/image.props";
 
-// "loader" is our internal prop not intended to show up in the props panel
-const { loader, ...publicProps } = props;
+const getImageProps = (): Record<string, PropMeta> => {
+  // "loader" is our internal prop not intended to show up in the props panel
+  const { loader, ...publicProps } = props;
 
-export const imageProps: Record<string, PropMeta> = {
-  ...publicProps,
-  src: {
-    type: "string",
-    control: "file",
-    label: "Source",
-    required: false,
-  },
+  return {
+    ...publicProps,
+    src: {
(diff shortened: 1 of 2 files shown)


packages/image/src/index.ts
```diff
@@ -4,15 +4,19 @@ export type { ImageLoader } from "./image-optimize";
 export * as loaders from "./image-loaders";
 import { props } from "./__generated__/image.props";
 
-// "loader" is our internal prop not intended to show up in the props panel
-const { loader, ...publicProps } = props;
+const getImageProps = (): Record<string, PropMeta> => {
+  // "loader" is our internal prop not intended to show up in the props panel
+  const { loader, ...publicProps } = props;
```
rpominov, line 9: Probably only this spread causes issue, because the other kind of spread we use in many other components.

If so, might be better to put only it in a function to make situation more clear

rpominov (approved):

packages/image/src/index.ts
TrySound: Nope, looks like spread affects bundle too

rpominov: Just strange that spread here for example is not a problem https://github.com/webstudio-is/webstudio-builder/blob/317edc0d79a2dbcea0b9684a7e8f48113911a77d/packages/react-sdk/src/components/link.ws.tsx#L37

Or maybe it's a problem too

kof:
Wow, is there a way to detect such issues? it's possible after changes we will just not notice and break it again

TrySound:
We may need to decouple meta from component on resolution level.
For example we gonna have libraries including internal one as separate packages.
We can use different entry points to access them.
For example "@webstudio-is/image" for component and "@webstudio-is/image/meta" for meta and props.

PR #1588, opened 2023-05-06
(feat): Change button element to embed to prevent nesting

JayaKrishnaNamburu:
## Description

Changes the button element to `embed`. So, it will prevent nesting of buttons. `embed` might not be correct wording for it. But, it seems `embeds` prevent adding child elements like for `images`, `videos` etc..

## Steps for reproduction

1. Add a button element to canvas
2. Select button element and try to add a new element into it. 

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] tested locally
- [ ] added tests


packages/react-sdk/src/components/button.ws.tsx
```diff
@@ -14,7 +14,7 @@ const presetStyle = {
 
 export const meta: WsComponentMeta = {
   category: "general",
-  type: "container",
+  type: "embed",
```
kof, line 17: this will prevent user from having an icon inside of a button

kof: I think we need a more hollistic approach to this problem, to look more generally at how we would prevent nesting wrong elements inside other wrong elements

JayaKrishnaNamburu: Since, we the studio always knows the elements that are supported in the platform. How about every element have a new attribute that says, what `type` of children are allowed to add 🤔 
Some element to element mappers of sort ?

JayaKrishnaNamburu: Good catch btw 👍

kof: We have been discussing this internally, @TrySound has something to say about it

TrySound: Describing all possible types might not be a good idea because the list can grow a lot. Maybe support forbiddenDescendants?

TrySound: Or even forbiddenAncestors? Easier to write.

kof: yes, but should we have a global map of forbidden descendants and forbidden ancestors?

JayaKrishnaNamburu: Yes, i am also thinking along those lines. Should i go ahead and try something out ?

TrySound: No worries I already started

JayaKrishnaNamburu: Ah, ok ok. I had some thing like this and use them in return to type the map and props meta and elements 😄 

```ts
export enum ELEMENTS {
  SLOT = 'Slot',
  FRAGMENT = 'Fragment',
  BOX = 'Box',
  BODY = 'Body',
  TEXT_BLOCK = 'Text Block',
  HEADINNG = 'Heading',
  PARAGRAPH = 'Paragraph',
  LINK = "Link",
  LINK_BLOCK = 'Link Block',
  RICH_TEXT_LINK = 'Link Text',
  SPAN = 'Span',
  BOLD = 'Blod',
  ITALIC = 'Italic',
  SUPERSCRIPT = 'Superscript',
  SUBSCRIPT = 'Subscript',
  BUTTON = 'Button',
  INPUT = 'Input',
  FORM = 'Form',
  IMAGE = 'Image',
  BLOCKQUOTE = 'Blockquote',
  LIST = 'List',
  LIST_ITEM = 'List Item',
  SEPERATOR = 'Separator',
  CODE = 'Code',
  LABEL = 'Label',
  SUCCESS_MESSAGE = 'Success Message',
  ERROR_MESSAGE = 'Error Message',
  TEXT_AREA = 'Text Area',
  RADIO_BUTTON_FIELD = "Radio Button Field",
  RADIO_BUTTON = 'Radio Button',
  CHECKBOX_FIELD = 'Checkbox Field',
  CHECKBOX = 'Checkbox'
}

```

packages/react-sdk/src/components/elements.ts
```diff
@@ -0,0 +1,34 @@
+export enum ELEMENTS {
```
kof, line 1: - we are not using ts enum
- this logic is gonna change a lot, because we want to make it all auto-generated, user only needs to add a component, the rest should be automated, meta will be optional

kof:
Closing this, lets discuss next time before investing time into fixing

PR #1590, opened 2023-05-07
Support remote label for next style sources

TrySound:
Ref https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=1037-32728&t=HT5zNo0zhXDCpIX8-0

Here show next style sources as remote too. And add alert when next style source has same style as local so user see something overrides it.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/property-name.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/property-name.tsx
@@ -4,6 +4,7 @@ import type { StyleProperty } from "@webstudio-is/css-data";
 import { toValue } from "@webstudio-is/css-engine";
 import {
   theme,
+  rawTheme,
   Button,
   Flex,
   Box,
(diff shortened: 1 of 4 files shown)


apps/builder/app/builder/features/style-panel/shared/property-name.tsx
```diff
@@ -62,6 +63,24 @@ const PropertyPopoverContent = ({
...
+              if (styleSource?.type === "local") {
+                name = "local style source";
+              }
+              if (styleSource?.type === "token") {
+                name = `"${styleSource.name}" token`;
+              }
+              return (
+                <Text key={property} color="subtle">
+                  The value of {property} is overriden by {toValue(value)} from{" "}
+                  {name}
```
kof, line 79: this could be better off as a single template literal

kof (approved):

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
TrySound: Probably 🤷

PR #1591, opened 2023-05-07
Fix link default and preset styles

TrySound:
Here made added browser defaults for link and set color as preset.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/css/presets.ts
+++ b/packages/react-sdk/src/css/presets.ts
@@ -18,3 +18,21 @@ export const borders: EmbedTemplateStyleDecl[] = [
     value: { type: "unit", value: 1, unit: "px" },
   },
 ];
+
+export const linkColors: EmbedTemplateStyleDecl[] = [
+  {
+    property: "color",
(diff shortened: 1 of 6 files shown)


packages/react-sdk/src/css/presets.ts
```diff
@@ -18,3 +18,21 @@ export const borders: EmbedTemplateStyleDecl[] = [
     value: { type: "unit", value: 1, unit: "px" },
   },
 ];
+
+export const linkColors: EmbedTemplateStyleDecl[] = [
```
kof, line 22: we should really rename EmbedTemplate because for example here it's not an embed any more

kof (approved):

packages/react-sdk/src/css/presets.ts
TrySound: We reuse same simplified type. It's just resolved differently. I can alias it on top of module to make it less confusing.

PR #1596, opened 2023-05-08
(feat)/outline-section in style-panel

### LINKED ISSUE #1595: Outline property section
Design https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=4512-49995&t=J7aaJzbOtv9QmvKe-4

Have a look at borders section implementation, all components needed already used there

Screenshot 2023-05-07 at 19 30 48

JayaKrishnaNamburu:
I am marking this as draft, as there is a issue related to this and the render right now. The `canvas` is using `outline` to display the borders for the elements.

The style-panel, set and reset values works fine. Need to discuss on how the canvas should be handled in displaying these. Since, the selectors are being overwritten by the styles that are applied to show the borders.
`data-ws-collapsed`

Screenshot 2023-05-08 at 10 51 20 AM

Closes #1595 

## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


JayaKrishnaNamburu:
And this. use case happens, is clear here
https://github.com/webstudio-is/webstudio-builder/blob/main/apps/builder/app/canvas/collapsed.ts#L206

Not sure, if it's considered as blocker for this one. As it happens only in specific conditions. When the node has any children or some sort. It's all ok

kof:
canvas outline helper would always overwrite user styles, its a very rare case though when box is collapsing otherwise

JayaKrishnaNamburu:
The canvas and the panel has the same behaviour as webflow. And yes, it's a very rare case. To have a empty block and then style it with outline. So opening it for review, as the rest of the functionality is working as expected. Let me know, if there are any changes that are needed.

TrySound (approved):

apps/builder/app/builder/features/style-panel/sections/outline/outline-style.tsx
```diff
@@ -0,0 +1,75 @@
...
+  { value: "none", Icon: SmallXIcon },
+  { value: "solid", Icon: DashBorderIcon },
+  { value: "dashed", Icon: DashedBorderIcon },
+  { value: "dotted", Icon: DottedBorderIcon },
+];
+
+export const OutlineStyle = (
+  props: Pick<
+    RenderCategoryProps,
+    "currentStyle" | "setProperty" | "deleteProperty" | "createBatchUpdate"
```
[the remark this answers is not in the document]
JayaKrishnaNamburu: Updated the typings 👍

apps/builder/app/builder/features/style-panel/sections/outline/outline-offset.tsx
```diff
@@ -0,0 +1,53 @@
...
+        label={"Offset"}
+        onReset={() => deleteProperty(property)}
+      />
+
+      <CssValueInputContainer
+        key={property}
+        property={property}
+        label={outlineOffsetStyleConfig?.label || ""}
+        styleSource={getStyleSource(currentStyle[property])}
+        keywords={[]}
```
kof, line 46: there are no keywords?

packages/css-data/bin/mdn-data.ts
```diff
@@ -56,7 +56,16 @@ const normalizedValues = {
...
-  if (keyword === "currentcolor") {
+  /*
+   * outline-color support both invert and currentColor properties.
+   * The default value is `invert` but that isn't supported in all browsers.
+   * So, going with currentColor for consistency.
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/outline-color#formal_definition
+   */
+  if (
+    keyword === "currentcolor" ||
+    (property === "outline-color" && keyword === "invertOrCurrentColor")
```
kof, line 67: shouldn't we in this case always use currentColor instead of invertOrCurrentColor for all properties, not only outline-color?

JayaKrishnaNamburu:
@kof  as i see, i couldn't pin point where the filtering for `keywords` from values is happening in border. But, if users manually enter `medium` or something like that. Border seems to accept those. 
Can you point me where to check for restricting `keywords` for certain styles like for `outlineColors`.

Example, when `auto-generated` for `borderTopWidth`
https://github.com/webstudio-is/webstudio-builder/blob/main/packages/css-data/src/__generated__/keyword-values.ts#L2649
It has all the keywords too. But, i think here is the check it's changing default values
https://github.com/webstudio-is/webstudio-builder/blob/main/apps/builder/app/builder/features/style-panel/sections/borders/border-property.tsx#L81

kof:
> But, if users manually enter medium. Border seems to accept those. Should they be restricted ?

No, in this case it should take it I think

apps/builder/app/builder/features/style-panel/shared/style-info.ts
```diff
@@ -60,6 +60,15 @@ type SourceProperties = {
...
 
+/*
+ * For some cases we are encouraging to use custom defaults, than
+ * `initial` values provided by browsers. This helps in controlling the behaviour
+ * of such properties
+ */
+const CUSTOM_DEFAULT_VALUES: Partial<Record<StyleProperty, StyleValue>> = {
+  outlineWidth: { value: 0, type: "unit", unit: "px" },
+};
+
```
JayaKrishnaNamburu, line 71: Heya, i took a little different approach here. Since the users are allowed to use `keywords` at any time. It is little tricky to always follow up with `keyword` default checks. So, if a element misses the style from `htmlSource`. The logic checks if there are any custom defaults are defined. And if they are missing from here too, then it fallbacks to the auto-generated. I thought of adding cases while generating `__generated__/properties`. But, that effects the actual sources. So

This helps in arranging all the custom defaults too in one place. Does that looks clear, open to suggestions on improving this.

kof: I think @istarkov knows this logic from borders section

kof: the actual default value is medium, so we would need to use 1 to approximate that

kof: @TrySound what's the best way to handle this and rely on our normalize/preset?

kof: same should be for borders

kof: @istarkov were you just using hardcoded 1px for border width or did you map it from medium to 1 once you got medium from data?

kof: I see borders are in normalize.ts, probably outlines should be too

JayaKrishnaNamburu: But presets needs to import `normaiize` right. Because `outline` can be applied on every element ? We should add to every element ? 🤔 

https://github.com/webstudio-is/webstudio-builder/blob/main/packages/react-sdk/src/components/body.ws.tsx#L2

kof: I would say it should be just like the border, let's wait for @TrySound to reply

TrySound: This is where borders set for every tag
https://github.com/webstudio-is/webstudio-builder/blob/main/packages/react-sdk/src/css/normalize.ts#L40

JayaKrishnaNamburu: Thanks for the tip, added as preset now 👍

kof:
needed a rebase to fix CI issues, merging it from this pr https://github.com/webstudio-is/webstudio-builder/pull/1630
also setting default for border width to 1, since browser's default is not 0, but medium

PR #1599, opened 2023-05-08
Optimize patch endpoint

TrySound:
We batching transactions though each transaction and each namespaace in it loads build data, applies patches and writes to db. This makes patches do a lot of connections to db within single request.

Here each patch request loads and updates build only once and apply patches of all transactions to intermediary state.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/routes/rest.patch.ts
+++ b/apps/builder/app/routes/rest.patch.ts
@@ -1,19 +1,38 @@
 import type { ActionArgs } from "@remix-run/node";
 import type { SyncItem } from "immerhin";
-import type { Build } from "@webstudio-is/project-build";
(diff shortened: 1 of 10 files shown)


istarkov:
Serious improvements locally, now 300-400ms vs 2s+

istarkov:
So most time is spent here 157ms of 159ms on save.

https://github.com/webstudio-is/webstudio-builder/blob/736cf219d647c3d22fb4c594a026ea20726fe59e/apps/builder/app/routes/rest.patch.ts#L189-L191

Let's think how to make it 1ms

istarkov (approved):
Super!

PR #1603, opened 2023-05-08
Add bug report button

kof:
## Description

We may need to rethink what we show when user hits report bug, e.g. ideally user would at least make a screenshot or create a video but for now as a simple option is to lead them directly into github discussions

## Steps for reproduction

1. see new bug button on the left bottom

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

TrySound (approved):

apps/builder/app/builder/features/sidebar-left/sidebar-left.tsx
```diff
@@ -59,6 +61,20 @@ export const SidebarLeft = ({ publish }: SidebarLeftProps) => {
             </SidebarTabsTrigger>
           ))}
         </SidebarTabsList>
+        <Box css={{ borderRight: `1px solid  ${theme.colors.borderMain}` }}>
+          <SidebarTabsTrigger
+            as={"button"}
```
istarkov, line 66: Just curious wny not link with `target="_blank"
As in case of something really bad with js, native link will work

kof: yep, problem was types of that button don't take an anchor and I didn't want to rewrite the button

PR #1607, opened 2023-05-09
Remove hover debounce

istarkov:
## Description

Debounce on hover makes me feel that everything is slow, you move mouse but hover is not completed.
Compare with immediate hover, feels like UI is superfast

For example Figma has immediate hover

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/instance-hovering.ts
+++ b/apps/builder/app/canvas/instance-hovering.ts
@@ -1,4 +1,3 @@
-import debounce from "lodash.debounce";
 import { idAttribute } from "@webstudio-is/react-sdk";
 import {
   hoveredInstanceSelectorStore,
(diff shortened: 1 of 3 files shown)


kof:
I think debounce there was because you can hover quickly over a bunch of items without triggering hover everywhere, which makes it look chaotic, but I also noticed the slowish feel, so maybe there is no way around.

When I hover over figma, those lines never skip, it shows everything you hover, so I am fine with the change.

apps/builder/app/canvas/instance-hovering.ts
```diff
@@ -65,7 +65,7 @@ export const subscribeInstanceHovering = () => {
       instanceId: instance.id,
       rect: getAllElementsBoundingBox(element),
     });
-  }, 50);
+  }, 0);
```
kof, line 68: do we still need the debounce at all?

kof (approved):

apps/builder/app/canvas/instance-hovering.ts
istarkov: Dont need just to show that it feels better

TrySound (approved):

PR #1618, opened 2023-05-11
Save project uploader when create file

TrySound:
We need to be able to track users who uploads files to prevent misuse of our storage.
Here added File.uploaderProjectId in db. This way we can always find project owner.
Saving user is less ideal because often can be no user for example when share edit permission by token.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/src/upload.ts
+++ b/packages/asset-uploader/src/upload.ts
@@ -45,16 +45,12 @@ export const createUploadName = async (
       OR: [
         {
           status: "UPLOADED",
-          assets: {
-            some: { projectId },
-          },
+          uploaderProjectId: projectId,
         },
         {
           status: "UPLOADING",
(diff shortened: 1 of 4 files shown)


packages/asset-uploader/src/upload.ts
```diff
@@ -32,25 +32,31 @@ export const createUploadName = async (
       "You don't have access to create this project assets"
     );
   }
+  const { userId } = context.authorization;
+  if (userId === undefined) {
+    throw new Error("The user must be authenticated to create a project");
```
istarkov, line 37: Wrong error. Probably we should not ask to auth to upolad file. If user has edit permission thorugh a token why not?
mb better to write initial ProjectId here

TrySound: Added uploaderProjectId instead

istarkov (approved):

PR #1624, opened 2023-05-13
Add html component

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1563

Looks like dangerouslySetInnerHtml is quite safe and do not allow to break outside layout.
Plus as important point it works with ssr.

To execute scripts while injecting used range hack described here https://ghinda.net/article/script-tags. innerHTML does not execute scripts.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/components/html.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export const Html = forwardRef<ElementRef<"div">, Props>((props, ref) => {
+  const { code, ...rest } = props;
+  return (
+    <div
+      {...rest}
+      ref={ref}
+      style={{ display: code?.trim().length !== 0 ? "contents" : "block" }}
+      dangerouslySetInnerHTML={{ __html: code ?? "" }}
```
kof, line 14: didn't we want to use insertAdjacentHTML or is react using it?

TrySound: Just first draft

packages/react-sdk/src/components/html.tsx
```diff
@@ -0,0 +1,19 @@
...
+  code?: string;
+};
+
+export const Html = forwardRef<ElementRef<"div">, Props>((props, ref) => {
+  const { code, ...rest } = props;
+  return (
+    <div
+      {...rest}
+      ref={ref}
+      style={{ display: code?.trim().length !== 0 ? "contents" : "block" }}
```
kof, line 13: why not using data-state attribute and based on it style over preset?

kof: actually you don't even need to ever use block, because when there is no user code pasted we can render fallback that will hold the spacing

packages/react-sdk/src/components/html.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+/**
+ * Scripts are executed when rendered client side
+ */
+const ExecutableHtml = (props: ChildProps) => {
+  const { code, innerRef, ...rest } = props;
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
istarkov, line 29: What prevent us to set `<script` tag on the server?

istarkov: I see, below

packages/react-sdk/src/components/html.tsx
```diff
@@ -0,0 +1,94 @@
...
+type ChildProps = {
+  innerRef: ForwardedRef<HTMLDivElement>;
+  // code can be actually undefined when prop is not provided
+  code?: string;
+};
+
+/**
+ * Scripts are executed when rendered client side
+ */
+const ExecutableHtml = (props: ChildProps) => {
```
istarkov, line 25: i would add. comment that we don't have a server rendering inside builder canvas, so the range trick (give a link to the trick article) is used

istarkov:
I have feeling that after we disable "execute ..."

image

We need to do a FULL rerender of the canvas document, (adding new key to the root element)
At least to avoid examples like above.

Probably somehow clearing all event listeners from document. Or even do a canvas refresh/reload.

packages/react-sdk/src/components/html.tsx
```diff
@@ -0,0 +1,94 @@
...
+ */
+const InnerHtml = (props: ChildProps) => {
+  const { code, innerRef, ...rest } = props;
+
+  return (
+    <div
+      {...rest}
+      ref={innerRef}
+      style={{ display: "contents" }}
+      dangerouslySetInnerHTML={{ __html: props.code ?? "" }}
```
istarkov, line 64: Probably I'm wrong but this will ruin layout

istarkov: I just tested with SSR and this ruins layout

image

This should not be used without html sanitizer which will enclose each tag

istarkov: I think it will be perfectly ok to use client-side injection here too, less error-prone.
But probably some scripts depending on load even will not work.

packages/react-sdk/src/components/html.tsx
TrySound: Nothing, this is the main point of this component.

apps/builder/app/builder/features/props-panel/controls/code.tsx
```diff
@@ -0,0 +1,36 @@
...
+      instanceId={instanceId}
+      meta={{
+        ...meta,
+        control: "text",
+      }}
+      prop={prop}
+      propName={propName}
+      onChange={(value) => {
+        if (value.type === "string") {
+          // sanitize html before saving
```
istarkov, line 24: mb some comment why this works ;)

PR #1626, opened 2023-05-13
Fix assets vercel uploading 

istarkov:
## Description

1. Add 404 instead of redirect
2. Fix vercel/remix upload

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/routes/$.tsx
+++ b/apps/builder/app/routes/$.tsx
@@ -1,6 +1,10 @@
 import { redirect } from "@remix-run/node";
 import type { LoaderArgs } from "@remix-run/node";
-import { useLoaderData, useRouteError } from "@remix-run/react";
+import {
+  isRouteErrorResponse,
+  useLoaderData,
+  useRouteError,
+} from "@remix-run/react";
 import { type Params, Root, getComponent } from "@webstudio-is/react-sdk";
 import env from "~/env/env.public.server";
 import { sentryException } from "~/shared/sentry";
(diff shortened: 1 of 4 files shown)


apps/builder/app/routes/$.tsx
```diff
@@ -9,6 +13,17 @@ import { ErrorMessage } from "~/shared/error";
 import { getBuildParams, dashboardPath } from "~/shared/router-utils";
 
 export const loader = async ({ request }: LoaderArgs) => {
+  const url = new URL(request.url);
+  // See remix.config.ts for the publicPath value
+  const publicPath = "/build/";
```
kof, line 18: shouldn't we get this value over some config or at least put it to some reusable place

TrySound (approved):

apps/builder/app/routes/$.tsx
istarkov: js and ts mix and moreover esm and cjs
between this file and remix.config.js

Looks like too many things to change is tsconfig, and usually causer runtime checks on exports as they are different in esm and cjs

istarkov: cc @TrySound do you know a simple way to share const between remix.config.js and routes/$.ts?

TrySound: I'm not sure. Some shit will happen for sure. Better hardcode. I'm sure this is temporary.

PR #1629, opened 2023-05-14
Fix types

istarkov:
## Description

Makes AllComponent type better

Found null bug.

Enforced for all components 
```
    [componentAttribute]: instance.component,
    [idAttribute]: instance.id,
```
to be non-optional. 

cc @TrySound should fix your issue with Html component

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
+++ b/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
@@ -79,7 +79,7 @@ export const SelectedInstanceConnector = ({
   instanceProps,
   instanceSelector,
 }: {
-  instanceElementRef: { current: undefined | HTMLElement };
+  instanceElementRef: { current: null | HTMLElement };
   instance: Instance;
   instanceStyles: StyleDecl[];
   instanceProps: undefined | Prop[];
@@ -93,7 +93,7 @@ export const SelectedInstanceConnector = ({
     setDataCollapsed(instance.id, true);
 
     const element = instanceElementRef.current;
-    if (element === undefined) {
+    if (element == null) {
       return;
     }
     const showOutline = () => {
--- a/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
+++ b/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
@@ -93,7 +93,7 @@ export const SelectedInstanceConnector = ({
     setDataCollapsed(instance.id, true);
 
     const element = instanceElementRef.current;
-    if (element == null) {
+    if (element === null) {
       return;
     }
     const showOutline = () => {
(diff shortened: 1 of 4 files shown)


apps/builder/app/canvas/shared/use-pointer-outline.ts
```diff
@@ -18,6 +18,7 @@ export const usePointerOutline = () => {
     ref.current = div;
     return () => {
       document.body.removeChild(div);
+      ref.current = undefined;
```
kof, line 21: if we get null anyways, i would use null consistently, here as well

apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
```diff
@@ -93,7 +93,7 @@ export const SelectedInstanceConnector = ({
     setDataCollapsed(instance.id, true);
 
     const element = instanceElementRef.current;
-    if (element === undefined) {
+    if (element == null) {
```
kof, line 96: if you use null, then we can go with strict equal

apps/builder/app/canvas/shared/use-pointer-outline.ts
```diff
@@ -18,6 +18,7 @@ export const usePointerOutline = () => {
     ref.current = div;
```
TrySound, line 18: Is this module even used?

kof: I used it in the past to debug the coordinates of the pointer, since we are doing recalculation based on the canvas scale and position, we basically need to adjust the pointer position that actually is on canvas during drag&drop

I used to turn this on to draw a rectangle for the pointer, we haven't had issues for a while, but it's an easy way to check where the pointer coordinate would be during drag to see if we are sending wrong coordinates to the canvas

kof: it could use a better place on the file system, like some kind of tools folder or something, if we want to keep it

apps/builder/app/canvas/shared/use-pointer-outline.ts
kof: oh I can see now why I used undefined, this ref is never passed to react, so we fully control it

apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
kof: checked again, after the change to initial null, it's always either null or html element, so non-strict check is not required

I personally don't care about this, since it's not doing any harm here, but it goes against the rfc and I don't want to make it confusing for other readers by making it inconsistent

kof (approved):
approving it, but I don't love the fact that we can't agree on a consistent usage, especially in cases where it is so simple to not have to support both undefined and null and yet you are using `==` checks where it's not needed

TrySound (approved):

PR #1632, opened 2023-05-15
Fix scroll reset in style panel

### LINKED ISSUE #1617: Style panel scroll resets when selected instance changes
See https://discord.com/channels/955905230107738152/1104421616005435412

- May also happen when background is added (although I couldn't reproduce)
- Also "Select an instance on the canvas" flashes for a brief moment

rpominov:
## Description

1. Closes #1617

Before this PR `selectedInstanceIsRendered` behaved like this:

Switching to instance that **is** on canvas: `true -> false -> true`
Switching to instance that **is not** on canvas: `true -> false`

Issue: we can't tell apart brief `false` in the first case, and final `false` in the second

After this PR:

Switching to instance that **is** on canvas: `true -> undefined -> true`
Switching to instance that **is not** on canvas: `true -> undefined -> false`

Now we can distinguish temporary `undefined` from permanent `false` and don't hide style panel during `undefined`

## Steps for reproduction

Case 1:

1. Select a Box instance
2. Scroll style panel
3. Select another Box instance
4. Scroll should not reset

Case 2:

1. Select a Box instance
2. Select a Success Message instance in a Form
3. Style panel should show "Select an instance on the canvas" message 

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
+++ b/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
@@ -188,6 +188,8 @@ export const SelectedInstanceConnector = ({
       unsubscribeScrollState();
       unsubscribeWindowResize();
       unsubscribeIsResizingCanvas();
+
+      // see webstudio-component.tsx for where it's set to `false`
       selectedInstanceIsRenderedStore.set(undefined);
     };
   }, [
--- a/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
+++ b/apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
@@ -11,7 +11,6 @@ import {
   selectedInstanceIntanceToTagStore,
   selectedInstanceUnitSizesStore,
   selectedInstanceIsRenderedStore,
-  selectedInstanceSelectorStore,
 } from "~/shared/nano-states";
 import htmlTags, { type htmlTags as HtmlTags } from "html-tags";
 import { getAllElementsBoundingBox } from "~/shared/dom-utils";
@@ -19,10 +18,7 @@ import { subscribeScrollState } from "~/canvas/shared/scroll-state";
 import { selectedInstanceOutlineStore } from "~/shared/nano-states";
 import type { UnitSizes } from "~/builder/features/style-panel/shared/css-value-input/convert-units";
 import { setDataCollapsed } from "~/canvas/collapsed";
-import {
-  areInstanceSelectorsEqual,
-  type InstanceSelector,
-} from "~/shared/tree-utils";
+import { type InstanceSelector } from "~/shared/tree-utils";
 
 const isHtmlTag = (tag: string): tag is HtmlTags =>
   htmlTags.includes(tag as HtmlTags);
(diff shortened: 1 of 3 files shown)


apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -142,6 +143,18 @@ export const WebstudioComponentDev = ({
...
+    if (instanceSelector.length === 1) {
+      // If by the time root is rendered,
+      // no selected instance renders and sets `selectedInstanceIsRendered` to `true`,
+      // then it's clear that selected instance will not render at all, so we set it to `false`
+      if (selectedInstanceIsRenderedStore.get() === undefined) {
+        selectedInstanceIsRenderedStore.set(false);
+      }
+    }
+  });
+
```
istarkov, line 157: I would describe cases then it happens. As I don't understand why this logic is placed on Root (Body?)
Why it should be rendered at all and why execute useEffect on every render. What if some optimisations will be applied and this rendering will never happen?

rpominov: Ah, yes. This actually relies on `useStore(selectedInstanceSelectorStore)` above in the component. We rely on re-render after `selectedInstance` change

apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
```diff
@@ -192,19 +188,7 @@ export const SelectedInstanceConnector = ({
...
-      // This rare issue is difficult to reproduce, occurring roughly once every 100 calls.
-      if (
-        areInstanceSelectorsEqual(
-          selectedInstanceSelectorStore.get(),
-          instanceSelector
-        ) === false
-      ) {
-        selectedInstanceIsRenderedStore.set(false);
-      }
+      selectedInstanceIsRenderedStore.set(undefined);
```
istarkov, line 191: I would give a link to webstudio-component useEffect here

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
rpominov: Added code comment

apps/builder/app/canvas/features/webstudio-component/selected-instance-connector.ts
rpominov: Addded

istarkov (approved):
I will accept, as of now we have no idea how component decide what to render.
But this looks super error prone. 
Any optimisation could kill current solution

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -142,6 +143,20 @@ export const WebstudioComponentDev = ({
...
+    // 1 means root
+    if (instanceSelector.length === 1) {
+      // If by the time root is rendered,
+      // no selected instance renders and sets `selectedInstanceIsRendered` to `true`,
+      // then it's clear that selected instance will not render at all, so we set it to `false`
+      if (selectedInstanceIsRenderedStore.get() === undefined) {
+        selectedInstanceIsRenderedStore.set(false);
+      }
+    }
+  });
```
kof, line 158: this is executed every single time, shouldn't it rather have dependencies, e.g. instanceSelector

rpominov: The code inside is very light. Checking dependencies can be even slower

apps/builder/app/shared/nano-states/nano-states.ts
```diff
@@ -233,8 +233,13 @@ export const selectedInstanceIntanceToTagStore = atom<
   undefined | Map<Instance["id"], HtmlTags>
 >();
 
-/** Whether or the selected instance is rendered on canvas */
-export const selectedInstanceIsRenderedStore = atom<boolean>(false);
+/**
+ * Whether or not the selected instance is rendered on canvas:
```
kof, line 237: This sentence is confusing, maybe the variable name is confusing with this semantics?

Kinda can't use "is" if its not boolean, you need to change the name

kof: seems like better name here would be: "selectedInstanceRenderState": "mounted" | "unmounted" | "something else" ?

rpominov: Ok

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
kof: the danger is that someone can add more to the code without realizing it's going to run every time

PR #1633, opened 2023-05-15
Create assets with patches

TrySound:

Successor of https://github.com/webstudio-is/webstudio-builder/pull/1376

Assets no longer created while uploading. Asset patch now tracks new assets to save in db.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/src/patch.ts
+++ b/packages/asset-uploader/src/patch.ts
@@ -1,5 +1,5 @@
 import { type Patch, applyPatches } from "immer";
-import type { Project } from "@webstudio-is/prisma-client";
+import { type Project, prisma } from "@webstudio-is/prisma-client";
 import {
   type AppContext,
   authorizeProject,
@@ -9,6 +9,10 @@ import { deleteAssets } from "./delete";
 import { loadAssetsByProject } from "./db/load";
 import type { AssetClient } from "./client";
 
+/**
+ * patchAssets can only delete or add assets
+ * update patches are ignored
+ */
 export const patchAssets = async (
   { projectId }: { projectId: Project["id"] },
   patches: Array<Patch>,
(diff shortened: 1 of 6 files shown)


packages/asset-uploader/src/patch.ts
```diff
@@ -40,4 +44,31 @@ export const patchAssets = async (
...
+    if (assets.has(assetId) === false) {
+      addedAssets.push(asset);
+    }
+  }
+  if (addedAssets.length !== 0) {
+    await prisma.asset.createMany({
+      data: addedAssets.map((asset) => ({
+        id: asset.id,
+        projectId: asset.projectId,
+        name: asset.name,
```
rpominov, line 64: This name can be arbitrary label now in theory? Doesn't link to file in store any more?

(Just curious)

rpominov (approved):

packages/asset-uploader/src/patch.ts
TrySound: It's foreign key for File table and used to address actual file in storage.

rpominov: Ah, I thought tables are linked by id because I saw this line, and thought that file id becomes asset id eventually

https://github.com/webstudio-is/webstudio-builder/blob/ad7fafe9cf915ec63721c9c1d522d8f4b37b5d5c/apps/builder/app/builder/shared/assets/use-assets.tsx#L191

Which I now realise wouldn't work because many assets can use the same file. So maybe it worth double checking the line above and how id gets passed from there.

TrySound: Yeah, maybe better rename to assetId

rpominov: Got it 👍

PR #1638, opened 2023-05-17
Mark files as deleted

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/193

To make undo of assets deletion work we need to stop deleting files from storage to restore assets without loosing actual file.

Here added File.isDeleted and file.updatedAt fields to mark as deleted. This way we can run scheduler to actually delete files which are marked for some time already. And undo of asset deletion now works.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/asset-uploader/src/patch.ts
```diff
@@ -63,6 +61,14 @@ export const patchAssets = async (
...
+    // restore file when undo is triggered
+    await prisma.file.updateMany({
+      where: { name: { in: Array.from(fileNames) } },
+      data: {
+        isDeleted: false,
+      },
+    });
+
     await prisma.asset.createMany({
       data: addedAssets
```
rpominov, line 73: Not related to this PR, but we can fix here: 

`projectId: asset.projectId` below is a vulnerability. Should be either `projectId: projectId`, or we can filter out assets where `asset.projectId !== projectId`

rpominov (approved):

istarkov (approved):

PR #1640, opened 2023-05-17
selectedInstanceIsRendered -> selectedInstanceRenderState

rpominov:
Fixes review comment: https://github.com/webstudio-is/webstudio-builder/pull/1632#discussion_r1196082482

### DIFF
--- a/apps/builder/app/shared/nano-states/nano-states.ts
+++ b/apps/builder/app/shared/nano-states/nano-states.ts
@@ -237,12 +237,12 @@ export const selectedInstanceIntanceToTagStore = atom<
 >();
 
 /**
- * Whether or not the selected instance is rendered on canvas:
- * `true` — rendered;
- * `false` — not rendered;
- * `undefined` — selected instance unmounted and we don't know yet whether a new one will mount.
+ * pending means: previous selected instance unmounted,
+ * and we don't know yet whether a new one will mount
  **/
-export const selectedInstanceIsRenderedStore = atom<boolean | undefined>(false);
+export const selectedInstanceRenderStateStore = atom<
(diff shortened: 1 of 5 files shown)


apps/builder/app/shared/nano-states/nano-states.ts
```diff
@@ -237,12 +237,12 @@ export const selectedInstanceIntanceToTagStore = atom<
...
- * Whether or not the selected instance is rendered on canvas:
- * `true` — rendered;
- * `false` — not rendered;
- * `undefined` — selected instance unmounted and we don't know yet whether a new one will mount.
+ * pending means: previous selected instance unmounted,
+ * and we don't know yet whether a new one will mount
  **/
-export const selectedInstanceIsRenderedStore = atom<boolean | undefined>(false);
+export const selectedInstanceRenderStateStore = atom<
+  "mounted" | "not-mounted" | "pending"
```
kof, line 244: I would expect camel case notMounted or unmounted

kof (approved):

apps/builder/app/shared/nano-states/nano-states.ts
rpominov: Changed to notMounted.

unmounted sounds wrong to me because it assumes it was mounted and then unmouted

kof: I think plain english unmounted doesn't imply it was previously mounted

rpominov: Even if not according to strict english rules, at least to me is sound like implies. Can be more people like me who may get confused

PR #1649, opened 2023-05-18
Make meta serializable

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1129 https://github.com/webstudio-is/webstudio-builder/issues/1545

Here put all components meta icons into strings and fixed usage with innerHTML. This allows us to load components only from canvas and send to builder.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

apps/builder/app/builder/features/props-panel/props-panel.tsx
```diff
@@ -59,7 +59,10 @@ const InstanceInfo = ({
     css={{ height: theme.spacing[13], color: theme.colors.foregroundSubtle }}
     align="center"
   >
-    <meta.Icon />{" "}
+    <div
+      style={{ width: 16, height: 16 }}
+      dangerouslySetInnerHTML={{ __html: meta.Icon }}
```
kof, line 64: if icon is a html string, then it shouldn't be upper case any more?

apps/builder/app/builder/features/sidebar-left/panels/components/components.tsx
```diff
@@ -110,7 +110,11 @@ export const TabContent = ({ publish, onSetActiveTab }: TabContentProps) => {
                             {...pressProps}
                             {...{ [dragItemAttribute]: component }}
                             label={meta.label}
-                            icon={<meta.Icon />}
+                            icon={
+                              <div
```
kof, line 114: maybe we could just move the rendering into the final component and pass the icon as-is as a string?

kof (approved):

apps/builder/app/builder/shared/meta-icon.tsx
```diff
@@ -0,0 +1,13 @@
+type Props = {
+  size?: number | string;
+  icon: string;
+};
+
+export const MetaIcon = ({ size = 16, icon }: Props) => {
+  return (
+    <div
+      style={{ width: size, height: size }}
```
kof, line 9: I would not use inline style, since you can have reusable css

kof: Would probably use a span, but no strong opiinion

kof: check out https://www.radix-ui.com/docs/primitives/utilities/accessible-icon

kof: depending on how icon is used, you may need to announce it accessibly and/or hide it from screen readers check out what radix does with icons

PR #1652, opened 2023-05-18
Initial CSS parser

ggurgone:
## Description

This PR moves the parse-css-value utility that was before part of the builder app to the css-data package. 

In this package I am adding an initial parse-css utility that will take CSS and convert to a format that we can work on within Webstudio. For now it is relatively simple and it supports only CSS classes parsing.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-data/package.json
```diff
@@ -12,15 +12,18 @@
...
+    "lint": "eslint ./src --ext .ts,.tsx --max-warnings 0",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest"
   },
   "devDependencies": {
     "@types/css-tree": "^2.0.0",
     "@webstudio-is/scripts": "workspace:^",
+    "@webstudio-is/jest-config": "workspace:^",
     "@webstudio-is/tsconfig": "workspace:^",
     "camelcase": "^6.3.0",
     "css-tree": "^2.3.1",
```
istarkov, line 24: Now it's not a devDependency
Also probably no refs on css-tree at builder left

patches/css-tree@2.3.1.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/package.json b/package.json
+index 934e699f19c6f1a265c1ef29121c6dde92cd25ed..a54e28a44de8e1adcf775be9564c2ad8b15b5c98 100644
+--- a/package.json
++++ b/package.json
+@@ -18,11 +18,11 @@
+     "validation"
+   ],
+   "type": "module",
+-  "module": "./lib/index.js",
++  "module": "./dist/csstree.esm.js",
```
ggurgone, line 10: patching css-tree to point to the built version so that packages (eg. the builder) that use `css-data` on the client don't fail:

Screen Shot 2023-05-22 at 10 51 28 AM

packages/css-data/src/property-parsers/background.ts
```diff
@@ -22,6 +21,28 @@ export const parseBackground = (
...
+  return {
+    backgroundImage: parseBackgroundImage(backgroundImage),
+    backgroundColor:
+      backgroundColor && backgroundColor.type === "rgb"
+        ? backgroundColor
+        : undefined,
+  };
+};
+
+export const backgroundToLonghand = (
```
istarkov, line 41: It's backgorundImage to longhand, here is not the full background parsing algo

packages/project-build/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "@webstudio-is/tsconfig/base.json",
-  "include": ["src"],
+  "include": ["src", "../../@types/**/*.d.ts"],
```
istarkov, line 3: cc @TrySound @Andarist Im not sure this will work

istarkov (approved):
Im pretty sure this `"../../@types/**/*.d.ts"` will ruin something with types 
But who knows

packages/css-data/src/property-parsers/background.ts
giuseppeg: I didn't add new logic for parsing per-se. I just split unwrapping the shorthand and parsing the longhand values

packages/project-build/tsconfig.json
ggurgone: `@types/css-tree` is not exposing default generated instances types of things like Lexer.

Up until this point `apps/builder` referenced these types for CSS Tree (because of the reason above). In short I didn't change anything so it shouldn't break

packages/css-data/src/property-parsers/background.ts
ggurgone: I didn't add new logic for parsing per-se. I just split unwrapping the shorthand and parsing the longhand values

packages/project-build/tsconfig.json
Andarist: Quite frankly, those kinds of options always confuse me - it gets really tricky when u mix similar-purpose options. I can't say with confidence that I know what kind of outcome this might have for your `.d.ts` generation and stuff. However, this was already configured like this for some packages so I guess that this might be roughly OK. I have a feeling though that this should use `typeRoots` and not `include` - but if it works... I wouldn't bother changing this right now.

ggurgone: > Quite frankly, those kinds of options always confuse me - it gets really tricky when u mix similar-purpose options

I agree with you! When `typeRoots` for some reason these are not picked up though. I created an issue about revisiting the types for css-tree https://github.com/webstudio-is/webstudio-builder/issues/1663

PR #1675, opened 2023-05-25
Add project domain validations

istarkov:
## Description

Services we need like `image-transform.wstd.io`, `local.wstd.io` 
can intersect with user-created domains. _in some cases that will cause disastrous results_

Do not allow to create domains we already use, and reserve `wstd-sys-` and `wstd_sys_` for future services we can create

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project/src/db/project-domain.ts
+++ b/packages/project/src/db/project-domain.ts
@@ -1,0 +1,79 @@
+import { customAlphabet } from "nanoid";
+import slugify from "slugify";
+
+const nanoid = customAlphabet("1234567890abcdefghijklmnopqrstuvwxyz");
+
+const slugifyOptions = { lower: true, strict: true };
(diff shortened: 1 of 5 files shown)


packages/project/src/db/project-domain.ts
```diff
@@ -0,0 +1,79 @@
...
+      return {
+        success: false,
+        error: `Domain ${domain} is reserved`,
+      } as const;
+    }
+
+    return {
+      success: true,
+      domain,
+    };
```
TrySound, line 62: Why not here and below as const?

TrySound: Ah not needed with returned type

TrySound (approved):

PR #1676, opened 2023-05-25
Replace meta.children with meta.template

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1569

Here we change template in meta to include current component instead of implicitly creating it.

This way we can provide insert composition of components instead of single one for example input+label. Or even wrap current component like label>input.

Another usecase is ability to set local styles and props for current component.

Note: review with hidden whitespaces

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
looking great

kof (approved):
looking great

kof (approved):
looking great

PR #1678, opened 2023-05-27
Search tree shift by selector

TrySound:
Here refactored horizontal shift in tree to search available placement by selector. This enables us to add restricting ancestors with requiredAncestorss and invalidAncestors options in meta (in the next PR).

The idea of this refactoring to do similar search for both left and right directions based on item selectors which means from deepest to root. So before searching I compute deepest expanded selector when all items at the bottom.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
I feel like there is new logic added that isn't covered by tests, haven't checked exactly

TrySound:
Not really new logic. The algo is splitted in two phases so the search can be done always from deepest to root.

PR #1680, opened 2023-05-28
Update inspector panels design

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1644

- added instance info to all panels
- fixed style sources design

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/combobox.tsx
+++ b/packages/design-system/src/components/combobox.tsx
@@ -95,10 +95,14 @@ export const ComboboxListboxItem = forwardRef(ListboxItemBase);
 
 export const Combobox = Popper;
 
+export const StyledPopperContent = styled(PopperContent, {
+  zIndex: theme.zIndices.max,
+});
+
 export const ComboboxContent = forwardRef(
   (props: ComponentProps<typeof PopperContent>, ref: Ref<HTMLDivElement>) => (
     <Portal>
-      <PopperContent ref={ref} {...props} />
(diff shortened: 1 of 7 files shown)


packages/design-system/src/components/combobox.tsx
```diff
@@ -95,10 +95,14 @@ export const ComboboxListboxItem = forwardRef(ListboxItemBase);
 
 export const Combobox = Popper;
 
+export const StyledPopperContent = styled(PopperContent, {
+  zIndex: theme.zIndices.max,
```
kof, line 99: It is already rendered inside of a portal, why do we need zIndex here?  I am afraid this will end up with another problem in the end, e.g. if a popover needs to be on top

TrySound: Portal does not change depth. Everything with zindex overlaps it.

kof: right, but we usually dont' set z-index on portals, the last one just wins, so they manage themselves automatically? did we end up having another portal with z-index?

TrySound: I added position: sticky for style source section which is overlapped by css inputs.

kof: then it should be just locally set, not globally in design system

kof (approved):

PR #1681, opened 2023-05-28
Sort components by order property in WsComponentMeta

kof:
## Description

We want to be able to specify the order in which components are shown in the panel.

Right now the order property is not set where needed, because I am waiting for feedback from @taylornowotny 

## Steps for reproduction

1. Open add components panel
2. See components are in the right order

## Code Review

- [ ] hi @taylornowotny  , I need you to do
  - give some thought to both components that need to be first which we set specific order to (the rest will go as is)
  - recheck the categories, I feel like our split into categories is not intentional

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

kof:
@taylornowotny slot component is completely missing

taylornowotny:
Added, thanks for the notice. We'll move Slot to the "Layout" category when we have that but we don't need it yet.

Screenshot 2023-05-30 at 1 56 18 PM

PR #1682, opened 2023-05-29
Prevent user from editing outdated builds

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1664

Here added Build.version which is updated on every patch. The client should provide latest received version to compare with stored in db. When version mismatched user is suggested to reload.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov:
IMO can be done without returning version on each update.
What if `version` is just uint counter.
On first sync client receive `version`.

Then on each patch client sends `version++`.

If server found that `currentVersion !== patchVersion - 1` it returns `resync` command

istarkov (approved):
I would use `version` as simple counter. 
In that case there is no need to return it to the client.

apps/builder/app/routes/rest.patch.ts
```diff
@@ -216,15 +235,12 @@ export const action = async ({ request }: ActionArgs) => {
...
-      data: dbBuildData,
-      where: {
-        id_projectId: { projectId, id: buildId },
-      },
-    });
-  }
+  await prisma.build.update({
+    data: dbBuildData,
+    where: {
+      id_projectId: { projectId, id: buildId },
```
istarkov, line 241: See you read serverVersion few async commands above, everything here is not inside transaction.
So race conditions are possible.
To avoid you must add version check into update command too.

```
where: {
   ....
   version: serverVersion
}
```

And gracefully catch error in case of race condition

istarkov (approved):

apps/builder/app/routes/rest.patch.ts
```diff
@@ -64,6 +70,13 @@ export const action = async ({ request }: ActionArgs) => {
     throw Error(`Build ${buildId} not found`);
   }
 
+  const serverVersion = build.version;
+  if (clientVersion !== serverVersion) {
+    return {
+      status: "version_mismatched",
```
kof, line 76: nit: why underscore separated?

apps/builder/app/builder/shared/sync/sync-server.ts
```diff
@@ -62,27 +65,53 @@ const useNewEntriesCheck = ({
...
+          const result = await response.json();
+          if (result.status === "ok") {
+            lastVersion += 1;
+            return { ok: true };
+          }
+          // when versions mismatched ask user to reload
+          // user may cancel to copy own state before reloading
+          if (result.status === "version_mismatched") {
+            const shouldReload = confirm(
+              "Someone else is edited the project. Please reload to get the latest state."
```
kof, line 93: "Project was edited in a different tab, browser or by another user. Please reload to get the latest state."

Notion's AI works great for stuff like this, I asked it to improve the text


Screenshot 2023-05-29 at 10 19 00

apps/builder/app/routes/rest.patch.ts
kof: btw maybe we should start organizing these statuses as consts, since we likely will add more in the future? and it would help prevent typos as this is not being typechecked

kof:
Screenshot 2023-05-29 at 10 30 52

got a diff after building

kof (approved):
Works nicely!

PR #1684, opened 2023-05-29
Add requiredAncestors and invalidAncestors to meta

TrySound:
Here added two component restrictions to meta.

requiredAncestors prevent inserting components outside of required ones here restricted ListItem to always be inside of List.

invalidAncestors prevent inserting components inside of invalid ones here prevented nested buttons

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/shared/instance-utils.test.ts
```diff
@@ -0,0 +1,242 @@
...
+      TextBlock: { ...base, type: "rich-text" },
+      Form: { ...base, type: "container" },
+      Box: { ...base, type: "container" },
+      Div: { ...base, type: "container" },
+      Body: { ...base, type: "container" },
+    } as const;
+    return new Map(Object.entries(configs)) as Map<string, WsComponentMeta>;
+  };
+
+  it("finds container", () => {
```
kof, line 33: we use "test()" everywhere

kof (approved):

apps/builder/app/shared/instance-utils.test.ts
TrySound: describe/it usually goes in pair. Here have 2 describe.

kof: that's fine, test() is the same as it(), having both is problematic, brings inconsistencies, you will end up having both all over the place. Ideally we need to even disable it() somehow

kof: there are some good answers, specifically this one: https://stackoverflow.com/questions/45778192/what-is-the-difference-between-it-and-test-in-jest#answer-56072272

tldr: it's just about semantics, test() is shorter ... I found in the past `it()` made certain things awkward to express in english and people don't use it properly over time, where test() doesn't enforce semantics

TrySound: Sure here used describe to group tests per utilities.

PR #1686, opened 2023-05-29
Hide props and styles tabs for html embed

TrySound:
Html embed has display: contents and should not be styled or have props.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/inspector/inspector.tsx
```diff
@@ -85,6 +85,12 @@ export const Inspector = ({ publish, navigatorLayout }: InspectorProps) => {
     );
   }
 
+  // functional components usually have display: contents
+  // and should not be styled or have props
+  const isFunctionalComponent =
```
kof, line 90: I feel like this should be not here hardcoded, but part of meta, because ultimately it depends on the fact how component behaves.

A component can potentially choose to not react to style changes for any reason.

I think we need a meta flag: disablePanels: ['styles']

TrySound: Settings probably should not be disabled.
Is there a case to have props but not styles or the opposite?

kof: right now I can't think of any use case to disable props/settings, but I feel like there will be something, potentially even some other panel, it would be easy to add a value to an enum over time, we could go with just one value for now 'styles'

kof: alternative is specific flag: `enableStylePanel` or something

kof: or more based on the component behavior: `canAcceptStyles`

kof: isPresentational or isFunctional technically works too, but sound a  bit more vague than specifically saying that the component accepts or doesn't accept styles, aka `isStylable` or just `stylable: true/false`

kof (approved):

PR #1689, opened 2023-05-30
Publish by buildId

istarkov:
## Description

The idea is to deploy by `buildId` instead of `projectId`.
Every build now can have `deployment` metadata  params (currently `domains` only, but in the future can be any additional metadata, like vars etc)

Publish is simplified (-2 workers), the same trpc as for other saas api is used to publish, see `cms.publish`
(also local development is simplified as edge-api at saas works on local dev perfectly)

isDev and isProd are not needed anymore. Any published build has `deployment` metadata, and various builds can be active now on different domains.

Build clone now seriously faster as no parse/serialize steps involved at all.

Partial index `CREATE UNIQUE INDEX "Build_dev_index" ON "Build" ("projectId") WHERE "deployment" IS NULL;` created so we could be sure that no more than 1 dev build exists.

createBuild and cloneBuild are now splitted, and seriously simplified

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/shared/db/canvas.server.ts
```diff
@@ -1,34 +1,33 @@
-import type { Project } from "@webstudio-is/project";
 import type { Data } from "@webstudio-is/react-sdk";
-import { loadBuildByProjectId } from "@webstudio-is/project-build/index.server";
-import { db as projectDb } from "@webstudio-is/project/index.server";
+import { loadById } from "@webstudio-is/project-build/index.server";
```
TrySound, line 2: loadBuildById?

apps/builder/app/routes/rest.build.$buildId.tsx
```diff
@@ -10,24 +10,23 @@ export const loader = async ({
...
 
-    const pagesCanvasData = await loadProductionCanvasData(
-      { projectId },
-      context
-    );
+    const pagesCanvasData = await loadProductionCanvasData(buildId, context);
 
-    const project = await projectDb.project.loadByParams(
-      { projectId },
+    const project = await projectDb.project.loadById(
```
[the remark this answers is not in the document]
istarkov: Already project exists `await projectDb.project.loadById` not intent of this PR to change that

packages/project-build/src/db/build.ts
```diff
@@ -130,100 +123,89 @@ const createNewPageInstances = (): Build["instances"] => {
...
-): Promise<void>;
-// eslint-disable-next-line func-style
-export async function createBuild(
-  props: {
-    projectId: Build["projectId"];
-    env: "dev" | "prod";
-    sourceBuild: Build | undefined;
   },
   _context: AppContext,
   client: Prisma.TransactionClient
```
TrySound, line 133: Context unused. Client maybe not needed? In the next function too.

istarkov: Its because of lazyness and because our security checks are not optimized, pretty sure we need to check rights inside Build. So we will use it 100% some day

apps/builder/app/shared/context.server.ts
```diff
@@ -24,15 +24,21 @@ const createAuthorizationContext = async (
...
-    domainEnv: {
-      BRANCH_NAME: env.BRANCH_NAME ?? "main",
-      PUBLISHER_TOKEN: env.PUBLISHER_TOKEN ?? "",
-      PUBLISHER_ENDPOINT: env.PUBLISHER_ENDPOINT,
+  };
+
+  return context;
+};
+
+const createCmsContext = (request: Request) => {
```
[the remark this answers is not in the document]
TrySound: Probably cc @kof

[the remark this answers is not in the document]
istarkov: Also I think we don;t need this services at all in the future ;-)
Table events would be much better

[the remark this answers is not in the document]
kof: wait wats the cms service? haha We don't technically have a cms

packages/trpc-interface/src/shared/cms.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Localhost implementation of the dashboard trpc interface
```
kof, line 2: cms name is very confusing here, dashboard isn't cms

kof: I think what you want is a "dashboard" service

istarkov: Wrong comment, see below

apps/builder/app/shared/context.server.ts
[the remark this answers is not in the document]
istarkov: renamed to `deployment`

PR #1692, opened 2023-05-30
Figcaption as a tag for Text

kof:
## Description

Closes https://github.com/webstudio-is/webstudio-builder/issues/1377

## Steps for reproduction

1. add Text
2. change tag prop to figcaption

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/components/text-block.tsx
```diff
@@ -1,10 +1,19 @@
...
 
 export const defaultTag = "div";
 
-export const TextBlock = forwardRef<
-  ElementRef<typeof defaultTag>,
-  ComponentProps<typeof defaultTag>
->((props, ref) => <div {...props} ref={ref} />);
+// We don't want to enable all tags because Box is usually a container and we have specific components for many tags.
+type Props = ComponentProps<typeof defaultTag> & {
+  tag?: "div" | "span" | "figcaption";
```
TrySound, line 12: Span is not block

kof: Yes, which is why we need to rename to Text, we already discussed it in another PR

kof: Sometimes you need a span

kof: e.g. when text is inside a label, you don't want a div https://github.com/webstudio-is/webstudio-builder/issues/1552

TrySound: Label is rich text though

packages/react-sdk/src/components/text-block.tsx
```diff
@@ -1,10 +1,19 @@
...
+};
 
-TextBlock.displayName = "TextBlock";
+export const TextBlock = forwardRef<ElementRef<typeof defaultTag>, Props>(
+  ({ tag = defaultTag, ...props }, ref) => {
+    return createElement(tag as string, { ...props, ref });
```
TrySound, line 17: Why as string?

TrySound: Seems like box works without it

kof: I think it used to not work without it, removing from both

packages/react-sdk/src/components/text-block.tsx
kof: Let's discuss in dm if its a problem, merging now

PR #1702, opened 2023-06-02
Vimeo

kof:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1688

## Description

Vimeo Component

## Steps for reproduction

1. Add vimeo
2. play with props

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/props.ts
```diff
@@ -3,7 +3,7 @@ import { computed } from "nanostores";
 import { useStore } from "@nanostores/react";
 import type { Instance, Page, Prop, Props } from "@webstudio-is/project-build";
 import type { Asset, Assets } from "@webstudio-is/asset-uploader";
-import { ReactSdkContext } from "./context";
+import { ReactSdkContext } from "@webstudio-is/react-sdk";
```
TrySound, line 6: It is react-sdk. Should not be needed.

packages/sdk-components-react/src/vimeo-preview-image.ws.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  type WsComponentMeta,
+  type WsComponentPropsMeta,
+} from "@webstudio-is/react-sdk";
+import { propsMeta as imagePropsMeta, meta as imageMeta } from "./image.ws";
+import { props } from "./__generated__/vimeo-preview-image.props";
+
+export const meta: WsComponentMeta = {
+  ...imageMeta,
+  category: undefined,
```
TrySound, line 10: Can it be omitted?

packages/sdk-components-react/src/vimeo.tsx
```diff
@@ -0,0 +1,407 @@
...
+      responsive,
+      speed,
+      title,
+      transparent,
+      interactiveParams,
+      color,
+    ]);
+
+    return (
+      <VimeoContext.Provider
```
TrySound, line 343: Context can warp only children

TrySound (approved):

taylornowotny:
Feedback:
- It works and feels really good!
- The play button being on-brand for vimeo is nice 👍
- It works with interactive elements (screenshot)
Screenshot 2023-06-04 at 10 04 38 AM

- It would feel much better if the preview image was on by default. As a user I was initially disappointed to add my URL and see a blank screen. Seeing the preview image would be a nice indication that the URL paste worked.
- Another moment that confused me is when I selected the Preview Image element. The URL for the image is there, but doesn't show until I go to the Vimeo element and switch it on. But if I upload an image then it shows right away. I understand why you think this preferable form a speed point of view, but the UX is not very smooth. This would be solved with preview on by default.
- When I play a video, the play button and preview image disappear for a moment before the video starts playing. Could we give all video players a black bg by default? If the image disappears and leaves a black screen, that feels natural. If it disappears and you see the background of the page, it feels like the video might have broken for a moment.

taylornowotny:
- I enabled Portrait, Title, and Byline but saw no effect. Are these working? I tested two videos from Vimeo's front page.
- What is "Background"? Can we make this property a little more clear from the name?

packages/sdk-components-react/src/vimeo-preview-image.ws.ts
kof: or alternatively add "none" for no category, so that's it's nicer to use?

packages/sdk-components-react/src/vimeo.tsx
kof: what do you mean?

kof:
@taylornowotny would using a better default background than what we have now make the situation a bit better? instead of semi transparent black. User would still have to replace it but maybe better?

I am super afraid of adding the api-based preview by default and make all sites slower by default

taylornowotny:
Understandable, it's a tough tradeoff.

I have some questions still:
It wouldn't make "all" sites slower, just sites with a vimeo/youtube video, right?
And this would still be faster on Webstudio than Webflow & Framer?
If preview is on by default and a user uploads an image, will the slowdown from the preview still be a factor? Could we automatically turn off the preview in this case to save the user the extra step?

Majority of users expect that video will slow down the page a little and they expect a visual indication that their video is successfully added. The preview image helps identify the video. And the type of user who wants the easiest method of adding videos will be the most reluctant to interact with the properties to turn the preview on.

On the other hand I don't think any users will want a blank preview. Those who want to optimize for speed can take the extra step of turning off the preview and uploading their own image. I predict that these users are the minority and they are willing to do a little more configuring, so I would rather ask these users to adjust the properties than vise versa.

taylornowotny:
All this being said, the consequences of this decision are minimal and I would like to gather user feedback on this eventually anyway.

kof:
> It wouldn't make "all" sites slower, just sites with a vimeo/youtube video, right?
yes

> And this would still be faster on Webstudio than Webflow & Framer?
yes

> If preview is on by default and a user uploads an image, will the slowdown from the preview still be a factor? Could we automatically turn off the preview in this case to save the user the extra step?

yes but the risk is that they are happy with automatic preview and never upload an image, we need to encourage the better pattern

kof:
>  Those who want to optimize for speed can take the extra step

This is what kills "fast by default", the thinking that someone will do the extra step to make it fast

kof:
>  I predict that these users are the minority and they are willing to do a little more configuring

Yes and that's not because they don't like fast website, it's because they don't know how much worse the experience will be. Making them make an extra step to make it fast is relying on their ability to grasp the consequence. This is a terrible way to think if you want really fast sites for everyone, no matter how well they understand.

kof:
> What is "Background"? Can we make this property a little more clear from the name?

Yep, was also thinking about renaming it to "Background Mode", sounds good?

kof:
> I enabled Portrait, Title, and Byline but saw no effect. Are these working? I tested two videos from Vimeo's front page.

They are but they are also depending on each other to work, I think maybe we need to simplify this for the user.

kof:
> When I play a video, the play button and preview image disappear for a moment before the video starts playing. Could we give all video players a black bg by default? If the image disappears and leaves a black screen, that feels natural. If it disappears and you see the background of the page, it feels like the video might have broken for a moment.

It will only feel natural if the video once loaded is dark, but if the video is light you get: click > black > white

I just did something I think is better, check it out, should be really smooth now

packages/sdk-components-react/src/vimeo-preview-image.ws.ts
kof: added category "hidden"

kof:
@taylornowotny I think after click on play button from placeholder we need to show a spinner, because it can take a lot of time on a weak connection to actually load the video and we have nothing to show as a reaction to "play clicked"

darindimitroff (commented):
LGTM overall, main comment being about not showing image preview by default. I understand the general idea we want to be super-performant, but omitting things like the video title and preview by default leads to lack of feature parity with the default Vimeo embed  and with what people are used to in most tools.

We can definitely launch like this and be fine for now, but I'm certain we will start getting feedback about this when Webstudio goes more mainstream.

kof:
@darindimitroff we can over time make the preview image from vimeo fast by caching it and not having the perf hit when using it, then we can enable it by default

taylornowotny:
Looks good except that the portrait, byline, and title seem to have no effect. 

https://github.com/webstudio-is/webstudio-builder/assets/86495896/d1575e37-baac-4399-8c1c-637378542e6f

Agree with the need for a spinner. Will work with Mark to get a universal spinner that will work on all buttons and the dashboard

taylornowotny (approved):
Looks and works well!

packages/sdk-components-react/src/vimeo-preview-image.ws.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const meta: WsComponentMeta = {
+  ...imageMeta,
+  category: "hidden",
+  label: "Preview Image",
+  requiredAncestors: ["Vimeo"],
+};
+
+export const propsMeta: WsComponentPropsMeta = {
+  props: { ...props, ...imagePropsOverrides },
```
TrySound, line 20: Props are 100% same. Just reexport image props without exposing these overrides.

packages/sdk-components-react/src/vimeo-spinner.ws.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const meta: WsComponentMeta = {
+  ...imageMeta,
+  category: "hidden",
+  label: "Spinner",
+  requiredAncestors: ["Vimeo"],
+};
+
+export const propsMeta: WsComponentPropsMeta = {
+  props: { ...props, ...imagePropsOverrides },
```
TrySound, line 20: same reuse box props

packages/sdk-components-react/src/vimeo-spinner.ws.ts
```diff
@@ -0,0 +1,22 @@
...
+} from "@webstudio-is/react-sdk";
+import {
+  propsOverrides as imagePropsOverrides,
+  propsMeta as imagePropsMeta,
+  meta as imageMeta,
+} from "./image.ws";
+import { props } from "./__generated__/vimeo-spinner.props";
+
+export const meta: WsComponentMeta = {
+  ...imageMeta,
```
TrySound, line 13: The component is not an image

packages/sdk-components-react/src/vimeo-spinner.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const VimeoSpinner = forwardRef<ElementRef<typeof defaultTag>, Props>(
+  (props, ref) => {
+    const vimeoContext = useContext(VimeoContext);
+
+    if (vimeoContext.status !== "initialized") {
+      return null;
+    }
+
+    return <Box {...props} ref={ref} />;
```
TrySound, line 20: Do we even need it to be Box?

packages/react-sdk/package.json
```diff
@@ -44,6 +44,7 @@
     "@webstudio-is/fonts": "workspace:^",
     "@webstudio-is/generate-arg-types": "workspace:^",
     "@webstudio-is/project-build": "workspace:^",
+    "colord": "^2.9.3",
```
TrySound, line 47: Should be added to sdk-components-react instead

packages/react-sdk/src/components/component-meta.ts
```diff
@@ -21,6 +21,7 @@ export const componentCategories = [
   "text",
   "media",
   "forms",
+  "hidden",
```
TrySound, line 24: Do we need it as special category? Omitted type worked well for us. All rich text child are hidden this way.

packages/sdk-components-react/src/vimeo-spinner.tsx
TrySound: Just a div would work. We don't need it to be semantic.

packages/react-sdk/src/components/component-meta.ts
kof: thought we could improve ergonomics: 

`const { category, ...linkMetaRest } = linkMeta;` this is creating a local category variable we don't actually use, it works but seems more like a workaround

packages/sdk-components-react/src/vimeo-spinner.tsx
kof: true

packages/sdk-components-react/src/vimeo-preview-image.ws.ts
kof: I was thinking that this will be a problem if someone changes the interface of that components, it would be very confusing to not see the changes and then find out its not using the actual props of that component

kof: another option to just copy/paste that bit and avoid dependency

pnpm-lock.yaml
```diff
@@ -14084,6 +14090,34 @@ packages:
     transitivePeerDependencies:
       - supports-color
 
+  /jest-cli@29.3.1:
```
TrySound, line 14093: Unrelated bump. Please regenerate lockfile.

packages/sdk-components-react/src/vimeo-spinner.ws.ts
```diff
@@ -0,0 +1,28 @@
...
+const presetStyle = {
+  div,
+} satisfies PresetStyle<"div">;
+
+export const meta: WsComponentMeta = {
+  type: "container",
+  icon: BoxIcon,
+  states: defaultStates,
+  presetStyle,
+  order: 0,
```
TrySound, line 20: Why order if hidden?

kof: forgottten

pnpm-lock.yaml
kof: happened automatically upon pnpm add

packages/sdk-components-react/src/box.tsx
```diff
@@ -5,7 +5,7 @@ import {
   type ComponentProps,
 } from "react";
 
-const defaultTag = "div";
+export const defaultTag = "div";
```
TrySound, line 8: not necessary anymore?

PR #1703, opened 2023-06-02
Rename some components

kof:
## Description

- [ ] Link Block  - remove it as we add ability to insert instances into Link
- [ ] Text Block > Text  - label and component
- [ ] Styled Text > Text - label, Span - component
- [ ] data migration

## Todo

Styled Text > Text Span and TextSpan

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Alternative is to keep the component names and only change labels - doesn't need migration

kof:
Also I was just thinking, maybe component names don't need to be equal labels, e.g. Link component and LinkBlock component are exactly the same, it should be just a reexport of a Link, on the component level there is nothing special to LinkText, it's just a Link, so maybe component name semantics should correspond to what it is, and only label would be different because that would correspond the meta config for the builder UI

PR #1708, opened 2023-06-02
Specify types with exports condition

TrySound:
Saas is swithing to moduleResolution=bundler.
Here I replaced types top level field with exoprts condition in all our packages. Thi should fix issues with new sdk components packages and simplify using additional entry points.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/package.json
+++ b/packages/asset-uploader/package.json
@@ -40,14 +40,15 @@
     ".": {
       "source": "./src/index.ts",
       "import": "./lib/index.js",
-      "require": "./lib/cjs/index.js"
+      "require": "./lib/cjs/index.js",
+      "types": "./lib/types/index.d.ts"
     },
     "./index.server": {
       "source": "./src/index.server.ts",
-      "import": "./lib/index.server.js"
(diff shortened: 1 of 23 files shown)


istarkov (approved):

packages/asset-uploader/package.json
```diff
@@ -40,14 +40,15 @@
     ".": {
       "source": "./src/index.ts",
       "import": "./lib/index.js",
-      "require": "./lib/cjs/index.js"
+      "require": "./lib/cjs/index.js",
+      "types": "./lib/types/index.d.ts"
```
Andarist, line 44: This should be moved right after the `source` condition everywhere. This currently works OK but relies on a TS bug (known as "🐛 Used fallback condition", you can learn more about it [here](https://arethetypeswrong.github.io/?p=%40webstudio-is%2Fasset-uploader%400.74.0)). 

`package.json#exports` are order-sensitive (even though they are defined using an object) and that rule should include `types` condition as well.

PR #1732, opened 2023-06-08
Experimental universal input

kof:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/props-panel/controls/text.tsx
+++ b/apps/builder/app/builder/features/props-panel/controls/text.tsx
@@ -12,6 +12,7 @@ import {
   VerticalLayout,
   HorizontalLayout,
 } from "../shared";
+import { useState } from "react";
 
 type ImplementationProps = {
   label: string;
@@ -64,7 +65,10 @@ const AsTextarea = ({
         <TextArea
           id={id}
           value={localValue.value}
-          onChange={(event) => localValue.set(event.target.value)}
+          onChange={(event) => {
+            const value = event.target.value;
+            localValue.set(value);
+          }}
           onBlur={localValue.save}
           rows={rows ?? 1}
           onKeyDown={(event) => {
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/props-panel/controls/text.tsx
```diff
@@ -1,80 +1,102 @@
...
+          return;
+        }
+        const isMultiline = rows > 1;
+        if (isMultiline === false) {
+          event.preventDefault();
+        }
+        // Insert the newline at the caret position.
+        if (event.shiftKey && isMultiline === false) {
+          const element = event.currentTarget;
+          if (element.selectionStart || element.selectionStart === 0) {
```
istarkov, line 82: selectionStart has `unsigned long` type based on spec, so this condition should be always true

apps/builder/app/builder/features/props-panel/controls/text.tsx
```diff
@@ -1,80 +1,102 @@
...
 import {
   type ControlProps,
   getLabel,
   useLocalValue,
   VerticalLayout,
   HorizontalLayout,
 } from "../shared";
+import { useState, type ComponentProps } from "react";
+
+const countLines = (value: string) => (value.match(/\n/g) || "").length + 1;
```
istarkov, line 11: to avoid +1 you can use `value.match(/^/mg).length`

istarkov (approved):

apps/builder/app/builder/features/props-panel/controls/text.tsx
kof: selectionStart is undefined if nothing is selected

kof: just checked, you are right, its always a number

taylornowotny:
Tested and it works as described for the most part.

Issues to consider, maybe separately:
- not all inputs are working. Test "Controls Color" in Vimeo properties
- when you have an input inline with a label and add new lines, then de-select and re-select that element, the multi-line input is now below the label. This is good, but would be better if it adapted right away. I'd like to talk about what we think the ideal UX is here.
- maybe adding a new line adds it at the end of the text and not before?
- maybe adding text that is larger than the container starts a new line automatically?

kof:
@taylornowotny please create a follow-up issue with those notes

P.S. color is working but expects a hex value e.g. "#ededed" need to make the control into color picker to prevent such issues

taylornowotny (approved):

kof:
>  This is good, but would be better if it adapted right away. I'd like to talk about what we think the ideal UX is here.

I tried adoption right away, felt terrible, we can talk about it again.

kof:
> maybe adding a new line adds it at the end of the text and not before?

Adding a new line is supposed to be adding it where you caret is, unless its a bug

kof:
> maybe adding text that is larger than the container adapts the size of the input to fit automatically?

Trouble is not technical, trouble is that sometimes a single-line input can be longer and it doesn't necessary mean you want to go multiline?

Maybe I am wrong on this, we can for sure make it auto-grow without having typed a newline

PR #1735, opened 2023-06-09
Allow inserting into editable containers

TrySound:
Here I removed "rich-text" type from components. Instead "container" can be edited when has only text or rich text children.

And any time user can drop into editablle container which will wrap all text and rich text children into "Span" component.

Had to implement position adjustment to consider text children which do not have representation in navigator tree and canvas dnd.

Span has temporary SpanContainer name to not conflict with existinig Span for styled text in rich text. Will rename in other PR when remove no longer necessary components.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):

kof (changes requested):
Is it expected?

1. Create a Box
2. type some text in it
3. drop in image
4. image replaces text, text is gone

packages/sdk-components-react/src/span-container.ws.tsx
```diff
@@ -0,0 +1,26 @@
...
+import type { defaultTag } from "./span-container";
+import { props } from "./__generated__/span.props";
+
+const presetStyle = {
+  span,
+} satisfies PresetStyle<typeof defaultTag>;
+
+export const meta: WsComponentMeta = {
+  type: "container",
+  label: "Span",
```
kof, line 18: Shouldn't the label be "Text"?

kof: I mean when we automatically wrap text with a span, in the navigator it probably should say "Text", same as we wanted to rename " Styled Text" to "Text" in rte when clicking on style button

kof (approved):

packages/sdk-components-react/src/span-container.ws.tsx
TrySound: Then we will have 3 text components, no?
Text which is div, text with is span, text which is rich-text-child span.

kof: Hmm, right now Text has tag property, so it's possible to have one component and render as different tags but I am not sure what's best here since we still deciding on the strategy for multi-tag components

PR #1736, opened 2023-06-09
Cli Authentication endpoint

andrasbacsai:
## Description

REST endpoint to authenticate in CLI. Uses session token from UI.

Super basic and could be improved security-wise with generated tokens/refresh tokens, but it works.

### DIFF
--- a/apps/builder/app/routes/rest.auth.cli.tsx
+++ b/apps/builder/app/routes/rest.auth.cli.tsx
@@ -1,0 +1,39 @@
+import { json, redirect, type LoaderArgs } from "@remix-run/node";
+import { sentryException } from "~/shared/sentry";
+
+export const loader = async ({ request }: LoaderArgs) => {
+  try {
+    const currentUrl = new URL(request.url);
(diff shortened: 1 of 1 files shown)


apps/builder/app/routes/rest.auth.cli.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const _session = cookies
+      ?.split(";")
+      .find((c) => c.trim().startsWith("_session="))
+      ?.split("=")[1];
+    if (!_session) {
+      return redirect(
+        "/login?returnTo=" + currentUrl.pathname + currentUrl.search
+      );
+    }
+    returnToUrl.searchParams.set("token", _session);
```
istarkov, line 24: From what I know in auth, you should never expose secrets in url used in browsers. In that case they leak through history, through some android apis etc.
I don't have enough knowledge how cli are usually do logins, but I think that OAUTH device flow is made exactly for this

istarkov (changes requested):
Should be closed imo

kof:
So there was some replacement foudn for this? closing...

PR #1737, opened 2023-06-09
Fix object fit, ignore loader

kof:
## Description

- I used object position instead of object fit by mistake
- Unrelated: ignore loader prop

## Steps for reproduction

1. add vimeo
2. select preview image
3. see it has object fit cover

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/image.ws.tsx
+++ b/packages/sdk-components-react/src/image.ws.tsx
@@ -53,5 +53,4 @@ export const propsMeta: WsComponentPropsMeta = {
     ...propsOverrides,
   },
   initialProps: ["src", "width", "height", "alt", "loading"],
-  ignoredProps: ["loader"],
 };
--- a/packages/sdk-components-react/src/image.ws.tsx
+++ b/packages/sdk-components-react/src/image.ws.tsx
@@ -37,9 +37,6 @@ export const meta: WsComponentMeta = {
   order: 0,
 };
 
-// "loader" is our internal prop not intended to show up in the props panel
-const { loader, ...publicProps } = props;
-
 // Automatically generated props don't have the right control.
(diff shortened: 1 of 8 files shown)


TrySound (approved):

packages/sdk-components-react/src/image.ws.tsx
```diff
@@ -37,9 +37,6 @@ export const meta: WsComponentMeta = {
   order: 0,
 };
 
-// "loader" is our internal prop not intended to show up in the props panel
-const { loader, ...publicProps } = props;
```
TrySound, line 41: I think we could have simpler solution. Why it is generated at all? It's prop of internal component.
```
type Props = Omit<ComponentPropsWithoutRef<typeof WebstudioImage>, 'loader'>;
```

kof: damn, true

kof: ah noo, this way you can't use the loader prop at all

kof: the problem is that it's a valid prop, one could use it if needed over code

TrySound: we don't have loader prop in props panel

kof: actually in this particular case, loader isn't a valid prop, it is an internal one, so your solution is correct here, but generally I think we will have a situation where we need ignoredProps

TrySound: Explicitly listing is hard. I'd just ignore internally on generation.

PR #1738, opened 2023-06-09
Entri

istarkov:
## Description

Implements entri service support

## Steps for reproduction

Wait for SaaS release. Or locally add following env variables to the `.env` file
```
ENTRI_APPLICATION_ID=webstudio
ENTRI_SECRET=$(bwget WS_ENTRI_SECRET)
```

image

image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,153 @@
...
+      {entriTokenData?.success === false && (
+        <Text color="destructive">{entriTokenData.error}</Text>
+      )}
+
+      {scriptLoadingError !== null && (
+        <Text color="destructive">{scriptLoadingError.message}</Text>
+      )}
+
+      <Button
+        disabled={isEntriOpen || isScriptLoading || scriptLoadingError !== null}
```
kof, line 127: isEntriOpen and isScriptLoading  could be single enum state? maybe even scriptLoadingError as well

apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,153 @@
...
+        color="neutral"
+        css={{ width: "100%", flexShrink: 0 }}
+        onClick={() => {
+          setIsEntriOpen(true);
+
+          entriTokenLoad(undefined, (data) => {
+            if (data.success) {
```
kof, line 134: would early return instead and the rest would have less indentation

apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,153 @@
...
+        onSuccess();
+      }
+    };
+
+    window.addEventListener("onEntriClose", handleOnEntriClose, false);
+
+    return () => {
+      window.removeEventListener("onEntriClose", handleOnEntriClose);
+    };
+  }, [domain, onSuccess]);
```
kof: would wrap this as part of entri wrapper, outside of react

apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+/**
+ * Our FloatingPanelPopover adds pointerEvents: "none" to the body.
+ * We open the entry dialog from the popover, so we need to allow pointer events on the entri dialog.
+ */
+const globalStyles = globalCss({
+  body: {
+    "&>#entriApp": {
```
kof, line 66: instead In would just query that element when entri renders and set it inline and do that in our sentri wrapper, outside of react, less dep on specificity or radix

packages/domain/src/trpc/domain.ts
```diff
@@ -7,7 +7,54 @@ import { createProductionBuild } from "@webstudio-is/project-build/index.server"
 
 const { router, procedure } = initTRPC.context<AppContext>().create();
 
+const EntryResponse = z.object({
+  auth_token: z.string(),
+});
+
 export const domainRouter = router({
+  getEntriToken: procedure.query(async ({ ctx }) => {
```
kof, line 15: entri specific logic should probably be separate module

packages/trpc-interface/src/context/context.server.ts
```diff
@@ -27,6 +27,14 @@ type DomainContext = {
   domainTrpc: TrpcInterfaceClient["domain"];
 };
 
+// https://developers.entri.com/docs/install
+type EntriContext = {
```
kof, line 31: could be imported from some central sentry module

apps/builder/app/builder/features/topbar/entri.tsx
istarkov: Its 3rd party API there is no way to detect when entry element will come to DOM. It can be async etc. This is good way to workaround the issue

apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,157 @@
+import useScript from "react-script-hook";
```
kof, line 1: lets move most of the logic for entri in a single package and keep here only whats specific about topbar

kof (changes requested):
lets reorg the code and put everything entri related into one folder inside domain package as well as wrap entri into js functions instead of coupling entri wrapper with react

apps/builder/app/builder/features/topbar/entri.tsx
kof: wouldn't couple it with radix still, e.g. can use react's style element directly

apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,157 @@
...
+        return;
+      }
+
+      setIsEntriOpen(false);
+
+      if (event.detail.success) {
+        onSuccess();
+        return;
+      }
+      onClose();
```
kof, line 106: onClose seems semantically unrelated from success/failure

onClose should happen in any case when something closes, success/failure can be status

apps/builder/app/builder/features/topbar/entri.tsx
istarkov: Don't get about radix, there is no radix here

kof: sorrry I meant stitches

apps/builder/app/builder/features/topbar/entri.tsx
```diff
@@ -0,0 +1,41 @@
+import { Button, Text } from "@webstudio-is/design-system";
+import {
+  useEntri,
+  entriGlobalStyles,
+  type DnsRecord,
+  type EntriCloseDetail,
+} from "~/shared/entri/entri";
```
kof, line 7: use index

apps/builder/app/shared/context.server.ts
```diff
@@ -2,6 +2,7 @@ import type { AppContext } from "@webstudio-is/trpc-interface/index.server";
 import env from "~/env/env.server";
 import { authenticator } from "~/services/auth.server";
 import { trpcClient } from "~/services/trpc.server";
+import { entryApi } from "./entri/entri-api.server";
```
kof, line 5: I thought we agreed that all entri stuff would be in one folder in one package, but it seems we now have a bunch of entri folders in various places

kof (approved):
I don't like the fact entri dependency is spread all over the place. Will have to move it myself then.

apps/builder/app/shared/context.server.ts
istarkov: Everything is in a single folder entri, whats wrong?

packages/trpc-interface/src/context/context.server.ts
```diff
@@ -27,6 +27,16 @@ type DomainContext = {
...
```
kof, line 31: entri specific, should come from entri

kof: removing entri, won't automatically make us remove this

PR #1739, opened 2023-06-09
Fix editing text inside button

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1719

Buttons has issue with unavoidable space which prevents entering spaces. Here bind lexical to temporary span inside of button. Now space works fine.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
what should we do about sites already using innerText?

kinda need a migration or keep supporting innerText

packages/prisma-client/prisma/migrations/20230610111903_button_children/migration.ts
```diff
@@ -0,0 +1,131 @@
...
+            const instancesList: InstancesList = JSON.parse(build.instances);
+            const propsList: PropsList = JSON.parse(build.props);
+            const newPropsList: PropsList = [];
+            const buttonInstanceIds = new Set<string>();
+            const buttonInnerTexts = new Map<string, string>();
+
+            for (const instance of instancesList) {
+              if (
+                instance.component === "Button" ||
+                instance.component === "VimeoPlayButton"
```
kof, line 84: btw that brings me to the question: how would we migrate someone else's components and my guess is we can't, if they upgrade component they have to redo them or the component author has to create a migration?
wdyt?

TrySound: I guess similar to what we planned with sdk. Library migrations.

kof: true, forgot about it

PR #1740, opened 2023-06-10
Css preview

kof:
## Description

This PR introduces a CSS preview

Screenshot 2023-06-10 at 03 48 55

## Steps for reproduction

1. select any instance, add some styles
2. see in navigator css preivew

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

packages/css-engine/src/core/css-engine.test.ts
```diff
@@ -408,7 +479,9 @@ describe("CssEngine", () => {
     rule.styleMap.delete("display");
     expect(engine.cssText).toMatchInlineSnapshot(`
       "@media all {
-        .c { background-image: url(foo.png) }
+        .c {
+          background-image: url(foo.png)
```
TrySound, line 483: This means our published site will have less optimized css. Maybe use method instead of getter with some "pretty" flag?

kof: yes but only spaces, which compress well over gzip, we actually need to optimize by converting to shorthands and such or by using atomic css, for now I think it's not worth adding that flag that can only remove some whitespaces

taylornowotny:
My feedback:

Looks great! In the future let's give this its own tab that will show/hide it. I will design this later.

Curious about the colors here. Which colors are used and for what purposes?

Please add a separator above the section title component so separate this from the Navigator visually.

It's not clear that the window is scroll-able. At first I missed the lower lines becuase I didn't hover this section, since there's nothing to click on I didn't naturally put my mouse here. Can we show scroll bar always in this section?

kof:
> Looks great! In the future let's give this its own tab that will show/hide it. I will design this later.

We discussed that this feature will be of HIGH value and needs to be usable in parallel to style panel and the space under navigator was your suggestion then

> Curious about the colors here. Which colors are used and for what purposes?

Colors come from one of the themes for the syntax highlighter, we can completely create our own theme, I can tell you all the positions that need a color, right now I  picked one that looked best with our colors, others were much worse.

> Please add a separator above the section title component to separate this from the Navigator visually.

- [ ] great, that was definitely missing

> It's not clear that the window is scroll-able. At first I missed the lower lines becuase I didn't see the scroll bar bc I didn't hover this section. Since there's nothing to click on I didn't naturally put my mouse here. Can we show scroll bar always in this section?

We will add a copy to buffer button to click, changing scrollbar pattern seems like suboptimal idea, skipping.

taylornowotny:
> Colors come from one of the themes for the syntax highlighter, we can completely create our own theme, I can tell you all the positions that need a color, right now I picked one that looked best with our colors, others were much worse.
Yea I don't have a problem with these colors but I would like to bring them in line with our color system. Please tell me everything that needs a color when you can.

kof:
https://github.com/webstudio-is/webstudio-builder/issues/1748

taylornowotny:
Feedback: A line can extend past the edge of the container without breaking so some of the content is not visible

PR #1746, opened 2023-06-11
New tooltips component

kof:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/style-source/style-source-badge.tsx
```diff
@@ -1,7 +1,7 @@
-import { styled, DeprecatedText2 } from "@webstudio-is/design-system";
+import { styled, Text } from "@webstudio-is/design-system";
 import { theme } from "@webstudio-is/design-system";
 
-export const StyleSourceBadge = styled(DeprecatedText2, {
+export const StyleSourceBadge = styled("div", Text, {
```
kof, line 4: no need for div

TrySound:
Maybe make max height bigger? Looks unusable with more than 1 property.
image

Also badge look weird with the font size
image

kof:
@taylornowotny things are not perfect but that's where we are right now, I think its better than before, gpt descriptions will come next

taylornowotny:
The text style is inconsistent on different tooltips. If there's just one line it should be the body (Regular text), not title (Title text).

Screenshot 2023-06-12 at 11 16 46 AM
Screenshot 2023-06-12 at 11 12 47 AM
Screenshot 2023-06-12 at 11 12 39 AM

taylornowotny:
https://github.com/webstudio-is/webstudio-builder/assets/86495896/c395ee20-1e01-4ea6-9705-027731c0a7ca

Scrolling the CSS content is not a good UX.

The CSS value is not meant to show in the property tooltip. In some cases it's harmless, in other cases like this, it leads to massive CSS in the tooltip. I suggest eliminating CSS values in tooltips as a start to solving this issue.

taylornowotny:
Tooltips to do:

- Use new tooltips for icon button and the options in option rows
- Always show breakpoint in “Value comes from” section
- Always show “Value comes from” section when a style is applied, even for “set” styles
- Add semicolon to the end of CSS values in the CSS name for correct syntax

Notes from our call:
- Title is the name of the trigger in Webstudio.
- CSS name for properties is the CSS property name + current value
- CSS name for values when applicable is the CSS property name + value (value on the trigger, not current value)
- Body is the description of the trigger. In majority of cases a description of the CSS property OR value will work here, but there are exceptions which we can fix individually later.

Design follow up after beta:
- A solution for long CSS names will be designed
- A copy-to-clipboard button will be added to the design
- A shortcut hint will be added to the “Reset” button

If you noticed anything I missed lmk

PR #1759, opened 2023-06-12
CSS data descriptions generation script

ggurgone:
## Description

Using ChatGPT, this scripts generates descriptions for CSS properties and declarations (property-value).
It uses `keywordValues` from [`src/__generated__/keyword-values.ts`](https://github.com/webstudio-is/webstudio-builder/blob/css-data-descriptions/packages/css-data/src/__generated__/keyword-values.ts) to get a list of all the properties and values.

The result is a TS module located at `./src/__generated__/property-value-descriptions.ts` which exports:
- properties
- declarations

By default it generates descriptions only for new properties/values, however you can use the `--forceRegenerate` flag to regenerate everything.

When it comes to generating descriptions for declarations whose value is a color eg. `background-color: orange` we generate a description for a generic `property: {color}` declaration (e.g. `background-color: {color}`) and GPT will generate a description template with a placeholder `{color}`.

The script then repeats the template for every color value, replacing `{color}`. This is to avoid generating a ton of similar description and wasting GPT tokens.

The script needs two env variables:
- process.env.OPENAI_KEY - Your OpenAI API key https://platform.openai.com/account/api-keys
- process.env.OPENAI_ORG - Your OpenAI Org ID https://platform.openai.com/account/org-settings

## Steps for reproduction

Run it 

```
OPENAI_KEY=your-key OPENAI_ORG=your-org pnpm build:descriptions
```

## Code Review

  - review sample descriptions

## Before merging

- [x] Get approval on sample descriptions
- [x] Generate all the descriptions


packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -0,0 +1,340 @@
...
+ * This is to avoid generating a ton of similar description and wasting GPT tokens.
+ *
+ * The script needs two env variables:
+ * - process.env.OPENAI_KEY - Your OpenAI API key https://platform.openai.com/account/api-keys
+ * - process.env.OPENAI_ORG - Your OpenAI Org ID https://platform.openai.com/account/org-settings
+ */
+
+const args = process.argv.slice(2);
+let forceRegenerate = false;
+if (args.includes("--forceRegenerate")) {
```
kof, line 30: nit: flag would be better lower case --force-regenerate

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -0,0 +1,340 @@
...
+let declarationsDescriptions: Record<string, string> = {};
+
+try {
+  ({
+    properties: propertiesDescriptions,
+    declarations: declarationsDescriptions,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Fix this else it'll complain that we cannot use top-level await.
+  } = await import(path.join(targetPath, fileName)));
+  if (!propertiesDescriptions) {
```
kof, line 48: we are having an internal rule, which may sound weird, but we are not using !before word, because in many fonts its unreadable `!i`, we use if (propertiesDescriptions === false) for example

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Properties descriptions
+ */
+const newPropertiesNames = Object.keys(keywordValues).filter(
+  (property) =>
+    forceRegenerate || typeof propertiesDescriptions[property] !== "string"
+);
+
+for (let i = 0; i < newPropertiesNames.length; ) {
```
kof, line 68: nit: we are not using single-word variables, e.g. in this case we would use `index` or even better for of

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -0,0 +1,340 @@
...
+  i += batchSize;
+}
+console.log("\n✅ Properties description generated!\n");
+
+/**
+ * Declarations descriptions
+ */
+const newDeclarationsDescriptions: Record<string, string> = {};
+const nonColorValues = [
+  "auto",
```
kof, line 138: is "none" missing in this list?

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -0,0 +1,340 @@
+/* eslint-disable no-console, func-style */
+import fetch from "isomorphic-fetch";
```
TrySound, line 2: Please use undid

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -0,0 +1,340 @@
...
+  if (!content) {
+    return [404, "No response"];
+  }
+
+  return content;
+}
+
+function grabDescriptions(message: string) {
+  const descriptions: string[] = [];
+  message.split("\n").forEach((line) => {
```
kof, line 324: maybe you even want tosplit by message.split(/\s/) just in case, I don't know if that's what you want

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -0,0 +1,406 @@
+// This file was auto-generated
+export const properties = {
```
kof, line 2: seems like there should be many more, what do I miss?

ggurgone: I only generated a few to show you an example. Will generate them all if you like the result. 

By the way I am regenerating these few in a sec and pushing the new ver.

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -0,0 +1,406 @@
...
+    "Sets the accent color to a specific antiquewhite",
+  "accentColor:aqua": "Sets the accent color to a specific aqua",
+  "accentColor:aquamarine": "Sets the accent color to a specific aquamarine",
+  "accentColor:azure": "Sets the accent color to a specific azure",
+  "accentColor:beige": "Sets the accent color to a specific beige",
+  "accentColor:bisque": "Sets the accent color to a specific bisque",
+  "accentColor:black": "Sets the accent color to a specific black",
+  "accentColor:blanchedalmond":
+    "Sets the accent color to a specific blanchedalmond",
+  "accentColor:blue": "Sets the accent color to a specific blue",
```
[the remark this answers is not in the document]
ggurgone: Here they are expanded but GPT generates only one description. Should we have only a generic without expanding and leaving webstudio builder/ui do the lookup at runtime?

```
"accentColor:{color}": "Sets the accent color to a specific {color}",
```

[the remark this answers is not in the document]
kof: this or using some wording that relies on user seeing the exact color somewhere else. User will see the color inside that tooltip and also in the style panel, so I think we could just have one generic, wihtout mentioning the color

ggurgone:
Had a chat with @taylornowotny who suggested to compile a list of selected properties-values to generate descriptions for. 

> I'll get this to you sometimes this week, but it's not necessarily part of the beta so today I need to focus on beta testing

I asked to compile this list in the same format as `keywordValues` so that once it is ready we can generate initial descriptions with GPT.

taylornowotny:
Discussed with @ggurgone 

We need to provide a manually written list of the properties and values that we want descriptions for. I will do this later in the week since this is not part of the scope of the beta.

Why do we need a manual list?
These descriptions are for tooltips. Tooltips will not be displayed for every possible value for every possible property. For example we don't need a description for every CSS color name. We also don't need descriptions for keyword values in comboboxes becuase those don't have tooltips. Only a few values will have descriptions. This list will be significantly shorter than the list of all properties with all possible values.

kof:
@taylornowotny colors will be excluded in general
comboboxes/select - we will have later a way to show a hint below the highlighted item, @ggurgone please keep everything generated

packages/css-data/bin/property-value-descriptions.ts
ggurgone: Why not standard fetch? The api is known by most devs, for example if I were to install undici I would need learn how it works

packages/css-data/src/__generated__/property-value-descriptions.ts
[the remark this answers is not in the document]
ggurgone: decided to ignore colors

packages/css-data/bin/property-value-descriptions.ts
kof: I had no idea we want to use undici, but I just had a quick glance, it seems to have massive perf improvements and has the standard fetch https://github.com/nodejs/undici#undicifetchinput-init-promise

kof (approved):

kof (approved):

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -0,0 +1,4574 @@
...
+  gridAutoRows: "Sets the height of an element",
+  gridColumnEnd: "Sets the character used to hyphenate words",
+  gridColumnStart: "Sets whether words can be hyphenated",
+  gridRowEnd: "Sets the orientation of an image",
+  gridRowStart: "Sets the algorithm used to render an image",
+  gridTemplateAreas: "Sets the resolution of an image",
+  gridTemplateColumns: "Sets the size of the first letter in a block of text",
+  gridTemplateRows: "Sets the starting position of a grid item's column",
+  hangingPunctuation: "Sets the ending position of a grid item's column",
+  height: "Sets the starting position of a grid item's row",
```
kof, line 180: well this one is a complete fail :)

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -0,0 +1,4574 @@
+// This file was auto-generated
+export const properties = {
```
kof, line 2: we are missign properties like "background", "border" etc because we are not using shortcuts in the data, but we have those as categories and need tooltips for them

ggurgone: I thought you didn't support shorthand

kof: we don't, but we show the tooltip for example on background category etc

packages/css-data/src/__generated__/property-value-descriptions.ts
ggurgone: the good thing is that we can tweak this manually. If we add properties to the source file that we use to generate the script will generate only new entries (unless you run it with --force)

kof: so we can add the values here directly?

ggurgone: technically yes

kof: alright, going to do some edits now

Issue #879, opened 2023-01-27
More style panel controls


kof:
## Todo

- [x] cursor
- [x] opacity
- [x] pointer events
- [x] Blend Mode
- [x] user select
- [ ] transition
- [ ] filters

taylornowotny:
+ Blend Mode

I am tracking the Effects sections and it will be done soon, it's just behind some other features on my to-do list like Position Section. I will update this issue with any questions or further discussion.

darindimitroff:
I think all of these would end up in the Effects section, correct?

taylornowotny:
Yes but most or all of these will be individual sections. 

See [the WIP designs in figma here](https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=4512-49938) 

<img width="275" alt="Screenshot 2023-06-23 at 8 56 41 AM" src="https://github.com/webstudio-is/webstudio-builder/assets/86495896/8a8f54dc-90bf-443b-af6e-76664b935cfd">

taylornowotny:
<img width="582" alt="Screenshot 2023-06-23 at 8 58 23 AM" src="https://github.com/webstudio-is/webstudio-builder/assets/86495896/e2988df1-00dd-4837-84fe-8c51baa9c916">

@darindimitroff notice that for properties like box-shadows, backgrounds, and Transitions, we have a pattern where the section title has an Add (+) button so the user can add multiple values for that property and arrange their order. This design uses the [CSS Value List Item component](https://www.figma.com/file/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?type=design&node-id=2002%3A6155&t=N1E9gl5gEc6NOqP1-1). 

Let me know if you have questions about this design pattern or if you think Transitions would be better with a different pattern.

kof:
closing this one, transition and filters are separate issues

PR #1760, opened 2023-06-12
Add effects section with opacity and cursor

istarkov:
## Description

ref #879

Known issues not fixed here.

Scrub on opacity doesn't work as expected
as depending on opacity dimension it has different min-max and step.

Also any numeric value of opacity is valid css value, and just clamped if it exceeds, so we can't consider values out of 0-1 range as percents too.

So couldn't be fixed fast.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
feel free to merge, adding @taylornowotny to check it out but fixes can be done later

PR #1770, opened 2023-06-13
Tooltips improvements

kof:
## Description

- Fixes a bunch of wrong logic for showing style source origin info in the tooltip
- Renders GPT descriptions

## Steps for reproduction

1. check tooltips in style panel

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/property-name.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/property-name.tsx
@@ -151,13 +151,11 @@ const TooltipContent = ({
     selectedStyleSource
   );
   const cssText = getCssText(properties, style);
-  let breakpoint = selectedBreakpoint;
-  if (styleValueInfo?.cascaded) {
-    const { breakpointId } = styleValueInfo.cascaded;
-    breakpoint = breakpoints.get(breakpointId);
-  }
-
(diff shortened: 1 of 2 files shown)


istarkov (approved):

taylornowotny:
The descriptions on property labels seem to be descriptions of the applied value, not descriptions of the property itself.
e.g. The tooltip for the "Display" property should describe the display property - NOT the display: flex value.

Screenshot 2023-06-13 at 11 33 20 AM

taylornowotny:
I warned against using chatGPT descriptions and I am going to again request that they be removed so we can define them manually.

The descriptions are consistently unhelpful, usually just describing what is already obvious about the property from the name, with no information about how to actually use the property or value.

Here are some examples. These are the rule, not the exception:

Screenshot 2023-06-13 at 11 54 00 AM
Oh really, so display: inline-block sets the display to inline-block? Very helpful.

Screenshot 2023-06-13 at 11 56 48 AM

Screenshot 2023-06-13 at 11 57 28 AM


We don't need tooltips that simply use the name of the peroperty in a sentence, we need tooltips that teach users how to use these properties to achieve their vision.

We need to write these ourselves. It's work, sure, but it's less work than writing separate docs, both for us and for the user.

PR #1773, opened 2023-06-13
Override inherited with html style

TrySound:
Html styles override inherited styles. This fixes the issue with link inheriting from body instead of showing predefined in browser style.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

istarkov:
image

Wrong color for visited shown

TrySound:
We decided to ignore visited and active because their behavior is hard to reimplement and most likely user won't see it as will redefine stateless link color.

PR #1785, opened 2023-06-14
Support shorthand property in tooltips

TrySound:
Added following shorthand property in tooltip configuration
- background
- borderWidth
- borderStyle
- borderColor
- borderRadius
- sizing

Also fixed height description.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/flex-child/flex-child.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/flex-child/flex-child.tsx
@@ -108,6 +108,7 @@ const FlexChildSectionSizing = (props: RenderCategoryProps) => {
       <PropertyName
         style={currentStyle}
         properties={["flexGrow", "flexShrink"]}
+        shorthandProperty="sizing"
         label="Sizing"
         onReset={() => {
           setSizing.deleteProperty("flexGrow");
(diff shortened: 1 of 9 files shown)


istarkov (approved):

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
```diff
@@ -120,40 +120,61 @@ const getBreakpointName = (
...
+    styleValueInfo?.nextSource?.value ??
+    styleValueInfo?.previousSource?.value ??
+    styleValueInfo?.cascaded?.value ??
+    styleValueInfo?.preset ??
+    styleValueInfo?.htmlValue ??
+    styleValueInfo?.inherited?.value
+  );
+};
+
 const getDescription = (
```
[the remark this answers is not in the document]
TrySound: Basically this is all logic `declarations[`property:value`] ?? properties['property']`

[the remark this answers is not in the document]
kof: yes, lets add that info, might be not obvious

apps/builder/app/builder/features/style-panel/sections/flex-child/flex-child.tsx
```diff
@@ -108,6 +108,7 @@ const FlexChildSectionSizing = (props: RenderCategoryProps) => {
       <PropertyName
         style={currentStyle}
         properties={["flexGrow", "flexShrink"]}
+        shorthandProperty="sizing"
```
kof, line 111: lets find a better name for shorthandProperty since sizing is not a real css shorthand, mb:
"group" or "category"?

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
```diff
@@ -120,40 +120,61 @@ const getBreakpointName = (
   return breakpoint?.minWidth ?? breakpoint?.maxWidth ?? "Base";
 };
 
+const getStyleValue = (styleValueInfo: undefined | StyleValueInfo) => {
```
kof, line 123: we need this vary same function in css preview as well as in useStyleInfo

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -140,6 +146,7 @@ export const properties = {
   flexGrow:
     "Specifies the ability of a flex item to grow to fill available space",
   flexShrink: "Specifies the ability of a flex item to shrink if necessary",
+  sizing: "Specifies the ability of a flex item to grow or shrink",
```
kof, line 149: maybe we should call it flexChildSizing or something to be clear in this global context

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -1962,6 +1977,9 @@ export const declarations = {
   "flexShrink:inherit":
     "Inherits the amount of space a flex item will shrink from its parent element",
   "flexShrink:unset": "Sets the ability for a flex item to shrink if necessary",
+  "sizing:0,0": "Flex item doesn't grow or shrink",
+  "sizing:1,0": "Flex item grows to fill available space",
+  "sizing:0,1": "Flex item shrinks if necessary",
```
kof, line 1982: same

packages/css-data/src/__generated__/property-value-descriptions.ts
kof: alternative is to keep this in a separate place because these are non-standard? by adding them here we are making no distinction if the properties here are correct css properties or there is some mistake, so we can't validate them

could be a separate file that exports just the cust custom things, then merges them upon export into one for the consumer

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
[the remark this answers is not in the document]
kof: after a long discussion with Taylor we need change this logic, DMd you

PR #1807, opened 2023-06-16
Tweak prompt and regenerate properties descriptions

### LINKED ISSUE #1793: Issues with Tooltips descriptions
I warned against using chatGPT descriptions and I am going to again request that they be removed so we can define them manually.

The descriptions are consistently unhelpful, at best just describing what is already obvious about the property from the name, with no information about how to actually use the property or value.

Here are some examples. These are the rule, not the exception:
image
image
image

We don't need tooltips that simply use the name of the property in a sentence, we need tooltips that teach users how to use these controls to achieve their vision.

We need to write these ourselves. It's work, sure, but it's less work than writing separate docs, both for us and for the user. I'd rather have no description than an AI description.

ggurgone:
Fixes #1793

## Description

I tweaked the prompt and generated new descriptions. 

@kof @taylornowotny The results are still not ideal but please take a look and decide whether we should merge these changes or not (probably still better than the previous generation).

PS: I tweaked the script a bit so now it creates: `propertiesOverrides` and `propertiesGenerated` (same for property:value descriptions). This way we can add overrides to `propertiesOverrides`


kof:
Can we organize the prompt in a way that is easy to copy it out and test it in chat interface? Basically probably using a template string without the code?

ggurgone:
@kof moved them to .md files

kof (approved):

PR #1810, opened 2023-06-17
Ignore complex props

TrySound:
Ref https://github.com/webstudio-is/webstudio/discussions/37

There are props like object and function we cannot express with UI. Even more there is style prop which if casted to string gives runtime error. So the best way would be to ignore such types.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/__generated__/vimeo.props.ts
```diff
@@ -2,10 +2,8 @@ import type { PropMeta } from "@webstudio-is/generate-arg-types";
 
 export const props: Record<string, PropMeta> = {
   slot: { required: false, control: "text", type: "string" },
-  style: { required: false, control: "text", type: "string" },
   title: { required: false, control: "text", type: "string" },
   defaultChecked: { required: false, control: "boolean", type: "boolean" },
-  defaultValue: { required: false, control: "text", type: "string" },
```
kof, line 8: defaultValue shouldn't be removed?

packages/sdk-components-react/src/__generated__/vimeo.props.ts
```diff
@@ -19,7 +17,6 @@ export const props: Record<string, PropMeta> = {
   accessKey: { required: false, control: "text", type: "string" },
   autoFocus: { required: false, control: "boolean", type: "boolean" },
   className: { required: false, control: "text", type: "string" },
-  contentEditable: { required: false, control: "text", type: "string" },
```
kof, line 22: not really necessary but isn't a complex type, so I think this indicates mistake in the logic

packages/sdk-components-react/src/__generated__/vimeo.props.ts
```diff
@@ -41,7 +38,6 @@ export const props: Record<string, PropMeta> = {
   about: { required: false, control: "text", type: "string" },
   content: { required: false, control: "text", type: "string" },
   datatype: { required: false, control: "text", type: "string" },
-  inlist: { required: false, control: "text", type: "string" },
```
kof, line 44: also seems type string

kof (approved):
check the things that this change got removed from props, seems like some of them are not complex object types, its also strange that style type doesn't work as a string, because you can totally set it as a string `document.body.style = ''color: red'`

packages/sdk-components-react/src/__generated__/vimeo.props.ts
TrySound: It's "Booleanish" type

packages/sdk-components-react/src/__generated__/vimeo.props.ts
TrySound: Seems like should be defined by component as html one is very generic.

packages/sdk-components-react/src/__generated__/vimeo.props.ts
kof: can we detect such types from the data?

TrySound:
Not strange. We generate not dom but react props.

packages/sdk-components-react/src/__generated__/vimeo.props.ts
kof: Yeah, I wonder if our inference is inaccurate or the types are actually inaccurate from the source

TrySound: Not sure why defaultValue is defined on all components even div
```
        defaultValue?: string | number | ReadonlyArray<string> | undefined;
```

packages/sdk-components-react/src/__generated__/vimeo.props.ts
TrySound: any in types

packages/sdk-components-react/src/__generated__/vimeo.props.ts
TrySound: The question is which type we should prefer
```
boolean | 'true' | 'false'
```

kof: form that choice this is clearly a boolean

packages/sdk-components-react/src/__generated__/vimeo.props.ts
kof: I can't even find an attribute inlist anywhere in the spec, how did we get there, how can we get rid of it and maybe some similar cases?

kof:
> Not strange. We generate not dom but react props.

Right, we basically have a fundamental mismatch, because not all attribute names map exactly 1:1 to react props. This is a fundamental problem and given our cross-framework direction for the future we need to think how we can make sure the html native attributes always map correctly to whatever framework is doing

kof:
Another such missmatch is "class". We need a way to solve this for all frameworks, any ideas?

packages/sdk-components-react/src/__generated__/vimeo.props.ts
TrySound: Here yes but
```
'aria-checked'?: boolean | 'false' | 'mixed' | 'true' | undefined;
'aria-current'?: boolean | 'false' | 'true' | 'page' | 'step' | 'location' | 'date' | 'time' | undefined;
```

packages/sdk-components-react/src/__generated__/vimeo.props.ts
TrySound: React.HTMLAttributes

TrySound:
There is no universal way if we allow user components. Component types is the only source of props we can have.

kof:
Right, but in our case we extending the html attributes in the type, maybe this is a bad approach since not all html attributes can be passed from react to the dom

packages/sdk-components-react/src/__generated__/vimeo.props.ts
kof: super weird, I can't find it in the spec at all

kof: we are already using ComponentProps<'div'> so we are probably already relying on React.HTMLAttributes

TrySound: Yes, this is what's inside of React.HTMLAttributes

TrySound:
1. we do not pass from react to dom
2. HTMLAttributes is React specific type, it matches perfectly
3. The issue was in casting everything including style object to text control

So, no casting - no issue

kof (approved):

PR #1812, opened 2023-06-17
Batch pending transactions

TrySound:
When user looses connection we rerun jobs when retry succeeds. Though retry creates multiple requests which may hang when there is too many.

To avoid this here I removed jobs queue and schedule single job which executes all pending and scheduled transactions as an alternative too queue.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
good job, one thing to think about: what happens if we aggregate multiple types of updates in one job and then one of them fails, will all of them fail? maybe something for the future

kof:
Probably unrelated, but I was looking into payloads and saw that now a single change of the "width" results in 3 changes in one transaction:

- styleSourceSelections - seems like its not updating to anything new, just the same style source selection, seems like it shouldn't be sent
- styleSources - no patches, shouldn't be there for sure
- styles - the only valid patch

Also wondering why we sent revisePatches to the backend, I thought we only need them on the frontend where the change happened

TrySound:
Yeah, we can optimize this more. In separate PR. Revise patches are sent to keep the format of transactions same as in collaboration.

istarkov:
The whole code looks like race condition.
We have exported scheduleJob, we have exported executeJob and nothing inside 
`executeJob` prevents it to be runned in parallel, no guards nothing.

istarkov:
I think failing test could be:
```
let firstJobDone = false;
scheduleJob(async () => {
sleep(1000);
firstJobDone=true;
});
scheduleJob(async () => {
// no sleep
if (firstJobDone === false) {
throw new Error()
}
})
excuteJob()
```

istarkov:
I also don't get why 2 methods? schedule and execute, for me this has no sense. 
Imo retry etc must be inside naked js code without React, useEffects etc. Current is unreadable and untestable imo

kof:
> Imo retry etc must be inside naked js code without React, useEffects etc. Current is unreadable and untestable imo

I agree and I think we need to generally write more plain functions and just wrap them into react specific code.

A very good pattern for effects is `const fn = () => { return () => {}}` where this function returns a cleanup function if needed.

Super portable pattern.

PR #1820, opened 2023-06-19
Fix ssr error

istarkov:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-vars/src/index.ts
+++ b/packages/css-vars/src/index.ts
@@ -1,14 +1,12 @@
-let counter = -1;
-
-const unique = <Name extends string>(name: Name) => {
-  return `${name}-${++counter}` as const;
-};
-
-const define = <Name extends string>(name: Name, unique = false) => {
-  if (unique) {
-    return `--${name}` as const;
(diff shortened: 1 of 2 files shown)


packages/css-vars/src/index.ts
```diff
@@ -1,14 +1,12 @@
-let counter = -1;
-
-const unique = <Name extends string>(name: Name) => {
-  return `${name}-${++counter}` as const;
-};
-
-const define = <Name extends string>(name: Name, unique = false) => {
-  if (unique) {
-    return `--${name}` as const;
+const WEBSTUDIO_CSS_VARIABLE_POSTFIX = "-webstudio-variable";
```
kof, line 1: css-vars package was so far generic, not webstudio specific, I would keep it that way as long as possible, one way is to just use genearted hash, or maybe some generic suffix or define the suffix from application layer ....

TrySound: By package name
```
-webstudio-css-vars
```

kof (approved):

PR #1823, opened 2023-06-19
Forms playground

### LINKED ISSUE #1806: Form method and action are not configurable; sends pre-defined email directly to account's email
Form properties `method` and `action` are not available in the properties panel.

All form submissions send to the WebStudio user's registered email address from a webstudio.email address that seems to be generated corresponding to the Page

WebFlow exposes `method` and `action` so it seems like a fair requirement to at least keep parity with their functionality on forms given how important they are. Also, perhaps a major selling point if the submissions allotments are greater/unmetered.

Marketers and low/no-coders will hard require these settings be exposed as they will have existing crm/smtp providers and not be easily willing to switch. Might get away with integrations + key setting for most popular.

Devs will also require configuration of the form submission event/data. A webhook setting if nothing else.

istarkov:
## Description

Closes #1806

Most changes in this PR is ability to test it locally with saas described here https://dev.to/istarkov/local-development-with-public-and-private-monorepos-3554

Documentation how to test saas is added to SaaS

`action` and `method` now available for adding
`hidden inputs` are now cleared by prefix from formData (to not send ws inputs anywhere)

`action` and `method` are now read at server using `getFormProperties`

## Steps for reproduction

Create form, add action and method, see form has hidden inputs 

image

## Code Review

- [x] hi @TrySound , I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
Code looks good, haven't tested

PR #1826, opened 2023-06-20
Support data source insert from template

TrySound:

- added data sources store
- added dataSource type to props
- insert data sources from template (prop.dataSourceReference allows to bind several props to the same data source)
- convert to static prop on copying (simplification until requirements are clear)

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/copy-paste/plugin-instance.ts
+++ b/apps/builder/app/shared/copy-paste/plugin-instance.ts
@@ -22,6 +22,7 @@ import {
   assetsStore,
   projectStore,
   registeredComponentMetasStore,
+  dataSourcesStore,
 } from "../nano-states";
 import {
   type InstanceSelector,
@@ -132,7 +133,8 @@ const getAssetsUsedInProps = (props: Prop[], foundAssetsIds = new Set()) => {
       prop.type === "string" ||
       prop.type === "boolean" ||
       prop.type === "page" ||
-      prop.type === "string[]"
+      prop.type === "string[]" ||
+      prop.type === "dataSource"
     ) {
       continue;
     }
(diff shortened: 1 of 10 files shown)


apps/builder/app/shared/copy-paste/plugin-instance.ts
```diff
@@ -159,9 +161,36 @@ const getTreeData = (targetInstanceSelector: InstanceSelector) => {
...
+      if (dataSource?.type === "boolean") {
+        return {
+          id: prop.id,
+          instanceId: prop.instanceId,
+          name: prop.name,
+          type: dataSource.type,
+          value: dataSource.defaultValue,
+        } satisfies Prop;
+      }
+      // ensure all data source types are mapped to props
```
kof, line 189: these two look nearly identical, I think you might want to avoid typing these out since there will be more types and work this around with typescript in some different way

packages/react-sdk/src/embed-template.ts
```diff
@@ -104,10 +110,34 @@ const createInstancesFromTemplate = (
...
+          if (dataSource === undefined) {
+            const id = nanoid();
+            const name = prop.dataSourceReference;
+            if (prop.type === "boolean") {
+              const { type, value: defaultValue } = prop;
+              dataSource = { id, name, type, defaultValue };
+            } else if (prop.type === "string") {
+              const { type, value: defaultValue } = prop;
+              dataSource = { id, name, type, defaultValue };
+            } else {
```
kof, line 128: same here regarding same code, assuming you wrote twice to satisfy ts but there might be another way so we don't end up with 10 of those later

kof (approved):

apps/builder/app/shared/copy-paste/plugin-instance.ts
TrySound: Typescript does not allow to put them in one condition because type and defaultValue become two unions so we get multiplication of 4 types

cc @istarkov

istarkov: The issue is that https://www.typescriptlang.org/play?#code/C4TwDgpgBAIghsOBlA9gVwE4GNoF4oDeokAXFAEQDOwGAlgHYDm5ANFAG5wA2aEZ1dJgF8oAH0LE+FemgC2AIwgZWHbrzIyFSoQG4AUHqwp61KKlkQAYmnpYo+ABQATBHDLxEqTDgCU9gHyEelAhUBgQwJj0QaGxUJJkLogAdJIswXEhnDxSSXDJ2bwZIUJ6ukA

Gives all  combination of type and value
image

Checking how to avoid

istarkov: That rare case when spread works like a charm
https://www.typescriptlang.org/play?#code/C4TwDgpgBAIghsOBlA9gVwE4GNoF4oDeokAXFAEQDOwGAlgHYDm5ANFAG5wA2aEZ1dJgF8oAH0LE+FemgC2AIwgZWHbrzIyFSoQG4AUJKgAFDCjBR8E8FKo0GzNpx5SB9tgAsIXLin53hYlak0nKKyo5qUpphHl4+foKMunp6WCj01FCoshAAYmj0WBZQABQAJghwZPCIqJg4AJQWAHyEelAdUGkZwFAYxYSdQ1Dtwx0AdJMViCyjY57evlAA5ACMy7NDyUMYEMCY9H36ukA

PR #1831, opened 2023-06-21
Add property tooltip to icon controls

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1798

Here added property tooltip to toggle button controls and menu controls. Tooltip is triggered only by hover because click on controls is functional.

image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/shared/property-name.tsx
```diff
@@ -215,6 +206,45 @@ const TooltipContent = ({
...
+  onReset?: undefined | (() => void);
+  children: ReactElement;
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  return (
+    <Tooltip
+      open={isOpen}
+      onOpenChange={setIsOpen}
+      onClick={() => setIsOpen(false)}
+      triggerProps={openWithClick ? { onClick: () => setIsOpen(true) } : {}}
```
kof, line 232: - not sure why there is both onClick and triggerProps.onClick
- is there a particular reason to not pass onClick with triggerProps always and then do nothing is openWithClick is false? it would read more straight forward

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
```diff
@@ -215,6 +206,45 @@ const TooltipContent = ({
...
+  style: StyleInfo;
+  onReset?: undefined | (() => void);
+  children: ReactElement;
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  return (
+    <Tooltip
+      open={isOpen}
+      onOpenChange={setIsOpen}
+      onClick={() => setIsOpen(false)}
```
kof, line 231: do I understand correctly that you want to close the tooltip when user clicks on it?

kof:
- [ ] clicked dropdown item then tooltip appeared https://share.descript.com/view/79pJEbMaoJo

kof:
- [ ] can't click reset any more because it closes the tooltip first ... those handlers you removed in there where handling it

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
TrySound: I wanted to isolated open state in tooltip to not add in every `<PropertyTooltip>` usage.

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
TrySound: Yes, but all click handlers on content will work

TrySound:
Hah, works in safari and when opened with hover in chrome

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
kof: I am not sure that clicking on a tooltip should close the tooltip at all ... did @taylornowotny describe this somewhere?

taylornowotny: Is there a reason why closing the tooltip on click improves the user experience?

A "see more" button has been tentatively planned for long CSS content, but we can go with some other pattern if there's a reason to.

taylornowotny: Other than the "see more" button I can't think of a reason to close or not close tooltips on click

taylornowotny:
- [ ] Clicking the tooltip activates an icon button (not toggle button)

https://github.com/webstudio-is/webstudio-builder/assets/86495896/4c21ca64-4efe-43ba-82a5-9f64e317ecb4

- [ ] Made a new issue for remaining tooltips #1832

taylornowotny:
Screenshot 2023-06-21 at 9 48 19 AM

taylornowotny (changes requested):
see comments

apps/builder/app/builder/features/style-panel/shared/property-name.tsx
kof: Radix closes tooltips by default on click because they are not interactive. Our tooltips have interactive elements, so if user misclicks a link or button tooltip would close.

It feels like when tooltip is interactive it shouldn't close when you click somewhere in the content that is not a button ...

taylornowotny:
That makes sense and I agree. Our tooltips should not close on click then.

TrySound:
Looks like there is a conflict between dropdown and tooltip in radix implementation. Both cannot co-exist on the same element or be nested. In the end we get either tooltip click triggers dropdown or dropdown select triggers tooltip.

kof:
@TrySound I have registered pointer event before, checked if option is pressed, if it is - preventDefault ... this logic already existed and worked perfectly fine

kof (approved):

taylornowotny:
Looks and works great now 👍

taylornowotny (approved):

PR #1838, opened 2023-06-23
Fix url hash

### LINKED ISSUE #1828: Hash links not escaped
When hash links contain spaces they don't work, example https://webstudio.is/blog/design-tokens-better-than-classes all the links in TOC don't work except of the first one

istarkov:
## Description

closes #1828 

## Steps for reproduction

Create links to the page with section with spaces, see its escaped properly

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react-remix/src/shared/remix-link.tsx
+++ b/packages/sdk-components-react-remix/src/shared/remix-link.tsx
@@ -23,9 +23,14 @@ export const wrapLinkComponent = (BaseLink: typeof Link) => {
 
     if (href?.type === "page") {
       let to = href.page.path === "" ? "/" : href.page.path;
+      const urlTo = new URL(to, "https://any-valid.url");
+      to = urlTo.pathname;
+
       if (href.hash !== undefined) {
-        to += `#${href.hash}`;
+        urlTo.hash = href.hash;
+        to = `${urlTo.pathname}${urlTo.hash}`;
       }
+
       return <RemixLink {...props} to={to} ref={ref} />;
     }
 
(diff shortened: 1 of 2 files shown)


packages/sdk-components-react-remix/src/shared/remix-link.tsx
```diff
@@ -23,9 +23,14 @@ export const wrapLinkComponent = (BaseLink: typeof Link) => {
 
     if (href?.type === "page") {
       let to = href.page.path === "" ? "/" : href.page.path;
+      const urlTo = new URL(to, "https://any-valid.url");
```
kof, line 26: any chance this could throw? why not simply encodeURIComponent()

kof (approved):

packages/sdk-components-react-remix/src/shared/remix-link.tsx
istarkov: Because almost no one can cook encodeURI vs encodeURIComponent, with URL we don't need even think about what's is right percent encoding

istarkov: Should not throw because of 2nd parameter

packages/sdk-components-react/src/link.tsx
```diff
@@ -20,9 +20,15 @@ export const Link = forwardRef<HTMLAnchorElement, Props>((props, ref) => {
...
+        url = href.page.path === "" ? "/" : href.page.path;
+        const urlTo = new URL(url, "https://any-valid.url");
+        url = urlTo.pathname;
+
+        if (href.hash !== undefined) {
+          urlTo.hash = href.hash;
+          url = `${urlTo.pathname}${urlTo.hash}`;
+        }
       }
       break;
```
TrySound, line 33: put break inside of block and format. Case block formatted nicely.

PR #1849, opened 2023-06-23
Support updating data source values

TrySound:

Added data source values store which can be updated from both components and props panel.

```js
const { setState } = useContext(ReactSdkContext)
const instanceId = getInstanceIdFromComponentProps(props)

setState(instanceId, propName, value)
```

## Code Review

  - test nothing is broken in props panel on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -1,7 +1,7 @@
 import { useMemo, useEffect } from "react";
 import { useStore } from "@nanostores/react";
 import { computed } from "nanostores";
-import type { Instances, Page } from "@webstudio-is/project-build";
+import type { DataSource, Instances, Page } from "@webstudio-is/project-build";
 import {
   type Params,
   type Components,
@@ -32,6 +32,8 @@ import {
   selectedPageStore,
   registerComponentMetas,
   registerComponentPropsMetas,
+  dataSourceValuesStore,
+  dataSourcesStore,
 } from "~/shared/nano-states";
 import { useDragAndDrop } from "./shared/use-drag-drop";
 import { useCopyPaste } from "~/shared/copy-paste";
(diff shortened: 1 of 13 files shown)


packages/react-sdk/src/context.tsx
```diff
@@ -35,11 +36,21 @@ export const ReactSdkContext = createContext<
     propsByInstanceIdStore: ReadableAtom<PropsByInstanceId>;
     assetsStore: ReadableAtom<Assets>;
     pagesStore: ReadableAtom<Pages>;
+    dataSourceValuesStore: ReadableAtom<Map<DataSource["id"], unknown>>;
+    setState: (
```
kof, line 40: should it be just setState? it may be better when more specific? e.g. setInstanceState?

TrySound: setDataSourceValue?

kof (approved):

packages/react-sdk/src/context.tsx
kof: yeah, that's good

apps/builder/app/canvas/canvas.tsx
```diff
@@ -102,6 +104,24 @@ const useElementsTree = (components: Components, params: Params) => {
...
+            newValues.set(dataSourceId, dataSource.value);
+          }
+          for (const [dataSourceId, dataSourceValue] of dataSourceValues) {
+            newValues.set(dataSourceId, dataSourceValue);
+          }
+          return newValues;
+        }
+      ),
+      onDataSourceUpdate: (dataSourceId, value) => {
+        const dataSourceValues = new Map(dataSourceValuesStore.get());
```
istarkov, line 121: Looks like its only to force recompute? To emulate immutable data structure? Mb use immer instead?

istarkov (approved):

apps/builder/app/canvas/canvas.tsx
TrySound: No big deal to use immer here

PR #1855, opened 2023-06-25
Migrate form to data sources

TrySound:

Instead of SuccessMessage and ErrorMessage components we can use Box with bound state from form.

For now this is only boolean state which should be set on Form component side. Soon we explore using expression as data sources to change only enum value instead of booleans.

Changes in template
- ErrorMessage and SuccessMessage are replaced with Box with "show" bound to expression
- ErrorMessage and SuccessMessage have text in own children
- the rest of Form content is wrapped with Box bound to expression
- Form got state prop bound to variable
- Form initialState prop is deleted

Migration just bind show props of all from direct children to data sources. Without additional wrapping.


packages/sdk-components-react-remix/src/form.ws.tsx
```diff
@@ -30,59 +32,103 @@ export const meta: WsComponentMeta = {
...
+              component: "Button",
+              children: [{ type: "text", value: "Submit" }],
             },
           ],
         },
+
         {
           type: "instance",
-          component: "ErrorMessage",
+          label: "SuccessMessage",
```
kof, line 96: any reason to use camel case instead  of 2 capitalized words?

packages/sdk-components-react-remix/src/form.ws.tsx
```diff
@@ -30,59 +32,103 @@ export const meta: WsComponentMeta = {
...
+              value: false,
+              dataSourceRef: {
+                type: "expression",
+                name: "formSuccess",
+                code: `formState === 'success'`,
+              },
+            },
+          ],
           children: [
+            { type: "text", value: "Thank you for getting in touch!" },
```
kof, line 111: seeems like this value is not rendered

packages/sdk-components-react-remix/src/form.ws.tsx
```diff
@@ -30,59 +32,103 @@ export const meta: WsComponentMeta = {
...
+              type: "boolean",
+              value: false,
+              dataSourceRef: {
+                type: "expression",
+                name: "formError",
+                code: `formState === 'error'`,
+              },
             },
           ],
+          children: [{ type: "text", value: "Sorry, something went wrong." }],
```
kof, line 131: same with this one, doesn't render the value

packages/sdk-components-react-remix/src/form.ws.tsx
TrySound: I guess no just mimicking previous names

kof (approved):

kof:
Tested locally, seems to work, it didn't wrap contents of the form into a box with this migration but I think this is what you expected
Screenshot 2023-07-02 at 12 53 02

istarkov (approved):

PR #1862, opened 2023-06-26
Migrate to storybook 7

TrySound:
See https://storybook.js.org/docs/react/migration-guide and https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700

- all configs are now ts/esm
- had to add hack (process.env.NODE_ENV) to workaround storybook bug
- webpack is gone
- renamed storybook:run to storybook:dev
- amount of packages reduced from 2456 to 2060


istarkov (approved):

kof (approved):

.github/workflows/chromatic.yml
```diff
@@ -87,6 +88,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 2 # we need to fetch at least parent commit to satisfy Chromatic
+          ref: ${{ github.event.pull_request.head.sha || github.sha }} # HEAD commit instead of merge commit
```
Andarist, line 91: I was using this kind of override in the past but lately, I reverted that decision in my projects. While the default behavior of using the merge commit is surprising at first, it actually might catch more pre-merge problems. YMMV, ofc.

apps/builder/.storybook/main.ts
```diff
@@ -0,0 +1,29 @@
...
+      ...config,
+      define: {
+        ...config.define,
+        // storybook use "util" package internally which is bundled with stories
+        // and gives and error that process is undefined
+        "process.env.NODE_DEBUG": "undefined",
+      },
+      resolve: {
+        ...config.resolve,
+        conditions: ["source", "import", "module", "browser", "default"],
```
Andarist, line 19: According to the docs you only have to include additional conditions here:
https://github.com/vitejs/vite/blob/f900acdf265b7be9b52b99f04c101e6cf62b51b4/docs/config/shared-options.md?plain=1#L148

.github/workflows/chromatic.yml
kof: we are using lostpixel for visual regression tests, chromatic is only storybook, no tests, but the same question is probably for lostpixel config

PR #1863, opened 2023-06-26
Rest endpoint to load by projectId

andrasbacsai:
## Description

Add REST endpoint to get the latest buildId based on projectId:
`/rest/buildId/<projectId>`

Should return buildId or null (which means the project is not published yet).

It is required for the new CLI.

## Steps for reproduction

1. click button
3. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/routes/rest.project.$projectId.tsx
```diff
@@ -0,0 +1,58 @@
...
+    if (projectId === undefined) {
+      throw json("Required project id", { status: 400 });
+    }
+
+    const context = await createContext(request);
+
+    const project = await projectDb.project.loadById(projectId, context);
+    const buildId = project.latestBuild?.buildId ?? undefined;
+
+    if (buildId === undefined) {
```
istarkov, line 29: And since now we have same logic as rest/build endpoint
https://github.com/webstudio-is/webstudio-builder/blob/381b5ae480e9c7ce564d33b8e7a4b36130870c33/apps/builder/app/routes/rest.build.$buildId.tsx#L18
And will always have, plz reuse

istarkov (changes requested):
plz reuse same code

apps/builder/app/routes/rest.project.$projectId.tsx
andrasbacsai: How can I reuse the same code if params.buildId does not exist on this endpoint? The only param is the projectId.

istarkov: Since you read buildId its same code

andrasbacsai: It does not make sense to reuse the code for me as most of the functions are already shared. The sequence is different in this endpoint.

istarkov: Nothing different, existing endppoint is function of buildId, all you need to read buildId here and use same logic as existing in case if we will need to add anything this will be in a single place

andrasbacsai: Based on our discussion, updated the endpoint.

istarkov (approved):
Endpoint name looks strange, we dont return project, plz change

PR #1864, opened 2023-06-26
Fix deprecated storybook args

TrySound:
These stories break on upgrade.

See
- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-controloptions
- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#no-longer-inferring-default-values-of-args

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/button.stories.tsx
+++ b/packages/design-system/src/components/button.stories.tsx
@@ -116,25 +116,19 @@ export const Button = ({
 );
 
 Button.argTypes = {
-  children: { defaultValue: "Button", control: "text" },
-  color: {
-    defaultValue: "primary",
-    control: { type: "inline-radio", options: colors },
-  },
-  prefix: {
-    defaultValue: "undefined",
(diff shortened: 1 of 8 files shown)


packages/design-system/src/components/button.stories.tsx
```diff
@@ -116,25 +116,19 @@ export const Button = ({
 );
 
 Button.argTypes = {
-  children: { defaultValue: "Button", control: "text" },
```
kof, line 119: are we not relying on defaultValue?

kof: ah its down below

kof (approved):

packages/design-system/src/components/button.stories.tsx
TrySound: It's set with Button.args

PR #1871, opened 2023-06-28
Add chevron + new hover animation to main menu

darindimitroff:
https://github.com/webstudio-is/webstudio-builder/assets/16368791/62ade843-659c-453d-8a32-bf7f95f4b860

**This PR:**
- implements the proposed change for adding a chevron to the main menu for better discoverability
- replaces the current hover animation with a subtle vertical transform
- removes some of the styling logic related to the previous design

Might need a little bit of cleanup if we decide to go this way.

**Design considerations**
Personally, I'm a fan of the new design. It feels more discoverable and the new animation isn't as jarring. The only downside I can see if I decide to play devil's advocate is the bounding box of the button on hover isn't square anymore, so it doesn't align vertically with the left bar. IMO this type of alignment isn't mandatory and, while it might look kind of nice, it doesn't look worse with the update, especially given we don't have visible bounds when not hovering.

### DIFF
--- a/apps/builder/app/builder/features/topbar/menu/menu-button.tsx
+++ b/apps/builder/app/builder/features/topbar/menu/menu-button.tsx
@@ -5,7 +5,7 @@ import {
   rawTheme,
   toggleItemStyle,
 } from "@webstudio-is/design-system";
-import { HamburgerMenuIcon, WebstudioIcon } from "@webstudio-is/icons";
+import { ChevronDownIcon, WebstudioIcon } from "@webstudio-is/icons";
 
 const size = rawTheme.spacing[9];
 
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/topbar/menu/menu-button.tsx
```diff
@@ -5,47 +5,42 @@ import {
...
+    transformOrigin: "bottom center",
+    transition: "all 150ms ease",
+    color: "$foregroundTextMoreSubtle",
+  },
+  "span:nth-child(1)": {
+    marginRight: "2px",
+  },
+  "&:hover span:nth-child(2), &[data-state=open] span:nth-child(2)": {
+    transform: "translateY(2px)",
+    color: "$foregroundContrastMain",
```
kof, line 26: please use the theme directly, gives us type safety


```

color: theme.colors.foregroundContrastMain
```

kof (approved):
looks and feels great, just one comment to address

kof:
Closing this because not finished

PR #1874, opened 2023-06-28
Add expression utilities

TrySound:

Here added two utilities:

validateExpression to parse syntax and replace identifier. For parsing used jsep package which is very limited subset of js.

executeExpression to execute the chain of expressions. The code is unoptimized and will not cache results between calls. But still should be fast enough because expressions are indented to be quite simple.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/expression.test.ts
+++ b/packages/react-sdk/src/expression.test.ts
@@ -1,0 +1,98 @@
+import { expect, test } from "@jest/globals";
+import { executeExpression, validateExpression } from "./expression";
+
+test("allow literals and array expressions", () => {
+  expect(
+    validateExpression(`["", '', 0, true, false, null, undefined]`)
+  ).toEqual(`["", '', 0, true, false, null, undefined]`);
+});
+
+test("allow unary and binary expressions", () => {
+  expect(validateExpression(`[-1, 1 + 1]`)).toEqual(`[-1, 1 + 1]`);
+});
+
+test("forbid member expressions", () => {
+  expect(() => {
+    validateExpression("var1 + obj.param");
+  }).toThrowError(/Cannot access "param" of "obj"/);
+});
+
+test("forbid this expressions", () => {
+  expect(() => {
+    validateExpression("var1 + this");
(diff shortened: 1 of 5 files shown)


packages/react-sdk/src/expression.test.ts
```diff
@@ -0,0 +1,98 @@
+import { expect, test } from "@jest/globals";
+import { executeExpression, validateExpression } from "./expression";
+
+test("allow literals and array expressions", () => {
```
kof, line 4: can we put that into describe('expression')?

TrySound: Jest already groups by module name. This will just add additional nesting without any other benefit.

kof (approved):

packages/react-sdk/src/expression.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  expect(
+    executeExpression("exp1", "$ws$dataSource$", variables, expressions)
+  ).toEqual(2);
+});
+
+test("execute expression dependent on variables", () => {
+  const variables = new Map([["var1", 5]]);
+  const expressions = new Map([["exp1", "$ws$dataSource$var1 + 1"]]);
+  expect(
+    executeExpression("exp1", "$ws$dataSource$", variables, expressions)
```
istarkov, line 65: Any explanation why we need `variablePrefix`
Why not 
```
 const expressions = new Map([["$ws$dataSource$exp1", "1 + 1"]]);
```
Hard to read 
```
  const expressions = new Map([
    ["exp1", "$ws$dataSource$exp0 + 1"],
    ["exp0", "$ws$dataSource$var1 + 2"],
  ]);
  expect(
    executeExpression("exp1", "$ws$dataSource$", variables, expressions)
```

istarkov: Somehow I should get from above that $ws$dataSource$exp0 is ($ws$dataSource$var1 + 2)

Issue #1858, opened 2023-06-26
Position section design


taylornowotny:
https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=5774-74974

Talk with me if there's any questions!

PR #1879, opened 2023-06-29
Position section

istarkov:
## Description

ref #1858

Position section without "relative to element" (will do in next PR):

image

Changed label fonts on fonts from design. IMO non monospaced fonts for numeric dynamic labels is disaster.
Here on 3-4 symbols difference in font length could be serious amouunt pf pixels like '5'

Made label margins like in design 2px instead of 1. 

Simlified label logic to use css only solution.

Having that a lot of code is shared with spacing secion some bugs can be in both

Next PR:
image

## Steps for reproduction

Goto spacing and position section, play with params.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
> Changed label fonts on fonts from design. IMO non monospaced fonts for numeric dynamic labels is disaster.

Is the new font used non-monospaced? Was this problem discussed?

istarkov:
Yep, wrote in discord

kof:
- [ ] position when set is showed with ellipsis - I bet it was either unexpected in design or something is wrong with implementation, we def. need to get rid of ellipsis there

kof:
Question: is this look expected @taylornowotny ?

Screenshot 2023-07-09 at 16 33 07

istarkov:
Label has ellipsis for cases where it doesnt fit. I wrote in discord that current design with changes on 1-2px cause unpredictable effects, especially having that some css/html rendering a little bit different from figma

istarkov:
Label has ellipsis for cases where it doesnt fit. I wrote in discord that current design with changes on 1-2px cause unpredictable effects, especially having that some css/html rendering a little bit different from figma

kof:
- [ ] probably out of scope here, but still it would be much nicer if we could start dragging scrub on "Auto" or anything else when there is a unit value possible, and it would just switch to a number and start changing it just like you did it with position. It's nice when position says Auto but I don't need to set 0 before I can start using scrub

kof:
- [ ] If user selects static and changes top/left/right/bottom and doesn't see any effect they will be confused. Same for other values that don't allow for offset positioning. What did we define for this in figma?

kof (changes requested):
tested, haven't read code yet, todos in comments

kof:
@taylornowotny sounds like design needs adjustments

istarkov:
> Question: is this look expected @taylornowotny ?
> 
> Screenshot 2023-07-09 at 16 33 07

image

The issue that font has different width's fo symbols, and for example for 111 it would fit, but internally I use width for `333` otherwise during scrub it starts jumps from 2 line to 3 line
We need control to be at least 10-20px wider to fit at least {xxx}px

istarkov:
> * [ ]  If user selects static and changes top/left/right/bottom and doesn't see any effect they will be confused. Same for other values that don't allow for offset positioning. What did we define for this in figma?

In design initial appearance is 
image

The issue and why I did like now is that zIndex can be set on static elements, but by design zIndex is going with position section.

kof:
@istarkov so what needs to be done?

istarkov:
https://discord.com/channels/955905230107738152/1048308525673238558/1127655984416948354

apps/builder/app/builder/features/style-panel/sections/shared/keyboard.ts
```diff
@@ -1,6 +1,7 @@
 import { useState, useRef } from "react";
-import type { FocusEvent, KeyboardEvent } from "react";
-import type { SpaceStyleProperty } from "./types";
+import type { FocusEvent, FocusEventHandler, KeyboardEvent } from "react";
+import { useDebouncedCallback } from "use-debounce";
+// import type { SpaceStyleProperty } from "./types";
```
TrySound, line 4: remove comment?

TrySound (approved):

kof (approved):

PR #1882, opened 2023-06-30
Support data source expressions

TrySound:

Here introduced two types of data sources into "variable" and "expression". First stores the value, the second only executable code.

Execute expressions now output results of all expressions so we could access values globally with "computed" store

Embed template references now allow to define both variable reference and expression.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/copy-paste/plugin-instance.ts
+++ b/apps/builder/app/shared/copy-paste/plugin-instance.ts
@@ -22,7 +22,7 @@ import {
   assetsStore,
   projectStore,
   registeredComponentMetasStore,
-  dataSourcesStore,
+  dataSourceValuesStore,
 } from "../nano-states";
 import {
   type InstanceSelector,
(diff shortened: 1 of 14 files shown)


apps/builder/app/shared/copy-paste/plugin-instance.ts
```diff
@@ -143,6 +143,22 @@ const getAssetsUsedInProps = (props: Prop[], foundAssetsIds = new Set()) => {
   return foundAssetsIds;
 };
 
+const getPropTypeAndValue = (value: unknown) => {
```
kof, line 146: why do we have this function twice?

kof (approved):

apps/builder/app/shared/copy-paste/plugin-instance.ts
TrySound: It is temporary. Will be gone once we add ui stuff etc

istarkov (approved):

PR #1888, opened 2023-07-02
Override browser default tag based styles that use EM units

kof:
## Description

EM seems to be challenging to use and people tend to use PX or REM instead ... I changed those values that come with EM by default from browsers to the values Webflow is using.

I am not sure why webflow is doing certain things: e.g. paragraphs have marginTop 0, h1-3 have marginTop 20; marginBottom: 10, h4- have marginTop 10 ; marginBottom 20

These seem to be inconsistent and weird choices.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/css/normalize.ts
+++ b/packages/react-sdk/src/css/normalize.ts
@@ -76,7 +76,7 @@ export const li = baseStyle;
 export const ul = baseStyle;
 export const ol = baseStyle;
 
-export const p = [...baseStyle, ...presets.margins];
+export const p = [...baseStyle, ...presets.verticalMargins];
 export const span = baseStyle;
 
 // @todo for now not applied to html, as we don't have html element
--- a/packages/react-sdk/src/css/normalize.ts
+++ b/packages/react-sdk/src/css/normalize.ts
@@ -16,7 +16,7 @@
  */
 
 // webstudio custom opinionated presets
-import { borders, outline } from "./presets";
+import * as presets from "./presets";
 import type { EmbedTemplateStyleDecl } from "../embed-template";
 
 export type Styles = EmbedTemplateStyleDecl[];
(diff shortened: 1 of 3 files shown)


istarkov (commented):
It doesn't look like normalize now. Its set of opinionated styles has nothing with initial behavior of normalize - makes browsers render all elements more consistently

istarkov:
I think if we do changes which are different from notmalize.css we need to do them in other place

kof:
> It doesn't look like normalize now. Its set of opinionated styles has nothing with initial behavior of normalize - makes browsers render all elements more consistently

yes which is why the opinionated styles are in a separate file, but we use them from normalize ... should probably  merge them together in a different way instead of using reset directly from normalize, but that's not new here .. this was done before ... may need to change this @TrySound

taylornowotny:
One issue - the font and height properties have an orange label with no Value source in the tooltip. Orange labels represent remote user-added values so they should always show the value source in the tooltip. But these orange labels show on a new project with no user styles. These properties should be gray labels (Webstudio default) instead.

Screenshot 2023-07-03 at 9 07 06 AM
Screenshot 2023-07-03 at 9 09 44 AM

kof:
After a discussion we decided to not have margins by default at all

packages/react-sdk/src/css/presets.ts
```diff
@@ -25,3 +25,102 @@ export const outline: Styles = [
...
+    value: { type: "keyword", value: "solid" },
+  },
+  {
+    property: "borderLeftColor",
+    value: { type: "rgb", r: 226, g: 226, b: 226, alpha: 1 },
+  },
+] satisfies Styles;
+
+export const h1 = [
+  ...margins,
```
TrySound, line 81: headers have only top and bottom margins set. Do we want to bloat styles with left right too?

packages/react-sdk/src/css/presets.ts
```diff
@@ -25,3 +25,102 @@ export const outline: Styles = [
...
+    value: { type: "unit", value: 0, unit: "number" },
+  },
+  {
+    property: "marginRight",
+    value: { type: "unit", value: 0, unit: "number" },
+  },
+  {
+    property: "marginBottom",
+    value: { type: "unit", value: 0, unit: "px" },
+  },
```
TrySound, line 41: why number and px?

TrySound:
> this was done before ... may need to change this @TrySound

let's do it later

packages/react-sdk/src/css/presets.ts
kof: typo, please change to number, I am off right now

TrySound (approved):

packages/react-sdk/src/css/normalize.ts
```diff
@@ -37,13 +37,21 @@ const boxSizing = {
...
 export const div = baseStyle;
 export const address = baseStyle;
 export const article = baseStyle;
 export const aside = baseStyle;
-export const figure = baseStyle;
+export const blockquote = [
+  ...baseStyle,
+  ...presets.blockquote,
+] satisfies Styles;
+export const figure = [...baseStyle, ...presets.margins] satisfies Styles;
```
kof, line 54: might not need margins, but only vertical margins, need to check browser defaults there

PR #1894, opened 2023-07-03
Support generating expressions code

TrySound:
Here added another expression utility to generate function body generateExpressionsComputation.

InstanceRoot now requires computeExpressions(variables) prop which can be generated build time.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/expression.ts
+++ b/packages/react-sdk/src/expression.ts
@@ -119,7 +119,7 @@ const sortTopologically = (
 ) => {
   for (const id of list) {
     if (explored.has(id)) {
-      return sorted;
+      continue;
     }
     explored.add(id);
     const deps = depsById.get(id);
--- a/packages/react-sdk/src/expression.ts
+++ b/packages/react-sdk/src/expression.ts
@@ -136,7 +136,7 @@ const sortTopologically = (
  * and outputing map of results
  */
 export const generateExpressionsComputation = (
-  variables: Map<string, unknown>,
+  variables: Set<string>,
   expressions: Map<string, string>
 ) => {
   const depsById = new Map<string, Set<string>>();
@@ -163,7 +163,7 @@ export const generateExpressionsComputation = (
   // generate code comoputing all expressions
   let generatedCode = "";
 
-  for (const id of variables.keys()) {
+  for (const id of variables) {
     generatedCode += `const ${id} = _variables.get('${id}');\n`;
   }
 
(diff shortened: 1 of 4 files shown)


istarkov:
I would test it now with verdaccio, then will approve/decline

packages/react-sdk/src/expression.ts
```diff
@@ -141,25 +145,37 @@ export const executeExpressions = (
     }
   );
 
-  // execute chain of expressions
```
istarkov, line 144: sortedExpressions  above seem like not have transitivity rule for sorting i.e.
if a<b and b<c then a<c
seems like doesn't work for algotithms above.
I think that transitivity required for most sorting algorithms. Not sure 100% but pretty sure that its something every fast algorithm will rely on.

packages/react-sdk/src/expression.ts
```diff
@@ -111,15 +111,42 @@ export const validateExpression = (
...
-  variables: Map<string, unknown>,
+const sortTopologically = (
+  list: Set<string>,
+  depsById: Map<string, Set<string>>,
+  explored = new Set<string>(),
+  sorted: string[] = []
+) => {
+  for (const id of list) {
+    if (explored.has(id)) {
+      return sorted;
```
istarkov, line 122: return looks suspicios, are you sure not continue here?
May be add test for this function?

istarkov, line 122: return looks suspicios, are you sure not continue here?
May be add test for this function?

packages/react-sdk/src/expression.ts
```diff
@@ -111,15 +111,42 @@ export const validateExpression = (
...
+    }
+    explored.add(id);
+    const deps = depsById.get(id);
+    if (deps) {
+      sortTopologically(deps, depsById, explored, sorted);
+    }
+    sorted.push(id);
+  }
+  return sorted;
+};
```
istarkov, line 132:
```
/*
0 -> 1 -> 2 -> 3
1 -> 3
2 -> 4
*/

console.log(
  sortTopologically(
    new Set(["3", "0", "2", "1", "4"]),
    new Map([
      ["3", new Set(["2", "1"])],
      ["1", new Set(["0"])],
      ["2", new Set(["1"])],
      ["4", new Set(["2"])],
    ])
  )
);
```
gives wrong result

istarkov (approved):
Tested, see latest saas PR commit. Looks fine

PR #1908, opened 2023-07-06
Support effectful expressions

TrySound:
Here added two new utilities for generation and executing individual expressions. Such expressions can use assignment operator to change any variable.

This give us less opinionated solution for "set" action. We can just execute any code our parser supports.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/expression.test.ts
+++ b/packages/react-sdk/src/expression.test.ts
@@ -170,7 +170,7 @@ test("make sure dependency exists", () => {
   const variables = new Map();
   const expressions = new Map([["exp1", "var1 + 1"]]);
   expect(() => {
-    executeComputingExpressions(variables, expressions);
+    executeComputingExpressions(expressions, variables);
   }).toThrowError(/Unknown dependency "var1"/);
 });
 
--- a/packages/react-sdk/src/expression.test.ts
+++ b/packages/react-sdk/src/expression.test.ts
@@ -2,8 +2,10 @@ import { expect, test } from "@jest/globals";
 import {
   decodeDataSourceVariable,
   encodeDataSourceVariable,
-  executeExpressions,
-  generateExpressionsComputation,
+  executeComputingExpressions,
+  executeEffectfulExpression,
+  generateComputingExpressions,
+  generateEffectfulExpression,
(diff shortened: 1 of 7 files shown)


packages/react-sdk/src/expression.test.ts
```diff
@@ -145,3 +183,59 @@ test("encode/decode variable names", () => {
...
 });
+
+test("generate effectful expression", () => {
+  expect(
+    generateEffectfulExpression(`var0 = var0 + var1`, new Set(["var0", "var1"]))
+  ).toMatchInlineSnapshot(`
+    "let var0 = _variables.get('var0');
+    let var1 = _variables.get('var1');
+    var0 = var0 + var1;
+    return new Map([
```
kof, line 194: just wondering why is this returning a map instead of array ['var0', var0]

packages/react-sdk/src/expression.test.ts
```diff
@@ -145,3 +183,59 @@ test("encode/decode variable names", () => {
...
+      ['var0', var0],
+    ]);"
+  `);
+});
+
+test("forbid not allowed variables in effectful expression", () => {
+  expect(() => {
+    generateEffectfulExpression(`var0 = var0 + var1`, new Set(["var0"]));
+  }).toThrowError(/Unknown dependency \"var1\"/);
```
kof, line 230: Ha, is github right?

kof (approved):

packages/react-sdk/src/expression.test.ts
TrySound: 🤷

packages/react-sdk/src/expression.test.ts
TrySound: Because map has more tools

kof: kinda asking because having less stuff in that generated code would potentially make it easier to change over time? maybe not, I don't know.

TrySound: Just a universal api. Map for input an map for output.

PR #1909, opened 2023-07-07
(feat): Add ability to edit name of a tree node

### LINKED ISSUE #222: Add an ability to edit labels in the navigator tree
Users should be able to edit labels of components in the navigator tree.

(SEE 6/28 UPDATE BELOW)

JayaKrishnaNamburu:
## Description

Users should be able to set the name of any node inside the tree-view. 

## Steps for reproduction

1. Select any node from the tree
2. Double click the node
3. Node enters into the edit mode.
4. Enter should save the name for the node.
5. The name should reflect everywhere in the application.

Things to take care of

- [x] The node should get into edit mode when you `double` click the tree node.
- [x] The node should enter edit mode when you click `ctrl+r`
- [x] The keyboard shortcuts like, `up`, `down` expand and other tree operations should work as expected.
- [x] `Esc` or loose of focus should set the name back.

closes #222 

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


JayaKrishnaNamburu:
https://github.com/webstudio-is/webstudio-builder/assets/11075561/d11b921c-4a75-4e15-a34c-574a87d09adc

JayaKrishnaNamburu:
Gave it a shot, tested the scenarios.

kof:
- [ ] if I type a very long name and hit enter, I only see "..." after reload it fixes itself, same long string over settings panel works
Screenshot 2023-07-08 at 09 49 24

kof:
@taylornowotny @JayaKrishnaNamburu @istarkov 

I just realized why I never was able to reload figma in the browser - cmd+r was preventing reload and was editing the element name.

I think this is kinda important not to lose. 

When you hit enter in figma when navigator or canvas element is focused - they go into text editing of the element on canvas in both cases
When you hit enter in webstudio when navigator is focused - nothing is happening

What if instead of cmd+r/ctrl+r we just use "enter" when navigator element is focused to edit name there and use enter when canvas element is focused to edit text as it is right now?

kof (changes requested):
tested, some problems found @TrySound can review the code

taylornowotny:
I checked figma and for them "Enter" moves the section from the current item to its children. This is kinda nice but I've never used it and I don't see it as essential.

I only have one concern with using "Enter" and that's how focus is managed in the style panel. If I want to change the name after using the style panel, I have to unfocus the style panel. The only way to do that is with a click - ESC does not remove focus. This can lead to moments where focus stays too long on my last used control so when I hit Enter to change the name it activates the control instead. The click to unfocus cannot be on the canvas or the workspace or else it changes my selection away from the instance that I wanted to rename.

If we use ESC to unfocus then problem solved. It's awkward without.

kof:
Even if you unfocus style panel control with Esc, who says the focus would
automatically go to the navigator item? I think that would be quite magical.


If we wanted to make it easy to focus between top, left, center and right
panels, we could make a special shortcut for this, so after clicking on
canvas, it would take 2 presses to start editing instance

1. E.g. shortcut to focus navigator item
2. Enter

taylornowotny:
Focus is not needed to rename. ENTER will rename the current selected item.

The only issue, as I described, is that ENTER will affect the focused control in the style panel and it's hard to remove that focus. When nothing is focused, ENTER will rename the current item. So there just needs to be a way to remove focus with the keyboard and ESC is my suggestion.

Yes to the need for a shortcut to move focus between panels but that doesn't affect this issue.

kof:
Focusing navigator by hitting escape is not logical to me.

There is no such thing as actively unfocusing something. You can actively
only focus something else.

taylornowotny:
> Focus is not needed to rename. ENTER will rename the current selected item.

You do not need to focus the Navigator to rename something...

kof:
Seems like we need a call for this.

taylornowotny:
https://share.descript.com/view/lvxtWOUGJxa

kof:
@JayaKrishnaNamburu me and @taylornowotny had a call, what came out is that we just need a different shortcut that doesn't override browser default for cmd+r and we found that cmd/ctrl+e is a good replacement.

This gives us ability to have a quick way to rename instances from keyboard without interfering with very basic browser behavior.

"Enter" is already used for editing text on canvas and can stay that way

apps/builder/app/builder/shared/tree/index.tsx
```diff
@@ -59,21 +65,39 @@ export const InstanceTree = (
     [instances]
   );
 
+  const updateInstanceLabel = useCallback((instanceId: string, val: string) => {
```
kof, line 68: please use full names
```suggestion
  const updateInstanceLabel = useCallback((instanceId: string, value: string) => {
```

apps/builder/app/builder/shared/tree/index.tsx
```diff
@@ -92,3 +120,48 @@ export const getInstanceLabel = (
...
+      {children}
+    </EditableTreeItemLabel>
+  );
+};
+
+const EditableTreeItemLabel = styled(TreeItemLabel, {
+  variants: {
+    isEditing: {
+      true: {
+        background: "white",
```
kof, line 156:
```suggestion
        background: theme.colors.backgroundControls,
```

apps/builder/app/builder/shared/tree/use-editable.tsx
```diff
@@ -0,0 +1,65 @@
+import {
+  useRef,
+  useEffect,
+  type KeyboardEventHandler,
+  type KeyboardEvent,
+  type FocusEvent,
+} from "react";
+
+export const useEditable = ({
```
kof, line 9: I just realized you used content editable .... why risking a lot of bugs instead of just using a plain input?

So much that can happen in editable, its a nightmare

apps/builder/app/builder/shared/tree/use-editable.tsx
```diff
@@ -0,0 +1,65 @@
...
+      requestAnimationFrame(() => {
+        element.focus();
+        getSelection()?.selectAllChildren(element);
+        lastValueRef.current = getValue();
+      });
+      return;
+    }
+
+    element.removeAttribute("contenteditable");
+  }, [isEditing, element]);
```
kof, line 37: there is no need to have ref in a dependency, linter allows and its meant to not be needed, you can just use elementRef directly inside the effect, just don't create a const element variable

apps/builder/app/shared/nano-states/instances.ts
```diff
@@ -2,13 +2,18 @@ import { atom, computed } from "nanostores";
...
 import { useSyncInitializeOnce } from "../hook-utils";
+import type { ItemId } from "node_modules/@webstudio-is/design-system/src/components/tree/item-utils";
 
 export const isResizingCanvasStore = atom(false);
 
 export const selectedInstanceSelectorStore = atom<undefined | InstanceSelector>(
   undefined
 );
 
+export const editInstanceNameSelectorStore = atom<undefined | ItemId>(
```
kof, line 13: why is it called selector? When we say selector in instance case, its an array that represents a tree path. in this case its just an id, no?

packages/design-system/src/components/tree/tree-node.tsx
```diff
@@ -356,20 +363,31 @@ export const TreeItemBody = <Data extends { id: string }>({
...
+  HTMLDivElement,
+  {
+    children: React.ReactNode;
+    prefix?: React.ReactNode;
+  }
+> = ({ children, prefix, ...restProps }, ref) => {
+  return (
+    <>
+      {prefix}
+      <DeprecatedText2
```
kof, line 376: could be migrated to the new Text component

packages/design-system/src/components/tree/tree.stories.tsx
```diff
@@ -84,6 +84,10 @@ export const StressTest = () => {
   return (
     <Flex css={{ width: 300, height: 500, flexDirection: "column" }}>
       <Tree
+        isEditingItemName={undefined}
+        onEditItemName={() => {
+          return;
```
kof, line 89: could log the name in storybook console

kof: e.g. 

```
Basic.args = {
  onChange: action("onChange"),
};

```

packages/design-system/src/components/tree/tree.tsx
```diff
@@ -44,6 +44,8 @@ export type TreeProps<Data extends { id: string }> = {
     dropTarget: { itemSelector: ItemSelector; position: number | "end" };
   }) => void;
   onCancel: () => void;
+  isEditingItemName: ItemId | undefined;
```
kof, line 47: confusing naming, you are passing an "id", but call property a "name", also isSomething is always boolean

suggestion: `editingItemId`

packages/design-system/src/components/tree/tree.tsx
```diff
@@ -44,6 +44,8 @@ export type TreeProps<Data extends { id: string }> = {
...
+  onEditItemName: (itemId: ItemId) => void;
```
kof, line 48: semantics of this function are not clear, this is from what I can see starting the editing mode:
`onItemEditingStart`

apps/builder/app/shared/nano-states/instances.ts
JayaKrishnaNamburu: I was trying to get used to the state management stuff. I kinda copied the above behaviour 😅. Will update the names 👍

apps/builder/app/builder/shared/tree/use-editable.tsx
JayaKrishnaNamburu: I had the same idea first, but then i looked at how tokens are being edited in the right side panel with contenteditable. And thought that's the preferred way. And changed it to use `contenteditable` 😅

kof: actually lets check with @TrySound maybe that's indeed better? I remember using contenteditable there because there was some issue, maybe because you can't use input inside a button and I didn't want change the rendering? 

I don't remember ...

apps/builder/app/builder/shared/tree/index.tsx
```diff
@@ -59,21 +65,42 @@ export const InstanceTree = (
...
         return <></>;
       }
+      const label = getInstanceLabel(props.itemData, meta);
       return (
         <TreeItemBody {...props} selectionEvent="focus">
-          <TreeItemLabel prefix={<MetaIcon icon={meta.icon} />}>
-            {getInstanceLabel(props.itemData, meta)}
-          </TreeItemLabel>
+          <TreeItem
+            isEditing={props.itemData.id === editingItemId}
```
TrySound, line 91: Why you expose this state to whole builder? Can be local

apps/builder/app/builder/shared/tree/use-editable.tsx
TrySound: I guess it's fine

apps/builder/app/builder/shared/tree/index.tsx
kof: @TrySound you propose passing editingItemId instead as a prop?

TrySound: I propose to maintain "editing" state inside TreeItem

packages/design-system/src/components/tree/tree.stories.tsx
```diff
@@ -84,6 +85,11 @@ export const StressTest = () => {
   return (
     <Flex css={{ width: 300, height: 500, flexDirection: "column" }}>
       <Tree
+        editingItemId={undefined}
+        onItemEditingStart={() => {
+          action("onItemEditingStart");
+          return;
```
TrySound, line 91: No need return

apps/builder/app/builder/shared/tree/index.tsx
TrySound: See the issue is that both react and react-hotkeys rely on document events. So they cannot prevent each other. You can workaround this by listening keydown on element itself in effect and stopping propagation. In long term having as less as possible of global states reduce complexity of the app so it's important to not fall into "keep everything in stores".

JayaKrishnaNamburu: Yes, for `hotkeys` it's fine. But the root `event` listener is intercepting all the `keyDown` events on the tree node when the item does into edit mode. So, moving the `edit` state into local creates the issues with the root listener.

kof: I see basically we use that to stop listening for those other events. Maybe the solution is to add a check in those events to not react when key is pressed in an input?

JayaKrishnaNamburu: Yeah, that could be one solution. We can try that. By checking the event origin and the stop reacting to those. I will give it a try 👍

packages/design-system/src/components/tree/tree.tsx
```diff
@@ -442,6 +456,10 @@ const useKeyboardNavigation = <Data extends { id: string }>({
     rootRef,
     handleKeyDown,
     handleClick(event: React.MouseEvent<Element>) {
+      if (editingItemId) {
```
TrySound, line 459: You can use `event.detail === 1` and `event.detail === 2` to distinct click and double click.

TrySound: ah, right. It's to prevent click while editing.
Can be bypassed to TreeItemBody. You can use preventDefault() in your handler and for example check event.defaultPrevented to avoid calling selection

apps/builder/app/builder/shared/tree/index.tsx
TrySound: space is not intercepted here, space triggers click event

JayaKrishnaNamburu: I might be looking in the wrong place. But, the `root` is handling the `" "` too for expanding the node here.
https://github.com/webstudio-is/webstudio-builder/blob/main/packages/design-system/src/components/tree/tree.tsx#L369

packages/design-system/src/components/tree/tree.tsx
JayaKrishnaNamburu: Did that, didn't help much in some edge cases 😅. Instead it have new issues. `Space` doesn't add space  in `contenteditable` with `preventDefault` As it's prevent default behaviour 😬.  Let me change with `Input` and see if that solves most of these cases 👍

apps/builder/app/builder/shared/tree/index.tsx
TrySound: Right, I missed that. Well, we anyway need to prevent these events while typing. Because labels may have spaces.

kof: Hmm shouldn't you be able to catch those events on the parent of a content editable and prevent propagation?

Though even in this case, its still kinda bad abstraction because it would have to know which events to stop propagate.

Its kinda better to let that other thing know we are editing text ... so it seems a prop or a store is unavoidable

JayaKrishnaNamburu: Yeah i kinda tried multiple variations of the same. And ended up coming to this. I tried stopping propagation too. But, some events are still bubbling as the `onKeyDown` is being triggered on the `child` node and `root` node at the same time :X

kof: yeah, so either prop like you did or a store that can be imported from multiple places @TrySound what's better?

JayaKrishnaNamburu:
So, i wrapped most of the comments and need to implement the `cmd + E` shortcut. But i am stuck on one particular issue when a  node to has a really long name in it. 
The text won't re-centre when the `contenteditable` is removed from it. I tried a lof of options. But turns out the `Stylesource` too have the same issue with long names when the tag goes into edit mode. I tried adopting the same `useForceRecalStyle` hook. But, the edge case is kinda very tricky. Any ideas on this one 🤔 

https://github.com/webstudio-is/webstudio-builder/assets/11075561/c1dc4b5f-f4c4-4b42-8898-13181090c340

kof:
@JayaKrishnaNamburu fixed it https://github.com/webstudio-is/webstudio-builder/pull/1931

JayaKrishnaNamburu:
> @JayaKrishnaNamburu fixed it #1931 and made it reusable for your case, please don't copy/paste these hacks

Yup, for sure. Just wanted to create a re-producable env in the PR 😄. Will revert the changes 👍

packages/design-system/src/components/tree/item-utils.ts
```diff
@@ -50,3 +51,26 @@ export const areItemSelectorsEqual = (
   }
   return left.join(",") === right.join(",");
 };
+
+export const useForceRecalStyle = <Element extends HTMLElement>(
```
kof, line 55: huh why is it back? I thought we wanted to delete this?

kof (approved):
please address final comments and merge

PR #1913, opened 2023-07-08
Refactor sync server with async iterators

TrySound:
Here the logic is following. We iterate over transactions polling and every set of transactions retry until request is complete.

The code becomes more straightforward and solves
https://github.com/webstudio-is/webstudio-builder/pull/1812 for free because nothing is scheduled until requests are complete and transactions are queued in array.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/shared/sync/sync-server.ts
```diff
@@ -1,87 +1,92 @@
...
-  useEffect(() => {
-    if (authPermit === "view") {
-      return;
+// polling is important to queue new transactions independently
+// from async iterator and batch them into single job
+const pollTransactions = async function* () {
+  while (true) {
+    const transactions = sync();
+    if (transactions.length > 0) {
+      queueStatus.set("running");
```
istarkov, line 36: Isn't that be better to put into retry as well
```
queueStatus.set("running");
queueStatus.set("idle");
```

So all statuses would be in a single place?
Anyway i have no strong opinion, looks better than setInterval logic.

istarkov (approved):
Linear logic is 10x better than setInterval

apps/builder/app/builder/shared/sync/sync-server.ts
TrySound: Not quite we could put the other logic before or after invkoking retry. Here is clear lifecycle.

PR #1914, opened 2023-07-09
Support effectful expressions in component actions

TrySound:
Here added support for callbacks or event handlers on components which basically can do only assignment to variables for now with extended (with assignment operator) syntax of expressions.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/props-panel/controls/combined.tsx
+++ b/apps/builder/app/builder/features/props-panel/controls/combined.tsx
@@ -19,6 +19,7 @@ export const renderControl = ({
     throw Error("Data source is not resolved");
   }
 
+  // @todo remove once ui for action is implemented
   if (prop?.type === "action") {
     return;
   }
--- a/apps/builder/app/builder/features/props-panel/controls/combined.tsx
+++ b/apps/builder/app/builder/features/props-panel/controls/combined.tsx
@@ -19,6 +19,10 @@ export const renderControl = ({
     throw Error("Data source is not resolved");
   }
 
+  if (prop?.type === "action") {
+    return;
+  }
+
   if (
     meta.control === "text" &&
     (prop === undefined || prop.type === "string")
--- a/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
+++ b/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
@@ -183,12 +183,19 @@ export const WebstudioComponentDev = ({
     onClick: (event: MouseEvent) => {
       event.preventDefault();
       if (event.currentTarget instanceof HTMLAnchorElement) {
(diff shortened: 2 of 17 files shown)


apps/builder/app/builder/features/props-panel/controls/combined.tsx
```diff
@@ -19,6 +19,10 @@ export const renderControl = ({
     throw Error("Data source is not resolved");
   }
 
+  if (prop?.type === "action") {
```
kof, line 22: We don't render control when its bound? maybe a comment for why/what

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -183,12 +183,19 @@ export const WebstudioComponentDev = ({
     onClick: (event: MouseEvent) => {
       event.preventDefault();
       if (event.currentTarget instanceof HTMLAnchorElement) {
+        // @todo use Navigation API once implemented everywhere
+        // https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API
         handleLinkClick(event);
+      } else if (typeof userProps.onClick === "function") {
```
kof, line 189: When is this the case? When the element is not a link and there is an onClick callback? What happens if its a link, can't it have an onClick?

packages/react-sdk/src/embed-template.test.ts
```diff
@@ -384,3 +381,86 @@ test("generate data for embedding from props bound to data source expressions",
...
+          ],
+          children: [
+            {
+              type: "instance",
+              component: "Box2",
+              props: [
+                {
+                  type: "action",
+                  name: "onClick",
+                  value: [{ type: "execute", code: `boxState = 'success'` }],
```
kof, line 408: code vs expression, can it be more than expression? If yes then code makes sense

kof (approved):

apps/builder/app/builder/features/props-panel/controls/combined.tsx
TrySound: we don't have ui for actions. That's all.

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
TrySound: Bypass event handlers to other components for example Button

istarkov (approved):
Linear logic is 10x better than setInterval

apps/builder/app/builder/features/props-panel/controls/combined.tsx
kof: Maybe good idea to have a todo with a link to an issue in such cases

PR #1937, opened 2023-07-14
use useMount instead of sync once

istarkov:
## Description

useSyncInitializeOnce is incompatibel with modern React especially in dev mode.
useUnmount can't be used with it and any hooks with cleanup would give the same issue.

Should affect only devmode

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/canvas/canvas.tsx
```diff
@@ -167,7 +167,7 @@ export const Canvas = ({ params }: CanvasProps): JSX.Element | null => {
   const components = new Map(
     Object.entries({ ...baseComponents, ...remixComponents })
   ) as Components;
-  useSyncInitializeOnce(() => {
+  useMount(() => {
```
kof, line 170: this is basically useEffect(effect, []);

istarkov: useMount useUnmount is std hooks in react-use 
they are more clear than `useEffect(() => {}, [])`
They do not interfere with eslint i.e. add dependency x into `[]`

kof:
I don't really understand how this is solving it in a different way

kof:
How was the previous version failing exactly?

kof (approved):

istarkov:
> I don't really understand how this is solving it in a different way

useSyncInitializeOnce  should not be used at all, as its not working with any react cleanup usually placed in useEffect return callback.

Because now in dev mode, every useEffect is called twice

istarkov:
Current issue that I added cleanup and killed dev mode.

TrySound (approved):

PR #1945, opened 2023-07-17
(fix): Re-focus the tree-node after completion of editing the name

JayaKrishnaNamburu:
## Description

## Steps for reproduction

1. Use keyboard to navigate between the tree elements.
2. Click `Ctrl + E` to enter into edit mode.
3. Edit the name of the tree-node and then press `Enter`.
4. Use the up/down arrows to navigate to other nodes in the tree.

## Code Review

- [ ] hi, I need you to do
  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/tree/tree.tsx
+++ b/packages/design-system/src/components/tree/tree.tsx
@@ -420,8 +420,6 @@ const useKeyboardNavigation = <Data extends { id: string }>({
 
   const hadFocus = useRef(false);
   const prevRoot = useRef(root);
-  const prevEditingItemId = useRef(editingItemId);
-
   useEffect(() => {
     const haveFocus =
       rootRef.current?.contains(document.activeElement) === true;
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/tree/tree.tsx
```diff
@@ -439,11 +441,25 @@ const useKeyboardNavigation = <Data extends { id: string }>({
     }
   }, [root, rootRef, selectedItemSelector, setFocus]);
 
+  // When we stop editing an item, we want to focus it back again.
+  // To continue keyboard navigation.
+  useEffect(() => {
```
istarkov, line 446: After clicking Enter we save item somewhere, so we have an event to save, we can focus in same event handler, without using useEffect IMO

JayaKrishnaNamburu: Oh is it, let me try that way then 👍

istarkov: The other way is to return focus inside useContentEditable, preserving it initially

istarkov: Anyway this is just ideas how to not create new useEffect

JayaKrishnaNamburu: Yes, the change in `use-content-editable` might not be the best option. Since it focusses the text node. But not the whole tree-node. Anyways i will try the options for doing it without `useEffect` 👍

istarkov (approved):

PR #1957, opened 2023-07-19
Add tooltip

istarkov:
## Description

Left
- [ ] - storybook (next PR)
- [ ] - default styles (next PR)
- [x] - Add property to edit tooltip
- [ ] - Contraints - do not allow delete Trigger and Content (next PRs)
- [ ] - Icons  (next PRs)

### Whats done to make tooltip to work.

1. Portals support through passing [selectorIdAttribute](https://github.com/webstudio-is/webstudio-builder/blob/20c3b9fdea4a6f8e7a2cc00211b6e4ca95db13a5/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx#L290) 
2. [Hidden ScreenReader elements detection](https://github.com/webstudio-is/webstudio-builder/blob/20c3b9fdea4a6f8e7a2cc00211b6e4ca95db13a5/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx#L129)
3. [New elements on Canvas detection]https://github.com/webstudio-is/webstudio-builder/blob/20c3b9fdea4a6f8e7a2cc00211b6e4ca95db13a5/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx#L179

- [x] Collapsed
- [x] Do composeEventHandlers 
- [x] Radix add spans of tooltip content 
To support screen readers
https://github.com/twbs/bootstrap/blob/master/scss/mixins/_screen-reader.scss
Radix here
https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L552-L556
Cloning our content
https://github.com/radix-ui/primitives/blob/main/packages/react/visually-hidden/src/VisuallyHidden.tsx 

## Steps for reproduction

1. click button
3. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/collapsed.ts
+++ b/apps/builder/app/canvas/collapsed.ts
@@ -9,6 +9,7 @@ import {
   breakpointsStore,
   instancesStore,
   registeredComponentMetasStore,
+  selectedPageStore,
   stylesIndexStore,
 } from "~/shared/nano-states";
 import { selectedBreakpointStore } from "~/shared/nano-states";
@@ -92,8 +93,17 @@ const getInstanceSize = (instanceId: string, tagName: HtmlTags | undefined) => {
   };
 };
 
+const MAX_SIZE_TO_USE_OPTIMIZATION = 50;
+
 const recalculate = () => {
-  const instanceIds = Array.from(instanceIdSet);
+  const rootInstanceId = selectedPageStore.get()?.rootInstanceId;
+
+  const instanceIds =
+    instanceIdSet.size < MAX_SIZE_TO_USE_OPTIMIZATION
+      ? Array.from(instanceIdSet)
+      : rootInstanceId !== undefined
+      ? [rootInstanceId]
+      : [];
   instanceIdSet.clear();
   if (instanceIds.length === 0) {
     return;
--- a/apps/builder/app/canvas/collapsed.ts
+++ b/apps/builder/app/canvas/collapsed.ts
@@ -98,12 +98,16 @@ const MAX_SIZE_TO_USE_OPTIMIZATION = 50;
 const recalculate = () => {
   const rootInstanceId = selectedPageStore.get()?.rootInstanceId;
 
+  // Below algorithm quickly finds the common ancestor of all elements with an instanceId.
+  // However, for a large number of elements, it's more efficient to calculate from the root.
+  // In almost all cases, if instanceIdsSet >= MAX_SIZE_TO_USE_OPTIMIZATION, it likely includes all elements.
   const instanceIds =
     instanceIdSet.size < MAX_SIZE_TO_USE_OPTIMIZATION
       ? Array.from(instanceIdSet)
       : rootInstanceId !== undefined
       ? [rootInstanceId]
       : [];
+
   instanceIdSet.clear();
   if (instanceIds.length === 0) {
     return;
--- a/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
+++ b/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
@@ -262,7 +262,12 @@ export const WebstudioComponentDev = forwardRef<
 
   const composedHandlers: ImplicitEvents = {};
 
-  // Compose radix implicit event handlers with user-defined ones (e.g. onClick, onSubmit)
+  /**
+   * We combine Radix's implicit event handlers with user-defined ones, such as onClick or onSubmit.
+   * For instance, a Button within a TooltipTrigger receives
+   * an onClick handler from the TooltipTrigger.
+   * We might also need an additional onClick handler on the Button for other purposes (setting variable).
+   **/
   for (const [key, value] of Object.entries(restProps)) {
     const propHandler = props[key as keyof typeof props];
     if (
@@ -288,7 +293,6 @@ export const WebstudioComponentDev = forwardRef<
           instanceProps={instanceProps}
         />
       )}
-      {/* @todo restProps are radix events and props, for event handlers we need to composeEventHandlers like in radix */}
       <Component
         {...restProps}
         {...props}
--- a/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
+++ b/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
@@ -157,7 +157,7 @@ const useIsVisuallyHidden = (ref: RefObject<HTMLElement>) => {
 };
 
 /**
- * For some components that are wrapped with Radix Slot like components (Triggers etc where asChild=true),
+ * For some components that are wrapped with Radix Slot-like components (Triggers etc where asChild=true),
  * Slots in react-aria https://react-spectrum.adobe.com/react-spectrum/layout.html#slots
  * events are passed implicitly. We aim to merge these implicit events with the explicitly defined ones.
  **/
--- a/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
+++ b/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
@@ -43,6 +43,7 @@ import { SelectedInstanceConnector } from "./selected-instance-connector";
 import { handleLinkClick } from "./link";
 import { mergeRefs } from "@react-aria/utils";
 import { composeEventHandlers } from "@radix-ui/primitive";
+import { setDataCollapsed } from "~/canvas/collapsed";
 
 const TextEditor = lazy(() => import("../text-editor"));
 
@@ -170,6 +171,24 @@ type ImplicitEvents = {
    **/
 };
 
+const existingElements = new Set<string>();
+
+/**
+ * We are identifying newly created instances like Tooltips and ensuring the calculation of 'collapsed' elements.
+ */
+const useCollapsedOnNewElement = (instanceId: Instance["id"]) => {
+  useEffect(() => {
+    if (existingElements.has(instanceId) === false) {
+      setDataCollapsed(instanceId);
+    }
+
+    existingElements.add(instanceId);
+    return () => {
+      existingElements.delete(instanceId);
+    };
+  }, [instanceId]);
+};
+
 // eslint-disable-next-line react/display-name
 export const WebstudioComponentDev = forwardRef<
   HTMLElement,
@@ -194,6 +213,8 @@ export const WebstudioComponentDev = forwardRef<
     instanceSelector
   );
   const isPreviewMode = useStore(isPreviewModeStore);
+
+  useCollapsedOnNewElement(instanceId);
 
   // Scroll the selected instance into view when selected from navigator.
   useEffect(() => {
--- a/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
+++ b/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
@@ -153,13 +153,22 @@ const useIsScreenReaderDescendant = (ref: RefObject<HTMLElement>) => {
   return isScreenReaderDescendant;
 };
 
-type HandlerName = `on${string}`;
-
 /**
  * For some components that are wrapped with Radix Slot components (where asChild=true),
  * events are passed implicitly. We aim to merge these implicit events with the explicitly defined ones.
  **/
-type ImplicitEvents = Record<HandlerName, (event: never) => void>;
+type ImplicitEvents = {
+  onClick?: undefined | ((event: never) => void);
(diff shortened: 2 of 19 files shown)


apps/builder/app/canvas/collapsed.ts
```diff
@@ -92,8 +93,17 @@ const getInstanceSize = (instanceId: string, tagName: HtmlTags | undefined) => {
...
 const recalculate = () => {
-  const instanceIds = Array.from(instanceIdSet);
+  const rootInstanceId = selectedPageStore.get()?.rootInstanceId;
+
+  const instanceIds =
+    instanceIdSet.size < MAX_SIZE_TO_USE_OPTIMIZATION
+      ? Array.from(instanceIdSet)
+      : rootInstanceId !== undefined
+      ? [rootInstanceId]
+      : [];
```
kof, line 106: I don't understand what we are doing here or why

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -108,12 +119,81 @@ type WebstudioComponentDevProps = {
...
+
+  useEffect(() => {
+    if (ref.current !== null) {
+      for (
+        let element: HTMLElement | null = ref.current;
+        element !== null;
+        element = element.parentElement
+      ) {
+        if (
+          element.style.overflow === "hidden" &&
```
kof, line 141: probably should compute the style, because it could have been applied via css as welll, if we want it to work universally, not just with radix

kof: can this condition be more universal in general, not just targeted for radix? Effectively you are trying to detect visually hidden element, right?

could it be visually hidden by clip and overflow hidden without position absolute and width/height?

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -108,12 +119,81 @@ type WebstudioComponentDevProps = {
...
+          element.style.position === "absolute" &&
+          element.style.width === "1px" &&
+          element.style.height === "1px"
+        ) {
+          setIsScreenReaderDescendant(true);
+          return;
+        }
+      }
+    }
+  }, [ref]);
```
kof, line 152: I don't quite get it but it shouldn't need a ref as a dependency, no? or is it just for linter, if only for linter, then I would rather make the disable eslint comment because it makes it clear

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -108,12 +119,81 @@ type WebstudioComponentDevProps = {
...
-  components,
-}: WebstudioComponentDevProps) => {
+/**
+ * Radix's VisuallyHiddenPrimitive.Root https://github.com/radix-ui/primitives/blob/main/packages/react/visually-hidden/src/VisuallyHidden.tsx
+ * component makes content from hidden elements accessible to screen readers.
+ * Using the same Content children, however, leads to duplicated React elements, breaking our 'isSelected' logic.
+ * To prevent this, we check if an instance is a descendant of VisuallyHiddenPrimitive.Root, and if so,
+ * we avoid rendering it.
+ */
+const useIsScreenReaderDescendant = (ref: RefObject<HTMLElement>) => {
```
kof, line 129: could be ` useIsVisuallyHidden`

kof: because the logic isn't specifically about screenreader, screenreader is how its used, yes, but its not what it does

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -108,12 +119,81 @@ type WebstudioComponentDevProps = {
...
+  }, [ref]);
+
+  return isScreenReaderDescendant;
+};
+
+/**
+ * For some components that are wrapped with Radix Slot components (where asChild=true),
+ * events are passed implicitly. We aim to merge these implicit events with the explicitly defined ones.
+ **/
+type ImplicitEvents = {
```
kof, line 161: is there any way to not have a logic that is specific to radix here? I don't know how, but maybe there is some way?

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -108,12 +119,81 @@ type WebstudioComponentDevProps = {
...
+   * We ignore the remaining events because we currently detect events using the 'on' prefix.
+   * This approach (defining type like above instead of Partial<Record<`on${string}`, Handler>>)
+   * is necessary due to our TypeScript settings, where 'no exactOptionalPropertyTypes' is set,
+   * which makes it impossible to define a Partial<Record<>> with optional keys.
+   *  (without exactOptionalPropertyTypes ts defines it as {[key: string]: Handler | undefined)}
+   *   instead of {[key: string]?: Handler | undefined)})
+   **/
+};
+
+const existingElements = new Set<string>();
```
kof, line 174: is this going to grow forever?

packages/react-sdk/src/tree/webstudio-component.tsx
```diff
@@ -57,13 +55,42 @@ export const WebstudioComponent = ({
...
   return (
-    <Component {...props}>
+    <Component {...props} ref={ref}>
       {renderWebstudioComponentChildren(children)}
     </Component>
   );
+});
+
+export const idAttribute = "data-ws-id" as const;
+export const selectorIdAttribute = "data-ws-parent-id" as const;
```
kof, line 65: its not just parent id, is it? I think its the path up to the root, so it would be data-ws-path and pathAttribute if I am correct

kof (changes requested):
some comments left, otherwise great job!

kof:
I noticed a few things
- [ ] side offset and align offset should be unit values, we are missing a value type that can have the units like in a style panel (can be improvement for later if its too hard) but its not obvious that the number is in px
- [ ] align doesn't have the current value selected by default
Screenshot 2023-07-23 at 21 36 29

kof:
Just noticed this in the demo, Open and Is Open 

Screenshot 2023-07-23 at 21 52 45

apps/builder/app/canvas/collapsed.ts
istarkov: Added comment

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
istarkov: react-aria has near same a little bit other set of props in `style` attribute. 

But overflow, height, width, position, and clip are common.

aria example:
```
border: 0px; clip: rect(0px, 0px, 0px, 0px); clip-path: inset(50%); height: 1px; margin: -1px; overflow: hidden; padding: 0px; position: absolute; width: 1px; white-space: nowrap;
```

Also using the `style` attribute intead of getComputedStyle make it less possible to interfere with user defined styles as in our system they are all classes.

What if real user will create visually hidden box for its own needs. We should not skip user defined elements.

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
istarkov: No, there is delete in effect unsubscribe.

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
kof: in this case maybe the approach is not great and we should search for a way to explicitely tell builder from component that this component has a visually hidden thing? e.g. a hint using a special attribute or something?

packages/react-sdk/src/tree/webstudio-component.tsx
kof: or data-ws-selector

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
istarkov: Added comment about slots in other framework, they all behave the same, so it's not radix specific but how every Slot like wrapper works like

istarkov:
we can't use open as is written in comment here
https://github.com/webstudio-is/webstudio-builder/blob/b41205843a365214c0fa04910b6e176955a05995/packages/sdk-components-react/src/radix-tooltip.tsx#L17-L24

So additional props for builder only behaviour is created.
On newly added tooltips only isOpen exposed by default and its our variable binding hack.

istarkov:
isOpen is not saved and has ephemeral state, please test everything on newly created tooltips. As tooltips from previous versions doesn't just had right template

istarkov:
About align and default values, the issue that we don't support in UI undefined properties. Like align can be undefined - so system decides what to do

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
istarkov: Wouldn't somehow improve things, nothing would prevent user to make such element inside tooltip content.
Im not even sure that react-aria adds VisuallyHidden elements implicitly like radix.
And that this is issue is common across most frameworks/components

kof (approved):

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -108,12 +119,84 @@ type WebstudioComponentDevProps = {
...
+ * Radix's VisuallyHiddenPrimitive.Root https://github.com/radix-ui/primitives/blob/main/packages/react/visually-hidden/src/VisuallyHidden.tsx
+ * component makes content from hidden elements accessible to screen readers.
+ * react-aria VisuallyHidden https://github.com/adobe/react-spectrum/blob/e4bc3269fa41aa096700445c6bfa9c8620545e6a/packages/%40react-aria/visually-hidden/src/VisuallyHidden.tsx#L32-L43
+ * The problem we're addressing is that the Radix reuses the same Content children for VisuallyHiddenPrimitive.Root within the Tooltip component.
+ * Using the same Content children, however, leads to duplicated React elements, breaking our 'isSelected' logic.
+ * To prevent this, we check if an instance is a descendant of VisuallyHiddenPrimitive.Root, and if so,
+ * we avoid rendering it.
+ */
+const useIsVisuallyHidden = (ref: RefObject<HTMLElement>) => {
+  const [isScreenReaderDescendant, setIsScreenReaderDescendant] =
```
kof, line 132: that needed renaming too consistently with the above

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
kof: but why do we have here only Click and Submit?

kof: why these 2 and nothing else?

istarkov: we overrided theese 2 in the code below to support links and forms.
Here are both just for simplicity of reading `composeEventHandlers` below, so coder can see that
we are merging 
implicit `onClick` from radix like component, with WebstudioComponentDev internal onClick which call userProps.onClick
But here can be any handler, as with our ts settings I was unable to define Partial on a record type. 
And comment below says about that `We ignore the remaining events because we currently detect events using the 'on' prefix.` and our inability to use Partial here

packages/sdk-components-react/src/props.ts
```diff
@@ -30,3 +30,8 @@ export { propsMeta as Vimeo } from "./vimeo.ws";
 export { propsMeta as VimeoPreviewImage } from "./vimeo-preview-image.ws";
 export { propsMeta as VimeoPlayButton } from "./vimeo-play-button.ws";
 export { propsMeta as VimeoSpinner } from "./vimeo-spinner.ws";
+export {
+  propsMetaTooltip as Tooltip,
+  propsMetaTooltipContent as TooltipContent,
+  propsMetaTooltipTrigger as TooltipTrigger,
+} from "./radix-tooltip.ws";
```
TrySound, line 37: Radix components should have a separate package

kof: I agree, ideally yes, is it easy to split now? the lets split, if not - lets wait until we build the "install package ui"

PR #1972, opened 2023-07-24
Add font smoothing by default to the body

kof:
## Description

Adding font smoothing by default to the Body component

https://discord.com/channels/955905230107738152/955907787299360788/1132882679189229649

```
  -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
```

## Steps for reproduction

1. open any project, see in dev tools these 2 added (or check visually)

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

taylornowotny:
Screenshot 2023-07-24 at 9 06 24 AM

Text looks much nicer now! 👍

The new code is showing in the CSS preview for every item on the canvas.

kof:
@taylornowotny I am not sure why we are showing inherited properties in the css preview, should we remove those?

kof:
On the second thought its actually correct to show it there, because once we have CSS editor, user should be able to change those things as well

kof:
I think we could in the future group properties into sections: 
- local
- inherited
- cascaded
so it will be clear where each value is coming from just like it is in the style panel

kof:
For now I can sort the properties, showing first the local values, then inherited, then cascaded from breakpoints

taylornowotny:
Here's my suggestion:

Show only local and cascaded styles.
Do not show inherited styles.
This is how Webflow is doing it. No need for categories.

Reason: This window is starting to be overloaded with content. Removing less relevant information helps the user find the more relevant information more quickly. In this case the inherited styles are already displayed on the parent elements so repeating that information on every child is not necessary.

kof:
@taylornowotny how do you justify showing cascaded styles but not showing inherited? When do you draw the line?

taylornowotny:
> @taylornowotny how do you justify showing cascaded styles but not showing inherited? When do you draw the line?

Inherited styles are on the parent so they are easily accessible without switching breakpoints. Furthermore, repeating inherited styles on every child element is redundant and makes it slower to find the info about local styles.

It would be fine to show local styles only, not cascaded, but if a user wants to see those cascaded styles then they would have to switch breakpoints, which may reset their canvas size, thus potentially disrupting their work flow.

kof:
I think we can fix the problem of "only interested in token styles" by showing them first and making sure its clear these are token styles, I am implementing right now grouping where each group will have an inline comment above.

This way we don't have to decide which styles that apply to an element we show and which we don't, does this work for you?

taylornowotny:
That works, but not as good as removing inherited styles.

It still leaves the CSS Preview window overloaded with content and the order of local -> cascaded -> inherited goes against the direction of the CSS cascade of top -> bottom

kof:
here is how it looks like

Screenshot 2023-07-24 at 20 42 07

kof:
Regarding the order you are correct, cascaded should go first.

I will leave them all in for now. I think we need a better decision later, if we go after the ui devtools have, then they all should be in there, maybe for a cleaner look we can later collapse inherited/cascaded styles

Screenshot 2023-07-24 at 20 46 44

taylornowotny:
Why do we need a better decision later? I think my suggestion was ideal.

With inherited styles gone, you wouldn't even need to label "Style Sources", only "Cascaded".

kof:
> Why do we need a better decision later? I think my suggestion was ideal.

Because I think user may equally want to know what's inherited, whats cascaded and we can't remove just inherited, since this would be inconsistent + we do show all of that info in the style panel, so it would be also inconsistent with the style panel to not show that avlue

apps/builder/app/builder/features/sidebar-left/navigator/css-preview.tsx
```diff
@@ -21,33 +22,62 @@ const preStyle = css(textVariants.mono, {
...
+    }
+    if (value.cascaded?.value) {
+      cascadedStyles.set(property, value.value);
+      continue;
+    }
+    if (value.inherited?.value) {
+      inheritedStyles.set(property, value.value);
+      continue;
+    }
+  }
```
istarkov, line 60: What if we get here? Should be some exchaustive check if something will change in the future

kof:
> We can remove just inherited because the user can find all those styles on the parent anyway.

Without knowing what has been actually inherited - no, because not every property is inheritable. They would have to do gymnastics to understand what are the values the child will possible inherit from the parent.

apps/builder/app/builder/features/sidebar-left/navigator/css-preview.tsx
kof: please explain

taylornowotny (approved):

apps/builder/app/builder/features/sidebar-left/navigator/css-preview.tsx
istarkov: You have a huge series of if's
In some places you explicitly do nothing like 
```
 if (value === undefined) {
      continue;
    }
```
At the end of series you or should cover all possibilities, and have check that code will not be occured there.
For example we will add additional style etc 
Example https://github.com/webstudio-is/webstudio-builder/blob/61f164d40e6c62bc98a0f9f5cb894984b2d93bd5/packages/css-engine/src/core/to-value.ts#L7

istarkov: or give a comment that other possibilities exists and it's not an error that you are doing nothing

kof: I am confused, ofc its its not an error, we are exiting early beause the value was handled, its a typical early return situation though with a continue ... I don't know what can be misunderstood

istarkov: Its not misunderstanding its error.
See example I gave with `assertUnreachable` every time we add new style type my code give ts error.
Your code if we will add additional source will never somehow alert us that new style source exists and must be fixed here

istarkov (approved):

kof:
@TrySound @istarkov I found a way to create a Utility function that does all 3 things for our error handling:
- throws in development runtime
- throws in typescript
- doesn't break production, but lets sentry catch it in the future and logs the error

packages/error-utils/README.md
```diff
@@ -0,0 +1,78 @@
+# CSS Vars utils
+
+This package helps you create uniquely named css variables as well as use them.
+
+## API
+
+```ts
+import { cssVars } from "error-utils";
```
TrySound, line 8: Wrong docs?

istarkov (approved):

TrySound (approved):

packages/error-utils/README.md
kof: yep

PR #1975, opened 2023-07-24
Radix tooltip local styles

istarkov:
## Description

Local styles from https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/default/ui/tooltip.tsx

## Steps for reproduction

Add NEW tooltip, see styles

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/package.json
+++ b/packages/sdk-components-react/package.json
@@ -54,6 +54,7 @@
     "@webstudio-is/icons": "workspace:^",
     "@webstudio-is/image": "workspace:^",
     "@webstudio-is/react-sdk": "workspace:^",
+    "@webstudio-is/css-data": "workspace:^",
     "colord": "^2.9.3",
     "shallow-equal": "^3.1.0"
   },
--- a/packages/sdk-components-react/package.json
+++ b/packages/sdk-components-react/package.json
@@ -49,14 +49,15 @@
   "dependencies": {
     "@radix-ui/react-tooltip": "^1.0.6",
     "@react-aria/utils": "^3.13.3",
+    "@webstudio-is/css-data": "workspace:^",
     "@webstudio-is/css-vars": "workspace:^",
     "@webstudio-is/generate-arg-types": "workspace:^",
     "@webstudio-is/icons": "workspace:^",
     "@webstudio-is/image": "workspace:^",
     "@webstudio-is/react-sdk": "workspace:^",
-    "@webstudio-is/css-data": "workspace:^",
     "colord": "^2.9.3",
-    "shallow-equal": "^3.1.0"
+    "shallow-equal": "^3.1.0",
+    "tailwindcss": "^3.3.3"
   },
   "devDependencies": {
     "@storybook/react": "^7.1.0",
--- a/packages/sdk-components-react/src/radix-tooltip.ws.tsx
+++ b/packages/sdk-components-react/src/radix-tooltip.ws.tsx
@@ -98,15 +98,8 @@ export const metaTooltip: WsComponentMeta = {
           children: [
             {
               type: "instance",
-              component: "Box",
-              props: [],
-              children: [
-                {
-                  type: "instance",
-                  component: "Text",
-                  children: [{ type: "text", value: "The text you can edit" }],
-                },
-              ],
+              component: "Text",
+              children: [{ type: "text", value: "The text you can edit" }],
             },
           ],
         },
--- a/packages/sdk-components-react/src/radix-tooltip.ws.tsx
+++ b/packages/sdk-components-react/src/radix-tooltip.ws.tsx
@@ -84,17 +84,17 @@ export const metaTooltip: WsComponentMeta = {
            *  z-50 overflow-hidden rounded-md border bg-popover px-3 py-1.5 text-sm text-popover-foreground shadow-md
            **/
           styles: [
-            ...tc.z(50),
-            ...tc.overflow("hidden"),
-            ...tc.rounded("md"),
-            ...tc.border(),
-            ...tc.bg("popover"),
-            ...tc.px(3),
-            ...tc.py(1.5),
-            ...tc.text("sm"),
-            ...tc.text("popoverForeground"),
-            ...tc.shadow("md"),
-          ],
+            tc.z(50),
+            tc.overflow("hidden"),
+            tc.rounded("md"),
+            tc.border(),
+            tc.bg("popover"),
+            tc.px(3),
+            tc.py(1.5),
+            tc.text("sm"),
+            tc.text("popoverForeground"),
+            tc.shadow("md"),
+          ].flat(),
           children: [
             {
               type: "instance",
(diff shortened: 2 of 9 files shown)


TrySound (approved):

kof (approved):
looking good, and yeah we need to change styles authoring format, lets just make sure we make a good decision there

packages/sdk-components-react/src/theme/tailwind-classes.ts
```diff
@@ -0,0 +1,203 @@
+/**
+ * Quik and dirty implementation of tailwind classes conversion to webstudio styles.
+ */
+import type { EmbedTemplateStyleDecl } from "@webstudio-is/react-sdk";
+import * as variables from "./radix-variables";
+import { theme } from "./tailwind-theme";
+import { parseCssValue } from "@webstudio-is/css-data";
+
+// https://github.com/tailwindlabs/tailwindcss/blob/master/src/css/preflight.css
+export const preflight = (): EmbedTemplateStyleDecl[] => [
```
kof, line 10: we already have our normalize and reset ... is this not the same? how can we avoid having both?

packages/sdk-components-react/src/theme/tailwind-classes.ts
```diff
@@ -0,0 +1,203 @@
...
+  },
+  {
+    property: "borderBottomLeftRadius",
+    value: variables.radius,
+  },
+];
+
+export const border = (borderWidth?: number): EmbedTemplateStyleDecl[] => {
+  const key = `${borderWidth ?? "DEFAULT"}`;
+  const valueString = theme("borderWidth")?.[key] ?? "1px";
```
kof, line 83: shoudln't the theme rather have an already parsed css value so you don't need to parse it below?

packages/sdk-components-react/package.json
```diff
@@ -49,13 +49,15 @@
...
+    "@webstudio-is/css-data": "workspace:^",
     "@webstudio-is/css-vars": "workspace:^",
     "@webstudio-is/generate-arg-types": "workspace:^",
     "@webstudio-is/icons": "workspace:^",
     "@webstudio-is/image": "workspace:^",
     "@webstudio-is/react-sdk": "workspace:^",
     "colord": "^2.9.3",
-    "shallow-equal": "^3.1.0"
+    "shallow-equal": "^3.1.0",
+    "tailwindcss": "^3.3.3"
```
TrySound, line 60: Looks like too huge dependency to install just for config. Maybe host it?

packages/sdk-components-react/src/radix-tooltip.ws.tsx
```diff
@@ -70,6 +80,22 @@ export const metaTooltip: WsComponentMeta = {
           component: "TooltipContent",
           label: "Tooltip Content",
           props: [],
+          /**
+           *  z-50 overflow-hidden rounded-md border bg-popover px-3 py-1.5 text-sm text-popover-foreground shadow-md
+           **/
+          styles: [
+            ...tc.preflight(),
```
TrySound, line 87: Maybe use object notation? Like { z: 50 }

TrySound: Or lists of classes like ['z-50']?

packages/sdk-components-react/src/theme/tailwind-classes.ts
istarkov: It's naked tailwind theme, can spend a lot of time parsing it fully. Here is easier as I use known props.

packages/sdk-components-react/src/radix-tooltip.ws.tsx
istarkov: Would be not a big deal to transform in following PRs. For now not stable at all

packages/sdk-components-react/src/theme/tailwind-theme.ts
```diff
@@ -0,0 +1,27 @@
...
+type ThemeResolved = {
+  [key in keyof Theme]: Extract<Theme[key]>;
+};
+
+export const theme = <T extends ThemeProperties>(name: T): ThemeResolved[T] => {
+  const value = config?.theme?.[name];
+
+  if (typeof value === "function") {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return
+    return value({ theme, colors: {} });
```
kof, line 23: colors is not part of the theme? is it the same in tw?

istarkov: Part of, I just didnt include to avoid hsl etc parsing

packages/sdk-components-react/src/radix-tooltip.ws.tsx
kof: or a `ts.merge(tc.preflight(), tc.z(50), ...)`

packages/sdk-components-react/src/theme/tailwind-theme.ts
istarkov: Gonna hoist config, in that case can extend it better

packages/sdk-components-react/src/theme/tailwind-classes.ts
kof: yeah, better move, otherwise this looks like a place for all sorts of resets, but also check if we don't some of it already in our preset/normalize

packages/sdk-components-react/src/theme/tailwind-classes.ts
kof: hmm, so we can't use any tailwind them or can we?

istarkov: We can use any, here just supersmall theme wrapper, without any deep knowledge. Like I know that we can parse borderWidth or paddings but not sure about everything, its like Im adding utils I see in radix styles

packages/sdk-components-react/src/theme/tailwind-theme.ts
kof: if you change the theme spec its gonna be confusing over time

packages/sdk-components-react/src/theme/tailwind-classes.ts
istarkov: moved

packages/sdk-components-react/src/radix-tooltip.ws.tsx
istarkov: Added `.flat()` native array method and seems like well typed with ts

packages/sdk-components-react/src/theme/tailwind-theme.ts
istarkov: Added some simle theme colors etc override

packages/sdk-components-react/package.json
istarkov: Decided to not hoist for now as types of config are huge in tailwind. Probably next iteration

PR #1986, opened 2023-07-26
Allow to disable components detaching

TrySound:
Here added meta.detachable flag which allows when set to true disable copying, dragging and deleting instances of this component without ts parent instance.

This is necessary to constrain components based on external code and prevent deleting hidden components user cannot add again.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/navigator-tree.tsx
+++ b/apps/builder/app/builder/shared/navigator-tree.tsx
@@ -1,7 +1,9 @@
 import { useCallback, useMemo } from "react";
 import { useStore } from "@nanostores/react";
 import { shallowEqual } from "shallow-equal";
+import { toast } from "@webstudio-is/design-system";
 import type { Instance } from "@webstudio-is/project-build";
+import { generateDataFromEmbedTemplate } from "@webstudio-is/react-sdk";
 import {
   hoveredInstanceSelectorStore,
   instancesStore,
@@ -18,9 +20,9 @@ import {
   findClosestDroppableComponentIndex,
   reparentInstance,
   type InsertConstraints,
+  isInstanceDetachable,
 } from "~/shared/instance-utils";
 import { InstanceTree } from "./tree";
-import { generateDataFromEmbedTemplate } from "@webstudio-is/react-sdk";
 
 export const NavigatorTree = () => {
   const selectedInstanceSelector = useStore(selectedInstanceSelectorStore);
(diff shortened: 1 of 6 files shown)


apps/builder/app/builder/shared/navigator-tree.tsx
```diff
@@ -139,6 +141,10 @@ export const NavigatorTree = () => {
       onSelect={handleSelect}
       onHover={hoveredInstanceSelectorStore.set}
       onDragItemChange={(dragInstanceSelector) => {
+        if (isInstanceDetachable(dragInstanceSelector) === false) {
+          toast.error("Cannot drag instance without its parent instance");
```
kof, line 145: need to check wording with @taylornowotny

apps/builder/app/canvas/shared/use-drag-drop.ts
```diff
@@ -242,6 +244,10 @@ export const useDragAndDrop = () => {
     },
 
     onStart({ data: dragInstanceSelector }) {
+      if (isInstanceDetachable(dragInstanceSelector) === false) {
+        toast.error("Cannot drag instance without its parent instance");
```
kof, line 248: need to check wording with @taylornowotny

kof (approved):

kof:
Should we also set in Form component: Success, Error and in Vimeo: Spinner and Play as detachable false?

apps/builder/app/canvas/shared/use-drag-drop.ts
taylornowotny: More semantic suggestion:
"Instance cannot be moved outside of its parent component."

This covers the fact that it can't be copied or deleted, not limited to drag.

Question - are we disabling drag entirely or just drag outside of the parent? User may still want to move things around within the parent container.

TrySound:
Form Success and Error are Boxes

kof:
@TrySound true, but what happens if user deletes them? The form becomes broken and they can't restore it visually atm

TrySound:
Yes. Though this is meta level configuration. We cannot apply it to specific instances.

istarkov (approved):
Super!!!! Tested works well

apps/builder/app/canvas/shared/use-drag-drop.ts
TrySound: I would figure this out once we have a use case. For now it is always used for cases where the order does not matter.

PR #1999, opened 2023-07-27
(fix): Optional chaining for dataSources in react-sdk

JayaKrishnaNamburu:
I was working with the cli, and when executing few projects. I found there are builds without `dataSources` in them and it seems to break when we start the project.

If it's something happens only in older projects. Please free to close.

Screenshot 2023-07-28 at 02 26 05

### DIFF
--- a/packages/react-sdk/src/tree/root.ts
+++ b/packages/react-sdk/src/tree/root.ts
@@ -70,7 +70,8 @@ export const InstanceRoot = ({
       (dataSourceVariables) => {
         // set vriables with defaults
         const dataSourceValues: DataSourceValues = new Map();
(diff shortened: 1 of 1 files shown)


TrySound:
This should be handled when data is loaded. Not in renderer.
Let's figure this out. Where this data comes from?

JayaKrishnaNamburu:
This happens here
https://github.com/webstudio-is/webstudio-cli/blob/new-cli/src/sync.ts#L33
When the data is loaded into a file and then read back to the renderer

TrySound:
Which command you use to get this error

JayaKrishnaNamburu:
Closing this, false alarm. It was a bug inside `cli` implementation 😅

PR #2003, opened 2023-07-30
Add radix collapsible component

TrySound:
Ref https://www.radix-ui.com/docs/primitives/components/collapsible

Added new sdk-components-react-radix package with collapsible component. It does not have any styles so no need to move tailwind stuff yet.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react-radix/README.md
+++ b/packages/sdk-components-react-radix/README.md
@@ -1,0 +1,3 @@
+# Webstudio SDK Components
+
+Webstudio SDK is a TypeScript API that lets you use your Webstudio site or some components in your custom codebase or just render a complete Remix Document.
--- a/packages/sdk-components-react-radix/README.md
+++ b/packages/sdk-components-react-radix/README.md
@@ -1,3 +1,4 @@
-# Webstudio SDK Components
+# Radix Components for Webstudio
 
-Webstudio SDK is a TypeScript API that lets you use your Webstudio site or some components in your custom codebase or just render a complete Remix Document.
+Radix Primitives is a low-level UI component library with a focus on accessibility and customization.
+Default styling is inspired by https://ui.shadcn.com/docs.
--- a/packages/sdk-components-react-radix/src/radix-collapsible.tsx
+++ b/packages/sdk-components-react-radix/src/radix-collapsible.tsx
@@ -1,3 +1,6 @@
+/* eslint-disable react/display-name */
+// We can't use .displayName until this is merged https://github.com/styleguidist/react-docgen-typescript/pull/449
+
 import {
   type ReactNode,
   type ForwardRefExoticComponent,
--- a/packages/sdk-components-react-radix/src/radix-collapsible.tsx
+++ b/packages/sdk-components-react-radix/src/radix-collapsible.tsx
@@ -1,0 +1,54 @@
+import {
+  type ReactNode,
+  type ForwardRefExoticComponent,
+  type ComponentPropsWithRef,
+  forwardRef,
+  Children,
+} from "react";
+import {
+  Collapsible,
+  CollapsibleTrigger,
+  CollapsibleContent,
+} from "@radix-ui/react-collapsible";
+import {
+  type WebstudioAttributes,
+  splitPropsWithWebstudioAttributes,
+} from "@webstudio-is/react-sdk";
+
+export const RadixCollapsible: ForwardRefExoticComponent<
+  Omit<ComponentPropsWithRef<typeof Collapsible>, "defaultOpen" | "asChild">
+> = Collapsible;
+
+/**
+ * CollapsibleTrigger is HTML-less components.
+ * To make them work in our system, we wrap their attributes with a div that has a display: contents property.
+ *
+ * These divs function like fragments, with all web studio-related attributes attached to them.
+ */
+const displayContentsStyle = { display: "contents" };
+
+/**
+ * We're not exposing the 'asChild' property for the Trigger.
+ * Instead, we're enforcing 'asChild=true' for the Trigger and making it style-less.
+ * This avoids situations where the Trigger inadvertently passes all styles to its child,
+ * which would prevent us from displaying styles properly in the builder.
+ */
+export const RadixCollapsibleTrigger = forwardRef<
+  HTMLDivElement,
+  WebstudioAttributes & { children: ReactNode }
+>(({ children, ...props }, ref) => {
+  const firstChild = Children.toArray(children)[0];
+  const [webstudioAttributes, restProps] =
+    splitPropsWithWebstudioAttributes(props);
+  return (
+    <div ref={ref} style={displayContentsStyle} {...webstudioAttributes}>
+      <CollapsibleTrigger asChild={true} {...restProps}>
+        {firstChild ?? <button>Add button</button>}
+      </CollapsibleTrigger>
+    </div>
+  );
+});
+
+export const RadixCollapsibleContent: ForwardRefExoticComponent<
+  Omit<ComponentPropsWithRef<typeof CollapsibleContent>, "asChild">
+> = CollapsibleContent;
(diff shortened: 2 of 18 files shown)


istarkov (approved):
plz add to saas next release

istarkov:
Also Im not sure how we would define at saas from what library to import. 
Using prefix doesn't look like a good approach. Not sure we need prefix radix at all. We can just import at rpebuild and see whats exists inside package

packages/sdk-components-react-radix/README.md
```diff
@@ -0,0 +1,3 @@
+# Webstudio SDK Components
+
+Webstudio SDK is a TypeScript API that lets you use your Webstudio site or some components in your custom codebase or just render a complete Remix Document.
```
kof, line 3: needs a description

packages/sdk-components-react-radix/src/radix-collapsible.tsx
```diff
@@ -0,0 +1,57 @@
...
+ */
+export const RadixCollapsibleTrigger = forwardRef<
+  HTMLDivElement,
+  WebstudioAttributes & { children: ReactNode }
+>(({ children, ...props }, ref) => {
+  const firstChild = Children.toArray(children)[0];
+  const [webstudioAttributes, restProps] =
+    splitPropsWithWebstudioAttributes(props);
+  return (
+    <div ref={ref} style={displayContentsStyle} {...webstudioAttributes}>
```
kof, line 47: shouldn't this div be part of builder instead of SDK? We could have a meta that says something like "tagless" in which case we wrap it into that div with display contents.

reason for this is to make it simpler providing components to builder, since this is purely builder specific thing, a published site doesn't need this

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
```diff
@@ -0,0 +1,106 @@
...
+          },
+        },
+        {
+          name: "onOpenChange",
+          type: "action",
+          value: [
+            {
+              type: "execute",
+              args: ["open"],
+              code: `collapsibleOpen = open`,
```
kof, line 38: question: is this variable going to potentially conflict once we have code components added by the user we don't know about?

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
```diff
@@ -0,0 +1,106 @@
...
+  propsRadixCollapsibleContent,
+  propsRadixCollapsibleTrigger,
+} from "./__generated__/radix-collapsible.props";
+
+export const metaRadixCollapsible: WsComponentMeta = {
+  category: "radix",
+  type: "container",
+  label: "Collapsible",
+  icon: RadioUncheckedIcon,
+  template: [
```
kof, line 17: Styles will be added later?

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
```diff
@@ -0,0 +1,106 @@
...
+    {
+      type: "instance",
+      component: "RadixCollapsible",
+      props: [
+        {
+          name: "open",
+          type: "boolean",
+          value: false,
```
kof, line 25: should open be false by default? My initial reaction when testing: "where is content?"

kof:
Not specific to this PR, but I noticed that error message is a bit misleading: "This instance can't be moved outside of it's parent component" because I am not moving it outside, I was moving within.

https://share.descript.com/view/f2t8L00Xu9f

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
TrySound: No, this is replaced with data source id

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
TrySound: There is no styles in shaden for collapsible. Purely functional component.

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
TrySound: Dialogs and tooltips are closed by default too

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
kof: shadcn has tyles there, no?
https://ui.shadcn.com/docs/components/collapsible

kof (approved):

packages/sdk-components-react-radix/src/radix-collapsible.ws.ts
TrySound: Those are example styles

packages/sdk-components-react-radix/src/radix-collapsible.tsx
istarkov: The issue that it's not builder related only. We use components + component wrapper on raw Canvas too. 
(And to support refs we forward refs on `WebstudioComponent` too)

So logic need to be moved on Canvas too. To avoid various `refs` issues. (Like ref can't be passed to functional component)

Would need aditional meta info, and will cause additional questions like:

Why we forward ref for this component and do not forward ref for that component. (answer because it will be wrapped somewhere else with div fragment)

Why this component receives webstudio id property and that component do not receive webstudio id property
(we can't assign webstudioId to components like triggers) (answer - because we will add div somewhere else with webstudio id)

Also rendering tree at the builder and on canvas will be different, so easy to lost something too, even its display: contents it will have some side-effects. (Events, etc)

Even now it's hard to think about this types of components, and then everything will be implcit I feel it will be harder.

istarkov: Also as of now I made `asChild` on triggers to be always `true`. That makes trigger `element less` component. But with `asChild=false` it become standalone element. 
So slots are dynamic by they nature.
If some day we will find that we need to support `asChild` - how we will pass that information through the meta. 

(I will need to remove asChild from props of other components for now as they will cause bug)

kof: 1. we wrap sdk component anyways implicitely, following your logic we should be wrapping explicitly too
2. having inline style around every wrapped component isn't ideal
3. ist still not clear what needs to be passed to that wrapper component and why. If I had to create my own SDK component, I wouldn't know what to do in this case either

It's a system behavior leaked into user sdk.

kof: Btw even though the wrapper here is made explicit, the need for it or how it works is implicit. It will require understanding a lot of stuff and a lot of comments.

kof: Also this will need to happen in every such sdk component vs. single place inside the WebstudioComponent

kof: More so if we need to change this thing in any way, this will need to happen in every such component. Having it centrally will make it fixable in one place.

istarkov: I will wait a little bit until all components will be +- stable. Will work on Canvas (not tested) have no issues etc. Im not even near sure that we have covered all cases now.
Also as I wrote above it’s not meta information, as because of refs will need knowledge at Canvas level too

istarkov: Probably must be solved other way. For example supporting in the builder html less components without wrapper. (like htmlless component need to select all children on selection etc)

kof: Yeah, we will definitely need to support domless components soon anyways.

kof:
> Not specific to this PR, but I noticed that error message is a bit misleading: "This instance can't be moved outside of it's parent component" because I am not moving it outside, I was moving within. cc @taylornowotny

If we want to differentiate between move at all or move outside  then we need to have separate messages, otherwise the message should basically say "The instance can't be moved."

PR #2014, opened 2023-07-31
Attempt to hover/select without element

istarkov:
## Description

Removes `div style="display:contents"` arount Radix etc Components.

Below in Implementation is written why SelectedInstanceConnector was moved into standalone root level hook.
In short
- [x] - No issues with visually hidden components.
- [x] - Allows us to not use refs for the builder support. (Now our components can be without refs)
- [ ] - cc @TrySound ContentEditable could also use getAllElementsBoundingBox from domUtils to get element ref this will allow to remove useRef from WebstudioDevComponent

Changes from previous behavior:

- Now on preview elements are not selected (now is like in webflow).
- On changing marign etc we do not hide outline as before

## Implementation: 

having `instanceSelector` we can 
- Find all elements with attribute `data-ws-selector` which ends with `instanceSelector`
- Then find "root/parent" elements of found elements. _(We can find few subtrees)_
- Then use that elements instead of wrapper element.

This works fine with tree hovering and selections.
This allows us to not use `refs` inside our WebstudioDevComponent at all.

### Below is why SelectedInstanceConnector was rewritten

The issue arises when such element is duplicated by radix into VisuallyHidden element. 

There is no way to detect now inside `WebstudioComponentDev` does current Component is inside  VisuallyHidden
(We have no reference element to detect that.)

So current approach makes any additional changes useless.
I'm going to close this. And leave just for example

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/canvas/canvas.tsx
```diff
@@ -52,6 +52,7 @@ import { subscribeInstanceSelection } from "./instance-selection";
 import { subscribeInstanceHovering } from "./instance-hovering";
 import { useHashLinkSync } from "~/shared/pages";
 import { useMount } from "~/shared/hook-utils/use-mount";
+import { useSelectedInstance } from "./features/webstudio-component/selected-instance-connector";
```
kof, line 55: shouldn't webstudio-component have an index and import from the index?

kof: it already has an index

apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
```diff
@@ -32,12 +32,11 @@ import {
   type InstanceSelector,
   areInstanceSelectorsEqual,
 } from "~/shared/tree-utils";
-import { SelectedInstanceConnector } from "./selected-instance-connector";
+// import { SelectedInstanceConnector } from "./selected-instance-connector";
```
TrySound, line 35: remove comment?

apps/builder/app/canvas/instance-hovering.ts
```diff
@@ -44,27 +45,41 @@ export const subscribeInstanceHovering = () => {
       hoveredInstanceSelectorStore.set(undefined);
       hoveredInstanceOutlineStore.set(undefined);
     }, 100);
+
+    // Fixes the bug, that new hover occures during timeout
```
TrySound, line 49: Not a bug if it is fixed

istarkov: ? The code Fixes the bug
What the issue?

apps/builder/app/canvas/instance-hovering.ts
```diff
@@ -44,27 +45,41 @@ export const subscribeInstanceHovering = () => {
...
-    if (instanceId === undefined) {
+  const updateHoveredRect = (
+    elements: Element[],
+    instanceSelector: Readonly<InstanceSelector>
+  ) => {
+    if (elements.length === 0) {
       return;
     }
+
+    if (instanceSelector.length === 0) {
```
TrySound, line 68: The utility protects from such case by returning undefined. No need to check here.

apps/builder/app/canvas/instance-hovering.ts
```diff
@@ -44,27 +45,41 @@ export const subscribeInstanceHovering = () => {
...
+      return;
+    }
+
+    const instanceId = instanceSelector[0];
```
TrySound, line 72: const [instanceId] = instanceSelector

apps/builder/app/canvas/instance-hovering.ts
istarkov: Until I see type 
type NonEmptyArray<T> = [T, ...T[]]
I would check length, as its some internal implementation detail of selector not exposed somehow

TrySound (approved):

PR #2018, opened 2023-08-01
Setup for `cli` to interact with the deployed projects

JayaKrishnaNamburu:
First and very initial support for the webstudio cli


apps/cli/.eslintrc.cjs
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  root: true,
+  extends: ["@webstudio-is/eslint-config-custom"],
```
TrySound, line 3: We have config in the root. This should not be necessary. You can tweak no-console with overrides in root config specifically for cli.

apps/cli/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "description": "Webstudio CLI",
+  "author": "Webstudio <github@webstudio.is>",
+  "homepage": "https://webstudio.is",
+  "type": "module",
+  "bin": {
+    "webstudio": "./bin/cli.js"
+  },
+  "scripts": {
+    "format": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "build": "pnpm format && pnpm clean && esbuild ./src/* --outdir=./lib --platform=node",
```
TrySound, line 13: We have build-package command in scripts package which covers edge cases

apps/cli/src/constants.ts
```diff
@@ -0,0 +1,33 @@
+import { join } from "node:path";
+import envPaths from "env-paths";
+
+const APP_CONFIG = envPaths("webstudio").config;
+const GLOBAL_CONFIG_FILE_NAME = "webstudio-config.json";
+export const GLOBAL_CONFIG_FILE = join(APP_CONFIG, GLOBAL_CONFIG_FILE_NAME);
+
+export const CLI_ARGS_OPTIONS = {
+  debug: {
+    type: <const>"boolean",
```
TrySound, line 10: We do not use such syntax caz it is conflicts with jsx. Better do this instead
`export const CLI_ARGS_OPTIONS = { ... } as const`

apps/cli/src/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+    short: "v",
+  },
+  help: {
+    type: <const>"boolean",
+    short: "h",
+  },
+};
+
+export const HELP = `Usage:
```
TrySound, line 24: Code looks a bit fragmented with all these constants, utils, types. Looks like .html,  .css and .js. Let's group logic related to parsing/help like args.ts with showHelp and parseCommands

apps/cli/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const commands: SupportedCommands = {
+  link,
+};
+
+export const main = async () => {
+  try {
+    await ensureFileInPath(GLOBAL_CONFIG_FILE, JSON.stringify({}));
+    const args = parseArgs({
+      args: process.argv,
```
TrySound, line 16: Usually process.argv.slice(2) is passed to parsers so output never contain system stuff.

pnpm-lock.yaml
```diff
@@ -16018,3 +16192,25 @@ packages:
...
+      chalk: 5.3.0
+      fs-extra: 11.1.1
+      fx: 28.0.0
+      globby: 13.2.2
+      minimist: 1.2.8
+      node-fetch: 3.3.1
+      ps-tree: 1.2.0
+      webpod: 0.0.2
+      which: 3.0.1
+      yaml: 2.3.1
```
TrySound, line 16215: ouch

apps/cli/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+      options: CLI_ARGS_OPTIONS,
+      allowPositionals: true,
+    });
+
+    $.verbose = args.values.debug || false;
```
TrySound, line 21: Looks like we do not yet use zx. Let's remove it and see if it really necessary in other commands.

apps/cli/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "@webstudio-is/tsconfig/base.json",
+  "include": ["src"],
+  "compilerOptions": {
+    "baseUrl": ".",
+    "customConditions": null,
+    "incremental": true
+  }
```
TrySound, line 8: Do we need any of these options?

pnpm-lock.yaml
TrySound: These are zx dependencies

apps/cli/bin/cli.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node --experimental-specifier-resolution=node
+/* eslint-disable-next-line */
+import { main } from "../lib/index";
+
+main().then(
```
istarkov, line 5: mb top level await?
```
await main()
```

apps/cli/src/utils.ts
```diff
@@ -0,0 +1,38 @@
+import { dirname } from "node:path";
```
istarkov, line 1: The whole file has different error catching strategy.
- `deleteFolderIfExists` catches rm and then just console that `Error while deleting the folder`

others do not catch 
`await writeFile`
`await mkdir`
And use try/catch for logic.

To make consistent I will re-throw error at `deleteFolderIfExists` instead of `console.error("Error while deleting the folder: \n", err);`

istarkov (approved):

apps/cli/src/index.ts
JayaKrishnaNamburu: We need it in `build` stages. But i am removing it. We can achieve the same using `child_proecess` something like that. `zx` seems to be overkill for just 1-2 commands 👍

apps/cli/src/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  name: string;
+  content: string;
+  encoding: "utf-8";
+}
+
+export interface Folder {
+  name: string;
+  files: File[];
+  subFolders: Folder[];
+}
```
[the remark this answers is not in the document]
JayaKrishnaNamburu: Ah they are being used in templates generation. Will remove now, and add back when adding support for templates.

apps/cli/tsconfig.json
JayaKrishnaNamburu: We don't need them, we just need `compilerOptions.module` to be set to `module` to support nodejs native `import` assertions for json improts.

kof: could this be a https://github.com/webstudio-is/webstudio-builder/tree/main/packages/tsconfig src/node ? so we have a standard ts config for all node native scripts?

.gitignore
```diff
@@ -51,6 +51,7 @@ packages/**/lib
 generated
 storybook-static
 tsconfig.tsbuildinfo
+lib/
```
TrySound, line 54: apps/cli/lib

apps/cli/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "bin": {
+    "webstudio": "./bin/cli.js"
+  },
+  "scripts": {
+    "dev": "build-package --watch",
+    "build": "build-package"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@webstudio-is/scripts": "workspace:^",
```
TrySound, line 17: dev dep

apps/cli/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "0.83.0",
+  "description": "Webstudio CLI",
+  "author": "Webstudio <github@webstudio.is>",
+  "homepage": "https://webstudio.is",
+  "type": "module",
+  "bin": {
+    "webstudio": "./bin/cli.js"
+  },
+  "scripts": {
+    "dev": "build-package --watch",
```
TrySound, line 12: No need dev here. We use it only for remix app dependencies.

apps/cli/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "@webstudio-is/tsconfig/base.json",
+  "compilerOptions": {
+    "module": "esnext"
```
TrySound, line 4: Base already specifies latest. Doesn't it work for you?

apps/cli/src/args.ts
```diff
@@ -0,0 +1,40 @@
+import { parseArgs } from "node:util";
+
+export const showHelp = () => console.log(HELP);
```
TrySound, line 3: I'd use console.info for output. log is often used for debug. At least in my experience.

apps/cli/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      console.log(packageJSON.version);
+      return;
+    }
+
+    if (args.values?.help) {
+      showHelp();
+      return;
+    }
+
+    const command = commands[args.positionals[2] as Commands];
```
TrySound, line 32: Wrong index. Already skipped system positionals above.

apps/cli/src/args.ts
kof: oh this is a nice catch, I didn't know about console.info, we should prevent console.log from linter, so debugging statements don't get commited by accident and only commit console.info where needed

.gitignore
kof: why adding here something that is specific to one app? it can be a local .gitignore file unless its literally needed everywhere

apps/cli/src/args.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export type Command = (
+  args: Pick<
+    ReturnType<
+      typeof parseArgs<{
+        options: typeof CLI_ARGS_OPTIONS;
+      }>
+    >,
+    "values"
```
kof, line 16: hard to read, better split it

apps/cli/src/args.ts
```diff
@@ -0,0 +1,40 @@
...
+    short: "v",
+  },
+  help: {
+    type: "boolean" as const,
+    short: "h",
+  },
+};
+
+export const HELP = `Usage:
+    $ webstudio commands [flags...]
```
kof, line 33: looks like too much indentation? or is it correct?

apps/cli/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+import { join } from "node:path";
+import envPaths from "env-paths";
+
+const APP_CONFIG = envPaths("webstudio").config;
+const GLOBAL_CONFIG_FILE_NAME = "webstudio-config.json";
```
kof, line 5: why not just webstudio.json ?

apps/cli/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { SupportedCommands } from "./args";
+import packageJSON from "../package.json" assert { type: "json" };
+
+const commands: SupportedCommands = {
+  link,
+};
+
+export const main = async () => {
+  try {
+    await ensureFileInPath(GLOBAL_CONFIG_FILE, JSON.stringify({}));
```
kof, line 15:
```suggestion
    await ensureFileInPath(GLOBAL_CONFIG_FILE, '{}');
```

apps/cli/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    const args = parseArgs({
+      args: process.argv.slice(2),
+      options: CLI_ARGS_OPTIONS,
+      allowPositionals: true,
+    });
+
+    if (args.values?.version) {
+      console.log(packageJSON.version);
```
kof, line 23: same, lets switch to console.info everywhere and also add a linter against console.log, separate pr ofc

apps/cli/src/link.ts
```diff
@@ -0,0 +1,47 @@
...
+  try {
+    const shareLinkUrl = new URL(shareLink);
+    const host = shareLinkUrl.origin;
+    const token = shareLinkUrl.searchParams.get("authToken");
+    const paths = shareLinkUrl.pathname.split("/").filter(Boolean);
+    if (paths[0] !== "builder" || paths.length !== 2) {
+      throw new Error("Invalid share link.");
+    }
+    const projectId = paths[1];
+    if (!token || !projectId || !host) {
```
kof, line 20: !i

apps/cli/src/link.ts
```diff
@@ -0,0 +1,47 @@
...
+      ...currentConfigJson,
+      [projectId]: {
+        host,
+        token,
+      },
+    };
+
+    await writeFile(GLOBAL_CONFIG_FILE, JSON.stringify(newConfig, null, 2));
+    rl.close();
+    console.log(`Saved credentials for project ${projectId}. \n
```
kof, line 37: - console.info
- looks like you are using a combination of \n and a newline

apps/cli/src/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+import { access, mkdir, writeFile, constants, rm } from "node:fs/promises";
+
+export const ensureFileInPath = async (filePath: string, content?: string) => {
+  const dir = dirname(filePath);
+
+  await ensureFolderExists(dir);
+
+  try {
+    await access(filePath, constants.F_OK);
+  } catch (err) {
```
kof, line 11:
```suggestion
  } catch (error) {
```

apps/cli/src/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    await writeFile(filePath, content || "", "utf8");
+  }
+};
+
+export const ensureFolderExists = async (folderPath: string) => {
+  try {
+    await access(folderPath, constants.F_OK);
+  } catch (err) {
```
kof, line 19:
```suggestion
  } catch (error) {
```

apps/cli/src/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    await mkdir(folderPath, { recursive: true });
+  }
+};
+
+export const deleteFolderIfExists = async (generatedDir: string) => {
+  try {
+    await rm(generatedDir, { recursive: true });
+    console.log("Folder deleted successfully (if it existed).");
+  } catch (err) {
```
kof, line 28:
```suggestion
  } catch (errror) {
```

kof (approved):
added some more stylistic comments, otherwise loooks good

didn't test

apps/cli/tsconfig.json
JayaKrishnaNamburu: Yeah, since it runs on  node@18 directly and we need `import-assertions` and ts needs the `module` to be `esnext` to enable it. The base config seems to be `es2020`

apps/cli/bin/cli.js
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node --experimental-specifier-resolution=node
+/* eslint-disable-next-line */
+import { main } from "../lib/index";
```
TrySound, line 3: bin.js

apps/cli/src/index.ts
```diff
@@ -0,0 +1,42 @@
+import { parseArgs } from "node:util";
```
TrySound, line 1: Cli.ts

apps/cli/package.json
JayaKrishnaNamburu: Yeah, but good to have when working on the `cli` to make changes and comes in handy :)

apps/cli/src/constants.ts
JayaKrishnaNamburu: This is for global config. On the local level for the project it can be just `webstudio.json`. Or both can be the same,

apps/cli/src/args.ts
JayaKrishnaNamburu: Yeah, when printed in console. It's good 👍

apps/cli/src/link.ts
```diff
@@ -0,0 +1,47 @@
...
+      ...currentConfigJson,
+      [projectId]: {
+        host,
+        token,
+      },
+    };
+
+    await writeFile(GLOBAL_CONFIG_FILE, JSON.stringify(newConfig, null, 2));
+    rl.close();
+    console.info(`Saved credentials for project ${projectId}.
```
kof, line 37: to prevent this not nicely formatted stuff I see 2 options:
1. using arrays then join
2. use https://www.npmjs.com/package/strip-indent

kof: I think stripIndent is nicer to read

PR #2022, opened 2023-08-01
Add radix tabs component

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/pull/1516

- https://www.radix-ui.com/docs/primitives/components/tabs
- https://ui.shadcn.com/docs/components/tabs

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react-radix/src/tabs.ws.ts
```diff
@@ -0,0 +1,197 @@
...
+  type: "container",
+  label: "Tabs List",
+  icon: RadioCheckedIcon,
+  requiredAncestors: ["Tabs"],
+};
+
+export const metaTabsTrigger: WsComponentMeta = {
+  category: "hidden",
+  type: "container",
+  label: "Tabs Trigger",
```
kof, line 153: Tab Trigger

packages/sdk-components-react-radix/src/tabs.ws.ts
```diff
@@ -0,0 +1,197 @@
...
+  icon: RadioCheckedIcon,
+  requiredAncestors: ["TabsList"],
+  invalidAncestors: ["TabsTrigger"],
```
kof, line 156: TabTrigger

kof:
Having Value as number on Tabs component and not having it anywhere else nor being able to select from a list is gonna be quite hard to understand, what can we do about it?


Screenshot 2023-08-02 at 16 36 42

kof:
- [ ] changing the value on trigger and content and then deleting the value seems to not restore the original functionality

packages/sdk-components-react-radix/src/tabs.ws.ts
TrySound: Tabs is a feature namespace. Called the same in radix. Consistent with all other components.

TrySound:
> Having Value as number on Tabs component and not having it anywhere else nor being able to select from a list is gonna be quite hard to understand, what can we do about it?

So you don't want index to be internal thing?

kof:
> So you don't want index to be internal thing?

The problem is that UI doesn't explain to user what to do. How will one know what index to set? Ideally its a select box, not an input

TrySound:
We don't have a way to use select. This would require some querying dsl.

TrySound:
What we could do is to have prop extension. This way we can show fallback when user clear the input and not loose the instance id when user change the value.

```
{
  type: 'string',
  value: '',
  fallback: {
    type: 'index',
    withinInstanceId: '...'
  }
}
```

istarkov (approved):

PR #2033, opened 2023-08-03
Sync project data to the local config file using cli's sync command

JayaKrishnaNamburu:
Ref https://github.com/webstudio-is/webstudio-builder/issues/1825

## Description

PR adds the `sync` command to the `cli`. This can run in two modes.

```sh
webstudio sync <projectId>
```

which will create a new project folder and loads the build data into `<project>/webstudio.json` file.

```sh
webstudio sync
```

This mode is to sync the existing project with the latest build data.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review


apps/cli/src/args.ts
```diff
@@ -31,12 +33,11 @@ export const CLI_ARGS_OPTIONS = {
...
-    Flags:
-    --help, -h                      Show this help message
-    --version, -v                   Show the version of this script
+export const HELP = `\n Usage:
+\t $ webstudio commands [flags...] \n
+\t Commands:
+\t\t link <shared link> \t Login to Webstudio with shared link \n
+\t Flags:
+\t\t --help, -h \t\t Show this help message
+\t\t --version, -v \t\t Show the version of this script
```
TrySound, line 42: what was wrong with previous output?

JayaKrishnaNamburu: @kof suggested to switch to this on the previous PR
https://github.com/webstudio-is/webstudio-builder/pull/2018#discussion_r1281579669

TrySound: But it is like same but with explicit space characters. Wouldn't say it is nicer to read. I'd keep it same as before.

Adding dependency just for this case is not good.

kof: oh this looks like a mess, its not how I meant it the idea is to use stripIndent to actually make it look nice:
```
stripIndent(`
  Usage:
  ...
`)

Comment (TrySound) [reply apps/cli/src/args.ts:42]: I wouldn't say this gives too much value. Top level help template is good as is. Error messages are fine with separate strings. Adding dependency would not benefit end user.

File: apps/cli/src/commands/sync.ts [hunk 2]
```diff
@@ -0,0 +1,70 @@
...
+    a webstudio project. And load the projectId from it to sync the data.
+  */
+
+  if (projectId === undefined) {
+    try {
+      const content = await readFile(
+        join(cwd(), LOCAL_APP_CONFIG_FILE_NAME),
+        "utf8"
+      );
+      const build: Data = JSON.parse(content);
```
TrySound, line 28: Why local app config has type of project data?

JayaKrishnaNamburu: Because cli just saves all the data that comes from the api to the local file. Maybe the name is mis-leading, should it be changed 🤔

TrySound: It's just weird we havee the same file name for root config and webstudio-project folder. Why data is stored in the root and not in webstudio-project. And what webstudio-project even for?

TrySound: Yep, description says the opposite.

apps/cli/src/constants.ts
```diff
@@ -1,6 +1,24 @@
...
+  GLOBAL_CONFIG_FILE_NAME
+);
+
+export const LOCAL_APP_NAME = "webstudio-project";
+export const LOCAL_APP_CONFIG_FILE_NAME = "webstudio.json";
+export const LOCAL_CONFIG_FILE = join(
+  process.cwd(),
+  LOCAL_APP_NAME,
+  LOCAL_APP_CONFIG_FILE_NAME
+);
```
TrySound, line 17: Let's expose only resolved paths and have separate names for json with data and json with config.

apps/cli/src/args.ts
JayaKrishnaNamburu: Should i switch back to the old way ?

kof: @TrySound there is a lot of ugly indentation going on when using plain template strings, but JK did the wrong thing, so now it looks even worse, the idea  with stripIndent is to be able to format everything just like you would normaly and this is not perf-sensitive case, dependency is tiny ... once you see how it has to look like, you will never want to go back to vanilla template strings for large blocks of text

kof: I used a similiar function for tests in JSS, without it, those strings would be ugly as hell

https://github.com/cssinjs/jss/blob/master/packages/jss/tests/integration/rules.js#L111

apps/cli/src/commands/sync.ts
TrySound: Token is stores globally no? Local one would be committed.

JayaKrishnaNamburu: Yes, if the command runs in a existing project. It opens the local config. Checks the `projectId` and then checks for `token` from the global config. And makes a call for the latest build.

apps/cli/src/constants.ts
```diff
@@ -1,6 +1,24 @@
 import { join } from "node:path";
 import envPaths from "env-paths";
 
-const APP_CONFIG = envPaths("webstudio").config;
+const GLOBA_APP_CONFIG = envPaths("webstudio").config;
```
TrySound, line 4: GlOBAL

apps/cli/src/constants.ts
TrySound: You can use `join(cwd(), webstudio.json)` here too. cwd does not change if you not call proces.chdir

apps/cli/src/commands/sync.ts
TrySound: The type says you store data in root/webstudio.json. Not in root/webstudio-project/webstudio.json.

JayaKrishnaNamburu: I am confused a bit, can you clarify the question once more please. Here is the behaviour with the data. 
When loaded for the first time, it stores in 

```js
export const LOCAL_APP_NAME = "webstudio-project";
export const LOCAL_APP_CONFIG_FILE_NAME = "build.json";
export const LOCAL_CONFIG_FILE = join(
  process.cwd(),
  LOCAL_APP_NAME,
  LOCAL_APP_CONFIG_FILE_NAME
);
```

`/root/user/...../webstudio-project/webstudio.json`

https://github.com/webstudio-is/webstudio-builder/blob/feat/sync-project-data/apps/cli/src/commands/sync.ts#L15

But when the projectId is null in the command. We assume that is being run inside a existing project and check if the json file is present and so the
https://github.com/webstudio-is/webstudio-builder/blob/feat/sync-project-data/apps/cli/src/commands/sync.ts#L25
which is `join(cwd(), LOCAL_APP_CONFIG_FILE_NAME)`
Because `cwd` here already constructs the path till `root/user/...../webstudio-project/` and so i am joining till `join(cwd(), LOCAL_APP_CONFIG_FILE_NAME)`


If there is any doubts on the token, the path can always be the same
```js
const GLOBA_APP_CONFIG = envPaths("webstudio").config;
const GLOBAL_CONFIG_FILE_NAME = "webstudio-config.json";
export const GLOBAL_CONFIG_FILE = join(
  GLOBA_APP_CONFIG,
  GLOBAL_CONFIG_FILE_NAME
);
```
that is going to be `root/...webstudio-nodejs/webstudio-config.json`. The package `env-paths` creates the `webstudio-nodejs` here

JayaKrishnaNamburu:
Changed the local file name where the project data is stored to `build.json` file to avoid confusion now 👍

apps/cli/src/args.ts
```diff
@@ -1,9 +1,22 @@
...
+import stripIndent from "strip-indent";
 
-export const showHelp = () => console.info(HELP);
+export const showHelp = () =>
+  console.info(
+    stripIndent(`
+Usage:
+  $ webstudio commands [flags...] \n
+Commands:
+  link <shared link> \t Login to Webstudio with shared link \n
```
TrySound, line 10: Need to describe sync here

apps/cli/src/commands/sync.ts
```diff
@@ -0,0 +1,70 @@
...
+import ora from "ora";
+import { loadProjectDataById } from "@webstudio-is/http-client";
+
+import type { Command } from "../args";
+import { ensureFileInPath, loadGlobalConfigFile } from "../fs-utils";
+import { LOCAL_APP_CONFIG_FILE_NAME, LOCAL_CONFIG_FILE } from "../constants";
+
+export const sync: Command = async (args) => {
+  const spinner = ora("Syncing project data").start();
+  let projectId = args.positionals[1];
```
TrySound, line 14: A bit confusing positionals start with 1. Maybe slice command name before pass here?

apps/cli/src/commands/sync.ts
TrySound: Now I get it. So we just have inconsistent behavior.

Now the question is why sync with project and without project id write into different folders? Should user do `cd webstudio-project && cli sync && cd ../`?

JayaKrishnaNamburu: Yes, my assumption was. Once the project is closed, there shouldn't be any need for running commands from outside. So, the `sync` and `build` should run in the current folder. So, anyone can have multiple projects in the same folder. And each can work independently. I am actually looking into `webstudio-project` folder name to be the actual project name.

apps/cli/src/commands/sync.ts
JayaKrishnaNamburu: good idea, as we don't need it inside anyways 👍

apps/cli/src/commands/sync.ts
TrySound: Not quite safe as title may have characters not compatible with fs

TrySound: Maybe something like this?
```
$ webstudio sync FOLDER_NAME PROJECT_ID
$ webstudio sync FOLDER_NAME
```

JayaKrishnaNamburu: Ok, this suggestion sounds good. Will make all the final changes then. Thanks for clarifying 👍

TrySound:
Discussed with Oleg. Let's make it like this

```
$ webstudio link
# prompt for url
# store `.webstudio/config.json` with `{ projectId: string }`
# store `envPaths('webstudio').config` with `{ [projectId]: { ... } }`
```

```
$ webstudio sync
# use `.webstudio/config.json` to get liked project id
# load project data and store in `.webstudio/data.json`
```

Each repo/workspace would have only one webstudio project. If need more than one user can setup monorepo.

kof:
`webstudio init` would make it clear thats you need to do as a first step instead of `link`

JayaKrishnaNamburu:
Sounds good, what's your take on renaming `link` to `init` @TrySound

TrySound:
Init sounds more like it should do all link/sync/build together so user could only start the app after.

So let's keep link for now and see if it fits into workflow.

apps/cli/src/args.ts
```diff
@@ -1,10 +1,21 @@
...
-  "link" = "link",
-}
+Commands:
+  link       Link to an existing webstudio project
+  sync       Sync the linked webstudio project with the latest build
+
+  Flags:
+  --help     -h     Show this help message
+  --version  -v     Show the version of this script
+`)
```
TrySound, line 17: you can add 6 spaces before each line to actually leverage strip indent here.

apps/cli/src/commands/link.ts
```diff
@@ -0,0 +1,57 @@
+import { stdin as input, stdout as output, cwd } from "node:process";
```
TrySound, line 1: input, output are quite generic name and often used as arguments. Better pass input:stdin to createInterface for readability.

apps/cli/src/commands/link.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ensureFileInPath } from "../fs-utils";
+
+export const link: Command = async () => {
+  const rl = readline.createInterface({ input, output });
+  const shareLink = await rl.question(`Paste share link (with build access): `);
+
+  const shareLinkUrl = new URL(shareLink);
+  const host = shareLinkUrl.origin;
+  const token = shareLinkUrl.searchParams.get("authToken");
+  const paths = shareLinkUrl.pathname.split("/").filter(Boolean);
```
TrySound, line 20: Do you know what filter(boolean) for?

apps/cli/src/constants.ts
```diff
@@ -1,6 +1,24 @@
...
+const GLOBAL_APP_CONFIG = envPaths("webstudio").config;
 const GLOBAL_CONFIG_FILE_NAME = "webstudio-config.json";
-export const GLOBAL_CONFIG_FILE = join(APP_CONFIG, GLOBAL_CONFIG_FILE_NAME);
+export const GLOBAL_CONFIG_FILE = join(
+  GLOBAL_APP_CONFIG,
+  GLOBAL_CONFIG_FILE_NAME
+);
+
+export const PROJECT_NAME = "webstudio-project";
+export const LOCAL_CONFIG_FILE_NAME = ".webstudio/config.json";
```
TrySound, line 12: join(cwd(), ".webstudio/config.json")

PROJECT_NAME not needed at all

apps/cli/src/constants.ts
```diff
@@ -1,6 +1,24 @@
 import { join } from "node:path";
 import envPaths from "env-paths";
 
-const APP_CONFIG = envPaths("webstudio").config;
+const GLOBAL_APP_CONFIG = envPaths("webstudio").config;
```
TrySound, line 4: GLOBAL_CONFIG_FOLDER

apps/cli/src/constants.ts
```diff
@@ -1,6 +1,24 @@
...
 const GLOBAL_CONFIG_FILE_NAME = "webstudio-config.json";
-export const GLOBAL_CONFIG_FILE = join(APP_CONFIG, GLOBAL_CONFIG_FILE_NAME);
+export const GLOBAL_CONFIG_FILE = join(
+  GLOBAL_APP_CONFIG,
+  GLOBAL_CONFIG_FILE_NAME
+);
+
+export const PROJECT_NAME = "webstudio-project";
+export const LOCAL_CONFIG_FILE_NAME = ".webstudio/config.json";
+export const LOCAL_BUILD_FILE_NAME = ".webstudio/data.json";
```
TrySound, line 13: LOCAL_DATA_FILE

apps/cli/src/constants.ts
```diff
@@ -1,6 +1,24 @@
 import { join } from "node:path";
 import envPaths from "env-paths";
```
TrySound, line 3: Btw this module can be called config.ts

apps/cli/src/commands/sync.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  spinner.text = "Loading project data from webstudio";
+  const project = await loadProjectDataById({
+    projectId: localConfig.projectId,
+    authToken: token,
+    host: host,
+  });
+
+  const localBuildFilePath = join(cwd(), LOCAL_BUILD_FILE_NAME);
+  await ensureFileInPath(localBuildFilePath);
```
TrySound, line 50: put this after message along with write

apps/cli/src/commands/link.ts
JayaKrishnaNamburu: No idea, i will re-check the logic then. Just copied it from before 👍

apps/cli/src/constants.ts
JayaKrishnaNamburu: Changing it, But that means, users should create a empty folder and then run these commands inside of it right. Unlike cli tooks like `cra` etc which creates the empty project and clones everything inside of it.

TrySound: See, we want single ws project per package.json. if user want multiple projects then it can be monorepo.

apps/cli/src/cli.ts
```diff
@@ -1,42 +1,49 @@
...
-      args: process.argv.slice(2),
+      args: argv.slice(2),
       options: CLI_ARGS_OPTIONS,
       allowPositionals: true,
     });
 
     if (args.values?.version) {
       console.info(packageJSON.version);
-      return;
+      exit(0);
```
TrySound, line 28: You already call exit(0) in bin.js

TrySound (approved):

kof:
🎉

PR #2035, opened 2023-08-04
Add ability to build templates from the templates folder for scaffolding projects in cli

JayaKrishnaNamburu:
## Description

This PR introduces a build step for cli to generate templates which is needed in the next stages for building projects from the `sync` command from #2033 

## Code Review

## Before requesting a review

- [x] made a self-review


JayaKrishnaNamburu:
Checking the build,  looks like should add ignore rules for the `templates` folder

apps/cli/src/__generated__/templates.ts
```diff
@@ -0,0 +1,84 @@
+/*
+
+This is a auto-generated file. Please don't change manually.
+If needed to make any changes. Add them to ./templates folder and run pnpm run build:templates
+
+*/
+
+import type { Folder, ProjectType } from "../utils";
+
+export const TEMPLATES: Record<ProjectType, Folder> = {
```
kof, line 10: personally i would go for 
```suggestion
export const templates: Record<ProjectType, Folder> = {
```

kof: @TrySound wdyt, whats our convention, its kinda weird because const is used a lot of times with camel case, I used to think when something is exported as a constant and is never changing, it should be all caps, but it might be also a bit annoying, so not sure here, I will go with whatever you and @istarkov think is best

apps/cli/src/templates.ts
```diff
@@ -0,0 +1,84 @@
+/*
```
kof, line 1: why are we generated into `__generated__` and into source?

apps/cli/src/utils.ts
```diff
@@ -1,6 +1,24 @@
 import { dirname } from "node:path";
 import { access, mkdir, writeFile, constants, rm } from "node:fs/promises";
 
+export interface File {
```
kof, line 4: do you plan to extend this interface? if not it should be a Type

apps/cli/src/utils.ts
```diff
@@ -1,6 +1,24 @@
...
+  encoding: "utf-8";
+}
+
+export interface Folder {
+  name: string;
+  files: File[];
+  subFolders: Folder[];
+}
+
+export enum ProjectType {
```
kof, line 16: we don't use enums at all :)

TS enums is a mistake, they are hiding whats behind them and TS team already admitted it was a mistake. They try to not introduce runtime code any more.

apps/cli/templates/build.ts
```diff
@@ -0,0 +1,85 @@
+import { readdirSync, lstatSync, readFileSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { type Folder, ProjectType, ensureFileInPath } from "../src/utils";
+
+const TEMPLATES_TO_BUILD = Object.keys(ProjectType);
+
+console.log("Building Templates...");
```
kof, line 7: we discussed to use console.info and add a linter for .log

apps/cli/templates/build.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const folder: Folder = {
+    name: folderName,
+    files: [],
+    subFolders: [],
+  };
+
+  const files = readdirSync(folderPath);
+
+  for (let i = 0; i < files.length; i++) {
```
kof, line 27: 2 problems: 
1. we say there should be no single char variables, otherwise people tend to use them where it becomes unreadable
2. why is this not a `for of` ?

apps/cli/templates/build.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+const TEMPLATES_JSON: Record<string, Folder> = {};
+const TEMPLATES_JSON_PATH = new URL(
+  "../src/__generated__/templates.ts",
+  import.meta.url
+).pathname;
+await ensureFileInPath(TEMPLATES_JSON_PATH);
+
+for (let i = 0; i < TEMPLATES_TO_BUILD.length; i++) {
```
kof, line 61: same here, please check everywhere else

apps/cli/templates/defaults/remix.config.js
```diff
@@ -0,0 +1,28 @@
+/** @type {import('@remix-run/dev').AppConfig} */
+module.exports = {
```
kof, line 2: I think this config is designed for the builder and you don't need in the cli a lot of these things

apps/cli/templates/build.ts
JayaKrishnaNamburu: This happens on the dev phase, not for the users side. We import those `.json` files in the CLI while using the template.

apps/cli/templates/vercel/remix.config.js
```diff
@@ -0,0 +1,31 @@
+/** @type {import('@remix-run/dev').AppConfig} */
```
kof, line 1: same here

kof: btw why do we 2 identical configs?

apps/cli/templates/build.ts
kof: so why do we need .log instead of .info?

apps/cli/templates/defaults/remix.config.js
JayaKrishnaNamburu: These are all config files. Which are bunched into the `template.json` file. These are again generated on the fly into the end build to create a project. Since the cli runs `npm install` `npm build` to generate the builds. These files need to sit along with the users projects.

apps/cli/src/templates.ts
JayaKrishnaNamburu: We import and use the populated `template.json` instead of read the files from FS. Should we keep these outside of `src`. But still they are imported into the code for populating files.

kof: I don't get it, we have now 2 identical templates.ts files:
`
src/templates.ts
src/__generated__templates.ts
`

JayaKrishnaNamburu: I should be deleting the `src/templates.ts` file. Committed it by mistake 👍

apps/cli/templates/vercel/remix.config.js
JayaKrishnaNamburu: Not needed, we should pick same files from the `defaults` folder 👍

apps/cli/templates/defaults/remix.config.js
kof: sure but I don't know why you need all that stuff inside of it, I think you copied it from the builder, but cli for published sites doesn't have these dependences and requirements

apps/cli/templates/defaults/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "dependencies": {
+    "@webstudio-is/react-sdk": "^0.82.0",
+    "@webstudio-is/sdk-components-react-remix": "^0.82.0",
+    "@webstudio-is/sdk-components-react": "^0.82.0",
+    "@webstudio-is/form-handlers": "^0.75.0",
+    "use-sync-external-store": "^1.2.0"
```
TrySound, line 7: What use-sync-external-store for? Let's remove everywhere. I would not support react < v18

apps/cli/templates/vercel/remix.config.js
JayaKrishnaNamburu: `vercel` seems to need their own config for server.

```js
server: process.env.NODE_ENV === "development" ? undefined : "./server.ts",
```

Not sure, if patching the default with this line is worth the time spent in dev setup at the moment. So, just left the duplicate file. Until there can be a scalable setup for this thing. To extend the default with specific template changes.

JayaKrishnaNamburu:
Since #2018  and #2033  adds support for `link` and `sync` this will add support for the `build` command. Which consumes the data that `snync` stores inside `.webstudio/data.json` and then merges the template and generates the files from the data. Basically the PR adds two steps in the CLI flow.

- Template management, where the local files are converted into a JSON that is consumed at runtime to merge into the project.
- Prebuild step where the data is consumed and files are generated and merged with the template files from the above step.

There are still more setups for the followups
- [ ] downloading original assets.
- [ ] Downloading any fonts that are being consumed.
- [ ] Install the deps once the project is build and then start the server if needed.

apps/cli/src/args.ts
```diff
@@ -45,4 +45,29 @@ export const CLI_ARGS_OPTIONS = {
...
+  encoding: "utf-8";
+};
+
+export type Folder = {
+  name: string;
+  files: File[];
+  subFolders: Folder[];
+};
+
+export type ProjectType = "defaults" | "vercel" | "remix-app-server";
```
kof, line 67: ProjectType is alright, but Destination or Platform or Framework would be more clear, kinda tryint to describe this thing more accurately. "defaults" is not very clear what it is but if that's the only option then maybe "default" is more correct. I would still prefer a name that is descriptive

apps/cli/src/config.ts
```diff
@@ -11,6 +11,8 @@ export const GLOBAL_CONFIG_FILE = join(
 export const LOCAL_CONFIG_FILE = ".webstudio/config.json";
 export const LOCAL_DATA_FILE = ".webstudio/data.json";
 
+export const ASSETS_BASE = "/cgi/asset/";
```
kof, line 14: @istarkov is it ok to hardcode this? or should it be reusable from somewhere?

kof (approved):

kof:
Approved, but haven't tested

apps/cli/src/config.ts
JayaKrishnaNamburu: At the moment this is just being used as a placeholder. Can someone point me to the URL's of the original assets. Then we can download and pass them to that location. Right now, this is just being used a `prefix` that goes under `public` folder 👍

kof: ah so this is your folder, not our public one, then its fine, I would have named it in a way that tells this: original-assets or just assets .... that cgi prefix made me think its a path to our cgi script

@istarkov how does one get the original files downloaded? I am guessing they aren't even available on the public domain? 

Does cli need to use http api and talk to R2 directly like we do during upload?

apps/cli/package.json
```diff
@@ -9,19 +9,27 @@
     "webstudio": "./bin.js"
   },
   "scripts": {
-    "dev": "build-package --watch",
-    "build": "build-package"
+    "build:templates": "tsx ./templates/build.ts",
+    "dev": "pnpm build:templates && build-package --watch",
```
TrySound, line 13: Why do we need dev command here?

apps/cli/src/prebuild.ts
```diff
@@ -0,0 +1,250 @@
...
+    path: string;
+    file: string;
+  }>;
+};
+
+export const prebuild = async () => {
+  const appRoot = "app";
+
+  const routesDir = join(appRoot, "routes");
+  await deleteFolderIfExists(routesDir);
```
TrySound, line 52:
```
await rm(routesDir, { recursive: true, force: true })
```

apps/cli/templates/defaults/remix.config.js
```diff
@@ -0,0 +1,24 @@
...
+    /@webstudio-is\/(?!prisma-client)/,
+    "pretty-bytes",
+    "nanoevents",
+    "nanostores",
+    "@nanostores/react",
+    /micromark/,
+    "decode-named-character-reference",
+    "character-entities",
+    /mdast-/,
+    "unist-util-stringify-position",
```
TrySound, line 15: Most of it is not used

apps/cli/templates/defaults/template.tsx
```diff
@@ -0,0 +1,140 @@
+// This file was generated by pnpm prebuild
+import {
+  V2_MetaFunction,
```
TrySound, line 3: You can avoid lint error with
```
V2_MetaFunction as MetaFunction
```

apps/cli/templates/defaults/template.tsx
```diff
@@ -0,0 +1,140 @@
+// This file was generated by pnpm prebuild
```
TrySound, line 1: Is it generated? Shouldn't __generated__/template be used for this?

apps/cli/templates/remix-app-server/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "devDependencies": {
+    "@remix-run/dev": "^1.17.1",
+    "@remix-run/eslint-config": "^1.17.1",
+    "@types/react": "^18.0.35",
+    "@types/react-dom": "^18.0.11",
+    "eslint": "^8.38.0",
+    "typescript": "^5.0.4"
+  },
+  "engines": {
+    "node": ">=14.0.0"
```
TrySound, line 29: Our minimum required node is 18

apps/cli/templates/defaults/template.tsx
JayaKrishnaNamburu: Actually not generated, it's a template file. That line should be removed 👍

apps/cli/templates/defaults/remix.config.js
JayaKrishnaNamburu: I didn't do a detailed check of each dep. But some of them like `@webstudio-is/` and `nanostores` and `nano-events` etc are sub-deps of some of them. So need to do a detailed check on what 's needed. Will check once by one

apps/cli/package.json
JayaKrishnaNamburu: I am actually using this, to test the `cli` while building itself by using `npm link`. Comes in handy while making changes and checking the same in real time.

apps/cli/templates/defaults/remix.config.js
JayaKrishnaNamburu: Did a check with all the packages and removed the ones that are not being used now 👍

apps/cli/package.json
TrySound: You could use tax for testing. No need in watcher.

TrySound:
What remix-app-server does? Any value for end user?

apps/cli/package.json
JayaKrishnaNamburu: Cool, removing it then. Didn't know about this one

JayaKrishnaNamburu:
> What remix-app-server does? Any value for end user?

That's the default template it's basically a remix project.

apps/cli/src/fs-utils.ts
```diff
@@ -47,3 +39,46 @@ export const loadJSONFile = async <T>(filePath: string): Promise<T | null> => {
     return null;
   }
 };
+
+export const scaffoldProjectTemplate = async (
```
TrySound, line 43: Used only in build. Wouldn't call it "utils". Kinda core logic of build command. Let's move into build.ts

apps/cli/templates/vercel/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@remix-run/vercel": "^1.17.1",
+    "@vercel/node": "^2.10.3"
```
TrySound, line 4: Tricky one. Vercel now maintains own `@vercel/remix` as a replacement for these two. `@remix-run/vercel` is gonna be deprecated.

apps/cli/templates/defaults/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "sideEffects": false,
+  "scripts": {
+    "build": "remix build",
+    "dev": "remix dev",
+    "typecheck": "tsc"
+  },
+  "dependencies": {
+    "@remix-run/css-bundle": "^1.17.1",
+    "@remix-run/node": "^1.17.1",
+    "@remix-run/react": "^1.17.1",
```
TrySound, line 12: Bump to 1.19.2 used everywhere else

apps/cli/templates/defaults/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@webstudio-is/react-sdk": "^0.82.0",
+    "@webstudio-is/sdk-components-react-remix": "^0.82.0",
+    "@webstudio-is/sdk-components-react": "^0.82.0",
+    "@webstudio-is/form-handlers": "^0.75.0",
```
TrySound, line 16: This is tricky too. Need to maintain these versions to match latest published ones.

apps/cli/templates/remix-app-server/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "scripts": {
+    "start": "remix-serve build"
```
TrySound, line 3: I would just merge defaults and remix-app-server templates

apps/cli/templates/vercel/package.json
JayaKrishnaNamburu: So, should this template adopt the `@vercel/remix` package then ?

apps/cli/templates/defaults/package.json
JayaKrishnaNamburu: In one way, in will help in making cli independent of the updates to the packages. And keep always be in sync. But in another context, this can actually help in tying up the `cli` version and the `sdk` version together. If there are any breaking changes in `sdk` the cli will pul the latest ones. And get's effected with the pre-exiting projects.
So, a new version of `cli` can be maintained every time the `@webstudio/` packages are released

apps/cli/templates/vercel/package.json
TrySound: Remix says we may not need remix template at all. Like vercel can install own stuff when remix is detected.

https://github.com/remix-run/remix/tree/main/templates/vercel

JayaKrishnaNamburu: ok, dropping vercel template then 👍

apps/cli/src/prebuild.ts
```diff
@@ -44,12 +44,12 @@ export const prebuild = async () => {
   const appRoot = "app";
 
   const routesDir = join(appRoot, "routes");
-  await rm(routesDir, { recursive: true, force: true });
-  await mkdir(routesDir, { recursive: true });
+  rm(routesDir, { recursive: true, force: true });
+  mkdir(routesDir, { recursive: true });
```
TrySound, line 48: These are callback based. Import promise based from node:fs/promises.

JayaKrishnaNamburu: yes yes, still making the changes in local 👍

JayaKrishnaNamburu:
Dropped vercel template, as vercel seems to auto-detect this. And one more important change is, while refactoring i noticed. That the `template.tsx` file is actually just as a placeholder. And not needed for the builds. But it is being written to the project, read back and deleted again. So, changed it to be a auto-generated file.
Two things comes from this.
- The `ASSET_BASE` becomes configurable.
- And we don't need to write the file, read it back and delete once again... :)

Let me know, if this can be merged 👍

apps/cli/src/config.ts
istarkov: http api, same as used for the builder

TrySound:
This will unblock data sources support in cli
https://github.com/webstudio-is/webstudio-builder/pull/2098

TrySound (approved):

PR #2064, opened 2023-08-07
Use stream consumers while uploading

TrySound:
Looks like node now provides nice utilities to convert stream into json, buffer, text etc. Super handy

https://nodejs.org/docs/latest-v18.x/api/webstreams.html#utility-consumers

Cloudflare node compat supports this as well
https://developers.cloudflare.com/workers/runtime-apis/nodejs/streams/

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

kof (approved):
I hope this is well tested?

- locally?
- preview with s3?

If we break this its gonna be bad

TrySound:
yep, works everywhere

PR #2069, opened 2023-08-08
Add component hooks

TrySound:
We want to control some builder behavior specific to some components. For example set open state on dialog, collapsible etc when select in navigator.

Here I added hooks to be called from builder and executed in canvas. For now added 2 hooks onNavigatorSelect and onNavigatorDeselect.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/navigator-tree.tsx
+++ b/apps/builder/app/builder/shared/navigator-tree.tsx
@@ -123,7 +123,7 @@ export const NavigatorTree = ({ publish }: { publish: Publish }) => {
           publish({
             type: "emitComponentHook",
             payload: {
-              name: "onNavigatorDeselect",
+              name: "onNavigatorUnselect",
               data: {
                 instanceSelection: previousInstanceSelector.flatMap((id) => {
                   const instance = instances.get(id);
--- a/apps/builder/app/builder/shared/navigator-tree.tsx
+++ b/apps/builder/app/builder/shared/navigator-tree.tsx
@@ -117,6 +117,7 @@ export const NavigatorTree = ({ publish }: { publish: Publish }) => {
         shallowEqual(selectedInstanceSelectorStore.get(), instanceSelector) ===
         false
       ) {
+        const instances = instancesStore.get();
         const previousInstanceSelector = selectedInstanceSelectorStore.get();
         if (previousInstanceSelector) {
           publish({
@@ -124,10 +125,10 @@ export const NavigatorTree = ({ publish }: { publish: Publish }) => {
             payload: {
               name: "onNavigatorDeselect",
               data: {
-                instanceSelector: previousInstanceSelector as [
-                  string,
-                  ...string[],
-                ],
+                instanceSelection: previousInstanceSelector.flatMap((id) => {
+                  const instance = instances.get(id);
+                  return instance ? [instance] : [];
+                }),
               },
             },
           });
@@ -140,7 +141,10 @@ export const NavigatorTree = ({ publish }: { publish: Publish }) => {
           payload: {
             name: "onNavigatorSelect",
             data: {
-              instanceSelector: instanceSelector as [string, ...string[]],
+              instanceSelection: instanceSelector.flatMap((id) => {
+                const instance = instances.get(id);
+                return instance ? [instance] : [];
+              }),
             },
           },
         });
--- a/apps/builder/app/builder/shared/navigator-tree.tsx
+++ b/apps/builder/app/builder/shared/navigator-tree.tsx
@@ -108,43 +108,46 @@ export const NavigatorTree = ({ publish }: { publish: Publish }) => {
     [setState]
   );
 
-  const handleSelect = useCallback((instanceSelector: InstanceSelector) => {
-    // TreeNode is refocused during "delete" hot key here https://github.com/webstudio-is/webstudio-builder/blob/5935d7818fba3739e4f16fe710ea468bf9d0ac78/packages/design-system/src/components/tree/tree.tsx#L435
-    // and then focus cause handleSelect to be called with the same instanceSelector
-    // This avoids additional rerender on node delete
-    if (
-      shallowEqual(selectedInstanceSelectorStore.get(), instanceSelector) ===
-      false
-    ) {
-      const previousInstanceSelector = selectedInstanceSelectorStore.get();
-      if (previousInstanceSelector) {
+  const handleSelect = useCallback(
+    (instanceSelector: InstanceSelector) => {
+      // TreeNode is refocused during "delete" hot key here https://github.com/webstudio-is/webstudio-builder/blob/5935d7818fba3739e4f16fe710ea468bf9d0ac78/packages/design-system/src/components/tree/tree.tsx#L435
+      // and then focus cause handleSelect to be called with the same instanceSelector
+      // This avoids additional rerender on node delete
+      if (
+        shallowEqual(selectedInstanceSelectorStore.get(), instanceSelector) ===
+        false
+      ) {
+        const previousInstanceSelector = selectedInstanceSelectorStore.get();
+        if (previousInstanceSelector) {
+          publish({
+            type: "emitComponentHook",
+            payload: {
+              name: "onNavigatorDeselect",
+              data: {
+                instanceSelector: previousInstanceSelector as [
+                  string,
+                  ...string[],
+                ],
+              },
+            },
+          });
+        }
+        selectedInstanceSelectorStore.set(instanceSelector);
+        textEditingInstanceSelectorStore.set(undefined);
+        selectedStyleSourceSelectorStore.set(undefined);
         publish({
           type: "emitComponentHook",
           payload: {
-            name: "onNavigatorDeselect",
+            name: "onNavigatorSelect",
             data: {
-              instanceSelector: previousInstanceSelector as [
-                string,
-                ...string[],
-              ],
+              instanceSelector: instanceSelector as [string, ...string[]],
             },
           },
         });
       }
-      selectedInstanceSelectorStore.set(instanceSelector);
-      textEditingInstanceSelectorStore.set(undefined);
-      selectedStyleSourceSelectorStore.set(undefined);
-      publish({
-        type: "emitComponentHook",
-        payload: {
-          name: "onNavigatorSelect",
-          data: {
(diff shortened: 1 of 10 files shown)


apps/builder/app/builder/shared/navigator-tree.tsx
```diff
@@ -115,9 +116,33 @@ export const NavigatorTree = () => {
       shallowEqual(selectedInstanceSelectorStore.get(), instanceSelector) ===
       false
     ) {
+      const previousInstanceSelector = selectedInstanceSelectorStore.get();
+      if (previousInstanceSelector) {
+        publish({
+          type: "emitComponentHook",
+          payload: {
+            name: "onNavigatorDeselect",
```
kof, line 124: nit: onNavigatorUnselect would be slightly better, it means one specific thing, where deselect is more like deselect everything

In the future we will have ability to select multiple items in navigator

apps/builder/app/canvas/canvas.tsx
```diff
@@ -193,9 +195,14 @@ export const Canvas = ({ params }: CanvasProps): JSX.Element | null => {
       components: radixComponents,
       metas: radixComponentMetas,
       propsMetas: radixComponentPropsMetas,
+      hooks: radixComponentHooks,
     });
   });
 
+  useEffect(() => {
+    return subscribeComponentHooks();
+  }, []);
```
kof, line 204: nit:
```suggestion
  useEffect(subscribeComponentHooks, []);
```

apps/builder/app/shared/nano-states/components.ts
```diff
@@ -5,11 +5,51 @@ import {
...
+  }[keyof Hook]
+>;
+
+declare module "~/shared/pubsub" {
+  export interface PubsubMap {
+    emitComponentHook: HookCommand;
+  }
+}
+
+export const subscribeComponentHooks = () => {
```
kof, line 31: interresting, so we have now a general way to emit events that are 1:1 mapped to hooks? or do I misunderstand something?

maybe worth a comment for what it is with an example of a problem its solving

packages/sdk-components-react-radix/src/collapsible.tsx
```diff
@@ -35,3 +40,70 @@ export const CollapsibleTrigger = forwardRef<
 export const CollapsibleContent: ForwardRefExoticComponent<
   Omit<ComponentPropsWithRef<typeof Content>, "asChild">
 > = Content;
+
+/* BUILDER HOOKS */
+
+const namespace = "@webstudio-is/sdk-components-react-radix";
+
+const onCollapsibleContentMatched = (
```
kof, line 48: is this something we can generalize? having sdk author to write this is nearly impossible task

kof (approved):

packages/sdk-components-react-radix/src/collapsible.tsx
istarkov, line 48: Lets add comment:
```
For each CollapsibleContent component within the selection, we identify its closest parent Collapsible component and execute the associated callback function.
```

apps/builder/app/builder/shared/navigator-tree.tsx
kof: maybe even better onNavigatorItemUnselect

istarkov (approved):
Interesting, instead of calling hook per component we are calling it once and just change the prop dataSource

apps/builder/app/builder/shared/navigator-tree.tsx
kof: or alternatively we will have to use argument to distinguish between unselecting/selecting multiple items vs one

packages/sdk-components-react-radix/src/collapsible.tsx
istarkov: Probably function name like - matchClosestParents would be better

apps/builder/app/builder/shared/navigator-tree.tsx
kof: anyways, up to you, just so you are aware of interactions we will need in the future: select one, select multiple, select all, and correspondingly unselect for each type

packages/sdk-components-react-radix/src/collapsible.tsx
TrySound: matchClosestParents would not work because here we cover the case with nested collapsibles so we need to continue lookup after finding parent.

istarkov: name is a little bit confusing, generally it's near same as html `closest`

istarkov: Probably deoptimizing it and splitting into iterate over selector and `closest()` method would look more readable

kof: I think we need to do something about it right away. Because we need to open up the api soon and we can't be scaring people with such logic that they need to write

packages/sdk-components-react-radix/src/collapsible.tsx
```diff
@@ -35,3 +36,41 @@ export const CollapsibleTrigger = forwardRef<
...
+/* BUILDER HOOKS */
+
+const namespace = "@webstudio-is/sdk-components-react-radix";
+
+// For each CollapsibleContent component within the selection,
+// we identify its closest parent Collapsible component
+// and update its open prop bound to variable.
+export const hooksCollapsible: Hook = {
+  onNavigatorUnselect: (context, event) => {
+    for (const instance of event.instanceSelection) {
```
kof, line 49: btw does this mean if we have a multiselect in navigator, we will open up all instances? :) 
I have a feeling user expectation would be to do that only when user clicked one particular

packages/sdk-components-react-radix/src/collapsible.tsx
```diff
@@ -35,3 +36,41 @@ export const CollapsibleTrigger = forwardRef<
...
+          `${namespace}:Collapsible`
+        );
+        if (collapsible) {
+          context.setPropVariable(collapsible.id, "open", false);
+        }
+      }
+    }
+  },
+  onNavigatorSelect: (context, event) => {
+    for (const instance of event.instanceSelection) {
```
kof, line 63: `instanceSelection` vs `selectedInstances`

If its a list of instances that are selected the later seems semantically closer to what it is

packages/sdk-components-react-radix/src/collapsible.tsx
```diff
@@ -35,3 +36,41 @@ export const CollapsibleTrigger = forwardRef<
...
+
+/* BUILDER HOOKS */
+
+const namespace = "@webstudio-is/sdk-components-react-radix";
+
+// For each CollapsibleContent component within the selection,
+// we identify its closest parent Collapsible component
+// and update its open prop bound to variable.
+export const hooksCollapsible: Hook = {
+  onNavigatorUnselect: (context, event) => {
```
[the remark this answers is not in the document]
istarkov: logic is functional approach to move context like params first i.e  fn(context)(event)

So this order is highly popular too

packages/react-sdk/src/hook.ts
```diff
@@ -0,0 +1,42 @@
...
+type NavigatorEvent = {
+  instanceSelection: InstanceSelection;
+};
+
+export type Hook = {
+  onNavigatorSelect?: (context: HookContext, event: NavigatorEvent) => void;
+  onNavigatorUnselect?: (context: HookContext, event: NavigatorEvent) => void;
+};
+
+export const getClosestInstance = (
```
kof, line 28: 1. utils like this should have a test, so changing it is obvious and tests get added right away
2. I am surprised we don't have this utility in our instance utils we use in the builder,  I think all functional utilities which have to do with the data structure should be at some point in one reusable package. 
Reason is this is a foundation for plugins to manipulate our data structure. We need to start thinking about these utils as a public api for the data

packages/sdk-components-react-radix/src/collapsible.tsx
TrySound: No, it's a path to root

packages/sdk-components-react-radix/src/collapsible.tsx
TrySound: instanceSelector -> instanceSelection
No the list of multi select

packages/react-sdk/src/hook.ts
TrySound: It's specific to hooks api. Internally we do not always map selector to instances.

packages/sdk-components-react-radix/src/collapsible.tsx
kof: yes, selectedInstances is a list of selected instances semantically

packages/sdk-components-react-radix/src/collapsible.tsx
kof: so instanceSelection is always one? I think I don't udnerstand since this is an array

packages/sdk-components-react-radix/src/collapsible.tsx
TrySound: selected instances semantically is multi select
here it is a path from current to root

kof: ah so instanceSelection is a path of ordered instances from current to the root? Got it now, wouldn't understand it from naming at all. Need a name that reflects exactly this

packages/sdk-components-react-radix/src/collapsible.tsx
kof: actually we already use instanceSelector ... seems like a good name

kof: Selection is wrong semantics here

kof: selector is understood as a path, but selection is a list of things, its not a path

kof: honestly instancePath is clearer to me than instanceSelector, because with css selectors and stuff it sometimes means a description of a selector, not an ordered items that represent a path, child-parent relationship

I would honestly rename it everywhere to be a "path"

kof: I am being extra nitpicky here because its a public api

PR #2070, opened 2023-08-08
Synchronisations fixes

istarkov:
## Description

-  Fixes sync bugs https://discord.com/channels/955905230107738152/1138073603036954664
- There is no sense to retry `fatal` errors, we can stop sync.
- Fatal errors state, show fatal error message.
- If user cancel confirmation goto fatal state too, as sync will never be fixed.
- Intercept error if project is loaded during synchronisation.
- Errors logged, as during investigation there were no way to get why `patch` requests are in failed state. 
- rest.patch are now wrapped with try/catch so we can distinguish api errors.
- toasts added too (and made infinite in case of fatal errors)
- exponential backoff + jitter added
- improved connectivity messages

Use for review:

image

### Todo,
- [ ] - test well.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/topbar/sync-status.tsx
+++ b/apps/builder/app/builder/features/topbar/sync-status.tsx
@@ -18,37 +18,60 @@ const subscribeIsOnline = () => {
 };
 
 export const SyncStatus = () => {
-  const status = useStore(queueStatus);
+  const statusObject = useStore(queueStatus);
   const isOnline = useStore(isOnlineStore);
   useEffect(subscribeIsOnline, []);
 
-  if (status !== "failed") {
+  if (
+    statusObject.status === "idle" ||
+    statusObject.status === "running" ||
+    statusObject.status === "recovering"
+  ) {
     return null;
   }
 
-  return (
-    <Flex align="center" justify="center">
-      <Tooltip
-        variant="wrapped"
-        content={
-          <>
-            Offline changes will be synced with Webstudio once you go online.
-            {isOnline ? (
-              ""
-            ) : (
-              <>
-                <br />
-                Please check your internet connection.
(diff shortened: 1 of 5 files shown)


apps/builder/app/builder/shared/sync/sync-server.ts
```diff
@@ -125,18 +303,18 @@ export const useSyncServer = ({
...
-  // save the initial version from loaded build
-  const initialized = useRef(false);
-  useEffect(() => {
-    if (authPermit === "view" || initialized.current === true) {
-      return;
-    }
-    initialized.current = true;
-    syncServer({ buildId, projectId, authToken, lastVersion: version });
-  }, [buildId, projectId, authToken, authPermit, version]);
+  useSyncProject({
```
TrySound, line 306: Why not export useSyncProject directly?

istarkov: to combine with `useBeforeUnload` and not expose this above

apps/builder/app/builder/features/topbar/sync-status.tsx
```diff
@@ -18,37 +18,64 @@ const subscribeIsOnline = () => {
...
+            color={rawTheme.colors.foregroundDestructive}
+          />
+        </Tooltip>
+      </Flex>
+    );
+  }
+
+  ((value: never) => {
+    /* exhaustive check */
+  })(statusObject);
```
TrySound, line 80:
```
statusObject satisfies never
```

TrySound (approved):

PR #2076, opened 2023-08-09
Fix if mouse leaves Tooltip.Content and moves over an iframe, the Radix Tooltip stays open

istarkov:
## Description

When the mouse leaves Tooltip.Content and moves over an iframe, the Radix Tooltip stays open.
This happens because Radix's internal grace area relies on the pointermove event, which isn't triggered over iframes.
The current workaround is to set pointer-events: none on the canvas when the tooltip is open.

Here are:
- [x] Tooltip component sets `--canvas-pointer-events` on the documentElement if any tooltip is open
- [x] Iframe uses `--canvas-pointer-events`
- [x] `isCanvasPointerEventsEnabledStore` in order to use above css variable
- [x] Color picker hack fully removed. Now just intercepts `open` state change, as it's dialog and  canvas is not interactive when it's open already so it's safe to disable pointer-events too.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/settings-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/settings-panel.tsx
@@ -2,6 +2,7 @@ import type { Publish } from "~/shared/pubsub";
 import type { Instance } from "@webstudio-is/project-build";
 import { SettingsSection } from "./settings-section/settings-section";
 import { PropsSectionContainer } from "./props-section/props-section";
+import { Tooltip } from "@webstudio-is/design-system";
 
 export const SettingsPanelContainer = ({
   selectedInstance,
@@ -13,6 +14,10 @@ export const SettingsPanelContainer = ({
   return (
     <>
       <SettingsSection />
+      <Tooltip defaultOpen={true} content="This is a tooltip">
+        <button>TEST</button>
+      </Tooltip>
+
       <PropsSectionContainer
         publish={publish}
         selectedInstance={selectedInstance}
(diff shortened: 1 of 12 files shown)


TrySound (approved):

apps/builder/app/builder/features/settings-panel/settings-panel.tsx
```diff
@@ -13,6 +14,10 @@ export const SettingsPanelContainer = ({
   return (
     <>
       <SettingsSection />
+      <Tooltip defaultOpen={true} content="This is a tooltip">
```
kof, line 17: is this forgotten?

apps/builder/app/builder/features/style-panel/shared/color-picker.tsx
```diff
@@ -88,35 +88,6 @@ const styleValueToRgbaColor = (value: CssColorPickerValueInput): RgbaColor => {
   };
 };
 
-// Dragging over canvas iframe with CORS policies will lead to loosing events and getting stuck in mousedown state.
```
kof, line 91: why is this removed? the problem is still there

kof:
- [ ] not created in this issue but the tooltip around color picker is broken

Screenshot 2023-08-09 at 19 43 07

apps/builder/app/builder/features/style-panel/shared/color-picker.tsx
kof: just tested, seems to work

kof (approved):
don't forget to remove the TEST thing

PR #2089, opened 2023-08-10
Fix pointer events

istarkov:
## Description

Fix bug with drag component etc

implemented disableCanvasPointerEvents as disposable resource. 
That makes it much easier to control order of execution (disable first then enable), be sure that counter is always >=0, 
and easier control about enable/disable events allowing to dispose only if resource was created.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/color-picker.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/color-picker.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from "react";
+import { useState } from "react";
 import { colord, extend, type RgbaColor } from "colord";
 import namesPlugin from "colord/plugins/names";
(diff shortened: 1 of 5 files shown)


kof (approved):

apps/builder/app/builder/features/style-panel/shared/color-picker.tsx
```diff
@@ -122,6 +121,7 @@ export const ColorPicker = ({
   property,
 }: ColorPickerProps) => {
   const [displayColorPicker, setDisplayColorPicker] = useState(false);
+  const enableCanvasPointerEventsRef = useRef<() => void>();
```
kof, line 124: why not export this as a hook from design system then?

TrySound (approved):

PR #2095, opened 2023-08-11
Improve s3 keys sanitizer

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/2091 https://github.com/webstudio-is/webstudio-builder/issues/2049

Here fixed uploading of files by sanitizing a few more characters which require special treatment in signature v4.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/package.json
+++ b/packages/asset-uploader/package.json
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "@aws-crypto/sha256-js": "^5.0.0",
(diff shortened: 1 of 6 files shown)


kof (approved):
Discussed: I believe native client supported those chars, so we should be encoding those instead of stripping. Its not gonna stop us from merging but I do think this is a regression and parenthesis in a filename is super popular use case

PR #2098, opened 2023-08-11
Add utilities generator to sdk

TrySound:
Here moved code generation of utilities from saas into sdk.

Now InstanceRoot accepts utils object with these 3 fields

- indexesWithinAncestors,
- executeComputingExpressions,
- executeEffectfulExpression,

Generation utility can be shared between cli and saas to simplify maintanance.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/jest-config/index.js
+++ b/packages/jest-config/index.js
@@ -15,4 +15,5 @@ module.exports = {
     "^~/(.*)$": "<rootDir>/src/$1",
   },
   extensionsToTreatAsEsm: [".ts", ".tsx"],
+  prettierPath: null,
 };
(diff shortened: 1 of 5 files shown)


packages/jest-config/index.js
```diff
@@ -15,4 +15,5 @@ module.exports = {
     "^~/(.*)$": "<rootDir>/src/$1",
   },
   extensionsToTreatAsEsm: [".ts", ".tsx"],
+  prettierPath: null,
```
TrySound, line 18: Jest seem does not support prettier v3 because it removed synchronous api. Jest seem less suitable for us now.

istarkov (approved):

Issue #2096, opened 2023-08-11
Add radix components


TrySound:
- [x] Tooltip (@istarkov)
- [x] Sheet (@istarkov)
- [x] Dialog (@istarkov)
- [x] Tabs (@TrySound)
- [x] Collapsible (@TrySound)
- [x] Popover (@istarkov)
- [x] Accordion - only single (@TrySound)
- [x] Label  (@istarkov)
- [x] Select (@TrySound)
- [x] Switch (@TrySound)
- [x] Radio Group (@TrySound)
- [x] Checkbox (@TrySound)
- [ ] Slider
- [ ] ~~Progress~~
- [ ] ~~Dropdown Menu~~

Removing
- [x] Button (@istarkov)
- [x] Input (@istarkov)
- [x] Textarea (@istarkov)
- [ ] Separator

Link for testing
https://webstudio-builder-git-main-webstudio-is.vercel.app/dashboard

kof:
Lets close this one, missing components can be added once we see demand

PR #2101, opened 2023-08-12
Input + Textarea + Label components

istarkov:
## Description

ref #2096

image

image

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react-radix/src/input.ws.ts
+++ b/packages/sdk-components-react-radix/src/input.ws.ts
@@ -25,14 +25,15 @@ export const meta: WsComponentMeta = {
     { selector: ":disabled", label: "Disabled" },
     { selector: "::placeholder", label: "Placeholder" },
     { selector: "::file-selector-button", label: "File Selector Button" },
+    /*
     { selector: ":valid", label: "Valid" },
     { selector: ":invalid", label: "Invalid" },
     { selector: ":required", label: "Required" },
     { selector: ":optional", label: "Optional" },
-    { selector: ":disabled", label: "Disabled" },
     { selector: ":enabled", label: "Enabled" },
     { selector: ":read-only", label: "Read Only" },
     { selector: ":read-write", label: "Read Write" },
+    */
   ],
   order: 1,
   template: [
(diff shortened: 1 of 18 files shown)


TrySound (approved):

packages/sdk-components-react-radix/src/input.ws.ts
```diff
@@ -0,0 +1,79 @@
...
+    { selector: "::placeholder", label: "Placeholder" },
+    { selector: "::file-selector-button", label: "File Selector Button" },
+    { selector: ":valid", label: "Valid" },
+    { selector: ":invalid", label: "Invalid" },
+    { selector: ":required", label: "Required" },
+    { selector: ":optional", label: "Optional" },
+    { selector: ":disabled", label: "Disabled" },
+    { selector: ":enabled", label: "Enabled" },
+    { selector: ":read-only", label: "Read Only" },
+    { selector: ":read-write", label: "Read Write" },
```
kof, line 35: We need to add  a submenu, @taylornowotny was supposed to provide some design in figma

- [ ] We discussed with @taylornowotny  to reduce the list for both input components, because its too much at glance and these are very rare
https://github.com/webstudio-is/webstudio-builder/issues/2029
- [ ]  For now I would comment these additional once out and make sure we reuse this list. This way once we have a submenu, it comes to both.

istarkov: commented for now

kof:
- [ ] Textarea border radius and color don't match, please check styling
- [ ] label has quite a few differences: margin, line-height

packages/sdk-components-react-radix/src/input.ws.ts
kof: we need the same on the other input why can't we reuse these states?

istarkov: Added comment

PR #2120, opened 2023-08-14
Fix code text component

TrySound:
Fixes
https://discord.com/channels/955905230107738152/1140570603372413039 https://discord.com/channels/955905230107738152/1140566358023745587

- renamed inline to data-inline
- renamed meta to data-meta
- replaced css custom property with attrubute state
- fixed pasting of code in markdown
- added `white-space: pre-wrap` style

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/copy-paste/plugin-markdown.ts
+++ b/apps/builder/app/shared/copy-paste/plugin-markdown.ts
@@ -35,9 +35,9 @@ const astTypeComponentMap: Record<string, Instance["component"]> = {
   // we need to either have RichTextImage or support Image inside RichText
   image: "Image",
   blockquote: "Blockquote",
-  code: "Code",
+  code: "CodeText",
   // @todo same problem as with image
-  inlineCode: "Code",
+  inlineCode: "CodeText",
   list: "List",
   listItem: "ListItem",
   thematicBreak: "Separator",
@@ -113,7 +113,7 @@ const toInstanceData = (
       props.push({
         id: generateId(),
         type: "boolean",
-        name: "inline",
+        name: "data-inline",
         instanceId: instance.id,
         value: true,
       });
@@ -122,13 +122,6 @@ const toInstanceData = (
       instance.children.push({
         type: "text",
         value: child.value,
-      });
-      props.push({
-        id: generateId(),
-        type: "boolean",
-        name: "inline",
-        instanceId: instance.id,
-        value: false,
       });
       if (child.lang) {
         props.push({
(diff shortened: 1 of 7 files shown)


TrySound:
@kof It is already public. But with custom prop. I can remove state if you want.
I tested, works.

TrySound:
image

kof:
Inline is effectively only needed when pasted from markdown to make it display inline and not block like by default, so its a state kindof, but I am not sure we need to expose it as such on style source and calling it Data Something is also weird, it should be just "Inline" I guess.

Another thought: if I am using this component visually, I would just go and set display: inline instead of using a prop. So its kinda weird right now that I would need to even use a prop

kof:
This inline thing is really a problem for how a markdown plugin can set display inline on css when creating this instance, the prop was just the only way I knew how to do it at the time. Maybe we can just set display inline now?

apps/builder/app/shared/copy-paste/plugin-markdown.test.ts
```diff
@@ -499,98 +499,91 @@ describe("Plugin Markdown", () => {
...
+      "id": "123",
+      "instanceId": "123",
+      "name": "lang",
+      "type": "string",
+      "value": "js",
+    },
+    {
+      "id": "123",
+      "instanceId": "123",
+      "name": "data-meta",
```
kof, line 580: is this supposed to be still here?

apps/builder/app/shared/copy-paste/plugin-markdown.test.ts
```diff
@@ -499,98 +499,91 @@ describe("Plugin Markdown", () => {
...
+      "component": "Paragraph",
+      "id": "123",
+      "type": "instance",
+    },
+  ],
+  "props": [
+    {
+      "id": "123",
+      "instanceId": "123",
+      "name": "data-inline",
```
kof, line 537: same question

kof (approved):

apps/builder/app/shared/copy-paste/plugin-markdown.ts
```diff
@@ -110,52 +114,38 @@ const toInstanceData = (
...
         props.push({
           id: generateId(),
           type: "string",
           name: "lang",
-          instanceId: instance.id,
+          instanceId,
           value: child.lang,
         });
       }
-      if (child.meta) {
```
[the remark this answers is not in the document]
TrySound: meta prop is not standard. Either data-meta or remove

PR #2125, opened 2023-08-15
Add tooltips for component properties in Settings tab

gsppe:
## Description

**Important** The PR adds tooltips only for HTML attributes.

Fixes https://github.com/webstudio-is/webstudio-builder/issues/2027

## Steps for reproduction

1. Add a few components to the canvas
3. Select one
4. Go to the Settings tab in the right sidebar
5. Hover over the labels

HTML attributes should show a tooltip with a description

## Code Review

- [ ] @taylornowotny please review the descriptions in the props-descriptions.ts file
   - test it on preview
- [ ] @istarkov check the first commit that is where the implementation and integration are done. The idea behind this is that we have a central file with HTML attributes descriptions and when generating components props we pick the descriptions from there if the property doesn't have a custom description. Essentially we use props-descriptions as a fallback. Do we need tests for this feature?
  

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


GoharAkram:
- [ ] Tooltip in all component props was tested

Added all components and checked.

The tooltip does not show on these properties in these components:

1. Tag in Heading, Box and Text
2. Source in Image
3. Prefetch and Href in LInk
4. Ordered in Link
5. For in Input Label, Checkbox Field and Radio

RADIX:
6. Variant in Button
7. For in Input Label
8. Model in Dialog, Sheet and Popover


@kof Please check

ggurgone:
@GoharAkram thank you for taking a look!

> For
> Prefetch and Href in LInk

Good catch, I will fix these right away.

Reminder that this PR adds tooltips only for HTML attributes, therefore it is normal that some of the ones that you mentioned like `Tag` are missing descriptions.

kof:
prefetch is remix's own thing https://remix.run/docs/en/1.19.3/components/link#prefetch

taylornowotny:
Unfortunately there are issues with many of these descriptions. I'm making a list of corrections which I will comment soon.

taylornowotny:
Here is a list of everything that needs to change. In most cases the tooltip content is not descriptive enough to help the user understand how to use the control.

I am tracking all of the properties that do not have tooltips content yet and I will finish those tomorrow. No need for you to do those yet @ggurgone 

1. **Auto Focus** tooltip is needlessly verbose in a way that is inconsistent with other tooltips. Remove everything before “indicating” and change that word to “Indicates”. Remove “the <dialog> that it is part of” and replace with “its parent component”.

2. The **Type** property is very different between elements so I think this should have a different description for different elements:
- Button Type: "Defines the behavior of the button, such as submitting a form or resetting form fields."
- Input Type: "Specifies the type of data that this input will accept and helps the browser provide appropriate validation and formatting for that input type."

3. Replace **Required** content with:
”Indicates whether this form element must be filled before the form can be submitted.”

4. Replace **Aria Label** content with:
Provides the accessible name that describes an interactive element if no other accessible name exists, for example in a button that contains an image with no text.

5. Replace **Width** content with:
”Defines the image’s width in pixels.”

6. Replace **Height** content with:
”Defines the image’s height in pixels.”

7. Replace **Alt** content with:
”Text description of the image, which is very important for accessibility and search engine optimization. Screen readers read this description to users so they know what the image means. Alt text is also displayed on the page if the image can't be loaded for some reason.”

8. Replace **Loading** content with:
”Determines whether the image will load as soon as possible (Eager), or when it scrolls into view (Lazy). Lazy loading is a great option for pages with many images because it can significantly reduce the time it takes for the page to load initially.”

9. Add tooltip to **Source** prop on :
”Used to specify the source (URL) of the image file that should be displayed on the web page. You can paste a URL or choose an uploaded image from your Assets.”

10. The Vimeo props are great, thank you for taking care of these! Just one thing needs to be corrected. On the URL prop, change the last sentence from “…see our introductory guide.” to “…see Vimeo’s introductory guide.” Obviously we don’t want to mislead users into thinking that this info is in a Webstudio guide somewhere.

taylornowotny:
DONE. If a tooltip that I wrote already has content, replace the old content with my new content.

Embed - I don’t think these need tooltips. The properties are as clear as they can be already.

State (Form): Use this property to reveal the Success and Error states on the canvas so they can be styled. The Initial state is displayed when the page first opens. The Success and Error states are displayed depending on whether the Form submits successfully or unsuccessfully. 

For (Label, Checkbox Field, Radio Field): Associates this Label with an Input. The value of the “For” attribute should match the ID attribute of the corresponding Input element.

Name (Input, Radio, Checkbox): This name is important when submitting form data to the server, as it identifies the data associated with the input. When multiple inputs share the same name attribute, they are treated as part of the same group (e.g., radio buttons or checkboxes).

Prefetch (Link): Controls when and if the link prefetches the resources that the next page needs to make loading faster. “Intent” will prefetch when the link is hovered or focused. “Render” will prefetch when the link is rendered. “Viewport” will prefetch when the link is in the viewport. “None” will not prefetch.

Ordered (List): Shows numbers instead of bullets when toggled. See the “List Style Type” property under the “List Item” section in the Style panel for more options.

Tag (Box and Text): Use this property to change the HTML tag of this element to semantically structure and describe the content of a webpage. This can be important for accessibility tools and search engine optimization.

Tag (Heading): Use HTML heading levels (h1-h6) to structure content hierarchically, with h1 as the main title and subsequent levels representing sub-sections. Maintain a logical order and avoid skipping levels to ensure consistent and meaningful organization.

Open (multiple Radix components): Show or hide the content of this component on the canvas. This will not affect the initial state of the component.

Side Offset (Popover Content, Tooltip Content): The distance in pixels between the Content and the Trigger.

Align Offset (Popover Content, Tooltip Content): The offset in pixels from the "start" or "end" alignment options.

Side (Popover Content, Tooltip Content): The preferred alignment against the Trigger. May change when collisions occur.

Delay Duration (Tooltip): The delay before the Tooltip shows after the Trigger is hovered, in milliseconds. If no value is specified, the default is 700ms.

Disable Hoverable Content (Tooltip): When toggled, prevents the Tooltip content from showing when the Trigger is hovered.

taylornowotny:
There is just one property that I do not understand:

"Value" as seen on Navigation Menu + Accordion + Accordion Item + Tabs + Tabs Content

I don't see this mentioned in Radix docs and I have no context for understanding how this property is intended to be used. Can you help us out @kof @TrySound @istarkov ?

taylornowotny (changes requested):
Please add or update tooltips based on my comments

taylornowotny:
@gsppe Thanks for that note about Auto Focus - go ahead and change it as you see best. I only wanted to make sure that it wasn't specific to the < dialog > component becuase this property appears on multiple components.

TrySound:
Related issue https://github.com/webstudio-is/webstudio-builder/issues/2130

istarkov (approved):

packages/generate-arg-types/src/props/add-descriptions.ts
```diff
@@ -0,0 +1,74 @@
...
+  componenName: string,
+  argTypes: ReturnType<typeof propsToArgTypes>,
+  customDescriptions: { [key in string]: string } = {}
+) => {
+  if (componenName && ignoreComponents.has(componenName)) {
+    return;
+  }
+
+  Object.entries(argTypes).forEach(([propName, meta]) => {
+    // if (typeof meta.description !== "string") {
```
kof, line 33: forgotten?

kof (approved):

packages/html-data/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "@webstudio-is/scripts": "workspace:^",
+    "@webstudio-is/tsconfig": "workspace:^",
+    "camelcase": "^7.0.1",
+    "html-tags": "^3.3.1",
+    "jest": "^29.6.2",
+    "mdn-data": "2.0.32",
+    "tsx": "^3.12.6",
+    "type-fest": "^4.2.0",
+    "typescript": "5.1.6",
+    "zod": "^3.21.4"
```
TrySound, line 28: I doubt we need all these packages here.

packages/generate-arg-types/src/cli.ts
```diff
@@ -54,50 +55,97 @@ if (componentFiles.length === 0) {
...
-  mkdirSync(generatedDir, { recursive: true });
-  writeFileSync(generatedPath, fileContent);
+    mkdirSync(generatedDir, { recursive: true });
+    writeFileSync(generatedPath, fileContent);
 
-  console.log(`Done generating argTypes for ${generatedPath}`);
-}
+    console.log(`Done generating argTypes for ${generatedPath}`);
+  }
+})();
```
TrySound, line 151: Why not top level await?

packages/html-data/package.json
```diff
@@ -0,0 +1,54 @@
...
+  ],
+  "license": "AGPL-3.0-or-later",
+  "private": false,
+  "sideEffects": false,
+  "dependencies": {
+    "colord": "^2.9.3",
+    "css-tree": "^2.3.1",
+    "hyphenate-style-name": "^1.0.4",
+    "openai": "^3.2.1",
+    "warn-once": "^0.1.1"
```
TrySound, line 52: And these

packages/generate-arg-types/src/cli.ts
TrySound: Error or warning? Caz we don't care about cli for end user.

istarkov: Error

packages/html-data/package.json
```diff
@@ -14,18 +14,8 @@
     "test": "NODE_OPTIONS=--experimental-vm-modules jest"
   },
   "devDependencies": {
-    "@types/css-tree": "^2.3.1",
     "@webstudio-is/jest-config": "workspace:^",
```
TrySound, line 17: no need jest either

TrySound (approved):

PR #2126, opened 2023-08-15
Add chevron icon to accordion

TrySound:
Here added chevron to accordion trigger
by using css custom properties for cascade.

We still do not support css custom properties in style panel though here used only for transform which does not have ui too.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react-radix/src/accordion.ws.ts
```diff
@@ -37,20 +39,83 @@ const presetStyle = {
...
+        tc.py(4),
+        tc.font("medium"),
+        tc.transition("all"),
+        tc.hover([tc.underline()].flat()),
+        custom("accordion-trigger-icon-transform", "rotate(0deg)"),
+        tc.state(
+          [custom("accordion-trigger-icon-transform", "rotate(180deg)")],
+          "[data-state=open]"
+        ),
+        // [&[data-state=open]>svg]:rotate-180
```
kof, line 76: can be removed?

packages/sdk-components-react-radix/src/accordion.ws.ts
```diff
@@ -37,20 +39,83 @@ const presetStyle = {
...
+            tc.w(4),
+            tc.shrink(0),
+            tc.transition("transform"),
+            tc.duration(200),
+          ].flat(),
+          children: [
+            {
+              type: "instance",
+              component: "HtmlEmbed",
+              label: "Chevron Down Icon",
```
kof, line 103: could be "Chevron Icon" I guess since you are just rotating

packages/sdk-components-react-radix/src/theme/tailwind-classes.ts
```diff
@@ -843,6 +850,15 @@ export const transition = (value: "none" | "all"): EmbedTemplateStyleDecl[] => {
   ];
 };
 
+export const duration = (ms: number): EmbedTemplateStyleDecl[] => {
```
kof, line 853: what do we do with these unparsed values once we actually have a transition duration that is parsed and such? Isn't it easier to already add a data type?

packages/sdk-components-react-radix/src/accordion.ws.ts
TrySound: This is tailwind classes syntax comment

packages/sdk-components-react-radix/src/theme/tailwind-classes.ts
TrySound: Style panel is able to parse these

kof: but do you want this data structure long term in the db? otherwise whats the point in all the data structures we have, they all could be unparsed?

TrySound: Migrations maybe
Cc @istarkov

istarkov: Isn't it just unit?

packages/sdk-components-react-radix/src/accordion.ws.ts
```diff
@@ -37,20 +39,83 @@ const presetStyle = {
...
+      styles: [
+        tc.flex(),
+        tc.flex(1),
+        tc.items("center"),
+        tc.justify("between"),
+        tc.py(4),
+        tc.font("medium"),
+        tc.transition("all"),
+        tc.hover([tc.underline()].flat()),
+        custom("accordion-trigger-icon-transform", "rotate(0deg)"),
```
istarkov, line 71: What type custom returns? All above returns array but whats here?

packages/sdk-components-react-radix/src/accordion.ws.ts
```diff
@@ -37,20 +39,83 @@ const presetStyle = {
 // border-b
 const accordionItemStyles: EmbedTemplateStyleDecl[] = [tc.borderB()].flat();
 
-// flex
-const accordionHeaderStyles: EmbedTemplateStyleDecl[] = [tc.flex()].flat();
+const custom = (name: string, value: string) =>
```
istarkov, line 42: Please add return type. I don't get what's this and how to find then in code

istarkov (approved):
tc.property is so beautfull

PR #2133, opened 2023-08-16
(cli): Download images to local build in cli projects

JayaKrishnaNamburu:
Hi

This PR, allows the cli to download `images` and `fonts` and build the projects from the local assets.


apps/cli/src/prebuild.ts
```diff
@@ -1,6 +1,9 @@
 import { join, relative, dirname } from "node:path";
-import { writeFileSync } from "node:fs";
-import { rm, mkdir } from "node:fs/promises";
+import { writeFileSync, createWriteStream } from "node:fs";
+import { rm, mkdir, access } from "node:fs/promises";
+import fetch from "node-fetch";
+import { pipeline } from "node:stream";
```
istarkov, line 5: https://nodejs.org/docs/latest-v16.x/api/stream.html#streams-promises-api

apps/cli/src/prebuild.ts
```diff
@@ -43,7 +51,30 @@ type RemixRoutes = {
   }>;
 };
 
+export const downloadAsset = async (url: string, name: string) => {
+  const assetPath = join("public", ASSETS_BASE, name);
+  try {
+    await access(assetPath);
```
istarkov, line 57: I would comment that we do not download fil if it already exists.

apps/cli/src/prebuild.ts
```diff
@@ -43,7 +51,30 @@ type RemixRoutes = {
...
+      if (!response.ok) {
+        throw new Error(`unexpected response ${response.statusText}`);
+      }
+
+      await streamPipeline(
+        response.body as NodeJS.ReadableStream,
+        createWriteStream(assetPath)
+      );
+    } catch (error) {
+      console.error(`Error in downloading file ${name} \n ${error}`);
```
istarkov, line 71: Best would be to make cleanup, as here file can be created but be broken

apps/cli/src/prebuild.ts
```diff
@@ -1,6 +1,9 @@
 import { join, relative, dirname } from "node:path";
-import { writeFileSync } from "node:fs";
-import { rm, mkdir } from "node:fs/promises";
+import { writeFileSync, createWriteStream } from "node:fs";
+import { rm, mkdir, access } from "node:fs/promises";
+import fetch from "node-fetch";
```
TrySound, line 4: Use native fetch instead

apps/cli/src/prebuild.ts
```diff
@@ -43,7 +51,30 @@ type RemixRoutes = {
...
+  } catch {
+    const streamPipeline = promisify(pipeline);
+    try {
+      const response = await fetch(url);
+      if (!response.ok) {
+        throw new Error(`unexpected response ${response.statusText}`);
+      }
+
+      await streamPipeline(
+        response.body as NodeJS.ReadableStream,
```
TrySound, line 67: cast should not be needed with native fetch

TrySound (approved):

apps/cli/src/prebuild.ts
istarkov: Also good thing to reduce various errors caused by cli termination etc is to download files initially to  newly created tmp dir (there is a command to get system tmp dir location). https://www.npmjs.com/package/tmp
and immediately after download move file to the destination.

What it gives.

Download is a long operation, every termination of process will leave you with partially downloaded file, that's bad because on resync we will not download it again.
Move is superfast.

With above idea if termination occur on download then nothing bad happens. It was download to tmp. And the probability of termination on download is huge, because of time.

Probability of termination on file move is near zero. Because it's near immediate system operation.

istarkov: There is an old lib for this https://www.npmjs.com/package/fs-write-stream-atomic
сс @TrySound can you recommend anything else?

TrySound: https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fspromisesmkdtempprefix-options

apps/cli/src/prebuild.ts
```diff
@@ -43,13 +51,55 @@ type RemixRoutes = {
...
+            if (response.statusCode !== 200) {
+              reject(new Error(`HTTP Error: ${response.statusCode}`));
+              response.resume();
+            }
+            resolve(response);
+          })
+          .on("error", (error) => {
+            reject(error);
+          });
+      });
```
TrySound, line 77: I meant specifically `fetch()`. It's available in node globally since v18.

apps/cli/src/prebuild.ts
```diff
@@ -43,13 +51,55 @@ type RemixRoutes = {
...
+          .get(url, (response) => {
+            if (response.statusCode !== 200) {
+              reject(new Error(`HTTP Error: ${response.statusCode}`));
+              response.resume();
+            }
+            resolve(response);
+          })
+          .on("error", (error) => {
+            reject(error);
+          });
```
kof, line 76: why not
```suggestion
          .on("error", reject);
```

apps/cli/src/prebuild.ts
kof: yea something likee this

```suggestion
fetch(url).then(res => new Promise((resolve, reject) => {
    const dest = fs.createWriteStream(fn);
    res.body.pipe(dest);
    res.body.on('end', () => resolve());
    dest.on('error', reject);
}));
```

TrySound:
```js
await pipeline(response.body, fs.createWriteStream(path))
```

apps/cli/src/prebuild.ts
```diff
@@ -139,21 +175,53 @@ export const prebuild = async () => {
...
+  for (const asset of siteData.assets) {
+    if (asset.type === "image") {
+      const image = getImageAttributes({
+        /*
+          TODO:
+          https://github.com/webstudio-is/webstudio-builder/issues/2135
+          There should be a option in the loader, that allows to download
+          original image instead of the processed one. Right now, we are using
+          the width from the asset meta
+        */
```
TrySound, line 191: The thing about loader is that you can provide own one. Here you just don't want any of its logic. Why complicate it with more options?

TrySound (approved):

PR #2147, opened 2023-08-20
Add component tokens support

TrySound:
Ref https://github.com/webstudio-is/webstudio-builder/issues/2137

Here added static tokens support to meta. Id of resulting style source become persistent like this "componentName:tokenName". Such token styles are inserted only if style source with same id not exist. This way we prevent breaking styles already changed by user.

Style source input now autocomplete global tokens and suggest to create predefined component tokens.

Also disabled spell check in style source input.

Screenshot 2023-08-20 at 17 18 37

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/style-source-section.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source-section.tsx
@@ -94,7 +94,7 @@ const $baseBreakpointId = computed(breakpointsStore, (breakpoints) => {
 
 // metas are rarely change so keep preset token styles computing
 // in separate store
-const $presetTokens = computed(
+export const $presetTokens = computed(
   [registeredComponentMetasStore, $baseBreakpointId],
   (metas, baseBreakpointId) => {
     const presetTokens = new Map<
--- a/apps/builder/app/builder/features/style-panel/style-source-section.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source-section.tsx
@@ -9,6 +9,7 @@ import {
   type StyleSourceToken,
   type StyleSourceSelections,
   getStyleDeclKey,
+  StyleDecl,
 } from "@webstudio-is/project-build";
 import {
   Flex,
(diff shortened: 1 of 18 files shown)


apps/builder/app/builder/features/style-panel/style-source-section.tsx
```diff
@@ -81,26 +86,88 @@ const createLocalStyleSourceIfNotExists = (
...
+            state: styleDecl.state,
+            property: styleDecl.property,
+            value: styleDecl.value,
+          });
+        }
+        presetTokens.set(styleSourceId, {
+          component,
+          styleSource: {
+            type: "token",
+            id: styleSourceId,
```
kof, line 131: do you think this logic could be extracted as a function and have tests?

kof: or maybe even without extraction

apps/builder/app/shared/instance-utils.test.ts
```diff
@@ -317,3 +329,91 @@ describe("find closest droppable target", () => {
...
+  );
+  insertTemplateData(
+    {
+      children: [{ type: "id", value: "box1" }],
+      instances: [
+        { type: "instance", id: "box1", component: "Box", children: [] },
+      ],
+      props: [],
+      dataSources: [],
+      styleSourceSelections: [{ instanceId: "box1", values: ["1", "2"] }],
```
kof, line 360: "styleSourceSelection" seems more correct, because its one selection that contains many items

kof (approved):
didn't test

istarkov (approved):

apps/builder/app/shared/instance-utils.test.ts
TrySound: It's an array of selections

PR #2149, opened 2023-08-20
Allow horizontal scrolling in navigator

kof:
## Description

This is a workaround because we don't have a resizable panel yet. With deep nested trees we need more space

## Steps for reproduction

1. nest radix components with a lot of depth
2. see that you can scroll both vertically and horizontally
3. horizontal bar is invisible, its a separate problem, not sure why

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov:
Enter loong name for the body
image

Vertical scrollbar appears in the Tree but horizontal is not visible at all

TrySound (approved):

Issue #2151, opened 2023-08-21
Release planned tasks


istarkov:
## Blocking launch
- [x] - https://github.com/webstudio-is/webstudio-builder/issues/2137
  - [x] - Button, Navigation menu styled
  - [ ] - ~~Add style sources to existing Components as of now all local (https://github.com/webstudio-is/webstudio-builder/pull/2158)~~
  - [x] revert button (https://github.com/webstudio-is/webstudio-builder/pull/2169)
  - [x] revert menu (https://github.com/webstudio-is/webstudio-builder/pull/2177)
- [x] - Bugs
  - [x] - Drag/Drop broken (https://github.com/webstudio-is/webstudio-builder/pull/2152)
  - [x] - Accordion opens when clicking on Item, must be opened on Content (https://github.com/webstudio-is/webstudio-builder/pull/2155)
  - [x] - Flex child section not shown if wrapped with radix unstylable component|s (Same should be with position section too) https://discord.com/channels/955905230107738152/1143135920267210832 Fixed here (https://github.com/webstudio-is/webstudio-builder/pull/2167)
- [x] - Tab index in preview makes it impossible to test keyboard interactions. (One of possible solutions https://github.com/webstudio-is/webstudio-builder/pull/2112) https://github.com/webstudio-is/webstudio-builder/pull/2179#pullrequestreview-1595667382
- [x]  Renamings https://www.notion.so/webstudiois/component-name-update-0a81ecde53cd4770bdbbbf248b167c4b
- [x] Changes from Taylor's feedback https://www.notion.so/webstudiois/feedback-on-components-9688bc9c04cb48f1af838f289c0aad40
- [x] Fix clickable last item in navigation (broken when added horizontal scroll, the scroll track is covering content) https://github.com/webstudio-is/webstudio-builder/pull/2181

## After launch (or before if time allows)

- [x] local style source at the end
- [ ] - https://github.com/webstudio-is/webstudio-builder/issues/2143
- [x] - Remove useless/bad props from all radix components https://github.com/webstudio-is/webstudio-builder/issues/2124#issuecomment-1679185884 (#2154)
- [ ] - Components planned https://github.com/webstudio-is/webstudio-builder/issues/2096
  - [ ] - Think about renaming component names i.e Content instead of Tooltip Content, Trigger instead of TootlipTrigger
  - [ ] - Think about naming  base styles of the components
- [ ] - Navigation Menu (@istarkov)
[...]

PR #2153, opened 2023-08-21
Nav menu internals

istarkov:
## Description

ref #2151

image

image

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

taylornowotny:
Feedback on the Navigator menu component:

- The user will want to style the Viewport Container while a menu content is open, but selecting the viewport container closes any open menus, making it harder to see the effect of any style changes. Can we fix this so the viewport container will not close an open content?

- Radix has a Nav Item that is not a dropdown, just a link. When a user wants a simple Nav menu without dropdowns this is what they will copy-paste. Why don't we have this? 

- Why is the Navigation Menu Content left-aligned? Better to center it. The user will try to center this and may have a hard time understanding how to do so. If the user prefers to left-align it, that’s much easier for them.

- The Nav Link content title should be named “Title” and use a Text component, not a Box

- change names:
    - Navigation Menu Link → Accessible Link Wrapper
    - Link → Nav Menu Link
    - Box → Title
    - Keep the name of the component as “Navigation Menu” but shorten all children names from “Navigation” to “Nav” to conserve space

istarkov:
> Why is the Navigation Menu Content left-aligned? Better to center it. The user will try to center this and may have a hard time understanding how to do so. If the user prefers to left-align it, that’s much easier for them.

It's default stylings, I suggested to create a template project with various menus inside.
With various aligns etc.

The more issues will be with animations as changing most "wrong" props will kill animations

taylornowotny:
> > Why is the Navigation Menu Content left-aligned? Better to center it. The user will try to center this and may have a hard time understanding how to do so. If the user prefers to left-align it, that’s much easier for them.
> 
> It's default stylings, I suggested to create a template project with various menus inside. With various aligns etc.
> 
> The more issues will be with animations as changing most "wrong" props will kill animations

Good idea, but better if we center-align this by default. It will be much better for the user.

PR #2154, opened 2023-08-21
Exclude props

istarkov:
## Description

ref https://github.com/webstudio-is/webstudio-builder/issues/2151

Exclude following props from radix components

```
-e asChild -e modal -e defaultValue -e defaultOpen
```

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/generate-arg-types/src/cli.ts
```diff
@@ -14,7 +15,24 @@ const options = {
...
+      multiple: true,
+      short: "e",
+    },
+  },
+  strict: true,
+} as const satisfies ParseArgsConfig;
+
+const cliArgs = parseArgs({ args: process.argv.slice(2), ...CLI_ARGS_OPTIONS });
+
+console.log(cliArgs);
```
TrySound, line 32: Remove console?

packages/generate-arg-types/src/cli.ts
```diff
@@ -14,7 +15,24 @@ const options = {
...
+
+const componentsGlobString = cliArgs.positionals.join(" ");
```
TrySound, line 34: `const globs = cliArgs.positionals`?

TrySound (approved):
Please remove explicit Omit in our components

kof:
why is this not done over type Omit like we did before?

istarkov:
Same props, same names, done only for radix lib. There is no need to wrap every component and omit there.

packages/generate-arg-types/src/cli.ts
istarkov: They are inside '' to not allow system expand, so same logic as before

istarkov:
With Omit is so easy to forget to remove them at all.

packages/generate-arg-types/src/cli.ts
TrySound: Right after this there is `componentsGlobString.split(' ')`. What system expand?

istarkov: system expands pathes not inside ''

istarkov: `echo */*` is not same as `echo '*/*'`

istarkov: https://github.com/webstudio-is/webstudio-builder/blob/ab903af1303180f178a6d464709742d7381539e6/packages/sdk-components-react-radix/package.json#L44

TrySound: Got it. Could be a one liner with comment `const globs = cliArgs.positionals.join(" ").split(" ")`

PR #2168, opened 2023-08-24
Update cli readme

JayaKrishnaNamburu:
## Description

This PR adds a readme for the `cli`

## Code Review

  - workflow overview


apps/cli/Readme.md
```diff
@@ -0,0 +1,59 @@
+## WebStudio CLI
+
+WebStudio CLI is designed to help you seamlessly link, sync, and build projects from your web development workspace. This README will guide you through the process of setting up, linking, syncing, building, and running your projects using the WebStudio CLI commands.
```
kof, line 3: Webstudio

apps/cli/Readme.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Linking a Project
+
+To link your project to the WebStudio CLI, follow these steps:
+
+Create a folder using
+
+```sh
+mkdir <project-name>
```
kof, line 12: this step shouldn't be necessary, we can automate this

apps/cli/Readme.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Now, `cd` into your folder. That we just created.
+
+```sh
+cd <project-name>
+```
+
+Run the following command to link the project from your workspace.
```
kof, line 21: doesn't say what it means "linking"

apps/cli/Readme.md
kof: or what are other clis doing? like vercel cli e.g.? are they requiring mkdir?

apps/cli/Readme.md
kof: I would describe what it does when you link

apps/cli/Readme.md
```diff
@@ -0,0 +1,57 @@
+## Webstudio CLI
+
+The Webstudio CLI is designed to help you seamlessly link, sync, and build projects from your web development workspace. This README will guide you through the process of setting up linking, syncing, building, and running your projects using the Webstudio CLI commands.
```
kof, line 3: here would be nice an overview of all commands

apps/cli/Readme.md
```diff
@@ -0,0 +1,57 @@
...
+
+Let's build the project. In this step, the `cli` creates the routes, pages, and scaffolds the entire app using the default remix template. All the assets like images and fonts will also be downloaded into the `assets` folder under the `public` directory.
+
+```sh
+webstudio build
+```
+
+The project is scaffolded now. You can run `npm install` and then `npm run dev` to run your app in development mode. If you want to build a production version of the app, you can run `npm run build`.
+
+## Commands
```
kof, line 41: I think this could be at the very top after intro

apps/cli/Readme.md
```diff
@@ -0,0 +1,57 @@
...
+
+- To use a local version of the `cli`, you need to clone the repository.
+
+```sh
+git clone https://github.com/webstudio-is/webstudio-builder
+```
+
+- Then run `pnpm install` to install all the dependencies.
+- Once the dependencies are installed, you can use `pnpm build` to build all the necessary packages and the CLI itself.
+- Now, change to the `app/cli` folder and run `yarn link` to create a global link for the package.
```
kof, line 56: I just learned that cli is in `apps` ...wait  why is it not a package?

apps/cli/Readme.md
```diff
@@ -0,0 +1,57 @@
...
+Now, `cd` into the folder that we just created.
+
+```sh
+cd <project-name>
+```
+
+Run the following command to link the project from your workspace.
+
+```sh
+webstudio link
```
kof, line 22: how do I even get webstudio binary?

apps/cli/Readme.md
```diff
@@ -24,36 +22,36 @@ Run the following command to link the project from your workspace.
...
 
-- `help` shows you all the commands available.
-- `version` displays the current version of the cli.
+- `help` shows you all the available commands.
+- `version` displays the current version of the CLI.
 
 ## Development setup
 
-- For using a local version of the `cli`. You need to clone the repository.
+- To use a local version of the `cli`, you need to clone the repository.
```
kof, line 48: Is the use case to contribute? If yes maybe the title should be "Contributing"

packages/cli/Readme.md
```diff
@@ -0,0 +1,57 @@
+## Webstudio CLI
+
+The Webstudio CLI is designed to help you seamlessly link, sync, and build projects from your web development workspace. This README will guide you through the process of setting up linking, syncing, building, and running your projects using the Webstudio CLI commands.
```
kof, line 3: > from your web development workspace

don't know what that means, sounds like its about checking out the entire repo but this is not the goal for the cli, somebody should be able to use the CLI without even downloading the builder repostority

packages/cli/Readme.md
```diff
@@ -0,0 +1,57 @@
...
+
+Let's build the project. In this step, the `cli` creates the routes, pages, and scaffolds the entire app using the default remix template. All the assets like images and fonts will also be downloaded into the `assets` folder under the `public` directory.
+
+```sh
+webstudio build
+```
+
+The project is scaffolded now. You can run `npm install` and then `npm run dev` to run your app in development mode. If you want to build a production version of the app, you can run `npm run build`.
+
+## Commands
```
kof, line 41: this is supposed to list all the commands in a short form, right? because sync/build are also commands

packages/cli/Readme.md
kof: unless you meant the "project folder" something ...

PR #2170, opened 2023-08-24
Fix new href property duplicated on unmount

istarkov:
## Description

On onmount outdated onSave were used. What cause creation of additional href property on instance.

- [x] - Fix duplication on unmount (1st commit)
- [x] - Fix save, to validate if multiple properties with same name exists. (2nd commit)
- [x] - Use latest prop in resolveUrlProp (3rd commit)

Use hide whitespaces.

image

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/props.ts
+++ b/packages/react-sdk/src/props.ts
@@ -147,54 +147,64 @@ export const resolveUrlProp = (
   if (instanceProps === undefined) {
     return;
   }
-  for (const prop of instanceProps) {
-    if (prop.name !== name) {
+
+  let prop = undefined;
+
+  // We had a bug that some props were duplicated https://github.com/webstudio-is/webstudio-builder/pull/2170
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/props.ts
```diff
@@ -147,54 +147,64 @@ export const resolveUrlProp = (
   if (instanceProps === undefined) {
```
kof, line 147: maybe worth adding a test, there are already some tests for this module

kof (approved):

TrySound (approved):

packages/react-sdk/src/props.ts
istarkov: Done, Thank you!

PR #2175, opened 2023-08-25
Fix font weight detection of uploaded fonts

TrySound:
Ref https://discord.com/channels/955905230107738152/1144505313358254120

Here turns out there is not standard in naming font subfamilies. To fix I expanded the list of weights wiith spaceless variants. Added "regular" keyword to detect normal font weight.

## Code Review

- [ ] hi @itsNintu, I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/fonts/src/font-weights.ts
+++ b/packages/fonts/src/font-weights.ts
@@ -1,52 +1,40 @@
 export const fontWeights = {
   "100": {
     label: "Thin",
-    name: "thin",
-    alt: "hairline",
+    names: ["thin", "hairline"],
   },
   "200": {
     label: "Extra Light",
-    name: "extra light",
-    alt: "ultra light",
+    names: ["extra light", "extralight", "ultra light", "ultralight"],
   },
   "300": {
     label: "Light",
-    name: "light",
-    alt: "light",
+    names: ["light"],
   },
   "400": {
     label: "Normal",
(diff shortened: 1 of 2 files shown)


packages/fonts/src/font-weights.ts
```diff
@@ -1,52 +1,40 @@
...
     label: "Thin",
-    name: "thin",
-    alt: "hairline",
+    names: ["thin", "hairline"],
   },
   "200": {
     label: "Extra Light",
-    name: "extra light",
-    alt: "ultra light",
+    names: ["extra light", "extralight", "ultra light", "ultralight"],
```
kof, line 8: I was trying to normalize these, why would you add this stuff that is essentially all the same?

kof: ah you added all possible names to just be able to match in the end with the number?

kof (approved):

packages/fonts/src/font-weights.ts
TrySound: Not quite. Some fonts have subfamily with spaces, some without spaces. Here I explicitly listed all.

TrySound: yes

PR #2181, opened 2023-08-25
Fix horizontal scroll and last element selection in the Tree

istarkov:
## Description

ref #2151 

https://discord.com/channels/955905230107738152/1143789794426638377

Also made that scrollbars are not intersected
image

## Steps for reproduction

Add a lot of elements into the tree.
Be sure last element has a long long label.

See it can be selected, 
See both scrolls work

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/dashboard/header.tsx
+++ b/apps/builder/app/dashboard/header.tsx
@@ -67,9 +67,9 @@ export const Header = ({ user }: { user: User }) => {
       className={containerStyle()}
     >
       <WebstudioIcon width={30} height={23} />
-      <div>
+      <Flex gap="1" align="center">
         <Menu user={user} />
-      </div>
+      </Flex>
     </Flex>
   );
 };
--- a/apps/builder/app/dashboard/header.tsx
+++ b/apps/builder/app/dashboard/header.tsx
@@ -67,9 +67,9 @@ export const Header = ({ user }: { user: User }) => {
       className={containerStyle()}
     >
       <WebstudioIcon width={30} height={23} />
-      <Flex gap="1" align="center">
+      <div>
         <Menu user={user} />
-      </Flex>
+      </div>
     </Flex>
   );
 };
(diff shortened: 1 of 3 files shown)


apps/builder/app/dashboard/header.tsx
```diff
@@ -67,9 +67,9 @@ export const Header = ({ user }: { user: User }) => {
       className={containerStyle()}
     >
       <WebstudioIcon width={30} height={23} />
-      <Flex gap="1" align="center">
+      <div>
```
kof, line 70: huh, why this change in dashboard?

packages/design-system/src/components/scroll-area.tsx
```diff
@@ -32,15 +50,31 @@ const ScrollAreaScrollbar = styled(Scrollbar, {
...
+      both: {
+        "&[data-orientation=vertical]": {
+          marginBottom: theme.spacing[4],
+        },
+        '&[data-orientation="horizontal"]': {
+          marginRight: theme.spacing[4],
+        },
+      },
+      horizontal: {},
+      vertical: {},
```
kof, line 76: souldn't you have the same thing as in both but for horizontal and vertical?

kof: ah no, all good, this is space for when both are visible

packages/design-system/src/components/tree/tree.tsx
```diff
@@ -288,6 +289,10 @@ export const Tree = <Data extends { id: string }>({
         onBlur={keyboardNavigation.handleBlur}
         onKeyDown={keyboardNavigation.handleKeyDown}
         onClick={keyboardNavigation.handleClick}
+        css={{
```
kof, line 292: shoudln't this somehow automatically be prevented when content is wrapped in scrollable? we don't do marginRight for the right scroll either ... you don't want to have to define margin every time you wan to use scroll

packages/design-system/src/components/scroll-area.tsx
istarkov: image
Only this case need to be solved

packages/design-system/src/components/tree/tree.tsx
istarkov: In most cases we don't need this margin. Here I added it only to improve a little bit last element selectability

istarkov: If you have nothing to select no need in margin

kof (approved):

PR #2189, opened 2023-08-26
(fix): Fix nested routes in the cli projects

JayaKrishnaNamburu:
## Description

With the previous version of the cli, the nested routes are not working. Now, the `prebuild` step is adopted from the `saas` project. And now, all the nested routes do work. And the `_` for the pages names is removed. Since the `remix-config` for the projects from `saas` have disabled the `v2_routeConfiguration` in it.

  - test it on preview


TrySound:
I don't think we should disable future flags. Remix v2 is coming soon. Easier to migrate now.

TrySound (approved):

PR #2197, opened 2023-08-27
Drag on canvas closest detachable

TrySound:
Before when start dragging we searched for closest rich editor under cursor and did nothing when the instance is not detachable (for example select value or switch indicator).

Here I changed the behavior to also look for closest detachable instance and drag whole thing. This way we do not need toast like in navigator and can canvas dnd just work.

Also fixed placeholder in "Select Value" component. Before was no way to change it.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
- [x] Click on select instance doesn't select it

PR #2219, opened 2023-08-31
Remove assets stubbing

TrySound:
We used to set undefined in assets map value to preserve the order between uploading and uploaded states. Now we started sorting uploaded assets by create time and stubbing no longer do anything useful.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):

TrySound:
Stubbing was used to preserve order of uploads and prevent images jumping. Though after reload the order could be different.

PR #2222, opened 2023-08-31
Add a default favicon for the projects generated using cli

JayaKrishnaNamburu:
## Description

This PR adds a default favicon for the projects generated using `cli`. Now there is a another build step, which holds all the default image assets. And stores them as `base64`. And writes back to the destination project.

The two things that are handled are

- favicon
- no-warnings flag to disable warnings when using cli

## Code Review

## Before requesting a review

- [x] made a self-review


packages/cli/templates/build.ts
```diff
@@ -114,3 +118,58 @@ export const getRouteTemplate = () => {
...
+
+    if (isDirectory) {
+      const subFolder = parseAssets(filePath);
+      if (subFolder !== undefined) {
+        folder.subFolders.push(subFolder);
+      }
+    }
+
+    if (isFile) {
+      const content = Buffer.from(readFileSync(filePath)).toString("base64");
```
TrySound, line 153: readFileSync without encoding argument is buffer already

TrySound (approved):

PR #2227, opened 2023-09-01
Add typecript typechecking  for the cli

JayaKrishnaNamburu:
## Description

This PR adds, typescript type checking for the cli, which runs in the CI pipeline too.

## Code Review

## Before requesting a review

- [x] made a self-review

### DIFF
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "extends": "@webstudio-is/tsconfig/base.json",
+  "exclude": ["./templates/*"],
   "compilerOptions": {
     "module": "esnext"
   }
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,6 +9,8 @@
     "webstudio": "./bin.js"
   },
   "scripts": {
+    "typecheck": "tsc",
+    "checks": "pnpm typecheck",
     "build:templates": "tsx ./templates/build.ts",
     "build": "build-package"
   },
@@ -31,6 +33,7 @@
     "@types/node": "^18.13.0",
     "@webstudio-is/scripts": "workspace:^",
     "@webstudio-is/tsconfig": "workspace:^",
-    "tsx": "^3.12.6"
+    "tsx": "^3.12.6",
+    "typescript": "5.1.6"
   }
 }
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -7,9 +7,8 @@ export const build: Command = async () => {
   try {
     await access(LOCAL_DATA_FILE);
     await prebuild();
-  } catch (error) {
-    console.log(error);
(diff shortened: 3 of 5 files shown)


packages/cli/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "extends": "@webstudio-is/tsconfig/base.json",
+  "exclude": ["./templates/*"],
```
TrySound, line 3: Why not check templates?

JayaKrishnaNamburu: Example, 
https://github.com/webstudio-is/webstudio/blob/add-typecheck-for-cli/packages/cli/templates/route-template.ts#L114

`ASSET_BASE` is not imported in this file. But the generated template from this can import `ASSET_BASE`. 

https://github.com/webstudio-is/webstudio/blob/add-typecheck-for-cli/packages/cli/src/__generated__/router.ts#L8

And many of the imports that are using relative path doesn't actually exist until the file is injected into users project. So, it's raising un-wanted errors.

TrySound: .ts is also wrong extension. Should be .tsx

JayaKrishnaNamburu: The extension to doesn't really matter, because we are taking it as a string in build phase. In `prebuild` when routes are actually created from this. They are marked as `tsx`

https://github.com/webstudio-is/webstudio/blob/main/packages/cli/src/prebuild.ts#L349

JayaKrishnaNamburu: This is the error, if we don't ignore the files under templates
Screenshot 2023-09-01 at 13 02 04

The files `../__generated_/index.css` won't exist in the cli. They are available only in the users repo. So, they are no actual ts bugs. SO ignored those files. We do, so the same for `eslint` too.
https://github.com/webstudio-is/webstudio/blob/main/.eslintrc.js#L9

TrySound: Let's still rename to .tsx caz it's hard to maintain utility changes when typescript is trashing module with syntax errors.

JayaKrishnaNamburu: ok, changed and pushed again 👍

packages/cli/templates/build.ts
```diff
@@ -101,7 +101,7 @@ writeFileSync(TEMPLATES_JSON_PATH, content, "utf-8");
   Building route template
 */
 const routeTemplateFile = readFileSync(
-  new URL("./route-template.ts", import.meta.url).pathname,
+  new URL("./route-template.tsx", import.meta.url).pathname,
```
TrySound, line 104: readFileSync accepts url, no need to extract pathname

TrySound (approved):

packages/cli/templates/build.ts
JayaKrishnaNamburu: Need to up my node game 😅

PR #2228, opened 2023-09-01
New package name for CLI

kof:
- We decided to use `npx webstudio` for simplicity and new package name is nicer
- Added the license


TrySound:
pnpm install is necessary

kof:
pnpm install didn't modify anything

TrySound (approved):

PR #2229, opened 2023-09-01
Enforce local style source at the end

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/1965

Tweaked logic and dnd to always put local style source in the end and tokens before it.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
@@ -104,7 +104,7 @@ export const useSortable = <Item extends { id: string; source: ItemSource }>({
     onStart({ data: itemId }) {
       const item = items.find((item) => item.id === itemId);
       if (items.at(-1) === item && item?.source === "local") {
-        toast.error("Local token is always last and can not be moved");
+        toast.error("Local style source is always last and can not be moved");
         useDragHandlers.cancelCurrentDrag();
         return;
       }
--- a/apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
@@ -8,7 +8,9 @@ import {
   useDrop,
   type DropTarget,
   computeIndicatorPlacement,
+  toast,
 } from "@webstudio-is/design-system";
+import type { ItemSource } from "./style-source";
 
 type UseSortable<Item> = {
   items: Array<Item>;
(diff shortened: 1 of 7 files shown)


kof:
- [ ] What should happen to the "backspace" it used to delete the last token, but it can't delete local, so it does nothing

apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
```diff
@@ -81,10 +94,24 @@ export const useSortable = <Item extends { id: string }>({
...
+
+      return itemId;
     },
-    onStart({ data }) {
-      setDragItemId(data);
+    onStart({ data: itemId }) {
+      const item = items.find((item) => item.id === itemId);
+      if (items.at(-1) === item && item?.source === "local") {
+        toast.error(
+          "Local style source should be last and cannot be reordered"
```
kof, line 108: @taylornowotny this wording ok? its shown when you try to reorder local

kof:
We have 2 options with backspace: 
1. delete this ability entirely
2. delete the token after the local

I am not sure 2 is not going to cause problems, because local looks visually like other tokens.

taylornowotny:
I suggest removing the backspace-delete functionality as it no longer makes sense with Local being at the end.

It would not be used in the redesigned Styles Sources UI either, if we ever end up building that.

apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
taylornowotny: Good idea @TrySound , wording can improve to: "Local token is always last and can not be moved"

kof:
> I suggest removing the backspace-delete functionality as it no longer makes sense with Local being at the end.
> 
> It would not be used in the redesigned Styles Sources UI either, if we ever end up building that.

alright

apps/builder/app/builder/features/style-panel/style-source/use-sortable.tsx
```diff
@@ -81,10 +94,22 @@ export const useSortable = <Item extends { id: string }>({
...
+      }
+
+      return itemId;
     },
-    onStart({ data }) {
-      setDragItemId(data);
+    onStart({ data: itemId }) {
+      const item = items.find((item) => item.id === itemId);
+      if (items.at(-1) === item && item?.source === "local") {
+        toast.error("Local token is always last and can not be moved");
```
kof, line 107: Local style, not token

taylornowotny: This is the wording that I chose. If you want we can change it to "Local style source" but not "style" by itself

taylornowotny: or perhaps more semantic: "Local styles are always last and can not be moved."

kof: works, thanks

kof (approved):

PR #2232, opened 2023-09-01
Add missing build files for `npm` publish for cli

JayaKrishnaNamburu:
This PR will revert the name of the package back to `@webstudio/cli`. And adds the `files` field in package.json. This will fix the publish of files to npm

## Code Review

## Before requesting a review
- [x] made a self-review with `pnpm pack` and `npm pack` it is bundling `lib` folder too now.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -29,6 +29,17 @@
     "p-limit": "^4.0.0",
     "strip-indent": "^4.0.0"
   },
+  "exports": {
+    "source": "./src/cli.ts",
+    "link": "./lib/commands/link.js",
+    "build": "./lib/commands/build.js",
+    "sync": "./lib/commands/sync.js"
+  },
+  "files": [
+    "lib/*",
+    "src/*",
+    "!*.test.*"
+  ],
   "devDependencies": {
     "@types/node": "^18.13.0",
     "@webstudio-is/scripts": "workspace:^",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -8,6 +8,11 @@
   "bin": {
     "webstudio": "./bin.js"
   },
+  "files": [
+    "lib/*",
+    "src/*",
+    "!*.test.*"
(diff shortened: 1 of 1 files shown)


packages/cli/package.json
```diff
@@ -29,6 +29,17 @@
     "p-limit": "^4.0.0",
     "strip-indent": "^4.0.0"
   },
+  "exports": {
+    "source": "./src/cli.ts",
+    "link": "./lib/commands/link.js",
+    "build": "./lib/commands/build.js",
+    "sync": "./lib/commands/sync.js"
+  },
```
TrySound, line 37: I'm not sure if we need any exports at all

packages/cli/package.json
```diff
@@ -29,6 +29,17 @@
...
+  "files": [
+    "lib/*",
+    "src/*",
+    "!*.test.*"
+  ],
```
TrySound, line 42: please put it before scripts field

JayaKrishnaNamburu: But in other packages like `css-data` `css-vars` the field is after the `scripts` one

packages/cli/package.json
JayaKrishnaNamburu: We actually don't need them. Add as it is mentioned in discord. Will remove them then

packages/cli/package.json
JayaKrishnaNamburu: Made the changes 👍

packages/cli/package.json
```diff
@@ -8,6 +8,11 @@
   "bin": {
     "webstudio": "./bin.js"
   },
+  "files": [
+    "lib/*",
+    "src/*",
+    "!*.test.*"
+  ],
```
TrySound, line 15: Now you can see bin is missing in the list

TrySound: templates too

TrySound (approved):

PR #2238, opened 2023-09-03
Prevent submit inside dialog in builder

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/2233

Capture mode fixes unexpected submit of form when placed inside radix dialog. Though didn't find a trace why this happen. There is `form.method = 'dialog'` which has special behavior inside `<dialog>` but it is not our case.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
doesn't it make sense to capture all of the events, just in case, it would make it more robust, no?

istarkov (approved):

PR #2250, opened 2023-09-05
Switch the  CLI to using yargs

istarkov:
## Description

Switch on yargs in cli parser.

Shows help on specific commands too

image

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/cli/src/commands/yargs-types.ts
```diff
@@ -0,0 +1,65 @@
+/*
```
kof, line 1: that's weird, why is this not in the package typings ... I guess rhetorical question

istarkov: In most cases this types are not needed.
Like
```
cmd.command(
  ['deploy [script]'],
  '🆙 Deploy your Worker for Platforms to Cloudflare.',
  // builder
  (yargs: Argv) =>
    yargs
      .positional('script', {
        describe: 'The path to an entry point for your worker',
        type: 'string',
        requiresArg: true,
        demandOption: true,
      })
      .option('name', {
        describe: 'Name of the worker',
        type: 'boolean',
        demandOption: true,
      }),
  // handler
  async (options) => {
    console.log(options.name);
  }
);
```
will be extracted automatically.
Cloudflare made 2 things:
1. Allow  command builder to be in the other file
2. Make handler more strict (removing [key:string]: unknown) from type so `options.anyRandomName` start giving an error (in original yargs it is not an error just `unknown`)

kof: thats intetresting info that would be nice in there

kof (approved):

JayaKrishnaNamburu (approved):
Looks good 💯

TrySound (approved):

PR #2252, opened 2023-09-06
Add guided flow using prompts for cli

JayaKrishnaNamburu:
## Description

This PR adds prompts, when the cli is ran using `webstudio` and without any commands. It requests the users if they want to create a folder. If yes, it will create a folder and changes the directory to the newly created ones.

This PR even allows `webstudio link` command to be executed without prompts using

```sh
webstudio link -l <link>
```

## Code Review

  - test it on preview
 
## Before requesting a review
- [x] made a self-review


packages/cli/src/commands/flow.ts
```diff
@@ -0,0 +1,50 @@
...
+import { join } from "node:path";
+import { link } from "./link";
+import { sync } from "./sync";
+import { build } from "./build";
+
+export const flow = async () => {
+  const isProjectCofnigured = await isFileExists(".webstudio/config.json");
+  const prompsList: PromptObject[] = [];
+
+  if (isProjectCofnigured === false) {
```
istarkov, line 13: typo configured

packages/cli/src/commands/link.ts
```diff
@@ -4,10 +4,29 @@ import * as readline from "node:readline/promises";
...
+
+export const link = async (
+  options: StrictYargsOptionsToInterface<typeof linkOptions> | { link?: string }
+) => {
+  let shareLink: string;
+  if (options.link) {
+    shareLink = options.link;
+  } else {
+    const rl = readline.createInterface({ input: stdin, output: stdout });
+    shareLink = await rl.question(`Paste share link (with build access): `);
```
istarkov, line 27: view access should be enough?

istarkov (approved):

packages/cli/src/commands/link.ts
kof: - view access is enough, so not worth mentioning at all, because its the lowest permission level anyways
- Better wording: `Please paste a link provided in the "Share" dialog in the Builder`

kof: updated wording again

kof (approved):

packages/cli/package.json
```diff
@@ -34,11 +34,13 @@
     "env-paths": "^3.0.0",
     "ora": "^7.0.1",
     "p-limit": "^4.0.0",
+    "prompts": "^2.4.2",
     "strip-indent": "^4.0.0",
     "yargs": "^17.7.2"
   },
   "devDependencies": {
     "@types/node": "^18.13.0",
+    "@types/prompts": "^2.4.4",
```
kof, line 43: that lib had a definition inside, they seem to have migrated the typedefs from def. typed to their package

kof: see https://github.com/terkelg/prompts

kof: idk whats the difference between those types, so just a hint

JayaKrishnaNamburu: My `zed` editor doesn't seem to recognise them. I will switch to `vscode` and check them 👍

JayaKrishnaNamburu: Looks like they are still not exposing the types in the same package when they publish to `npm`
https://www.npmjs.com/package/prompts?activeTab=code

PR #2263, opened 2023-09-07
Show spinners after the prompts are responded and close the prompts on `ctrl + c` in the cli

JayaKrishnaNamburu:
## Description

This PR
- Allows the prompts to be cancelable at any time.
- Prompts the user if the dependencies can be installed, when the project is being connected for the first time.
- Shows the spinner once all the prompts are responded. Which was previously clearing the prompts to be displayed.

## Code Review

  - test it on preview

## Before requesting a review
- [x] made a self-review

### DIFF
--- a/packages/cli/src/commands/init-flow.ts
+++ b/packages/cli/src/commands/init-flow.ts
@@ -1,4 +1,3 @@
-import prompts from "prompts";
 import { ensureFolderExists, isFileExists } from "../fs-utils";
 import { chdir, cwd, stdout as shellOutput } from "node:process";
 import { spawn } from "node:child_process";
@@ -7,6 +6,7 @@ import ora from "ora";
 import { link } from "./link";
 import { sync } from "./sync";
 import { build, buildOptions } from "./build";
+import { prompt } from "../prompts";
 import type { StrictYargsOptionsToInterface } from "./yargs-types";
 
 export const initFlow = async (
@@ -16,37 +16,19 @@ export const initFlow = async (
   let shouldInstallDeps = false;
 
   if (isProjectConfigured === false) {
-    const { createFolder } = await prompts([
-      {
-        type: "confirm",
-        name: "createFolder",
-        message: "Do you want to create a folder",
-        initial: true,
-        onState: (state) => {
-          if (state.aborted) {
-            process.nextTick(() => {
-              process.exit(0);
-            });
-          }
-        },
-      },
-    ]);
+    const { shouldCreateFolder } = await prompt({
+      type: "confirm",
+      name: "shouldCreateFolder",
+      message: "Do you want to create a folder",
+      initial: true,
+    });
 
-    if (createFolder === true) {
-      const { folderName } = await prompts([
-        {
-          type: "text",
-          name: "folderName",
-          message: "Enter a project name",
-          onState: (state) => {
-            if (state.aborted) {
-              process.nextTick(() => {
-                process.exit(0);
-              });
-            }
-          },
-        },
-      ]);
+    if (shouldCreateFolder === true) {
+      const { folderName } = await prompt({
+        type: "text",
+        name: "folderName",
+        message: "Enter a project name",
+      });
 
       if (folderName === undefined) {
         throw new Error("Folder name is required");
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/init-flow.ts
```diff
@@ -13,46 +13,71 @@ export const initFlow = async (
...
   if (isProjectConfigured === false) {
-    prompsList.push(
+    const { createFolder } = await prompts([
       {
         type: "confirm",
-        name: "folder",
+        name: "createFolder",
         message: "Do you want to create a folder",
         initial: true,
+        onState: (state) => {
```
istarkov, line 24: Mb wrap onState logic into own propmts wrapper
```
import prompts from './promts';// <-- our wrapper
await prompts({ type, name, message, initial });
```

packages/cli/src/commands/init-flow.ts
```diff
@@ -13,46 +13,71 @@ export const initFlow = async (
...
+    }
+    await link({ link: projectLink });
   }
 
-  /*
-    If the project is already set up, we can sync and build it.
-  */
   await sync();
   await build(options);
+  const spinner = ora().start();
```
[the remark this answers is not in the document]
JayaKrishnaNamburu: They have their own spinner states inside, when the commands are ran independently.

packages/cli/src/commands/init-flow.ts
```diff
@@ -13,58 +13,90 @@ export const initFlow = async (
   options: StrictYargsOptionsToInterface<typeof buildOptions>
 ) => {
   const isProjectConfigured = await isFileExists(".webstudio/config.json");
-  const prompsList: PromptObject[] = [];
-  const spinner = ora().start();
+  let shouldInstallDeps = false;
 
   if (isProjectConfigured === false) {
-    prompsList.push(
+    const { createFolder } = await prompts([
```
TrySound, line 19: shouldCreateFolder is better name, createFolder implies function

packages/cli/src/commands/init-flow.ts
```diff
@@ -7,64 +6,68 @@ import ora from "ora";
...
-        name: "folder",
-        message: "Do you want to create a folder",
-        initial: true,
-      },
-      {
-        type: (prev) => (prev === true ? "text" : null),
+    const { shouldCreateFolder } = await prompt({
+      type: "confirm",
+      name: "shouldCreateFolder",
+      message: "Do you want to create a folder",
```
kof, line 22: "Would you like to create a folder?"

packages/cli/src/commands/init-flow.ts
```diff
@@ -7,64 +6,68 @@ import ora from "ora";
...
         message: "Enter a project name",
-      },
-      {
-        type: (prev) => (prev === undefined ? null : "text"),
-        name: "projectLink",
-        message: "Enter a project link",
+      });
+
+      if (folderName === undefined) {
+        throw new Error("Folder name is required");
```
kof, line 34: why? if they don't enter it, it goes to "./" like we discussed

packages/cli/src/commands/init-flow.ts
```diff
@@ -7,64 +6,68 @@ import ora from "ora";
...
-    await link({ link: response.projectLink });
+    if (projectLink === undefined) {
+      throw new Error(`Project Link is required`);
+    }
+    await link({ link: projectLink });
+
+    const { installDeps } = await prompt({
+      type: "confirm",
+      name: "installDeps",
+      message: "Do you want to install dependencies",
```
kof, line 54: Would you like to install dependencies?

kof: We need to treat the cli messaging like we are dealing with a customer in a UI

kof: Please check all messages, run them through grammarly or notion ai for correctness

packages/cli/src/commands/init-flow.ts
JayaKrishnaNamburu: It does go, this is if they selected the option to create a folder and still don't enter 👍

PR #2266, opened 2023-09-07
Update text inside cli init flow

JayaKrishnaNamburu:
## Description

This PR updates the text from the cli's init flow. The changes are request from the #2263

## Code Review

### DIFF
--- a/packages/cli/src/commands/init-flow.ts
+++ b/packages/cli/src/commands/init-flow.ts
@@ -19,7 +19,7 @@ export const initFlow = async (
     const { shouldCreateFolder } = await prompt({
       type: "confirm",
       name: "shouldCreateFolder",
-      message: "Do you want to create a folder",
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/init-flow.ts
```diff
@@ -62,7 +62,7 @@ export const initFlow = async (
 
   if (shouldInstallDeps === true) {
     const spinner = ora().start();
-    spinner.text = "Installing dependencies";
+    spinner.text = "Installing Dependencies";
```
TrySound, line 65: Why "Dependencies"? Not a title or heading, just a sentence.

JayaKrishnaNamburu: Changed it, i was using `D` in the same file. Now changed in both the places 👍

TrySound (approved):

PR #2285, opened 2023-09-08
Fix local install of cli binaries

istarkov:
## Description

Fixes local installation of cli 
The issue was that at `pnpm install` 
`./lib/bin.js` wasn't exists.

Solution is to create it and forward all calls to `./lib/bin`

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/cli/bin.js
+++ b/packages/cli/bin.js
@@ -1,4 +1,3 @@
 #!/usr/bin/env node --experimental-specifier-resolution=node --no-warnings
-"use strict";
 // eslint-disable-next-line import/no-internal-modules
 import "./lib/bin";
--- a/packages/cli/bin.js
+++ b/packages/cli/bin.js
@@ -1,0 +1,4 @@
+#!/usr/bin/env node --experimental-specifier-resolution=node --no-warnings
+"use strict";
+// eslint-disable-next-line import/no-internal-modules
+import "./lib/bin";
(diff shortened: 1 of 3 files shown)


packages/cli/bin.js
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node --experimental-specifier-resolution=node --no-warnings
+"use strict";
+// eslint-disable-next-line import/no-internal-modules
+import "./lib/bin";
```
TrySound, line 4: Don't forget extension

kof (approved):

packages/cli/bin.js
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node --experimental-specifier-resolution=node --no-warnings
+"use strict";
```
TrySound, line 2: no need use strict it is default

packages/cli/bin.js
istarkov: esbuild doesn't generate extensions on imports, so do I

istarkov: See ./lib folder

TrySound: Please add it. I'll figure out what to do with generated modules.

PR #2288, opened 2023-09-08
Escape special characters in route names in cli projects

JayaKrishnaNamburu:
## Description

Add `[]` to escape special characters in route names.

## Code Review

## Before requesting a review

- [x] made a self-review


kof (approved):

istarkov:
I would give a link to docs https://remix.run/docs/en/main/file-conventions/route-files-v2#escaping-special-characters

And its time to add test as even logic is simple now.

Even this logic `/x/y => [x].[y]._index.ts` needs explanation and test.
Must be explained in comment why we are using nested route syntax with `_index` postfix, (to avoid layout nesting blabla)

Some day we could change this, to use more beautiful routes, to use $lang features, use folders to group huge projects etc. So test and good comments and doc links would be very useful

JayaKrishnaNamburu:
The `_index` is kinda mandatory at the moment. Let's say there is a route `/test.one.tsx` and then there is a `/text.tsx` route. Remix doesn't pick the `/text.tsx` by default unless we mention the `_index` at the end. 

Or else it's just picking the first route that matches the `/test` and was quite not reliable. So, by default marking the page as a route with `_index`. So deep nested ones works as expected.

istarkov:
Write this in code comments, also see I add in my comment alternative syntax showing that its not the only way to do that and 1 line shorter ;) in code

JayaKrishnaNamburu:
@istarkov added comments and updated fixtures.

istarkov (approved):

PR #2289, opened 2023-09-08
Throw error message if the project config is missing in the users env when using cli

JayaKrishnaNamburu:
…when using cli

## Description

Adds a error message when the project config like, `host` and `token` are missing from the root.

## Code Review

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -38,7 +38,7 @@ export const main = async () => {
         alias: "help",
         type: "boolean",
       })
-      .scriptName("webstudio")
+      .scriptName("webstudio-cli")
       .usage(
         `Webstudio CLI (${packageJson.version}) allows you to setup, sync, build and preview your project.`
       );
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/sync.ts
```diff
@@ -36,7 +36,16 @@ export const sync = async () => {
     return;
   }
 
-  const { host, token } = globalConfig[localConfig.projectId];
+  const projectConfig = globalConfig[localConfig.projectId];
+
+  if (projectConfig === undefined) {
+    spinner.fail(
+      `Project config is not found. Please make sure the project is linked. \n Use webstudio link command to link your project`
```
kof, line 43: `Project config is not found, please run "webstudio-cli link".

Also would be good to consistently use color for commands, example from init: `      `Run ${pc.dim("npx vercel")} to publish on Vercel.`,`

packages/cli/src/cli.ts
```diff
@@ -38,7 +38,7 @@ export const main = async () => {
         alias: "help",
         type: "boolean",
       })
-      .scriptName("webstudio")
+      .scriptName("webstudio-cli")
```
JayaKrishnaNamburu, line 41: @kof @istarkov i changed this to `webstudio-cli` since the bin is changed to `webstudio-cli` too in package.json

Or it's going to trigger for both ?

PR #2314, opened 2023-09-13
feat: AI Copywriter

gsppe:
## Description

Implement AI copywriter (no final UI yet) behind feature flag. The feature consists of:

- a `copywriter` chain to generate the copy with AI
- a /rest/ai/copy endpoint to handle generation requests. The response is streamed as plain text
- the temporary UI which is a tab in the right sidebar. 
 
For every selected instance it allows to generate copy for it and its subtree. A `collectTextInstances` utility creates an array of text nodes objects with the following shape:

```ts
{
  instanceId: string;
  index: number;
  type: "Heading" | "Paragraph";
  text: string; // current text or empty
}
```

The AI responds with an array of these where the text property contains the generated text. The frontend parses incrementally and updates a Webstudio instance by `instanceId`.

The `index` is used to replace the proper text node in the `instance.children` array. When all the children are text nodes these are merged into a single text node, otherwise the AI would generate `children.length` copy. We identify this case by setting the `index` to `-1`.

The following tasks will be done in a follow-up PR when the design is ready

- [ ] retry logic and errors handling
- [ ] design and final UX from @taylornowotny 
- [ ] rate limiting logic in apps/builder/app/routes/rest.ai.copy - will implement this in a separate PR with Upstash
- [ ] tests

### DIFF
--- a/apps/builder/app/builder/features/ai/copywriter.tsx
+++ b/apps/builder/app/builder/features/ai/copywriter.tsx
@@ -23,10 +23,7 @@ const patchTextInstance = (textInstance: copywriter.TextInstance) => {
   store.createTransaction([instancesStore], (instances) => {
     const currentInstance = instances.get(textInstance.instanceId);
 
-    if (
-      currentInstance === undefined ||
-      currentInstance.children.length === 0
-    ) {
+    if (currentInstance === undefined) {
       return;
     }
 
@@ -61,10 +58,23 @@ const onChunk = (completion: string) => {
     }
 
     patchTextInstance(currenTextInstance);
-  } catch (error) {
+  } catch {
     /**/
   }
 };
+
+const $textInstances = computed(
+  [instancesStore, selectedInstanceStore],
+  (instances, selectedInstance) => {
+    if (selectedInstance) {
+      return copywriter.collectTextInstances({
+        instances: instancesStore.get(),
+        rootInstanceId: selectedInstance.id,
+      });
+    }
+    return [];
+  }
+);
 
 export const Copywriter = () => {
   const [isLoading, setIsLoading] = useState(false);
@@ -72,20 +82,8 @@ export const Copywriter = () => {
 
   const project = useStore(projectStore);
 
-  const textInstances = computed(
-    [instancesStore, selectedInstanceStore],
-    (instances, selectedInstance) => {
-      if (selectedInstance) {
-        return copywriter.collectTextInstances({
-          instances: instancesStore.get(),
-          rootInstanceId: selectedInstance.id,
-        });
-      }
-      return [];
-    }
-  );
-
-  const textInstancesCount = textInstances.get().length;
+  const textInstances = useStore($textInstances);
+  const textInstancesCount = textInstances.length;
 
   return (
     <Box>
@@ -161,7 +159,7 @@ export const Copywriter = () => {
             <input
               type="hidden"
               name="textInstances"
-              value={JSON.stringify(textInstances.get())}
+              value={JSON.stringify(textInstances)}
             />
 
             <Button type="submit" disabled={isLoading}>
--- a/packages/ai/src/chains/copywriter/index.ts
+++ b/packages/ai/src/chains/copywriter/index.ts
@@ -19,7 +19,11 @@ import type { Instance, Instances } from "@webstudio-is/sdk";
 export const TextInstanceSchema = z.object({
   instanceId: z.string(),
   index: z.number(),
-  type: z.union([z.literal("Heading"), z.literal("Paragraph")]),
+  type: z.union([
+    z.literal("Heading"),
+    z.literal("Paragraph"),
+    z.literal("Text"),
+  ]),
   text: z.string(),
 });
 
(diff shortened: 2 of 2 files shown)


TrySound:
Probably something to do with non main base branch

apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -0,0 +1,187 @@
...
+    if (replaceAll) {
+      if (currentInstance.children.every((child) => child.type === "text")) {
+        currentInstance.children = [{ type: "text", value: textInstance.text }];
+      }
+      return;
+    }
+
+    if (currentInstance.children[textInstance.index].type === "text") {
+      currentInstance.children[textInstance.index].value = textInstance.text;
+      instances.set(textInstance.instanceId, currentInstance);
```
giuseppeg, line 47: is this necessary or mutating `currentInstance` will just work? I am asking because I realized that at line 40 I don't do an `instance.set` after the update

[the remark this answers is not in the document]
TrySound: you can remove, should work

apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+export const Copywriter = () => {
+  const [isLoading, setIsLoading] = useState(false);
+  const abort = useRef<AbortController | null>(null);
+
+  const project = useStore(projectStore);
+  const instance = useStore(selectedInstanceStore);
+
+  const textInstances = useMemo(() => {
```
TrySound, line 77: This can be a local computed store
```
computed([instancesStore, selectedInstanceStore], (instances, selectedInstance) => {
  ...
})
```

taylornowotny:
Designs are ready! https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=7500-40455&mode=design

I'm available to help if you have any questions

apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -0,0 +1,191 @@
...
+  Label,
+  Text,
+} from "@webstudio-is/design-system";
+import {
+  instancesStore,
+  projectStore,
+  selectedInstanceStore,
+} from "~/shared/nano-states";
+import { useStore } from "@nanostores/react";
+import { collectTextInstances } from "node_modules/@webstudio-is/ai/src/chains/copywriter";
```
TrySound, line 19: bad import

pnpm-lock.yaml
```diff
@@ -6319,13 +6322,20 @@ packages:
       '@remix-run/web-stream': 1.0.4
       web-encoding: 1.1.5
 
-  /@remix-run/web-fetch@4.3.6:
-    resolution: {integrity: sha512-ifadyJS+/7W6LhKyA8tIR9fBIPwLCFVpl1YCYg5i0ikykiXxE3IWtPVB1G51AJFghc1YgR7rq8BRJLsJeUbE5Q==}
+  /@remix-run/web-blob@3.1.0:
```
TrySound, line 6325: Is bumping remix deps necessary here? Caz they seem to be duplicated.

apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  const textInstances = computed(
+    [instancesStore, selectedInstanceStore],
+    (instances, selectedInstance) => {
+      if (selectedInstance) {
+        return collectTextInstances({
+          instances: instancesStore.get(),
+          rootInstanceId: selectedInstance.id,
+        }).map((instance) => {
+          if (instance.text === "Text you can edit") {
```
TrySound, line 84: Very flaky assumption

kof: how do we detect these? maybe we should have used a placeholder property with an empty text property? or something

kof: Otherwise we need to extract text to some const or meta and compare to see if its a default

kof: I am sure @gsppe just doesn't have a solution for doing this well and we could help out

kof: could be instance.text || instance.placeholder for rendering, then this check would be basically `if (instance.text)`

kof: we are discussing it with @TrySound, ignore it for now

TrySound: We decided to remove placeholders from data so you can safely remove this conversion too.
https://github.com/webstudio-is/webstudio/pull/2399

apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -0,0 +1,184 @@
...
+      .parse(JSON.parse(untruncateJson(completion)));
+
+    const currenTextInstance = jsonResponse.pop();
+
+    if (currenTextInstance === undefined) {
+      return;
+    }
+
+    patchTextInstance(currenTextInstance);
+  } catch (error) {
```
TrySound, line 64: (error) can be omitted

apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -0,0 +1,184 @@
...
+        return copywriter.collectTextInstances({
+          instances: instancesStore.get(),
+          rootInstanceId: selectedInstance.id,
+        });
+      }
+      return [];
+    }
+  );
+
+  const textInstancesCount = textInstances.get().length;
```
TrySound, line 88: Move into module level to not recreated on every render and access with `useStore($textInstances)` to subscribe

apps/builder/app/builder/features/ai/copywriter.tsx
TrySound: `try {} catch {}`?

PR #2323, opened 2023-09-16
Pregenerate forms properties

TrySound:
Here decoupled actions from props data in runtime by generating prepared forms data. It will let us get rid of props after migrating to jsx generator.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
I barely understand what this is about ...

TrySound:
Before correct method and action for specific form were inferred in remix action from props data. Now we regenerate map with only form info to not bloat runtime (both client and server) with props data.

istarkov (approved):

PR #2325, opened 2023-09-17
fix: Stop scaling instance labels

kof:
Closes https://github.com/webstudio-is/webstudio/issues/1990

## Description

We were scaling instance labels along with instaces

## Steps for reproduction

1. select instance
2. resize canvas to scale it up and down
3. labels get smaller, but should stay same size while instance scales
4. outline position should stay correct

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/workspace/canvas-tools/outline/use-outline.ts
```diff
@@ -0,0 +1,23 @@
+import { useStore } from "@nanostores/react";
+import type { Instance } from "@webstudio-is/sdk";
+import type { Atom } from "nanostores";
+import { scaleStore } from "~/builder/shared/nano-states";
+
+export const useOutline = (
```
TrySound, line 6: Can be computed

TrySound: Ah, got it. You pass different stores with same shape.

TrySound (approved):

apps/builder/app/builder/features/workspace/canvas-tools/outline/use-outline.ts
kof: yes

kof:
Found a bug with scaling during d&d, fixing ...

GoharAkram (approved):
- [ ] Tested by adding components and scaling the canvas up and down
Works fine now and does not decrease instance label scaling below 100%

kof:
@GoharAkram additionally you needed to test if adding instance outlines the target instance correctly along with placement indicator

TrySound (approved):

TrySound:
Getting shifted outline
image

kof:
@TrySound good catch
@GoharAkram missed it

apps/builder/app/builder/features/workspace/canvas-tools/outline/label.tsx
```diff
@@ -16,24 +16,22 @@ type LabelRefCallback = (element: HTMLElement | null) => void;
...
     (element: null | HTMLElement) => {
-      if (element === null || instanceRect === undefined) {
+      if (element === null || rect === undefined) {
         return;
       }
       const labelRect = element.getBoundingClientRect();
       let nextPosition: LabelPosition = "top";
-      if (labelRect.height > instanceRect.top) {
-        nextPosition = instanceRect.height < 250 ? "bottom" : "inside";
+      if (labelRect.height > rect.top) {
```
istarkov, line 29: Why we check `height` here? And not the similar dimension like bottom or top?

istarkov (approved):

apps/builder/app/builder/features/workspace/canvas-tools/outline/label.tsx
kof: - didn't change this logic and actually forgot to revert the change I made in renaming
- tested logic works correctly, the change you are proposing breaks a use case

GoharAkram:
Hey @kof This is just a final review of the initial PR or has something specifically changed since the last review?

GoharAkram (approved):
- [ ] This was tested. 

Added instances and checked that scaling canvas above or below 100% doesn't affect the labels. It only scales the component instances.

also the selected instance's outline is restricted to canvas now, This fixes the earlier issue

PR #2328, opened 2023-09-17
Preview on canvas when style panel select value is highlighted

kof:
https://github.com/webstudio-is/webstudio/issues/2324

https://github.com/webstudio-is/webstudio/assets/52824/7489689f-92bb-4cd5-888b-748bcae2369c

## Description

With this PR all select controls get a preivew when user hovers with the mouse or uses arrow keys.
We use select control in:
- position
- blend mode
- font-weight
- background size
- many others

In many cases preview won't be visible because cursor doesn't change unless you actually have mouse over that element

## Steps for reproduction

1. hover over position value
2. when mouse out preview should be gone
3. arrow keys do the same
4. if value wasn't applied and select box closed - preview should reset

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

GoharAkram (approved):
- [ ] tested in blend and position settings as per instructions.. it works.

kof:
Added font weight and background size previewing as well as these were separate code paths before.

GoharAkram:
@kof Tested on font weight.. Works fine.

 What do you mean exactly by background size.. which section is this value in?

GoharAkram (approved):
- [ ] Tested on font weight and bg size.. it works fine

TrySound (approved):

PR #2337, opened 2023-09-18
[Refactor] Replace Node eval Hack with Dynamic Import

istarkov:
## Description

Use dynamic import instead of `node --eval`
`pathToFileURL` is used in import to support windows

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/cli/src/prebuild.ts
+++ b/packages/cli/src/prebuild.ts
@@ -43,8 +43,8 @@ import { LOCAL_DATA_FILE } from "./config";
 import { ensureFileInPath, ensureFolderExists, loadJSONFile } from "./fs-utils";
 import merge from "deepmerge";
 import { createImageLoader } from "@webstudio-is/image";
-import { $ } from "execa";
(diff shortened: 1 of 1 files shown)


packages/cli/src/prebuild.ts
```diff
@@ -197,14 +197,11 @@ export const prebuild = async (options: {
     await copyTemplates(options.template);
   }
 
-  const constantsJson =
-    await $`node --input-type=module --eval ${`import * as consts from './app/constants.mjs'; console.log(JSON.stringify(consts))`}`;
+  const constants: typeof sharedConstants = await import(
+    pathToFileURL(join(cwd(), "/app/constants.mjs")).href
```
TrySound, line 201: leading slash in the path look a little confusing. Should be safe to remove.

TrySound (approved):

PR #2341, opened 2023-09-20
perf: generate html react props from html spec

TrySound:
Here scraped html spec to generate attributes per tag and added converter to out props meta format.

In the next PR will reuse for settings panel and remove html attributes from generated comomponent props.

Having html attributes not tied to specific framework will let us generate formats compatible with any framework.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/html-data/src/find-html-attributes.ts
```diff
@@ -0,0 +1,119 @@
+import { writeFile, mkdir } from "node:fs/promises";
+import { Parser, defaultTreeAdapter } from "parse5";
+import type { Attribute, AttributesByTag } from "./types";
+import { htmlPropsDescriptions } from "./descriptions";
+
+// scrub attributes data from https://html.spec.whatwg.org/multipage/indices.html#attributes-3
```
kof, line 6: wasn't this available in mdn-data package? or some other package? I know entire HTML spec is in jsons at this point, even the official spec has a repo, scraping that stuff seems fragile

PR #2347, opened 2023-09-20
build: use webstudio export condition for tools

TrySound:
This fixes the issue with eventsource-parser package having source export condition with different setup.

Now everywhere "webstudio" is used to reference our sources.

## Code Review

- [ ] hi @gsppe, I need you to do

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/.storybook/main.ts
+++ b/apps/builder/.storybook/main.ts
@@ -16,7 +16,7 @@ export default {
       },
       resolve: {
         ...config.resolve,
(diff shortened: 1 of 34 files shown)


TrySound:
> Are we required to pass --conditions=webstudio every time we use tsx now?
You had to before too. Node relies on import condition by default.

istarkov (approved):

PR #2356, opened 2023-09-22
perf: generate asset prop for links and images

TrySound:
Before links and images required logic to traverse raw props data and assets to resolve href and src. This can potentially affect rendering performance and requires additional data to be stored in runtime.

Here pregenerated props with "asset" type as a first step. Images still have to be aware of asset base url to trim it from resolved src but no longer access to assets store. Links now interpret it as string url.

See fixtures for examples.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov:
So now this
image

istarkov (approved):

PR #2358, opened 2023-09-23
chore: Add checks for template names for build command in cli

JayaKrishnaNamburu:
## Description

Small changes that adds checks for templates and hints for templates in the `--help` command.

## Steps for reproduction

```sh
webstudio-cli --help
```

## Code Review


packages/cli/src/commands/build.ts
```diff
@@ -21,7 +21,9 @@ export const buildOptions = (yargs: CommonYargsArgv) =>
     .option("template", {
       type: "string",
       default: "vercel",
-      describe: "[Experimental] Template to use for the build",
+      describe: `[Experimental] Template to use for the build
+Templates Available
+${PROJECT_TEMPALTES.join(`\n`)}`,
```
TrySound, line 26: You can try `choices: ['xs', 's', 'm', 'l', 'xl']` field

packages/cli/src/prebuild.ts
```diff
@@ -198,6 +198,16 @@ export const prebuild = async (options: {
    **/
   template?: string;
 }) => {
+  if (
+    options.template !== undefined &&
+    options.template.startsWith("./") === false &&
```
TrySound, line 203: Just `.`. User may refer to parent folder.

packages/cli/src/commands/build.ts
JayaKrishnaNamburu: We can't use the `choices` because local templates will fail the check. As `yargs` itself is throwing error if users pass
```sh
webstudio-cli build --template ./template
```

something like this.

packages/cli/src/commands/init-flow.ts
```diff
@@ -119,3 +112,7 @@ export const initFlow = async (
       .join("\n")
   );
 };
+
+export const convertToTitleCase = (str: string) => {
```
kof, line 116: we are using in builder import { titleCase } from "title-case";

kof: and we also have a humanizeString utility, but I guess you only need title-case

kof (approved):

JayaKrishnaNamburu:
Renamed `netlify-edge-functions` to `netlify`. So people can use edge as default. And the same reflects in `init-flow` too. The options will be

```sh
[ ] Vercel
[ ] Netlify
[ ] Netlify Functions
```

packages/cli/src/commands/init-flow.ts
JayaKrishnaNamburu: But the package seems to change the `netlify-functions` to `
Netlify-Functions` but we are displaying like `Netlify Functions`. Is it fine ? I don't have much of a opinion on how to show. But the package claims to use english rules. So, sounds promising, changing it now 👍

packages/cli/src/config.ts
```diff
@@ -51,3 +51,5 @@ export const jsonToGlobalConfig = (json: unknown) => {
 };
 
 export type GlobalConfig = z.infer<typeof zGlobalConfig>;
+
+export const PROJECT_TEMPALTES = ["vercel", "netlify", "netlify-functions"];
```
TrySound, line 55: Were both functions? The distinction was in "edge", no?

.github/workflows/cli-test.yml
```diff
@@ -69,17 +69,17 @@ jobs:
         run: pnpm webstudio-cli build --template netlify-functions --preview && pnpm prettier --write ./app/ ./package.json
 
         # Testing fixtures for netlify-edge-functions template
-      - name: Test cli link command for netlify-edge-functions
-        working-directory: ./fixtures/webstudio-remix-netlify-edge-functions
+      - name: Test cli link command for netlify
```
istarkov, line 72: 1st lets move this into fixture  packages
i.e.
```json
"scripts": {
    "fixture:link": "webstudio-cli link --link 'https://webstudio-builder-git-main-webstudio-is.vercel.app/builder/d845c167-ea07-4875-b08d-83e97c09dcce?authToken=e9d1343f-9298-4fd3-a66e-f89a5af2dd93",
    "fixture:sync": "blabla",
    "fixture:build": "blablablabla",
...
}
    ```

then using pnpm just run from the root 
```bash
pnpm --filter='./fixtures/*' run --parallel fixtures:link
pnpm --filter='./fixtures/*' run --parallel fixtures:sync
pnpm --filter='./fixtures/*' run --parallel fixtures:build
```

to never touch test.yaml again

packages/cli/src/config.ts
JayaKrishnaNamburu: Yes, and i think the decisions was `edge` is cheaper and efficient make it default. We can add the differences between these in docs.

istarkov (commented):
plz do so we would able to resync rebuild relink all fixtures from the root

packages/cli/src/config.ts
istarkov, line 55: Isn't this just a folder names, use them

istarkov (approved):
plz do

packages/cli/src/config.ts
istarkov: Just any time we will add a new template to not search why its not shown

JayaKrishnaNamburu:
`custom-template` fixture is broken. Not it's corrected. See the `projectId` inside the `cli-test.yml` file is not matching with the projectId inside the fixture. Corrected it and updated the fixture too. 
Screenshot 2023-09-25 at 19 48 34
Screenshot 2023-09-25 at 19 48 52

https://github.com/webstudio-is/webstudio/pull/2358/files#diff-69532840426f93d7d6b29deb0ff774e32affeb262dc7c7145e1f395d5c69513a

https://github.com/webstudio-is/webstudio/pull/2358/files#diff-5fcc33269241517d79034d41d1a2f0ce649c35812ea21326ee55a45ee3413f0cL87

packages/cli/src/config.ts
TrySound: I mean netlify-functions does not describe at all. Both netlify options are functions

JayaKrishnaNamburu: Switched back to `netlify-functions` and `netlify-edge-functions` from their platform

TrySound (approved):

PR #2360, opened 2023-09-24
perf: generate pages paths

TrySound:
Here generated a list of pages paths. It is used in link component to switch to remix link internally when the link is app route.

Providing whole pages list can bloat runtime when there are many pages or when there are many seo data so having single page per route should be enough.

The list is always empty in builder because links are managed with global interceptor. Ideally frameworks should uses navigation api and distinct routes from other paths so we would not need to generate anything.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
didn't test

PR #2371, opened 2023-09-25
feat: Add helper functions for Tailwind support: shorthand expansion and variable substitution

istarkov:
## Description

Gonna merge this before next step.

Helper functions to parse tailwind.

1. `expandTailwindShorthand` - Expands shorthand CSS properties into their long-form equivalents. _'mx' is expanded to 'mr' (margin-right) and 'ml' (margin-left)._
2. `substituteVariables` - Resolves CSS variables by replacing their references with actual values. _Transforms `.class { --property: 1px; margin: var(--property); }` into `.class { margin: 1px; }`._

## Steps for reproduction

See tests.

## Code Review

- [ ] hi @kof, I need you to do
  - conceptual review (architecture, feature-correctness)
  - detailed review (read every line)
  - test it on preview

## Before requesting a review

- [ ] made a self-review
- [ ] added inline comments where things may be not obvious (the "why", not "what")

## Before merging

- [ ] tested locally and on preview environment (preview dev login: 5de6)
- [ ] updated [test cases](https://github.com/webstudio-is/webstudio/blob/main/apps/builder/docs/test-cases.md) document
- [ ] added tests
- [ ] if any new env variables are added, added them to `.env.example` and the `builder/env-check.js` if mandatory

PR #2378, opened 2023-09-26
feat: AI Sections

gsppe:
## Description

This PR adds two [AI chains](https://github.com/webstudio-is/webstudio/tree/main/packages/ai/src/chains#readme) and endpoints for the following two tasks:

1. chains/sections: Given a user prompt, break it into section descriptions
2. chains/template-generator: Give a section description generate a Webstudio Embed Template for it. To do so we pass the prompt and a list of available components to the AI and ask to generate a JSX tree using Tailwind CSS for styling. Once we get the result we transform the JSX tree into a Webstudio Embed Template.

- [ ] Extracted the JSX-to-Webstudio utils into a separate PR. #2396 Merge that and remove those from this PR.

Currently we are not converting `className` (tailwind styles) to Webstudio `EmbedStyleDecl` because we need #2371 

Companion rest endpoints for the above are:

1. rest.ai.sections.ts
2. rest.ai.template-generator.ts

### UI flow (potential usage)

A UI is implemented in followup PR (#2379) but the idea is that in the frontend we will:

- Call the Sections chain via rest.ai.sections.ts with the user prompt 
- Display the result as a list of editable sections description with `name` and `description`
- The user presses a generate all or individual generate buttons 
- We call the Template Generator chain via rest.ai.template-generator.ts
- We get back a Webstudio Embed Template which we insert on page

### To review

1. Start by reviewing the chains's index.ts file and looking at their system and user prompts.
2. Review `jsxToWsEmbedTemplate` which converts a JSX string to Webstudio Embed Template

## Code Review

- [ ] hi @kof - review the prompts (markdown files) and the UI flow (potential usage) above
- [ ] somebody else should do a general code review

## Before requesting a review

- [ ] made a self-review

### DIFF
--- a/packages/ai/src/utils/tw-to-ws.ts
+++ b/packages/ai/src/utils/tw-to-ws.ts
@@ -1,6 +1,6 @@
 import type { WsEmbedTemplate } from "@webstudio-is/react-sdk";
 import { traverseTemplateAsync } from "./traverse-template";
-import { parseCss, parseTailwindToCss } from "@webstudio-is/css-data";
+import { parseTailwindToWebstudio } from "@webstudio-is/css-data";
 
 export const tailwindToWebstudio = async (template: WsEmbedTemplate) => {
   return traverseTemplateAsync(template, async (node, parent) => {
@@ -9,8 +9,7 @@ export const tailwindToWebstudio = async (template: WsEmbedTemplate) => {
         (prop) => prop.name === "className"
       );
       if (classNameProp && classNameProp.type === "string") {
-        const css = await parseTailwindToCss(classNameProp.value);
-        const styles = Object.values(parseCss(css)).flat();
+        const styles = await parseTailwindToWebstudio(classNameProp.value);
 
         if (Array.isArray(node.styles)) {
           // Merge with existing styles
--- a/packages/css-data/src/index.ts
+++ b/packages/css-data/src/index.ts
@@ -16,7 +16,10 @@ export * from "./property-parsers/index";
 export * from "./parse-css-value";
 export * from "./parse-css";
 
-export { parseTailwindToCss } from "./tailwind-parser/parse";
+export {
+  parseTailwindToCss,
+  parseTailwindToWebstudio,
+} from "./tailwind-parser/parse";
 
 import { properties as generatedProperties } from "./__generated__/properties";
 


packages/feature-flags/package.json
```diff
@@ -12,7 +12,12 @@
     "build": "rm -rf lib && esbuild 'src/**/*.ts' 'src/**/*.tsx' --outdir=lib",
     "dts": "tsc --project tsconfig.dts.json"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@babel/parser": "^7.23.0",
+    "@babel/types": "^7.23.0",
+    "json5": "^2.2.3",
+    "type-fest": "^4.3.1"
+  },
```
TrySound, line 20: Are you sure this deps supposed to be in this package?

packages/ai/src/utils/tw-to-ws.ts
```diff
@@ -0,0 +1,32 @@
...
+export const tailwindToWebstudio = async (template: WsEmbedTemplate) => {
+  return traverseTemplateAsync(template, async (node, parent) => {
+    if (node.type === "instance" && node.props && node.props.length > 0) {
+      const classNameProp = node.props.find(
+        (prop) => prop.name === "className"
+      );
+      if (classNameProp && classNameProp.type === "string") {
+        const styles = await parseTailwindToWebstudio(classNameProp.value);
+
+        if (Array.isArray(node.styles)) {
```
TrySound, line 14: Why is array check? Isn't it instance node? type is always array or undefined.

packages/ai/src/utils/tw-to-ws.ts
```diff
@@ -0,0 +1,32 @@
...
+          // Merge with existing styles
+          const currentStyles = new Set(
+            node.styles.map(({ property }) => property)
+          );
+          styles.forEach((styleDecl) => {
+            if (currentStyles.has(styleDecl.property) === false) {
+              node.styles!.push(styleDecl);
```
TrySound, line 21: if you use for/of this non null cast would not be necessary

packages/ai/src/utils/tw-to-ws.ts
kof: we usually check for undefined so we know what's being checked explicitly in this case

kof:
Just tried to check out this branch, build fails with this:

```
                                        ^
apps/builder dev: Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/oleg/work/webstudio/webstudio-builder/node_modules/camelcase/index.js from /Users/oleg/work/webstudio/webstudio-builder/apps/builder/build/index.js not supported.
```

PR #2379, opened 2023-09-26
feat: AI Sections UI

gsppe:
## Description

This PR adds a baseline UI for #2378 (depends on this)

This UI implements the following flow

- Call the Sections chain via rest.ai.sections.ts with the user prompt 
- Display the result as a list of editable sections description with `name` and `description`
- The user presses a generate all or individual generate buttons 
- We call the Template Generator chain via rest.ai.template-generator.ts
- We get back a Webstudio Embed Template which we insert on page

The final UI will be implemented once we have some design and UI from @taylornowotny 

## Code Review

- [ ] @TrySound please review this implementation. The trickiest part is in the SectionsEditor where there is logic to generate all the sections in parallel or one by one, with the option to abort.

## Before requesting a review

- [ ] made a self-review

## Before merging


packages/ai/bin/build-icons.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (typeof icons === "object" && type in icons === false) {
+    icons[type] = {};
+  }
+
+  icons[type][name] = icon.replace("<svg", '<svg style="height: 1.1em;"');
+
+  return icons;
+}, {});
+
+const baseDir = path.dirname(iconFiles[0]).split("/").slice(0, -1).join("/");
```
TrySound, line 28: What are these magical paths parsings?

kof: @TrySound suggest a better way doing things right away to save time

PR #2385, opened 2023-09-28
feat: Parse tailwind to ordinary css compatible with webstudio

istarkov:
## Description

- apply tailwind transform, expand and substitute to get compatible css with webstudio
- media and pseudo classes temporarily removed from variable substitution (need specificity calculations for both, we don't support media in templates)

## Steps for reproduction

see tests

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-data/src/tailwind-parser/parse.ts
+++ b/packages/css-data/src/tailwind-parser/parse.ts
@@ -1,0 +1,25 @@
+import { UnoGenerator, createGenerator } from "@unocss/core";
+import presetMini, { type Theme } from "@unocss/preset-uno";
+import { expandTailwindShorthand } from "./shorthand";
+import { substituteVariables } from "./substitute";
+import warnOnce from "warn-once";
+
+let unoLazy: UnoGenerator<Theme> | undefined = undefined;
+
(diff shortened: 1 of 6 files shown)


packages/css-data/src/tailwind-parser/parse.ts
```diff
@@ -0,0 +1,25 @@
+import { UnoGenerator, createGenerator } from "@unocss/core";
+import presetMini, { type Theme } from "@unocss/preset-uno";
+import { expandTailwindShorthand } from "./shorthand";
+import { substituteVariables } from "./substitute";
+import warnOnce from "warn-once";
+
+let unoLazy: UnoGenerator<Theme> | undefined = undefined;
+
+const uno = () => {
+  unoLazy = createGenerator({
```
TrySound, line 10: Why lazy? Is it expensive to init?

istarkov: Have no idea, just to not init on start

TrySound (approved):

PR #2389, opened 2023-09-28
refactor: add commands emitter

TrySound:
Added new commands emitter utility to manage hotkeys across plugins and separate a lot of logic from canvas.

Here migrated only togglePreview as an example.

Later we can build commands panel and hotkeys overrides on top of this utility.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/shared/commands.ts
+++ b/apps/builder/app/canvas/shared/commands.ts
@@ -1,0 +1,5 @@
+import { createCommandsEmitter } from "~/shared/commands-emitter";
+
+export const { emitCommand, subscribeCommands } = createCommandsEmitter({
+  commands: [],
+});
--- a/apps/builder/app/canvas/shared/commands.ts
+++ b/apps/builder/app/canvas/shared/commands.ts
@@ -1,5 +1,6 @@
 import { createCommandsEmitter } from "~/shared/commands-emitter";
 
 export const { emitCommand, subscribeCommands } = createCommandsEmitter({
+  source: "canvas",
   commands: [],
 });
--- a/apps/builder/app/shared/commands-emitter.ts
+++ b/apps/builder/app/shared/commands-emitter.ts
@@ -9,7 +9,7 @@ type CommandMeta<CommandName extends string> = {
   defaultHotkeys?: string[];
 };
 
-type CommandHandler = (source: string) => void;
+type CommandHandler = () => void;
 
 /**
  * Command can be registered by builder, canvas or plugin
@@ -68,8 +68,8 @@ export const createCommandsEmitter = <CommandName extends string>({
    * Execute handlers in app where defined whenever command is emitted
    */
   const subscribeCommands = () => {
-    const unsubscribePubsub = subscribe("command", ({ source, name }) => {
-      commandHandlers.get(name)?.(source);
+    const unsubscribePubsub = subscribe("command", ({ name }) => {
+      commandHandlers.get(name)?.();
     });
     const handleKeyDown = (event: KeyboardEvent) => {
       const commandMetas = $commandMetas.get();
--- a/apps/builder/app/shared/commands-emitter.ts
+++ b/apps/builder/app/shared/commands-emitter.ts
@@ -28,8 +28,10 @@ type Command<CommandName extends string> = CommandMeta<CommandName> & {
 export const $commandMetas = atom(new Map<string, CommandMeta<string>>());
 
 export const createCommandsEmitter = <CommandName extends string>({
+  source,
   commands,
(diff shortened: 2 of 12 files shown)


kof (approved):

apps/builder/app/shared/commands-emitter.ts
```diff
@@ -0,0 +1,104 @@
...
+  const commandMetas = new Map($commandMetas.get());
+  const commandHandlers = new Map<string, CommandHandler>();
+  for (const { handler, ...meta } of commands) {
+    commandMetas.set(meta.name, meta);
+    commandHandlers.set(meta.name, handler);
+  }
+
+  if (commands.length > 0) {
+    onMount($commandMetas, () => {
+      // use the next tick after subscription started
```
istarkov, line 48: Why use next tick? What if I start immediately

apps/builder/app/shared/commands-emitter.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  handler: CommandHandler;
+};
+
+/*
+ * expose command metas to synchronize between builder, canvas and plugins
+ */
+export const $commandMetas = atom(new Map<string, CommandMeta<string>>());
+
+export const createCommandsEmitter = <CommandName extends string>({
```
istarkov, line 30: IMO emitter is something this `on(cmd, handler)`  and `emit` command, here the result type is not even similar

apps/builder/app/canvas/shared/commands.ts
```diff
@@ -0,0 +1,6 @@
+import { createCommandsEmitter } from "~/shared/commands-emitter";
+
+export const { emitCommand, subscribeCommands } = createCommandsEmitter({
```
istarkov, line 3: emitCommand seems like has no typings support, I can emit anything
emitCommand("anything");

apps/builder/app/shared/commands-emitter.ts
TrySound: Seems like the listener in sync-stores is executed after mount callback. So here scheduled to the end of execution so .set would log store as initialized. These all are hacks and need to figure out better way. Maybe with immerhin but without sending requests to server.

apps/builder/app/shared/commands-emitter.ts
TrySound: There is subscribe and emit. Actual handlers are moved into config. Idk dispatcher will work for you?

apps/builder/app/shared/commands-emitter.ts
```diff
@@ -0,0 +1,104 @@
...
+      type: "command",
+      payload: {
+        source,
+        name,
+      },
+    });
+  };
+
+  /**
+   * Execute handlers in app where defined whenever command is emitted
```
istarkov, line 68: I would add more words, what is app here having that canvas would be standalone app.
Would I receive subscribe event on canvas If I would emitCommand from the canvas. etc

apps/builder/app/canvas/shared/commands.ts
TrySound: I can't
image

apps/builder/app/shared/commands-emitter.ts
istarkov: Can you add this with example into comment, so everyone could get a problem. As the whole code looks strange

```js
// Read store
const commandMetas = new Map($commandMetas.get());

// on first subscriber read again?
onMount($commandMetas, () => {
    // use the next tick after subscription started
    Promise.resolve().then(() => {
      // @todo use patches to avoid race when both builder and canvas send commands
      $commandMetas.set(commandMetas);
    });
  });
```

istarkov: So onMount is needed for the 1st subscriber. And init for consequent.

apps/builder/app/shared/commands-emitter.ts
TrySound: I don't like words. Words suck 👴

istarkov (approved):

apps/builder/app/canvas/shared/commands.ts
istarkov: image

apps/builder/app/shared/commands-emitter.ts
TrySound: Yes, this is why we need to switch to patching instead of sending whole state. But in another PR. For now canvas just doesn't send empty commands

PR #2391, opened 2023-09-29
feat: AI Operations  

gsppe:
Implements an operation-based system to accomplish a number of edit tasks on a JSX+CSS representation of a selected Webstudio tree.

Every JSX element has the instance id as data-ws-id and therefore the LLM knows what to edit.

The idea is to create a number of operations (aiOperations) specs/schemas with Zod describing unique edit tasks, make a discriminated union, convert them to JSON Schema. We then inject this information and the current JSX+CSS in the LLM prompt asking the model to generate an array of operations.

The LLM responds with an array of operations (aiOperations) to fulfill the user request. We then process these doing the necessary transformations to convert these to a Webstudio-friendly format (wsOperations) and send them to the client where we apply them (see applyOperations).

Some of these operations are going to be "intermediate" operations for now. 

Generating quality user interfaces might be challenging with LLMs.

Having to do so within a generic operations framework like the one described above makes it even more challenging.

For this reason when it comes to generating user interfaces, the Operation chain instead offers an operation called generateTemplatePrompt that provides information about insertion point in the existing UI and a prompt, possibly enhanced.

Another chain can use these information and a more sophisticated prompt and model to then generate the user interface and return a generateInsertTemplate wsOperation. Webstudio does exactly this in a rest endpoint, using the template-generator chain with gpt4 instead of gpt-3.5-turbo.

### How to review 

- take a look at the operations README.md for an explanation of the architecture
- take a look at the rest.ai.op.ts endpoint to see how the operations chain is used
- review everything else

### DIFF
--- a/packages/ai/src/chains/operations/generate-insert-template.ts
+++ b/packages/ai/src/chains/operations/generate-insert-template.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 import { WsEmbedTemplate } from "@webstudio-is/react-sdk";
-import { jsxToTemplate } from "../../../utils/jsx-to-template";
+import { jsxToTemplate } from "../../utils/jsx-to-template";
 
 // Currently this operation is used is a separate LLM call after the main one has returned an insert-instance operation. This is to produce better results.
 // Effectively the aiOperation from this module is not used. The separate chain will call the LLM and we use the resulting completion to construct an aiOperation by hand
--- a/packages/ai/src/utils/request.ts
+++ b/packages/ai/src/utils/request.ts
@@ -1,10 +1,5 @@
 import { createChunkDecoder } from "ai";
-import type {
-  Chain,
-  ErrorResponse,
-  LlmResponse,
-  SuccessResponse,
-} from "../types";
+import type { LlmResponse } from "../types";
 import { createErrorResponse } from "./create-error-response";
 import { StreamingTextResponse } from "./streaming-text-response";
 


apps/builder/app/builder/features/ai/commands-operations.ts
```diff
@@ -0,0 +1,131 @@
...
+            if (candidateStyleSource && candidateStyleSource.type === "local") {
+              styleSource = candidateStyleSource;
+              styleSourceId = candidateStyleSource.id;
+              break;
+            }
+          }
+        }
+
+        if (styleSourceId === "") {
+          styleSourceId = uuid();
```
TrySound, line 106: We use nanoid for data id

apps/builder/app/builder/features/ai/commands-operations.ts
```diff
@@ -0,0 +1,131 @@
...
+        if (styleSourceSelection === undefined) {
+          styleSourceSelections.set(instanceId, {
+            instanceId,
+            values: [styleSourceId],
+          });
+        }
+
+        for (const embedStyleDecl of newStyles) {
+          const id = uuid();
+          styles.set(id, {
```
TrySound, line 122: We use `getStyleDeclKey(styleDecl)` to as style id

packages/ai/src/utils/jsx-to-template.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const postprocessTemplate = (
+  template: WsEmbedTemplate,
+  components: string[]
+) => {
+  traverseTemplate(template, (node) => {
+    if (node.type === "instance") {
+      if (node.component.startsWith("Radix.")) {
+        node.component =
+          "@webstudio-is/sdk-components-react-radix:" + node.component.slice(6);
```
TrySound, line 24: .slice('Radix.'.length)

apps/builder/app/shared/copy-paste/plugin-instance.ts
```diff
@@ -71,7 +71,7 @@ const InstanceData = z.object({
 
 type InstanceData = z.infer<typeof InstanceData>;
 
-const findTreeStyleSourceIds = (
+export const findTreeStyleSourceIds = (
```
TrySound, line 74: Export not needed

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,219 @@
...
+          const scope = createScope();
+
+          const jsx = generateJsxElement({
+            scope: createScope(),
+            instance,
+            props,
+            dataSources,
+            indexesWithinAncestors,
+            children: generateJsxChildren({
+              scope,
```
TrySound, line 174: Scope should be reused between root element and children

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+        const formData = new FormData(event.currentTarget);
+
+        const prompt = formData.get("prompt");
+        let data = formData.get("data");
+
+        if (typeof data === "string") {
+          try {
+            data = JSON.parse(data);
+          } catch (error) {
```
TrySound, line 63: optional catch binding

apps/builder/app/builder/features/ai/apply-operations.ts
```diff
@@ -0,0 +1,144 @@
+import store from "immerhin";
```
TrySound, line 1: Please rebase. I just replaced default store with serverSyncStore from ~/shared/sync

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,219 @@
...
+      >
+        {isLoading ? "stop" : "go"}
+      </Button>{" "}
+      {isLoading ? <SpinnerIcon /> : null}
+      {error ? <Text>{error}</Text> : null}
+    </form>
+  );
+};
+
+const useInstanceToJsx = (rootInstanceId: Instance["id"]) =>
```
TrySound, line 134: Github lost my comment.

This whole thing can be a top level computed with selectedInstanceSelector as input instead of passing it with prop.

packages/ai/src/chains/operations/index.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    const systemMessage: ModelMessage = [
+      "system",
+      formatPrompt(
+        {
+          operationsSchema: JSON.stringify(operationsSchema),
+          components: components
+            .map((name) =>
+              name.replace(
+                "@webstudio-is/sdk-components-react-radix:",
```
kof, line 59: Uhh this belongs somwere else, radix is just a package, like any user-installable one, it shouldn't know directly about it

packages/ai/src/chains/operations/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      wsOperations = await Promise.all(aiToWs(aiOperations));
+
+      // Clean up template and ensure valid components in templates
+      wsOperations.forEach((operation) => {
+        for (let index = 0; index < wsOperations.length; index++) {
+          if (operation.operation !== "insertTemplate") {
+            return;
+          }
+
+          postprocessTemplate(operation.template, components);
```
kof, line 130: postProcessTemplate?

packages/ai/src/chains/operations/index.ts
```diff
@@ -0,0 +1,151 @@
...
+        }
+      });
+    } catch (error) {
+      return {
+        success: false,
+        type: "parseError",
+        status: 500,
+        message: `Failed to convert operations. ${
+          process.env.NODE_ENV === "development" ? (error as Error) : ""
```
kof, line 139: do you want to use an empty error in prod? are you sure its better than showing the actual error or showing something meaningful? at least so user can report.

Also we should log the error even if its not being sent, right?

packages/ai/src/chains/operations/operations/edit-styles.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from "zod";
+import { EmbedTemplateStyleDecl } from "@webstudio-is/react-sdk";
+import { parseTailwindToWebstudio } from "@webstudio-is/css-data";
+
+const wsId = z
+  .string()
+  .describe("The element's ws-id that must get the new styles");
+
+export const aiOperation = z.object({
+  operation: z.literal("editStylesWithTailwindCSS"),
```
kof, line 10: nit: editWithTailwind seems clear and shorter

kof: oh I see there is a consistency with `generateInstanceWithTailwindStyles`

in this case it could be editStylesWithTailwind, generateInstanceWithTailwind

packages/ai/src/chains/template-generator/README.md
```diff
@@ -0,0 +1,114 @@
...
+  type ErrorResponse,
+} from "@webstudio-is/ai";
+
+function UiComponent() {
+  const [error, setError] = useState(null);
+  const [template, setTemplate] = useState({});
+
+  useEffect(() => {
+    console.log(template);
+  }, [template]);
```
kof, line 64: nit: this is generally a bad example for using effect, better to log it where it was received in the handler

packages/ai/src/chains/template-generator/README.md
```diff
@@ -0,0 +1,114 @@
...
+        <textarea name="prompt" />
+      </label>
+
+      <input
+        type="hidden"
+        name="components"
+        defaultValue='["Box","Heading","Image","Text"]'
+      />
+
+      <button disabled={isLoading}>Generate Section</button>
```
kof, line 110: nit: isLoading is always a bad small, even though its just example
better: status === 'loading'

apps/builder/app/builder/features/ai/apply-operations.ts
```diff
@@ -0,0 +1,143 @@
...
+  const dropTarget: DroppableTarget = {
+    parentSelector: [operation.addTo],
+    position: operation.addAtIndex + 1,
+  };
+
+  insertTemplateData(templateData, dropTarget);
+  return rootInstanceIds;
+};
+
+const deleteInstance = (operation: operations.deleteInstance.wsOperation) => {
```
kof, line 81: would be better to have a more precise name than to have the same name but different interfaces and prefixing them, e.g. this seems to delete instance by operation, could be deleteInstanceByOp

apps/builder/app/builder/features/ai/apply-operations.ts
```diff
@@ -0,0 +1,143 @@
...
+const applyStyles = (operation: operations.editStyles.wsOperation) => {
+  serverSyncStore.createTransaction(
+    [
+      instancesStore,
+      styleSourceSelectionsStore,
+      styleSourcesStore,
+      stylesStore,
+      breakpointsStore,
+    ],
+    (instances, styleSourceSelections, styleSources, styles, breakpoints) => {
```
kof, line 94: this function contains so much logic that it needs a few unit tests to make sure its not gonna break

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,203 @@
...
+  styleSourceSelectionsStore,
+  stylesStore,
+} from "~/shared/nano-states";
+import { applyOperations } from "./apply-operations";
+import { useStore } from "@nanostores/react";
+import { restAiOperations } from "~/shared/router-utils";
+
+export const Operations = () => {
+  const abort = useRef<AbortController>();
+  const [isLoading, setIsLoading] = useState(false);
```
kof, line 32: status: "loading" | "error" | "idle"

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,203 @@
...
+    styles,
+    styleSourceSelections
+  ) => {
+    if (selectedInstanceSelector === undefined) {
+      return null;
+    }
+
+    const [rootInstanceId] = selectedInstanceSelector;
+    const instance = instances.get(rootInstanceId);
+    if (instance == null) {
```
kof, line 154: this makes it look like instance can be null or undefined, I haven't looked, but I think it can only be one of these, we are trying to be more explicit everywhere

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const styleRules = getStyleRules(styles, treeStyleSourceSelections);
+    for (const { breakpointId, instanceId, state, style } of styleRules) {
+      engine.addStyleRule(`[${idAttribute}="${instanceId}"]${state ?? ""}`, {
+        breakpoint: breakpointId,
+        style,
+      });
+    }
+
+    return `<style>{\`${engine.cssText.replace(/\n/gm, " ")}\`}</style>${jsx
+      .replace(/data-ws-component="[^"]+"/g, "")
```
kof, line 200: we do have a constant for this attribute name, in case we want to change it in the future, search for "componentAttribute", maybe we need a better place than react-sdk for those attributes?

@TrySound

apps/builder/app/routes/rest.ai.op.ts
```diff
@@ -0,0 +1,145 @@
...
+import { isFeatureEnabled } from "@webstudio-is/feature-flags";
+
+import env from "~/env/env.server";
+import { createContext } from "~/shared/context.server";
+
+const RequestSchema = operations.ContextSchema;
+
+export const maxDuration = 120;
+
+export const action = async function action({ request }: ActionArgs) {
```
kof, line 17: no need to use named functions, const is enough to have a fn.name in stack traces

apps/builder/app/shared/router-utils/path-utils.ts
```diff
@@ -164,3 +164,4 @@ export const getPublishedUrl = (domain: string) => {
 };
 
 export const restAiCopy = () => `/rest/ai/copy`;
+export const restAiOperations = () => `/rest/ai/op`;
```
kof, line 167: could reuse one function:

`export const restAi = (action: 'copy' | 'op') => `/rest/ai/${action}`;`

apps/builder/remix.config.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
     "node-fetch",
     "immerhin",
     "strip-indent",
+    "camelcase",
```
kof, line 39: strange, we used this package before in many cases, but never added to this config, just to check if needed

packages/ai/src/chains/operations/README.md
```diff
@@ -0,0 +1,140 @@
...
+    }
+  });
+
+  if (response.success === false) {
+    return response;
+  }
+
+  const promptOperations = getGenerateTemplatePromptsWsOperations(response.data);
+
+  if (promptOperations.lengt > 0) {
```
kof, line 86:
```suggestion
  if (promptOperations.length > 0) {
```

packages/jsx-utils/src/heroicons/__generated__/heroicons.ts
```diff
@@ -2,1122 +2,1122 @@
 export const heroicons = {
```
kof, line 2: why do we need an entire icons library here? in particular as part of jsx-utils

packages/ai/src/chains/template-generator/index.ts
```diff
@@ -0,0 +1,108 @@
+import { z } from "zod";
```
TrySound, line 1: I would not put so much code into index.ts. It's better use as reexport or not use at all because very hard to search navigate.

packages/ai/src/chains/operations/operations/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./all";
```
TrySound, line 1: Why do we have operations/operations? Is "operations" a chain or can be moved upper level? I personally prefer to use flat structure to be an able to see everything with ls -la. Or at least as flat as possible.

packages/ai/src/chains/operations/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    try {
+      wsOperations = await Promise.all(aiToWs(aiOperations));
+
+      // Clean up template and ensure valid components in templates
+      for (const wsOperation of wsOperations) {
+        if (wsOperation.operation === "insertTemplate") {
+          postProcessTemplate(wsOperation.template, components);
+        }
+      }
+      // wsOperations.forEach((operation) => {
```
TrySound, line 118: No longer necessary or will be used in other PR?

packages/ai/src/chains/operations/README.md
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+Client side:
+
+```tsx
+import { operations, request, type ErrorResponse } from "@webstudio-is/ai";
+import { applyOperations } from "./apply-operations";
+
+request<operations.Response>(
```
TrySound, line 123: This name gives me shivers. Removed "request" package so many times. 😄

apps/builder/remix.config.js
TrySound: Always was in CLI. Now comes from tailwind parser in runtime.

apps/builder/app/builder/features/ai/operations.tsx
```diff
@@ -0,0 +1,205 @@
...
+    instances,
+    props,
+    dataSources,
+    metas,
+    breakpoints,
+    styles,
+    styleSourceSelections
+  ) => {
+    if (selectedInstanceSelector === undefined) {
+      return null;
```
TrySound, line 151: Better undefined or empty string

apps/builder/app/builder/features/ai/apply-operations.ts
```diff
@@ -0,0 +1,151 @@
...
+      const newStyles = [...operation.styles.values()];
+
+      const breakpointValues = Array.from(breakpoints.values());
+      const baseBreakpoint =
+        breakpointValues.find(isBaseBreakpoint) || breakpointValues[0];
```
TrySound, line 107: ??

packages/ai/src/chains/template-generator/index.ts
TrySound: Or just chain.ts

apps/builder/app/builder/features/ai/apply-operations.ts
kof: yes, but they read less nicely, we don't do prefixes generally, instead we use semantic naming to denote the difference

apps/builder/app/builder/features/ai/operations.tsx
kof: we re generally avoiding using null us much as possible, basically using it in cases where there is no other way only, otherwise undefined, in this case you can just 
```suggestion
      return;
```

apps/builder/app/builder/features/ai/apply-operations.ts
kof: const baseBreakpoint = breakpointValues.find(isBaseBreakpoint);

packages/ai/src/chains/operations/README.md
kof: call it `fetchSomething` where something is specific to what we do there

packages/ai/src/chains/template-generator/index.ts
kof: we use index only for exporting, nothing else, no logic whatsoever

kof: it removes the problem of having many index files open

packages/ai/src/chains/operations/README.md
kof: what is it specifically you are fetching?

packages/ai/src/chains/operations/operations/index.ts
TrySound: I don't like here operations/operations/ specifically, don't know how to fix though. Maybe flatten into operations/

packages/ai/src/chains/operations/README.md
kof: this sseems to be tightly coupled with openai, its not just a generic fetch for any models we might have, right? so maybe fetchOpenAI

kof: yeah, maybe better to split into a few clear functions

TrySound (approved):

PR #2396, opened 2023-10-01
feat: JSX utils

gsppe:
## Description

Adds utils to work with JSX and transform it to Webstudio Embed Template as well as converting Tailwind styles to Webstudio Embed Template style declarations and a fake `<HeroIcon>` component to `HtmlEmbed` with inline SVG taken from heroicons (we will need this for ai).

I split the PR in commits to it is easier to review, so you may want to review this PR per-commit. I need to merge it as soon as possible because I need these utils to build the AI edits feature.

Somebody later on should add tests.

### DIFF
--- a/packages/jsx-utils/src/heroicons/index.ts
+++ b/packages/jsx-utils/src/heroicons/index.ts
@@ -1,5 +1,5 @@
 import type { WsEmbedTemplate } from "@webstudio-is/react-sdk";
-import { traverseTemplate } from "..";
+import { traverseTemplate } from "../traverse-template";
 import { getIcon } from "./get-icon";
 
 export const heroiconsToSvgEmbed = (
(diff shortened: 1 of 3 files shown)


packages/jsx-utils/src/heroicons/index.ts
```diff
@@ -0,0 +1,52 @@
+import type { WsEmbedTemplate } from "@webstudio-is/react-sdk";
+import { traverseTemplate } from "..";
```
TrySound, line 2: `../traverse-template` to avoid circular modules

packages/jsx-utils/src/jsx.ts
```diff
@@ -0,0 +1,139 @@
...
+const parseProps = function parseProps(props: JsonProp[]) {
+  const p: EmbedTemplateProp[] = [];
+  props.forEach(({ name, value }) => {
+    if (typeof value === "string") {
+      p.push({ type: "string", name, value });
+    }
+    if (typeof value === "number") {
+      p.push({ type: "number", name, value });
+    }
+    // @todo add support for other prop types.
```
TrySound, line 136: boolean and string[] are missing from base types

TrySound (approved):

PR #2400, opened 2023-10-01
fix: use proper heading props for dialog and sheet components

### LINKED ISSUE #2206: Add Tag property to Headings in components
In the Dialog and Sheet there is a Heading component used for the title that does not have the Tag property that you would expect to see on a Heading component. This is an h2, but the user has no way of seeing that it's an h2 and without the Tag property they have no way of changing the heading level to fit their document.

If user tries to add the tag property from the props list, it is not listed.

- [ ] Add the Tag property to the initial props for the Dialog and Sheet titles. Or just replace the title with our Heading component which already has the Tag property.

ditorodev:
## Description

Fixes #2206 
Copies what the `sdk-components-react` package does for the Headings so users can control whats being rendered

## Steps for reproduction

1. Click the + button
2. Add a Radix Sheet or Radix Dialog component
3. Select the root on the layers
4. On the right side panel Toggle the `Open` to true
5. Select the Sheet/Dialog Title
6. On the right side panel go to `Settings`
7. Should be able to see a Tag input with a predefined value of `h1`

## Code Review

- [ ] @kof @TrySound Need you guys to check that this is how you imagined it working and that the current implementation follows the expected codebase pattern

## Before requesting a review
- [x] made a self-review

## Before merging
- [x] tested locally
- [ ] tested on preview environment (preview dev login: 5de6) -- First needs access to be able to deploy on vercel my preview env cc @kof

### DIFF
--- a/packages/sdk-components-react-radix/package.json
+++ b/packages/sdk-components-react-radix/package.json
@@ -41,7 +41,7 @@
   "scripts": {
     "dev": "rm -rf lib && esbuild 'src/**/*.ts' 'src/**/*.tsx' --outdir=lib --watch",
     "build": "rm -rf lib && esbuild src/components.ts src/metas.ts src/props.ts src/hooks.ts --outdir=lib --bundle --format=esm --packages=external",
-    "build:args": "NODE_OPTIONS=--conditions=webstudio generate-arg-types './src/accordion.tsx !./src/*.stories.tsx !./src/*.ws.tsx' -e asChild -e modal -e defaultValue -e defaultOpen -e defaultChecked && prettier --write \"**/*.props.ts\"",
+    "build:args": "NODE_OPTIONS=--conditions=webstudio generate-arg-types './src/(accordion|sheet|dialog).tsx !./src/*.stories.tsx !./src/*.ws.tsx' -e asChild -e modal -e defaultValue -e defaultOpen -e defaultChecked && prettier --write \"**/*.props.ts\"",
     "build:tailwind": "tsx scripts/generate-tailwind-theme.ts && prettier --write src/theme/__generated__",
     "build:stories": "webstudio-sdk generate-stories && prettier --write \"src/__generated__/*.stories.tsx\"",
     "dts": "tsc --project tsconfig.dts.json",
--- a/packages/sdk-components-react-radix/package.json
+++ b/packages/sdk-components-react-radix/package.json
@@ -41,7 +41,7 @@
   "scripts": {
     "dev": "rm -rf lib && esbuild 'src/**/*.ts' 'src/**/*.tsx' --outdir=lib --watch",
     "build": "rm -rf lib && esbuild src/components.ts src/metas.ts src/props.ts src/hooks.ts --outdir=lib --bundle --format=esm --packages=external",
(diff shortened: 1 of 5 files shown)


ditorodev:
@kof @TrySound Not sure what kind of tests should be added here? Maybe some snapshots fro storybook? Are there snapshot testing for the props?

ditorodev:
cc @kof @TrySound 

I think this is ready, missing some checkboxes regarding preview envs testing and just checking with you guys id codewise PR looks good! 

Im not adding tests nor new test cases as I dont think this applies. One thing that would be great to add in the future is Snapshot testing for the `__generated__` folders. I really like how it work in https://github.com/biomejs/biome

packages/sdk-components-react-radix/src/dialog.tsx
```diff
@@ -50,7 +51,18 @@ export const DialogOverlay = forwardRef<
 
 export const DialogContent = DialogPrimitive.Content;
 export const DialogClose = DialogPrimitive.Close;
-export const DialogTitle = DialogPrimitive.Title;
+
+type Tag = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+const defaultTag = "h1";
+export const DialogTitle = forwardRef<
+  HTMLHeadingElement,
+  React.ComponentProps<typeof DialogPrimitive.DialogTitle> & { tag?: Tag }
```
istarkov, line 59: `React.ComponentProps` => `ComponentProps`

istarkov (changes requested):
Thank you. 
2 small consistency issues.

packages/sdk-components-react-radix/src/dialog.tsx
ditorodev: Done @istarkov

istarkov (approved):
plz fix 'createElement' is declared but its value is never read.

ditorodev:
> plz fix 'createElement' is declared but its value is never read.
ups! done!

packages/sdk-components-react-radix/package.json
```diff
@@ -41,7 +41,7 @@
   "scripts": {
     "dev": "rm -rf lib && esbuild 'src/**/*.ts' 'src/**/*.tsx' --outdir=lib --watch",
     "build": "rm -rf lib && esbuild src/components.ts src/metas.ts src/props.ts src/hooks.ts --outdir=lib --bundle --format=esm --packages=external",
-    "build:args": "NODE_OPTIONS=--conditions=webstudio generate-arg-types './src/accordion.tsx !./src/*.stories.tsx !./src/*.ws.tsx' -e asChild -e modal -e defaultValue -e defaultOpen -e defaultChecked && prettier --write \"**/*.props.ts\"",
+    "build:args": "NODE_OPTIONS=--conditions=webstudio generate-arg-types './src/(accordion|sheet|dialog).tsx !./src/*.stories.tsx !./src/*.ws.tsx' -e asChild -e modal -e defaultValue -e defaultOpen -e defaultChecked && prettier --write \"**/*.props.ts\"",
```
kof, line 44: why is there special condition now for specific components? doesn't seem right

ditorodev: @kof it was already there, it just built for accordion before, I just added extras for sheet and dialog

TrySound (approved):

PR #2408, opened 2023-10-03
fix: fix undo/redo

TrySound:
Commands meta store introduced in latest changes polutted immerhin history. Here created separate store for client only synchronization.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/commands-emitter.ts
+++ b/apps/builder/app/shared/commands-emitter.ts
@@ -1,7 +1,7 @@
 import { isHotkeyPressed } from "react-hotkeys-hook";
 import { atom, onMount } from "nanostores";
-import store from "immerhin";
 import { $publisher, subscribe } from "~/shared/pubsub";
+import { clientImmerhinStore } from "~/shared/sync";
 
 type CommandMeta<CommandName extends string> = {
   // @todo category, description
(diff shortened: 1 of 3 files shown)


apps/builder/app/shared/commands-emitter.ts
```diff
@@ -61,11 +61,14 @@ export const createCommandsEmitter = <CommandName extends string>({
       // schedule store.set to the next tick
       // so store.listen is executed after store.set below
       Promise.resolve().then(() => {
-        store.createTransaction([$commandMetas], (commandMetas) => {
-          for (const { handler, ...meta } of commands) {
-            commandMetas.set(meta.name, meta);
+        clientImmerhinStore.createTransaction(
```
kof, line 64: could be just clientOnlyStore?

kof: or clientSyncStore

kof: Or: apiSyncstore, canvasSyncStore, pluginSyncStore

kof: Wait a second, isn't it that you only want to avoid undo/redo getting synced?

kof: Actually now as I think about it I don't understand why you use immerhin for shorthands at all, these things are ephemeral, real persistent things are going to happen in response to shorthands

istarkov (approved):

apps/builder/app/shared/sync/sync-stores.ts
```diff
@@ -53,11 +53,13 @@ declare module "~/shared/pubsub" {
     sendStoreChanges: {
       // distinct source to avoid infinite loop
       source: SyncEventSource;
+      storeType: "client" | "server";
```
kof, line 56: alternative names for storeType: "destination", "location", "origin"

kof:
I tested - works now

apps/builder/app/shared/sync/sync-stores.ts
TrySound: Will get rid of it soon anyway.

PR #2415, opened 2023-10-04
feat: Switch on viewport-based srcSets for width-defined Images

istarkov:
## Description

Previous PR: #2414

Now, by default, we set the width attribute when adding an image. This change could cause mobile screens to load oversized images—specifically, width * 2. To mitigate this, we're adopting the viewport width descriptor for srcSets. This approach ensures optimized image loading on smaller screens and maintains previous behavior on screens wider than the image's actual size.

## Steps for reproduction

Add images of various sizes, see srcSet

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/image/src/image-optimize.ts
+++ b/packages/image/src/image-optimize.ts
@@ -21,7 +21,7 @@
  *   source size value is 70vw  equal to 800px * 0,7 = 560px
  *
  *   browser internal srcset will be (we divide `w` descriptor by source size value):
- *   photo-small.jpg 320/560x, photo-medium.jpg 640/560x, photo-huge.jpg 1280/560x =>
+ *   photo-small.jpg 320w/560px, photo-medium.jpg 640w/560px, photo-huge.jpg 1280w/560px =>
  *   photo-small.jpg 0.57x, photo-medium.jpg 1.14x, photo-huge.jpg 2.28x
  *
  *   Finally same rules as for pixel density descriptor 'x' are applied.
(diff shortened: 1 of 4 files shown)


kof (approved):

packages/image/src/image-optimize.ts
```diff
@@ -131,6 +130,19 @@ const getWidths = (
     return { widths: deviceSizes, kind: "w" };
   }
 
+  const MAX_DEVICE_PIXEL_RATIO = 2;
+
+  let index = allSizes.findIndex((p) => p >= MAX_DEVICE_PIXEL_RATIO * width);
```
kof, line 135:
```suggestion
  let index = allSizes.findIndex((size) => size >= MAX_DEVICE_PIXEL_RATIO * width);
```

packages/image/src/image-optimize.ts
```diff
@@ -131,6 +130,19 @@ const getWidths = (
     return { widths: deviceSizes, kind: "w" };
   }
 
+  const MAX_DEVICE_PIXEL_RATIO = 2;
```
kof, line 133: is this running serverside? in browser there is window.devicePixelRatio

istarkov: Yep serverside so we use max. This is for srcSet generation so we don't need real pixel-ratio, just max - which is 2 (in real can be 3,4 etc but for images there is not need > 2)

kof: good info, added as a comment

istarkov: done

PR #2417, opened 2023-10-04
refactor: Rework LLM client types and implementations

gsppe:
Refactor the LLM client, responses shape and web client to support streaming and regular json request in a single helper.

This is needed so that I can later unify the streaming copywriter endpoint with the operations endpoint which returns JSON (no streaming)

## Before requesting a review

- [ ] made a self-review

## Before merging

### DIFF
--- a/apps/builder/app/builder/features/ai/copywriter.tsx
+++ b/apps/builder/app/builder/features/ai/copywriter.tsx
@@ -1,6 +1,10 @@
 import { z } from "zod";
 import untruncateJson from "untruncate-json";
-import { copywriter, request } from "@webstudio-is/ai";
+import {
+  copywriter,
+  request,
+  type StreamingTextResponseType,
+} from "@webstudio-is/ai";
 import { restAiCopy } from "~/shared/router-utils";
 import {
   Box,
@@ -102,7 +106,7 @@ export const Copywriter = () => {
 
               abort.current = new AbortController();
               setIsLoading(true);
-              request(
+              request<StreamingTextResponseType>(
                 [
                   restAiCopy(),
                   {
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -1,8 +1,11 @@
-import type { StreamingTextResponse } from "ai";
+import type { StreamingTextResponseType } from "./utils/streaming-text-response";
 
 /**
  * Generic Response types used both by Models and Chains.
  */
+
+// @todo Convert the response types to Zod
+// so that responses can be parsed and validated on the client.
 
 export type Tokens = {
   prompt: number;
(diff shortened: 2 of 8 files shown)


apps/builder/app/builder/features/ai/copywriter.tsx
```diff
@@ -134,12 +116,30 @@ export const Copywriter = () => {
...
+                    }
+                  },
                 }
               ).then((result) => {
                 abort.current = null;
-                if (typeof result !== "string") {
-                  alert("Error " + result.type);
+                if (result.success === false) {
+                  // eslint-disable-next-line no-console
+                  console.log(result);
```
TrySound, line 141: Forgotten? If not better use info/warn/error

TrySound (approved):

apps/builder/app/builder/features/ai/copywriter.tsx
kof: yep, please don't disable that lint rule, we should use log only for debugging so we don't forget to remove

packages/ai/src/types.ts
```diff
@@ -1,3 +1,48 @@
...
+  success: true;
+  tokens: Tokens;
+  data: ResponseData;
+};
+
+export type StreamingSuccessResponse = {
+  type: "stream";
+  success: true;
+  stream: StreamingTextResponse;
+};
```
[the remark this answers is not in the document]
TrySound: 🤷

TrySound (approved):

packages/ai/src/utils/streaming-text-response.ts
```diff
@@ -0,0 +1,37 @@
...
+    super(res, init);
+    this.getRequestHeaders();
+  }
+
+  getRequestHeaders() {
+    return addRawHeaders(this.headers);
+  }
+}
+
+const addRawHeaders = function addRawHeaders(headers: Headers) {
```
kof, line 19: we normally don't use named functions/functions, just arrow functions

packages/ai/src/utils/streaming-text-response.ts
```diff
@@ -0,0 +1,37 @@
...
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  headers.raw = function () {
```
kof, line 22: same

packages/ai/src/utils/streaming-text-response.ts
```diff
@@ -0,0 +1,37 @@
+import { StreamingTextResponse as _StreamingTextResponse } from "ai";
+
+export type StreamingTextResponseType = StreamingTextResponse;
+
+// vercel/ai's StreamingTextResponse does not include request.headers.raw()
+// which @vercel/remix uses when deployed on vercel.
+// Therefore we use a custom one.
+export class StreamingTextResponse extends _StreamingTextResponse {
```
kof, line 8: same thing, using underscore doesn't explain why we have classes with the same name, it just raises questions, I would rather have them more explicit names that explain why they exist

kof: yes, but reflecting it in the name is important still, e.g.
```ts
export class StreamingTextResponse extends RemixStreamingTextResponse
```

kof: even better now its even clear what wasn't before, that you are creating a remix interface that extends the vercel interface ... I completely didn't know that before, even with comment

PR #2420, opened 2023-10-05
fix: fix various text editor bugs

TrySound:
- add missing placeholder to Text component
- fix wrapping buttons and links with span
- preserve node content when entering editing mode if instance has no children
- hide "link" button in text toolbar when edit link component to prevent nested links

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Unrelated, but I just realized we allow inserting Paragraph into Text ... especially if Text can become span and p would be inside ... not sure its legal

kof (approved):

PR #2444, opened 2023-10-10
refactor: Split AI commands detection and execution

gsppe:
As per title the PR splits command detection and execution in two endpoints.

We will first make a request to the /rest/ai/detect endpoint which will return an array with one or multiple detected commands for the prompt. Once we have that we execute them in parallel. 

I rewrote the temporary UI completion handler to do all the above.  Error handling will still have to be polished to account for special cases like rate limiting etc.

## Code Review


apps/builder/app/routes/rest.ai.detect.ts
```diff
@@ -0,0 +1,115 @@
...
+import { isFeatureEnabled } from "@webstudio-is/feature-flags";
+
+import env from "~/env/env.server";
+import { createContext } from "~/shared/context.server";
+
+export const RequestParamsSchema = z.object({
+  prompt: z.string().max(1200),
+});
+
+export const maxDuration = 20;
```
kof, line 20: in what metric? seconds? worth a comment

packages/ai/src/chains/operations/chain.ts
```diff
@@ -21,6 +21,8 @@ import { createErrorResponse } from "../../utils/create-error-response";
  * it generates a series of edit operations to fulfill an edit request coming from the user.
  */
 
+export const name = "operations";
```
kof, line 24: isthis like a chain name? maybe chainType?

packages/ai/src/chains/operations/chain.ts
```diff
@@ -72,7 +72,7 @@ export const createChain = <ModelMessageFormat>(): Chain<
     const messages = model.generateMessages(llmMessages);
 
     const completion = await model.completion({
-      id,
+      id: name,
```
kof, line 75: mapping id to a non-unique name is a bit weird, would call it what it is

packages/ai/src/chains/operations/edit-styles.ts
```diff
@@ -2,9 +2,11 @@ import { z } from "zod";
 import { EmbedTemplateStyleDecl } from "@webstudio-is/react-sdk";
 import { parseTailwindToWebstudio } from "@webstudio-is/css-data";
 
+export const name = "edit-styles";
+
 const wsId = z
   .string()
-  .describe("The element's ws-id that must get the new styles");
+  .describe("The data-ws-id value of the element being edited.");
```
kof, line 9: could use the const for data-ws-id

istarkov (approved):

kof (approved):

PR #2454, opened 2023-10-11
refactor: AI Commands Bar

gsppe:
## Description

Polished and simplified a bit the AI Commands Bar UI, adding aborting and a few other missing things.

We are still missing proper styling for the `toast` components. I bet Ivan was going to style this in a followup PR.

### DIFF
--- a/apps/builder/app/builder/features/ai/ai-command-bar.tsx
+++ b/apps/builder/app/builder/features/ai/ai-command-bar.tsx
@@ -26,12 +26,7 @@ import {
   StopIcon,
   LargeXIcon,
 } from "@webstudio-is/icons";
-import {
-  useRef,
-  useState,
-  type MouseEvent,
-  type ComponentPropsWithoutRef,
-} from "react";
+import { useRef, useState } from "react";
 import { $isAiCommandBarVisible } from "~/shared/nano-states";
 import { useMediaRecorder } from "./hooks/media-recorder";
 import { useLongPressToggle } from "./hooks/long-press-toggle";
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/ai/ai-command-bar.tsx
```diff
@@ -26,57 +26,49 @@ import {
...
+  const { start, stop, cancel } = useMediaRecorder({
     onComplete: async (file) => {
-      setIsAudioTranscribing(true);
-      guardIdRef.current++;
-      const guardId = guardIdRef.current;
-      const text = await fetchTranscription(file);
-      if (guardId !== guardIdRef.current) {
-        return;
+      setStatus("transcribing");
+      abortController.current = new AbortController();
```
istarkov, line 56: This is so so code for async env. You are trying to use the fact that `onComplete` will never be called again until all async code below will be completed. It's not true. 
`onCompleted` doesn't give this guarantee, see `guardIdRef.current++` is not for fun. 
Even every button below is disabled "immediately" to prevent parallel execution of anything, it's React `setState` is not synchronous, there is not guarantee then button couldn't be clicked again before disabled state.

`abortController.current = new AbortController();` everytime you do this having assumptions above you need to check that `abortController.current` is disposed

apps/builder/app/builder/features/ai/ai-command-bar.tsx
```diff
@@ -26,57 +26,49 @@ import {
...
+      try {
+        let prompt = await fetchTranscription(
+          file,
+          abortController.current.signal
+        );
+        prompt = prompt.trim().replace(/\.$/, "");
```
istarkov, line 62: We should not validate/change our api response

apps/builder/app/builder/features/ai/ai-command-bar.tsx
```diff
@@ -26,57 +26,49 @@ import {
...
-      const guardId = guardIdRef.current;
-      const text = await fetchTranscription(file);
-      if (guardId !== guardIdRef.current) {
-        return;
+      setStatus("transcribing");
+      abortController.current = new AbortController();
+      try {
+        let prompt = await fetchTranscription(
+          file,
+          abortController.current.signal
```
istarkov, line 60: There is an error again this code.
again assumption that `onComplete` can't be executed in parallel. It can. 
So you need to check after function execution before using result that controller wasn't aborted. (fetchTanscription can use other async methods in the future not fetch only)

apps/builder/app/builder/features/ai/ai-command-bar.tsx
istarkov, line 62: We should not validate/change our api response

istarkov, line 62: We should not validate/change our api response

apps/builder/app/builder/features/ai/ai-command-bar.tsx
```diff
@@ -26,57 +26,49 @@ import {
...
+        );
+        prompt = prompt.trim().replace(/\.$/, "");
+        setPrompt(prompt);
+        handleAiRequest(prompt);
```
istarkov, line 64: Error, previous code used not the prompt but combined result.

apps/builder/app/builder/features/ai/ai-command-bar.tsx
```diff
@@ -86,132 +78,52 @@ export const AiCommandBar = () => {
     },
   });
 
-  const longPressToggleProps = useLongPressToggle({
+  const recordButtonProps = useLongPressToggle({
     onStart: () => {
       start();
       disableCanvasPointerEvents();
+      setStatus("recording");
```
istarkov, line 85: No its not recording here. User can disallow mic usage

apps/builder/app/builder/features/ai/ai-command-bar.tsx
```diff
@@ -247,51 +159,89 @@ export const AiCommandBar = () => {
...
-              onChange={setValue}
-              onKeyDown={(event) => {
+              disabled={status !== "idle"}
+              placeholder={
+                status === "recording"
+                  ? "Recording..."
+                  : status === "transcribing"
+                  ? "Transcribing recording..."
+                  : "Enter value..."
+              }
```
istarkov, line 169: Are you think I moved all this `? :` into linear imperative `if's` for fun ;-) The code above is unsupportable. Every new state would make this grow grow grow. See in previous code, new state just single `if` in render was added

istarkov (changes requested):

apps/builder/app/builder/features/ai/ai-command-bar.tsx
istarkov: You can type text then record

apps/builder/app/builder/features/ai/ai-command-bar.tsx
kof: move it to a function that resolves the state, then its easier to read, even better if its a state machine

kof:
this can be closed? @istarkov

kof:
Afaik we agreed to close this

PR #2455, opened 2023-10-12
fix: sync commands after builder and canvas initialized

TrySound:
Found we have race condition because of handshake
which complicates synchronizatio of commands.

Here removed handshake. Instead iframe now rendered client side after first effect and canvas sends "connect" event so builder start sending updates.

In the future this can be simplified even more with computed updates from current state like yjs does.

## Testing steps

- try to delete instance after selecting on canvas
- try to press enter to edit text after selecting from navigator

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/canvas.tsx
+++ b/apps/builder/app/canvas/canvas.tsx
@@ -228,7 +228,7 @@ export const Canvas = ({
   const instances = useStore(instancesStore);
   const elements = useElementsTree(components, instances, params, imageLoader);
 
-  const [initialized, setInitialized] = useState(false);
+  const [isInitialized, setInitialized] = useState(false);
   useEffect(() => {
     setInitialized(true);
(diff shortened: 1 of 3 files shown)


kof (approved):

apps/builder/app/canvas/canvas.tsx
```diff
@@ -233,6 +228,11 @@ export const Canvas = ({
   const instances = useStore(instancesStore);
   const elements = useElementsTree(components, instances, params, imageLoader);
 
+  const [initialized, setInitialized] = useState(false);
```
kof, line 231:
```suggestion
  const [isInitialized, setIsInitialized] = useState(false);
```

istarkov (approved):

PR #2459, opened 2023-10-12
feat: Enable features via url

gsppe:
Allows to enable features from the /dashboard endpoint by setting it via `features` url param

### DIFF
--- a/apps/builder/app/shared/use-set-features.ts
+++ b/apps/builder/app/shared/use-set-features.ts
@@ -1,5 +1,5 @@
 import { useSearchParams } from "@remix-run/react";
-import { setLocal } from "@webstudio-is/feature-flags";
+import { parse, readLocal, setLocal } from "@webstudio-is/feature-flags";
 import { useEffect } from "react";
 
 // Allows to set feature flags via URL parameter features=name1,name2,name3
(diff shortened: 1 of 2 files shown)


apps/builder/app/routes/dashboard._index.tsx
```diff
@@ -51,6 +56,15 @@ export const ErrorBoundary = () => {
 };
 
 const DashboardRoute = () => {
+  // Allows to set feature flags via URL parameter features=name1,name2,name3
```
kof, line 59: - why do we need to depend on react for this?
- why is this only in dashboard? what if URL is directly to a project?

I think we could solve both by simply checking this somewhere in feature-flags module

kof: you are right, adding aside effect to the module is bad, I guess then we have to move this to a hook and use it in both dashboard and builder

kof: ah I know what you mean now, but that's bad too, so lets just use what you did as a hook from feature-flags and call it in both dashboard and builder

kof (approved):

apps/builder/app/routes/dashboard._index.tsx
kof: looking good, you are right, we need to merge them, otherwise it could be more surprises

PR #2468, opened 2023-10-13
fix: ai instance deletion

gsppe:
Fixes instance deletion issues. Depends on https://github.com/webstudio-is/webstudio/pull/2467

### DIFF
--- a/apps/builder/app/shared/instance-utils.test.ts
+++ b/apps/builder/app/shared/instance-utils.test.ts
@@ -10,9 +10,11 @@ import {
   findClosestEditableInstanceSelector,
   insertTemplateData,
   type InsertConstraints,
+  deleteInstance,
 } from "./instance-utils";
 import {
   instancesStore,
+  registeredComponentMetasStore,
   styleSourceSelectionsStore,
   styleSourcesStore,
   stylesStore,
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/features/ai/ai-fetch-result.ts
```diff
@@ -85,6 +65,31 @@ export const fetchResult = async (
...
+    components: availableComponentsNames,
+    projectId: project?.id,
+    instanceId: selectedInstanceSelector?.[0],
+    prompt,
+  };
+
+  // @todo Future commands might not require all the requestParams above.
+  // When that will be the case, we should revisit the validatin below.
+  if (requestParams.instanceId === undefined) {
+    return ["You must select an element."];
```
kof, line 84:
```suggestion
    return ["Please select an instance on the canvas."];
```

kof (approved):

apps/builder/app/builder/features/ai/ai-fetch-result.ts
```diff
@@ -85,6 +65,31 @@ export const fetchResult = async (
...
+  }
+  if (
+    RequestParamsSchema.omit({ command: true }).safeParse(requestParams)
+      .success === false
+  ) {
+    return ["Invalid prompt data"];
```
kof, line 90: not sure when this is the case, but if its possible to get this as a user-facing issue, we might want a better message

giuseppeg: do you have any message suggestion for when (for whatever reason) the requestParams is not valid?

kof: well if we don't know anything that we can tell the user then I can't improve it

TrySound (approved):

PR #2472, opened 2023-10-13
fix: AI Command Bar error processing

istarkov:
## Description

Rate Limiting toasts.
Mic enable toast.
Improved Error messages toasts for debugability reasons. 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

- [ ] hi @kof, I need you to do
  - conceptual review (architecture, feature-correctness)
  - detailed review (read every line)
  - test it on preview

## Before requesting a review

- [ ] made a self-review
- [ ] added inline comments where things may be not obvious (the "why", not "what")

## Before merging

- [ ] tested locally and on preview environment (preview dev login: 5de6)
- [ ] updated [test cases](https://github.com/webstudio-is/webstudio/blob/main/apps/builder/docs/test-cases.md) document
- [ ] added tests
- [ ] if any new env variables are added, added them to `.env.example` and the `builder/env-check.js` if mandatory

PR #2474, opened 2023-10-14
feat: Add slash commands

gsppe:
Adds support for / slash commands to execute operations directly without going through command detection (it is skipped).

Available commands (with shortcuts)

/text /t
/style /s
/ui /u
/delete /d

The branch is on top of #2472 because that PR changes a lot in the same file

### DIFF
--- a/apps/builder/app/builder/features/ai/ai-fetch-result.ts
+++ b/apps/builder/app/builder/features/ai/ai-fetch-result.ts
@@ -39,32 +39,24 @@ const unknownArray = z.array(z.unknown());
 export const fetchResult = async (
   prompt: string,
   abortSignal: AbortSignal
-): Promise<string[]> => {
+): Promise<void> => {
   const commandsResponse = await handleAiRequest<commandDetect.Response>(
     fetch(restAi("detect"), {
       method: "POST",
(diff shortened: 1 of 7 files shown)


kof:
- [ ] We probably need to add info about it to the menu

apps/builder/app/builder/features/ai/ai-fetch-result.ts
```diff
@@ -34,29 +34,56 @@ import untruncateJson from "untruncate-json";
 import { traverseTemplate } from "@webstudio-is/jsx-utils";
 import { RequestParamsSchema } from "~/routes/rest.ai._index";
 
+const slashCommands = {
+  text: copywriter.name,
+  t: copywriter.name,
```
kof, line 39: idk regarding shortcuts, do we need them? is it popular?

istarkov:
I thought to add slash support through a textarea menu and pass command as standalone param, without text parsing in a fetch

kof:
We need to think more about this feature, we have a few more ideas and implementation details, but this seems to be a mixture of the feature and other changes, I hope by closing this we are not missing out on some other important fixes @gsppe

PR #2475, opened 2023-10-14
fix: Commands detection and small bugs

gsppe:
Attempt to improve commands detection and adds a few other smaller fixes.


kof:
Closing it according to @gsppe

PR #2479, opened 2023-10-15
fix: bundle esm dependencies in templates

TrySound:
Fixes https://github.com/webstudio-is/webstudio-community/discussions/62

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
how do we make sure this won't happen again? seems like we have no test that makes sure another dependency added will not break it again

PR #2482, opened 2023-10-15
fix: prevent image delete when close image dialog

TrySound:
Looks like "disabled" state triggered "delete" button near "Choose source".

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Bug still fully reproducible, after replacing the URL with image, closing image picker and selecting any instance on canvas other than that image will result in a reset

kof (approved):
works now!

PR #2487, opened 2023-10-15
fix: Instruct model to rewrite text when necessary

gsppe:
This could improve how the model writes, instructing it to ignore the current value


istarkov:
Super, it works, changes text, can translate it make a poem from text!!!!

istarkov (approved):

PR #2504, opened 2023-10-18
feat: Quick settings dialog

kof:
## Description

We haven't discussed the design, so this is a quick experimental implementation to see how hard it is and what the questions are.

The reason we need this is for components like Link, Image and others where component needs to be immediately configured after insertion, its practical to discover the settings right away like WF does

Screenshot 2023-10-18 at 13 33 37

## Todo

- [ ] make a list of components which should have settings open automatically when instance is created
- [ ] decide how to express settings auto-show via meta
- [ ] decide on the exact layout we show
- [ ] decide on the properties we show (maintaining a separate list is impractical, most likely we will just show all the initialProps) 
- [ ] what if we show the entire settings panel, would it be bad? should we only show initial props? wouldn't it be practical to also see instance label and show?
- [ ] if we only show initial props, we should rename it to QuickPropsDialog
- [ ] decide how image picker would open up

## Steps for reproduction

1. add an instance
2. click the temp icon on the label

## Code Review

- [ ] hi @taylor, I need you to do
 - design review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/workspace/canvas-tools/canvas-tools.tsx
+++ b/apps/builder/app/builder/features/workspace/canvas-tools/canvas-tools.tsx
@@ -76,7 +76,7 @@ export const CanvasTools = ({ publish }: CanvasToolsProps) => {
       {isPreviewMode === false && (
         <>
           <div className={containerStyle({ overflow: "hidden" })}>
-            <SelectedInstanceOutline />
+            <SelectedInstanceOutline publish={publish} />
             <HoveredInstanceOutline />
             <CollaborativeInstanceOutline />
           </div>
--- a/apps/builder/app/builder/features/workspace/canvas-tools/outline/label.tsx
+++ b/apps/builder/app/builder/features/workspace/canvas-tools/outline/label.tsx
@@ -6,6 +6,8 @@ import { theme } from "@webstudio-is/design-system";
 import { MetaIcon } from "~/builder/shared/meta-icon";
 import { registeredComponentMetasStore } from "~/shared/nano-states";
 import { getInstanceLabel } from "~/shared/instance-utils";
+import { QuickSettings } from "./quick-settings";
+import type { Publish } from "~/shared/pubsub";
 
 type LabelPosition = "top" | "inside" | "bottom";
 type LabelRefCallback = (element: HTMLElement | null) => void;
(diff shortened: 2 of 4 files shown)


kof:
@taylornowotny  Image picker panel is a problem, since we are showing it as a floating panel next to the right sidebar, where is it supposed to go in this case? I am guessing it would either show instead of quick settings dialog and then when image selected quick settings dialog is back, or it shows next to it,

But either way, we would need  to make image picker aware of the context where its been called from (settings vs quick settings)

taylornowotny:
I did a review went through all of ours and Webflow's components.

UX basics
- this is a completely separate menu from the context menu
- opens automatically when certain instances are added to the canvas
- those instances also have a settings button on the canvas label which opens this menu
- only the initial props should show in this menu, nothing more. This menu should be compact. There's no need to include addition settings or re-create the settings panel - the settings panel is perfectly useable for that.

Notes
- No quick settings for radix bc those have many layers.  It’s usually not the parent that has most of the settings.
- Images settings needs a design update, including the asset picker UX and the "choose source" UI/UX.

List of components
- all media
- links
- headings
- lists, not list items
- forms
- all inputs
- labels, maybe

Discussion points
- Include a button to "Show all settings" at the bottom?
- Let's discuss possible improvements to settings. I have a list of ideas.
- Should we include ID or not? Webflow doesn't. I'm not opinionated but excluding it would conserve vertical space.

kof:
If we want to avoid including certain properties there, we will probably rather need to have a separate properties list for each component that we want to show there, maybe this is even better if we don't want to show all initial properties in some cases?

For example Vimeo component has a lot of initial properties.

taylornowotny:
> If we want to avoid including certain properties there, we will probably rather need to have a separate properties list for each component that we want to show there, maybe this is even better if we don't want to show all initial properties in some cases?
> 
> For example Vimeo component has a lot of initial properties.

We can do this, but I don't think it's necessary. The only inotial prop I'd omit is "ID", but like I said I'm not opinionated about it.

Vimeo has a lot of initial props yes, but as long as the quick settings has a max height, is scrollable, and the most important settings are at the top, I don't think that's a problem. This is actually better than opening the settings panel automatically so we can revert that behavior.

kof:
@taylornowotny I was just thinking that we should probably just pick 3-4 most important props for vimeo in the quick settings dialog, if this. is what we want, then we need to have a separate list, we can have initialProps by default too

taylornowotny:
ok lets do that

PR #2511, opened 2023-10-20
feat: Add css transitions

JayaKrishnaNamburu:
## Description

Adds support for `css-transitions` when users are in `local` state. In any other state than local, the CRUD operations on transitions are disabled.


JayaKrishnaNamburu:
@taylornowotny @kof the transitions section is in usable state. The UX is very similar to `box-shadow`. So give it a try and see the changes needed.

kof:
@JayaKrishnaNamburu testing

kof:
- [ ] I think it makes sense to have WF defaults when adding transition:

Screenshot 2023-10-22 at 10 11 09

kof:
Is the feature still behind a feature flag? because we can't merge if user will see it as-is right away

apps/builder/app/builder/features/style-panel/shared/use-intermediate-layer.ts
```diff
@@ -0,0 +1,46 @@
+import type { InvalidValue, LayersValue } from "@webstudio-is/css-engine";
+import { useState } from "react";
+
+type IntermediateValue = {
+  type: "intermediate";
+  value: string;
+};
+
+export const useIntermediateContent = ({
```
istarkov, line 9: First this hook is written with bug. It copypasted from buggy version. And this bug are in few places.
intermediate value should be used only during edit. And should be reseted after onComplete. Otherwise any changes to style from other places (like AI changes) will never be shown.

2nd I don't think that this hook need to be shared. It has no value to be shareable. I don't even like its a hook, as its just handlers and state.

packages/css-data/bin/mdn-data.ts
```diff
@@ -233,6 +233,22 @@ const filteredProperties: FilteredProperties = (() => {
     const config = properties[property];
     const isSupportedStatus =
       config.status === "standard" || config.status === "experimental";
+
+    /*
```
istarkov, line 237: use GPT to fix english errors ;)  prperties -> .... etc

istarkov: Also I don;t understand comment, example will be better

packages/css-data/bin/mdn-data.ts
```diff
@@ -233,6 +233,22 @@ const filteredProperties: FilteredProperties = (() => {
...
+      But, the transition property from the designs allows users to set all at once so we need the
+      property to be available from the generated list
+      https://www.figma.com/file/EfzZ2w94KpmUviZJSBgxCd/%F0%9F%92%AD-Webstudio-Design-Work?type=design&node-id=6424-30607&mode=design
+
+      The values all, 0s, ease, 0s are all combination of defaults for the rest of the values
+      https://developer.mozilla.org/en-US/docs/Web/CSS/transition#formal_definition
+
+    */
+    if (property === "transition") {
+      config.initial = "all 0s ease 0s";
```
istarkov, line 249: Why we use config initial here, what it solves?

packages/css-data/src/property-parsers/transition.ts
```diff
@@ -0,0 +1,105 @@
...
+const cssTreeTryParseValue = (input: string) => {
+  try {
+    const ast = csstree.parse(input, { context: "value" });
+    return ast;
+  } catch {
+    return undefined;
+  }
+};
+
+export const parseTransition = (
```
istarkov, line 19: useless without tests no one will get what this function do

packages/css-data/src/property-parsers/transition.ts
```diff
@@ -0,0 +1,105 @@
...
+  if (parsed.error) {
+    return {
+      type: "invalid",
+      value: transition,
+    };
+  }
+
+  const layers: TupleValue[] = [];
+
+  csstree.walk(cssAst, (node) => {
```
istarkov, line 54: Imo underlying algorithm should be simple - simple.
It's just splitting children array on Operator `,` but too much logic for this simple thing.
like this comparisons `children.last === child` looks useless

packages/css-data/src/property-parsers/transition.ts
kof: definietely needs tests like other similar functions have

JayaKrishnaNamburu: Yeah, completely forgot adding them when opening for review. Will take a look at all the comments and the PR once again 👍

taylornowotny (changes requested):
Tested and works for me!

UI feedback:
- (bug) click an item to open the menu, then close the menu. You'll see the "remove" button stays visible when it should be hidden.
- There should be no space between items.

packages/css-data/src/property-parsers/transition.ts
JayaKrishnaNamburu: The first iteration of design has individual display like, so the parser was little complicated to pick everything. Looks like the design is changed now, and the parser also can be just simple one now.

JayaKrishnaNamburu:
Fixes the `disabled` state @taylornowotny. But regarding the second one

> There should be no space between items.
It's just the `CSSValueListItem` component. So, if it's changed here. It reflects everywhere. I am not sure, if need to be changed everywhere. Can you confirm on this once.

And regarding rest of all the comments, @istarkov handled a bunch of them. Added tests and added new comments in the `mdn-data` file for the `config.initial` changes

taylornowotny:
Yes Jaya, all CSSValueListItems should not have spaces between items.
Thanks!

JayaKrishnaNamburu:
Done @taylornowotny

taylornowotny:
Thanks @JayaKrishnaNamburu space is good now.

Here's a video of that bug I mentioned in case it wasn't clear

https://github.com/webstudio-is/webstudio/assets/86495896/ec899303-e90b-4697-b469-6c7b52dab983

apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const layers = parseTransition(intermediateValue.value);
+    if (layers.type === "invalid") {
+      setIntermediateValue({
+        type: "invalid",
+        value: intermediateValue.value,
+      });
+      return;
+    }
+
+    props.onEditLayer(props.index, layers);
```
istarkov, line 46: I think we need to reset intermediate value to undefined in handleComplete. Imagine you have edited it, then AI will change store value, but in control you will always see the local state

istarkov (changes requested):
Plz see my comments

taylornowotny:
Floating panel UI is done and documented!

Please read the documentation very carefully to see all the details.

https://github.com/webstudio-is/webstudio/issues/1939#issuecomment-1778141375

JayaKrishnaNamburu:
Please don't do any code / ux reviews. I started with the UX. So, takes some time.

taylornowotny (changes requested):
Transitions panel UI not yet implemented

Remove button bug that I pointed out is still there

JayaKrishnaNamburu:
@taylornowotny i just updated the upstream. Did't push any chaanges yet. Will let you know

JayaKrishnaNamburu:
And @taylornowotny i just tested it. It's not bug, but more of a UX thing maybe ?
Because once the floating panel is closed. The focus is moved back to the element that is just being edited. So, the selector is kicking and marking them as visible. 

https://github.com/webstudio-is/webstudio/blob/main/packages/design-system/src/components/css-value-list-item.tsx#L121

https://github.com/webstudio-is/webstudio/assets/11075561/7b6113c4-6896-4540-8d5d-3bd57009c71c

taylornowotny:
@JayaKrishnaNamburu 

After the menu closes, the item should not be focused.

Does that answer your question?

kof:
@taylornowotny I just checked, its the same behavior in production, so this pr didn't change it, I would suggest always try on main branch to see if the change was implemented here.

What seems wrong about the current state is not the focus behavior itself, but the outline. When closing panel, the item is still focused, but there is no outline, that makes it seem like its not focused but it is and icons are also still visible.

Do you think we should fix the outline and make it visible after closing? focus stays on the item in that case as it was before panel was opened. So user can keep navigating using arrow keys from where they left

JayaKrishnaNamburu:
@taylornowotny we moved the bug to a new issue here. And will be handled separately. https://github.com/webstudio-is/webstudio/issues/2543
Let's move the discussion on this bug there, so all the discussions will be in one place there 👍

istarkov (approved):

taylornowotny (approved):
Bug moved to a separate issue.

UI not implemented - I'm creating a separate issue for this. #2546 

Approving just so I'm not blocking this.

PR #2515, opened 2023-10-21
refactor: unify pubsub and support multiple iframes

TrySound:
Here refactored our pubsub. Now iframes are registered instead of bounding to specific usePublish. This let us broadcast actions to multiple plugins with global publish call.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/pubsub/create.ts
+++ b/apps/builder/app/shared/pubsub/create.ts
@@ -1,5 +1,5 @@
 import { createNanoEvents } from "nanoevents";
-import { useCallback, useEffect, useRef } from "react";
+import { useEffect } from "react";
 import { batchUpdate } from "./raf-queue";
 
 export const createPubsub = <PublishMap>() => {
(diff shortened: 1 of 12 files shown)


apps/builder/app/shared/pubsub/create.ts
```diff
@@ -27,32 +27,36 @@ export const createPubsub = <PublishMap>() => {
     );
   }
 
+  const iframes = new Set<HTMLIFrameElement>();
+  const registerIframe = (element: HTMLIFrameElement) => {
```
kof, line 31: would it make sense to not couple iframe and pubsub and instead just register "an object" that has a postMessage method or even a function that returns that object

would make sense if we already know we want to be able to subscribe also without using an iframe

kof: or even registering directly a postMessage function that would be called

kof (approved):

kof:
Didn't test, because I assume if it didn't work, nothing would work on  canvas and you clearly tested something on canvas is working

apps/builder/app/shared/pubsub/create.ts
TrySound: No guarantee postMessage is bound and can be called without "this".

kof: yeah, can be a callback registerPostMessage((...args) => {
  iframe.contentWindow.postMessage(...args)
})

kof: would call that something else than registerPostMessage though

kof: could be just .addEventListener or .on((...args) => {
   iframe.contentWindow.postMessage(...args)
})

PR #2527, opened 2023-10-26
feat: Add SiteSettings Dialog

istarkov:
## Description

Dialogs and saving of Site Settings and Page Settings.

Bugs & Fixes
- [x] - Use page name on swap to set path i.e. `/home`

Next PRs.

- CLI support (generating head title links and metas)
- image
- image
- Not sure why we need upload button at all, anyway in next PRs image

---

image

image

image

- [x] - Make homePage is not working

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/seo/site-settings.tsx
```diff
@@ -0,0 +1,210 @@
...
+const SiteSettingsView = () => {
+  const [value, setValue] = useState(
+    pagesStore.get()?.meta ?? {
+      siteName: "",
+      faviconAssetId: "",
+      code: "",
+    }
+  );
+
+  const open = useStore($isSiteSettigsOpen);
```
kof, line 153: would be nice to use isSomething for booleans (I know radix doesn't follow that) 
```suggestion
  const isOpen = useStore($isSiteSettigsOpen);
```

apps/builder/app/builder/features/seo/site-settings.tsx
```diff
@@ -0,0 +1,210 @@
...
+      }}
+    >
+      <DialogContent
+        onBlur={handleSave}
+        // Left Aside panels (e.g., Pages, Components) use zIndex: theme.zIndices[1].
+        // For a dialog to appear above these panels, both overlay and content should also have zIndex: theme.zIndices[1].
+        css={{
+          zIndex: theme.zIndices[1],
+        }}
+        overlayCss={{
```
kof, line 180: this is quite nasty, one idea to work this around is to check if css.zIndex specifically is defined and pass it to overlay inside DialogContent

kof: to avoid having to pass overlayCss at all

apps/builder/app/builder/features/seo/site-settings.tsx
```diff
@@ -0,0 +1,210 @@
...
+        {/* Title is at the end intentionally,
+         * to make the close button last in the tab order
+         */}
+        <DialogTitle>Site Settings</DialogTitle>
+      </DialogContent>
+    </Dialog>
+  );
+};
+
+export const SiteSettings = () => {
```
kof, line 194: We normally call this *Container so that the SiteSettingsView can be just SiteSettings
Container would only have some state/data logic but no rendering

apps/builder/app/builder/features/seo/site-settings.tsx
istarkov: Imo this should not exists at all and caused by poor aside panel implementation. Here I think explicit css is better vs implicit as most css systems do not allow to check existance of specific properties passed from top (as its usually function or classnames) I dont want to stick on stitches implementation

apps/builder/app/builder/features/seo/site-settings.tsx
istarkov: Its the only exported class, and should have name related to its usage. When I use any imported component in React I dont mind is it just data class or container. Containers etc are good for internal implementations, but exposing this information out of component scope does give nothing

apps/builder/app/builder/features/seo/image-control.tsx
```diff
@@ -0,0 +1,30 @@
...
+  onAssetIdChange: (assetId: string) => void;
+  children: ReactElement;
+}) => {
+  return (
+    <FloatingPanel
+      title="Images"
+      contentCss={{
+        // Left Aside panels (e.g., Pages, Components) use zIndex: theme.zIndices[1].
+        // For a panel to appear above these panels, both overlay and content should also have zIndex: theme.zIndices[1].
+        zIndex: theme.zIndices[1],
```
kof, line 17: if this is the only thing we need to pass over to content, might just pass zIndex or something

kof (approved):

kof:
Couldn't test if this works in terms of rendering

apps/builder/app/builder/features/seo/site-settings.tsx
kof: I agree, I think the fundamental problem is the composition, the way radix does this - it allows to not have these problems, but we created a higher level component that has the composition built-in. In theory that higher level component should have z-index hardcoded.

I also agree z-index shouldn't be needed but if we take z-index abstract as a need to add any CSS override, still the problem is composition

Issue #2388, opened 2023-09-28
Use mono text to display code in CLI and HTML Embed


taylornowotny:
Code should be displayed in mono font. We have a "mono" text token for this purpose.

Example:
Design:
<img width="318" alt="Screenshot 2023-09-28 at 12 04 29 PM" src="https://github.com/webstudio-is/webstudio/assets/86495896/598def9d-8c8e-4854-885e-a3f851ea2c05">
Prod:
<img width="314" alt="Screenshot 2023-09-28 at 12 04 58 PM" src="https://github.com/webstudio-is/webstudio/assets/86495896/5c7a1331-78b2-4654-a4a7-f7fc423f65cf">

Please use mono text in these places:
- [x] CSS preview
- [x] CLI text inputs
- [x] Code property text area on the Embed component

If there are other places where code is displayed but mono is not used, please fix those too.

faisal004:
hi @kof @TrySound can I work on this issue

kof:
@faisal004 sure, take it, we have theme.fonts.mono in stitches.config and it can be passed to those inputs over a CSS prop

PR #2534, opened 2023-10-28
fix: mono text added to display code in cli

faisal004:
## Description

This PR fixes the issue number #2388 
I have added font mono to display text of cli

Screenshot 2023-10-28 185213

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/input-field.tsx
+++ b/packages/design-system/src/components/input-field.tsx
@@ -31,7 +31,7 @@ export const inputFieldColors = ["placeholder", "set", "error"] as const;
 
 const inputStyle = css({
   all: "unset",
-  ...textVariants.regular,
+  ...textVariants.mono,
   color: theme.colors.foregroundMain,
   flexGrow: 1,
   flexShrink: 1,
--- a/apps/builder/app/dashboard/header.tsx
+++ b/apps/builder/app/dashboard/header.tsx
@@ -48,7 +48,7 @@ const Menu = ({ user }: { user: User }) => {
           </Flex>
         </Button>
       </DropdownMenuTrigger>
-      <DropdownMenuContent>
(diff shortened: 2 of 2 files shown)


packages/design-system/src/components/input-field.tsx
```diff
@@ -31,7 +31,7 @@ export const inputFieldColors = ["placeholder", "set", "error"] as const;
 
 const inputStyle = css({
   all: "unset",
-  ...textVariants.regular,
+  ...textVariants.mono,
```
kof, line 34: can't do that here, the input field is used everywhere, you need to use a CSS prop in specific places where you need that font

kof: alternative would be to add a variant "text" and the value would be textVariants, similar to how  Text component has 
```js
export const textStyle = css({
  margin: 0, // in case it's used with <p>
  WebkitFontSmoothing: "antialiased",
  variants: {
    variant: textVariants,
    color: {
```

but in this case it would be 

```js
variants: {
  text: textVariants
}
```

kof: I think a text variant is a great idea @istarkov wdyt?

faisal004: @kof sure I'll fix that

faisal004: Hi @kof I have been trying fixing it 
Changes input as you said but somehow its not changing
Screenshot 2023-10-31 224828
Screenshot 2023-10-31 224908

kof:
Please reopen once ready.

Issue #1939, opened 2023-07-15
Transitions section


kof:
Design is in draft stage, many thingsa are not finished there

https://www.figma.com/file/EfzZ2w94KpmUviZJSBgxCd/%F0%9F%92%AD-Webstudio-Design-Work?type=design&node-id=6424-30607&mode=design&t=LXNO2FwknePYv6qR-0

- [ ] Reuse the functional logic between background layers and transitions
- [ ] Render the same sortable list as with backgrounds, same accessibility
- [ ] Render floating panel
- [ ] Instead of "Type", it should be "Property". The control is combobox, it should support all properties, not just randomly selected once. User can type it, can autocomplete it, can select "All"
- [ ] Duration is CssValueInput with a fixed units list, default is 200ms
- [ ] Easing is a combobox with all possible easing options from https://easings.net/ as well as ability to type a custom one
- [ ] Delay is the same as duration, default is 0

JayaKrishnaNamburu:
I looked into this, i have several questions on this.

- In the first iteration, is there any plans to allow users itself to add transitions as they can with `box-shadow` and `background`. Because i see that in figma description but missing from the design.
![Screenshot 2023-10-09 at 17 14 33](https://github.com/webstudio-is/webstudio/assets/11075561/7e1063f0-ed88-4fde-91e6-baf3cfb260a9)

- Do we have a list of all the css attributes that are needed to show in the list. Or the list contains the time-definition-property from the actual spec. 
![Screenshot 2023-10-09 at 17 15 32](https://github.com/webstudio-is/webstudio/assets/11075561/35b55e49-9e0c-4672-8acc-2c7c6ce78e42)

- If users are allowed to use custom transition properties. Then the icons here are difficult to represent those. Eg, there should be icons for `linear`, `steps`, `cubic-beizer` etc. Or else there should be independent icons for these, `ease` `ease-in`, `ease-out`, `linear`, `ease-in-out`. Because the rest of them needs user inputs eg: `step(x, y)`, `cubic-beizer(..)`
![Screenshot 2023-10-09 at 17 16 54](https://github.com/webstudio-is/webstudio/assets/11075561/6005b3e9-94f5-4f2b-8a6d-a59b59719da5)

The best way seems to be, to show the full layer as text as it shown for box-shadow. It's additional meta for users to remember and relate the icons with the respective transition(But that's personal opinion). 

[...]

PR #2538, opened 2023-10-30
feat: Add UI element to control transition property.

JayaKrishnaNamburu:
## Description

Adds support for `transition-property` UI for the transition section. Related to #2546


taylornowotny:
Posting this here and closing the other issue

Part of #1939
Related to #2511

This issue is for the Transitions Section menu UI. Please read the documentation very carefully to see all the details.
Screenshot 2023-11-01 at 9 14 22 AM

Figma: https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=8085-44296&mode=design

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
```diff
@@ -0,0 +1,129 @@
...
+]);
+
+const comboBoxStyles = css({ zIndex: theme.zIndices[1] });
+
+export const TransitionProperty = ({
+  property,
+  onPropertySelection,
+}: TransitionPropertyProps) => {
+  const [inputValue, setInputValue] = useState(property?.value ?? "all");
+  useEffect(() => setInputValue(property.value), [property]);
```
kof, line 46: might want to track the value specifically

```suggestion
  useEffect(() => setInputValue(property.value), [property.value]);
```

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
```diff
@@ -0,0 +1,129 @@
...
+    value: { name: inputValue, label: inputValue },
+    selectedItem: undefined,
+    itemToString: (value) => value?.name ?? "",
+    onItemSelect: (prop) => {
+      setInputValue(prop.name);
+      onPropertySelection({ type: "keyword", value: prop.name });
+    },
+    onInputChange: (value) => {
+      if (value === undefined) {
+        return "";
```
kof, line 70: this return looks weird, I don't think the return value is used, did you mean to just return; ?
```suggestion
        return;
```

JayaKrishnaNamburu:
The rest of the inputs will be added in another PR's. Because `transition-timing` alone has another level of timing functions handling and the PR is becoming big. So, this one adds only handling `transition-property`.

kof:
Screenshot 2023-11-03 at 18 33 13

I can see on the video: `opacity 200ms ease 0s`, why is there 0s? isn't 200ms and 0s mutually exclusive?

JayaKrishnaNamburu:
> Screenshot 2023-11-03 at 18 33 13
> I can see on the video: `opacity 200ms ease 0s`, why is there 0s? isn't 200ms and 0s mutually exclusive?

Yup, one refers to the transition `duration` and one refers to the `delay` of the timing function

kof:
Right, didn't think of it, in the design we don't show the last parameter, but also probably @taylornowotny didn't think of delay, I guess its fine

kof:
-[ ] I guess parse errors are not handled yet, its fine to merge if not, just noting for the next one

kof:
- [ ] typed random text and it took it 

image

kof:
- [ ] something weird happens when using arrow keys, highlight jumps randomly

apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
```diff
@@ -43,34 +69,87 @@ export const TransitionContent = (props: TransitionContentProps) => {
...
+            Code
+            <Tooltip
+              variant="wrapped"
+              content={
+                <Text>
+                  Paste CSS code for a transition or part of a transition, for
+                  example:
+                  <br />
+                  <br />
+                  opacity 200ms ease;
```
kof, line 131: shouldn't this be a stringified version of the value we get as a default to sync these things

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
```diff
@@ -0,0 +1,17 @@
+import type { TupleValue } from "@webstudio-is/css-engine";
+
+export const convertTupleToStringValue = (layer: TupleValue): string => {
```
kof, line 3: isn't toValue() converting tuple value to css string? just not sure what's the difference

apps/builder/app/builder/features/style-panel/style-sections.tsx
```diff
@@ -42,7 +43,7 @@ export const categories = [
   "boxShadows",
   "outline",
   "effects",
-  "transitions",
+  // "transitions",
```
kof, line 46: should probably move it to the right place? or is it for convenience while in dev?

packages/css-data/bin/mdn-data.ts
```diff
@@ -397,6 +405,7 @@ const keywordValues = (() => {
 writeToFile("units.ts", "units", units);
 writeToFile("properties.ts", "properties", propertiesData);
 writeToFile("keyword-values.ts", "keywordValues", keywordValues);
+writeToFile("animatable.ts", "animatableProperties", animatableProperties);
```
kof, line 408: I would call the file animatable-properties.ts for clarity

packages/css-data/src/property-parsers/transition-property-extractor.ts
```diff
@@ -0,0 +1,55 @@
+import type {
+  KeywordValue,
+  TupleValue,
+  UnitValue,
+} from "@webstudio-is/css-engine";
+
+const isTimingFunction = (timing: string) => {
+  const regex =
+    /^(ease(-in-out|-in|-out)?|linear|cubic-bezier\((-?\d+(\.\d+)?, ?){3}-?\d+(\.\d+)?\)|steps\(\d+(-?(start|end))?\))$/gm;
```
kof, line 9: this is scary, what are your thoughts about using CSS parser vs this?

packages/css-data/src/property-parsers/transition-property-extractor.ts
```diff
@@ -0,0 +1,55 @@
...
+  return regex.test(timing);
+};
+
+export const extractTransitionProperties = (
```
kof, line 13: deserves some unit tests

kof: ah found them, nice

kof (approved):
approving, there are a bunch of questions/bugs, but feel free to move them to the next issue/pr, please don't forget to transfer them

taylornowotny:
omitting a delay and 0s delay are the same

kof:
> omitting a delay and 0s delay are the same
- [ ] true, lets omit when 0

apps/builder/app/builder/features/style-panel/style-sections.tsx
JayaKrishnaNamburu: Yup, need to be changed. Was moved just for dev convenience 👍

packages/css-data/src/property-parsers/transition-property-extractor.ts
JayaKrishnaNamburu: We are using css parser for initially parsing the value and finding individual values. But it just returns `Identfier` which can be anything there since transition is a complex value made up of two keyword values. Used this regex to find out the property from the kyewords.

apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
JayaKrishnaNamburu: We seem to use the same pattern in all the places. Buti tried `stip-indent`. It seems to collapse all spaces in jsx,

JayaKrishnaNamburu:
> * [ ]  typed random text and it took it as property name
> * [ ]  with that random text I can't click on property combobox arrow button, shows nothing, what's the right behavior when there is no match? showing nothing? or showing default list?
> 
> image

Looks like ike it is trying to show matched proeprties. And yeah the parser seems to accept since random string words as custom `transition-property` and seems to accept it. I will add a check in the parsre as thre is no support for custom proeprties at the moment 👍

packages/css-data/src/property-parsers/transition-property-extractor.ts
kof: there are css value parsers, I have a feeling we need one, this thing is gonna bite us badly

kof: also there needs to be tests for all of the things this regex covers if we are not using a library, but  I still think we should use a proper library to parse this @istarkov wdyt?

kof:
- [ ] doesn't look like 0 delay is being omited in the code

kof:
- [ ] doesn't show error state

 
Screenshot 2023-11-04 at 16 35 45

kof:
- [ ] arrow keys on combobox still not working

packages/css-data/src/property-parsers/transition-property-extractor.ts
istarkov: I would use css-tree and take function name to parse
image

JayaKrishnaNamburu:
> * [ ]  doesn't look like 0 delay is being omited in the code

I am adding actually in default state as a browser default. I will remove it by default now.

packages/css-data/src/property-parsers/transition-property-extractor.ts
JayaKrishnaNamburu: Yup, but it only works for `cubic-beizer` `setps`. For `ease`, `ease-in` etc it again shows as Identifiers. Which need to be verified again as both transition-property and timing functions are regarded as `identifiers` in those case. 

Screenshot 2023-11-04 at 21 42 07


So, still need to maintain a record of identifers and function names which are inside regex.  
And it will be two step process, regex-parse + name check from a list.
So i don't see much upside by moving from regex check to parsing using a tree and still comparing with a array. Anything specific with regex or it more code clarity ? 🤔

kof (approved):
- [ ] selecting property from combobox uses wrong index

packages/css-data/src/property-parsers/transition-property-extractor.ts
JayaKrishnaNamburu: Resolving here, moving the discussion to the transition-timing PR. Since these are related and it's a function swap anyways 👍

PR #2547, opened 2023-11-01
feat: add basic code editor

TrySound:

Here added code editor with variables completion support for bindings UI. Simple textarea would not work because same names can be used for different variables
to support seamless dnd for user. `$` is used to trigger completion, in the future can be replaced with tab or something like this. Current implementation is mostly a copy paste from lexical playground.

image

To test try CodeEditor component in storybook

## Use case

1. Our variables can have the same names for easier copy-pastability
To address specific variable they should be chosen from available list.

2. renaming variable should rename it everywhere to not break bindings.
So variables behave as symbol in text editor and removed completely on backspace.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/code-editor.tsx
+++ b/apps/builder/app/builder/features/settings-panel/code-editor.tsx
@@ -1,0 +1,382 @@
+import { useCallback, useEffect, useState } from "react";
+import { createPortal } from "react-dom";
+import {
+  type DOMConversionMap,
+  type DOMConversionOutput,
+  type DOMExportOutput,
+  type EditorConfig,
+  type LexicalNode,
+  type NodeKey,
+  type SerializedTextNode,
+  $applyNodeReplacement,
+  TextNode,
+  $getRoot,
+  $isElementNode,
+  createEditor,
+  type LexicalEditor,
+} from "lexical";
+import { $generateNodesFromDOM } from "@lexical/html";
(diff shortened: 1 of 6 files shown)


apps/builder/app/builder/features/settings-panel/code-editor.tsx
```diff
@@ -0,0 +1,383 @@
+import { useCallback, useEffect, useState } from "react";
```
kof, line 1: is this code editor supposed to be any code editing or specifically expressions editor?

kof (approved):

apps/builder/app/builder/features/settings-panel/code-editor.tsx
TrySound: Any code editing + completion of variables. Variables can be for js and css with custom properties.

kof: I just wonder how to distinguish this from a generic code editor we will have for react components at some point. And also this is gonna be needed in style panel too so settings-panel is not the right place soon enough

TrySound: I'd keep it here before it become suitable for css. Would be reimplemented anyway to support more stuff like highlighting.

taylornowotny (approved):

PR #2572, opened 2023-11-06
feat: Transition timing function UI

JayaKrishnaNamburu:
## Description

Adds `transition-timing` function UI to the transition content panel.

Screenshot 2023-11-06 at 16 51 35
Screenshot 2023-11-06 at 16 51 48


apps/builder/app/builder/features/style-panel/sections/transitions/transition-timing.tsx
```diff
@@ -0,0 +1,185 @@
...
+    selectedItem: undefined,
+    itemToString: (value) => value?.name ?? "",
+    onInputChange: (value) => setInputValue(value ?? ""),
+    onItemSelect: (prop) => {
+      setInputValue(prop.name);
+      onTimingSelection({ type: "keyword", value: prop.value });
+    },
+  });
+
+  const defaults = useMemo(() => {
```
kof, line 82: seems like quite  bit of code to filter and the items in each category, I have a feeling this could be rewritten in a more compact way and extracted into utility function  (same module) to make the component easier to read

kof (approved):
works! great job

apps/builder/app/builder/features/style-panel/sections/transitions/transition-timing.tsx
JayaKrishnaNamburu: I will extract the filtering alone into a function and then make the changes. Thought of it, but missed it refactoring. Will change and push. And then move to the rest two inputs and we can close the transition section 👍

taylornowotny (approved):
Approving since this is already approved, but there are some thing to point out:

- Code text not saving unless your hit ENTER (noted in discord)
- UI is not consistent with the design. I will not review the design until you tell me it's ready, but please make sure every detail is consistent with the design before you say it's ready, to avoid a lot of feedback.

PR #2573, opened 2023-11-06
feat: add variables to settings panel

TrySound:

Here integrated variables into props. Now it can be tested considering instances tree.

Screenshot 2023-11-06 at 13 10 33

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/controls/boolean.tsx
+++ b/apps/builder/app/builder/features/settings-panel/controls/boolean.tsx
@@ -1,10 +1,11 @@
-import { Grid, Switch, theme, useId } from "@webstudio-is/design-system";
+import { Box, Grid, Switch, theme, useId } from "@webstudio-is/design-system";
 import {
   type ControlProps,
   getLabel,
   Label,
   RemovePropButton,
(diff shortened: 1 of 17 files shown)


kof:
- [ ] should we use `==` in expressions? I guess its a little more user friendly from high level, since we compare primitives there is probably no problem?

kof:
- [ ] shouldn't we somehow allow the string value to be without quotes?

Screenshot 2023-11-06 at 11 17 33

TrySound:
>  shouldn't we somehow allow the string value to be without quotes?

This is expression without dependencies. Can be any json-compatible type. Will add objects/arrays support later.

TrySound:
> should we use == in expressions? I guess it's a little more user friendly from high level, since we compare primitives there is probably no problem?

both are supported

kof:
> both are supported

I know, I think its better to show usage of == in our cases too, so when they look into how our components work, people are less confused about what === means

kof (approved):

PR #2581, opened 2023-11-07
feat: Add UI handling for transition timing and delay attributes

JayaKrishnaNamburu:
- Adds `transition-duration` and `transition-delay` input boxes.
- Adds tooltips for all the UI actions for transitions.


JayaKrishnaNamburu:
@taylornowotny and @istarkov you test and review the UI for the changes.

kof (approved):
haven't checked if everything is according figma design, but I tested and haven't found bugs

taylornowotny:
Great work @JayaKrishnaNamburu ! Really appreciate your help here. It's almost ready but there are a few issues.

- Easing property is initially focused, but initial focus should be on the first property which is Property.
- Padding should be 16px horizontal and 8px vertical (screenshot)
Screenshot 2023-11-08 at 9 42 24 AM

- Gap between rows is 4px, not 8
- Controls + Labels should be wrapped in a 32px fixed height row and vertically centered like in the design.
Screenshot 2023-11-08 at 9 54 27 AM
Screenshot 2023-11-08 at 9 43 36 AM

- Controls are 128px fixed width in the design
Screenshot 2023-11-08 at 9 55 02 AM

- Menu items should not have a dash between words 
Screenshot 2023-11-08 at 9 46 39 AM

- No scroll bar in select menu - see other select menus for example of what to do here 
Screenshot 2023-11-08 at 9 48 24 AM

- The “Easing” control should be a select button, not a combo box.
- Labels should not extend past the length of the word. Fixing this will make the tooltips align better over the word.
Screenshot 2023-11-08 at 9 55 57 AM

taylornowotny:
(bug) If you type a value that has no results in a combo box, you can’t open the select menu any more.


https://github.com/webstudio-is/webstudio/assets/86495896/cb3133ca-c0be-40e7-bf2f-3ab8883a2981

taylornowotny (changes requested):
See UI feedback

Otherwise, great work! 🙏

JayaKrishnaNamburu:
Thanks for the detailed review on design Taylor, I need to be honest. Missed a lot of spacing stuff, which I was not able to recognise unless you pointed it out 😅.

taylornowotny:
> Thanks for the detailed review on design Taylor, I need to be honest. Missed a lot of spacing stuff, which I was not able to recognise unless you pointed it out 😅.

Not a problem, that's why we have feedback :) Thanks again!

taylornowotny:
Yes it should be a Select box. If the user wants custom easing, that’s why we have the text area.

Please add a “Custom” option to the bottom of the easing menu (no category) and show that in the select box when the user modifies the easing via the text area. If that wasn’t in the docs I apologize for overlooking it. 

Make sense?

JayaKrishnaNamburu:
Yeah sounds good, will make the changes 👍🏽

kof:
@JayaKrishnaNamburu Basically when custom easing doesn't match any of the predefined one - then we say "Custom" in the select box

JayaKrishnaNamburu:
@taylornowotny i replaced the timing with select now. And fixed the spacings. Let me know if things are looking good now 😄

taylornowotny:
- Hmm, Custom option looks like its part of the Ease In Out category when it's at the very bottom of the list. Please put Custom at the bottom of the Default category instead, thank you.

- As for the property menu, these menu items should be checkbox items just like the Easing menu. As you can see in the screenshot, "Background" option is selected but you don't see a check on the Background menu item.
Screenshot 2023-11-13 at 10 48 30 AM

- These buttons are misaligned on Property, Duration, and Delay
Screenshot 2023-11-13 at 10 50 07 AM
Screenshot 2023-11-13 at 10 50 57 AM

- Remove tooltips on inputs. Only include the tooltips from the design please.
Screenshot 2023-11-13 at 10 51 46 AM

- gaps and padding are not consistent with the design. Following up about this in DMs.

taylornowotny:
Also, the Easing menu has this nice scroll thing at the bottom - why doesn't the Property menu have this?
Screenshot 2023-11-13 at 10 54 33 AM
Screenshot 2023-11-13 at 10 55 21 AM

JayaKrishnaNamburu:
- gaps are correct as much as i can notice from the design
- The scroll down button is only available in `radix`. But the combobox is from downshift which doesn't give something similar to that unfortunately.
- tooltips comes by default for all the inputs by default. Should they still be removed ?

taylornowotny:
> * gaps are correct as much as i can notice from the design

I have updated the layout to make it easier to read. The Anatomy section of the docs is changed. Please **inspect the layout carefully in figma using dev mode**, then correct the spacing and gaps.  

>     * The scroll down button is only available in `radix`. But the combobox is from downshift which doesn't give something similar to that unfortunately.
Do we have a solution for this @TrySound @kof ? Whether we use a scroll down button or a scroll bar doesn't matter to me but we should be consistent.

>     * tooltips comes by default for all the inputs by default. Should they still be removed ?
Yes please, thank you. We never show tooltips on inputs.

kof:
@JayaKrishnaNamburu @taylornowotny in case of combobox it will have to be a scroll bar

taylornowotny:
> @JayaKrishnaNamburu @taylornowotny in case of combobox it will have to be a scroll bar

In that case can we change the select box menu to scroll bar for consistency please? I'll make an issue if you need it

kof:
@taylornowotny technically possible, I wouldn't do it though, select has specific behavior designed by radix for a reason, also our select boxes should be rather consistent than trying to make consistent completely different components. Combobox and select are technically so different there is maybe 1% similarity

taylornowotny:
@kof These are two different patterns solving the same problem - definitely not a good thing. Works against principles of simplicity and ease of use. 

The main functionality of these components doesn't need to be affected. Just need to use the same scroll pattern.

I understand if it's not a priority, but it should be fixed eventually.

kof:
@taylornowotny I didn't have time to dig into the decision making from radix in case of select, to understand why they use a button instead of a scroll, so yeah, first needs to be fully understood.

General difference is combobox is searchable and allows giant amount amount of items, select is for short lists. I have a feeling radix is solving particular ux issues with that button instead of scroll. Until we looked into that, we shouldn't make any assumptions.

taylornowotny:
@JayaKrishnaNamburu I gave the new UI its own space. Now it should be easier to inspect.


This is what you should see in dev mode now:
Screenshot 2023-11-15 at 10 50 49 AM

taylornowotny (approved):
Perfectly matching the design now! Thanks @JayaKrishnaNamburu for your work on this and for sticking with it until the end 🚀🙏⭐️

Issue #2523, opened 2023-10-24
Box Shadow UI


taylornowotny:
<img width="485" alt="Screenshot 2023-10-24 at 11 10 28 AM" src="https://github.com/webstudio-is/webstudio/assets/86495896/5f3cb148-e1b2-419a-ae3a-b50426191d9e">

The Box Shadow docs have been updated. There is now a design for the menu so users can adjust their box shadows without code. Please read the new docs carefully to notice all the changes.

Most important: "The code and the UI should be synced. This will be very valuable for the user. If valid code is pasted, the UI should change to reflect the pasted values. If the UI values are changed, the code text should change to reflect the UI values."

Figma: https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=4512-49938&mode=design

kof:
@lalitkumawat1m this is a quite complex one, requires knowledge of many design system parts + CSS parser. Would you like to start contributing with something simpler?

PR #2587, opened 2023-11-09
feat: Box shadow property extractor

JayaKrishnaNamburu:
## Description

This is the first one from the `box-shadow` UI series. Starting with the extracting individual resources from the `TupleValue` 

Related to #2523


packages/css-data/src/property-parsers/box-shadow-property-extractor.test.ts
```diff
@@ -0,0 +1,125 @@
+import { describe, expect, test } from "@jest/globals";
+import { extractBoxShadowProperties } from "./box-shadow-property-extractor";
+
+describe("extractBoxShadowProperties", () => {
+  test("parses and extracts values from 5em 5em 2em 2em #0000", () => {
+    const { inset, blur, spread, color } = extractBoxShadowProperties({
```
TrySound, line 6: This is more like parser and that thing producing tuple is tokenizer

JayaKrishnaNamburu: Should i change the name to `tokenizeBoxShadowTuple` or something else ?

packages/css-data/src/property-parsers/box-shadow-property-extractor.ts
```diff
@@ -0,0 +1,51 @@
...
+        return false;
+      }
+
+      return true;
+    }
+  );
+
+  const [offsetX, offsetY, blur, spread] = properties.filter(
+    (item): item is UnitValue => item.type === "unit"
+  );
```
istarkov, line 41: Based on spec 
```
<shadow> = 
  <color>?                                   &&
  [ <length>{2} <length [0,∞]>? <length>? ]  &&
  inset?                                     
```

So color can't be between length's same as inset

This `2px 2px rgba(255, 0, 0, 1) 2px 1px ` should not be parsed as 
color = rgba(255, 0, 0, 1)

istarkov (approved):
Anyway even algorith is wrong I don't think it's somehow seriously affects the final result ;-) Can be fixed, can be merged as is
In case of fix please add tests

packages/css-data/src/property-parsers/box-shadow-property-extractor.ts
JayaKrishnaNamburu: Yes, but in this case. The `proeprty-parser/box-shadow` take care of all such errors and generates this tuple. So, it's safe the tuple that is generated here is always a valid one. Or else the step where users enter such invalid `2px 2px rgba(255, 0, 0, 1) 2px 1px ` entries will get errors.

This step was just to assign them independently 👍

istarkov: The idea of functions is to not assume how the system around are working ;-)

JayaKrishnaNamburu: In real-word use-case. It can be something like

```js
const { blur, spread } = extractBoxShadowProeprties(parseBoxShadow("2px 2px rgba(255, 0, 0, 1) 2px 1px)
```

Something like this.

JayaKrishnaNamburu: I get it, we need to make them pure without assumptions. But, just curios. If i can find the spec with index matching and throw errors 👍

istarkov: specs are unreadable disaster ;-)

TrySound (approved):

PR #2605, opened 2023-11-12
feat: support generating collection

TrySound:

Here first part for collection support. Collections are generated as

```tsx
list.map((item, index) => <Fragment key={index}></Fragment>)
```

Added new "parameter" prop type to represent "item" which always refers to variable. "parameter" reflects same semantics as js functions where in function call it is argument `fn(1, 2, 3)` and in function definition it is parameter `function fn(a, b, c) {}`.

Collection children are always wrapped with fragment to let having more than one child and always pass key to avoid react warnings.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/component-generator.test.ts
+++ b/packages/react-sdk/src/component-generator.test.ts
@@ -12,6 +12,7 @@ import {
   generateJsxElement,
   generatePageComponent,
 } from "./component-generator";
+import { collectionComponent } from "./core-components";
 
 const clear = (input: string) =>
   stripIndent(input).trimStart().replace(/ +$/, "");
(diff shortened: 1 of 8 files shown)


packages/react-sdk/src/component-generator.test.ts
```diff
@@ -548,3 +620,64 @@ test("generate page component with variables and actions", () => {
...
+          instanceId: "list",
+          name: "item",
+          type: "parameter",
+          value: "dataSourceItem",
+        }),
+      ]),
+      indexesWithinAncestors: new Map(),
+    })
+  ).toEqual(
+    clear(`
```
kof, line 668: why not to use just inline snapshot?

kof (approved):

packages/react-sdk/src/component-generator.test.ts
TrySound: Until prettier v3 is supported very annoying to fix formatting manually all the time.

PR #2607, opened 2023-11-13
fix: Blurry canvas on windows

kof:
## Description

We have experimented quite a bit before and now apparently on windows the canvas is still blurry and removing all the hacks apparently solves it. All of this is according to @itsNintu because he is the only one with windows on the team, I don't see a good way to test it, because the blur is very subtle and you won't clearly see it remote browsers.

## Steps for reproduction

1. open builder, resize canvas within workspace (100% scale)
2. resize canvas bigger than workspace, so that scale is lower than 100%
3. see that at no point in time canvas is blurry

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/workspace/workspace.tsx
+++ b/apps/builder/app/builder/features/workspace/workspace.tsx
@@ -27,7 +27,6 @@ const workspaceStyle = css({
 
 const canvasContainerStyle = css({
   position: "absolute",
-  //transformStyle: "preserve-3d",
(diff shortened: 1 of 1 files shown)


itsNintu:
I can confirm this version renders a sharper canvas than what's currently live in production.

istarkov (approved):

PR #2619, opened 2023-11-16
test: Custom validator css input

JayaKrishnaNamburu:
## Description

The `InputField` is using `css-tree` to parse and validate the inputs by default using css properties. When using the `CSSValueInput` there is a property prop that takes care of it when passed.

```js
value = parseIntermediateOrInvalidValue(property, {
  type: "intermediate",
  value: `${value.value}`,
  unit: value.unit,
});
```
https://github.com/webstudio-is/webstudio/blob/main/apps/builder/app/builder/features/style-panel/shared/css-value-input/css-value-input.tsx#L31

But from the latest designs of `transitions` and `box-shadows`. The UI allows users to set the values individually. This worked best during transitions. Because, for every grouped transition property. They are independent way to set them like  `transitionDuration`, `transitionDelay` etc. For `box-shadow` it needs a minimum of two values.
https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#values

And from the design proposal for box-shadow. 
shadow
 users should be able to set them individually. Which becomes invalid when we try to set `offsetX` and `offsetY` individually. The other way i could think of is to pass a custom validator instead of using a default one. And this is the one of the approach. To pass a custom validator to test the theory and some inputs 😄 

## Testing
- cd `app/builder`
- `pnpm run storybook:dev`
- You can check the custom validator story. Here is a localhost link (http://localhost:6006/?path=/story/builder-features-style-panel-shared-css-value-input--with-custom-validator)
- This will allow to pass single `offsetX`.
- Here is the story details https://github.com/webstudio-is/webstudio/blob/custom-validator-css-input/apps/builder/app/builder/features/style-panel/shared/css-value-input/css-value-input.stories.tsx#L54-L85


apps/builder/app/builder/features/style-panel/shared/css-value-input/css-value-input.stories.tsx
```diff
@@ -1,15 +1,91 @@
...
+            type: "unit",
+            value: parseFloat(value),
+            unit,
+          };
+        }
+
+        return {
+          type: "invalid",
+          value,
+        };
```
JayaKrishnaNamburu, line 83: @istarkov @kof this is something i am just giving a try. What are your thoughts on this. The main reason is, `box-shadow` property doesn't work independently without multiple values. But the UI allows users to set them independently. Should there be a custom component, or re-use the existing one by passing custom validators something like this ??

Opened PR to test the theory. If sounds good, i can continue. Or else let's close and explore the ideas 👍

apps/builder/app/builder/features/style-panel/shared/css-value-input/unit-select-options.ts
```diff
@@ -29,6 +30,17 @@ const preferedSorting = [
 
 const visibleLengthUnits = ["px", "em", "rem", "dvw", "dvh"] as const;
 
+/*
+The boxShadow property is interdependent on other values of the proeprty,and cannot function in isolation.
```
istarkov, line 34: I would just add property to css-value-input to not show/add keywords
like 
showKeywords  = false

istarkov: This will remove the need of this comment and boxShadow check

apps/builder/app/builder/features/style-panel/shared/css-value-input/css-value-input.stories.tsx
```diff
@@ -1,15 +1,91 @@
...
+      }
+      parseIntermediateOrInvalidValue={(
+        _: StyleProperty,
+        styleValue: IntermediateStyleValue | InvalidValue
+      ): StyleValue => {
+        const value = styleValue.value.trim();
+        const unit = "unit" in styleValue ? styleValue.unit ?? "px" : "px";
+
+        /* Box shaodw needs atleast two values, so we are passing custom validator */
+        let styleInput = parseCssValue("boxShadow", `${value}${unit} 0`);
```
istarkov, line 62: can't we just check any unit value here like `width` using default parseBlabla... ?

istarkov: And if yes, we can just pass fake property name instead of all the code above ? 
No need in abstraction like above in a single edge case

JayaKrishnaNamburu: That's actually the first idea in my mind, but i was just not sure if passing a `fake` property with same defaults is allowed 😄. If allowed, yeah that's the easiest, cleanest approach 🚀

JayaKrishnaNamburu:
Closing this, as the idea of using a fake property to enable the checks for individual css properties is easiest, and cleanest now. Here is an example that is in progress.
https://github.com/webstudio-is/webstudio/pull/2706/files#diff-a437f51c104036bb3a2a1e33c5f5613402335d26290a759587068aa6869c2fa8R152-R161

```jsx
<CssValueInputContainer
  key="boxShadowBlur"
  property="borderTopWidth"
  label="BoxShadow Blur"
  styleSource="local"
  keywords={[]}
  value={blur ?? { type: "unit", value: 0, unit: "px" }}
  setValue={(value) => handlePropertyChange({ blur: value })}
  deleteProperty={() => {}}
/>
```

PR #2628, opened 2023-11-18
refactor: compute props contextually

TrySound:

Here refactored props values computing to reuse in both canvas and settings panel and support future collections which need to provide different values for same variables.

Grouped whole logic and covered with tests.

## Testing

- check image src is working
- check link switch page in preview

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/props-section/props-section.tsx
+++ b/apps/builder/app/builder/features/settings-panel/props-section/props-section.tsx
@@ -16,10 +16,12 @@ import {
 } from "@webstudio-is/design-system";
 import { decodeDataSourceVariable } from "@webstudio-is/react-sdk";
 import {
+  $propValuesByInstanceSelector,
   dataSourceVariablesStore,
   dataSourcesStore,
   propsIndexStore,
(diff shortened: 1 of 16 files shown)


apps/builder/app/builder/features/settings-panel/props-section/props-section.tsx
```diff
@@ -207,6 +209,20 @@ export const PropsSection = (props: PropsSectionProps) => {
   );
 };
 
+const getPropTypeAndValue = (value: unknown) => {
```
kof, line 212: if value is always just the value, you could just have a function that: 
```js
const type = inferValueType(value)
const smth = {type, value}
```

kof (approved):

apps/builder/app/builder/features/settings-panel/props-section/props-section.tsx
TrySound: need a pair to pass to union

PR #2630, opened 2023-11-18
fix: breakpoints are not registered for undo/redo

scr2em:
## Description

https://github.com/webstudio-is/webstudio/issues/2613

## Steps for reproduction

1. change the breakpoint.
2. try to undo/redo. 

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/breakpoints/breakpoints-popover-toolbar-button.tsx
+++ b/apps/builder/app/builder/features/breakpoints/breakpoints-popover-toolbar-button.tsx
@@ -30,14 +30,14 @@ export const BreakpointsPopoverToolbarButton = forwardRef<
   const scale = useStore(scaleStore);
   const breakpoint = useStore(selectedBreakpointStore);
   const [canvasWidth] = useCanvasWidth();
-  if (breakpoint === undefined || canvasWidth === undefined) {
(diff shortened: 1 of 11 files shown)


kof:
Seems like there was some kind of misunderstanding. Upon more testing we just discovered that undo/redo actually works if you look at the top bar indication. The problem is with the breakpoints editor input. Additionally we discovered undo-redo doesn't work in style panel inputs correctly in 2 cases:
1. when using arrow up/down, then undo
2. when using scrub, then undo

@TrySound is looking into this right now

scr2em:
@kof ops, what about the topbar ? when I select another breakpoint and  cmd + z, it doesn't undo. what am I missing here ?
image

kof:
> @kof ops, what about the topbar ? when I select another breakpoint and cmd + z, it doesn't undo. what am I missing here ?

We had a communication issue, when you click on a breakpoint, we never intended to make it undoable, undoable are only things that result in a change of data. Selecting a different breakpoint doesn't change data. 

I was testing the behavior when changing  the value of the breakpoint, e.g. 1280 change to 1230. And there I saw that undo didn't work but the reason is a bug in the undo/redo behavior we have with the inputs

https://github.com/webstudio-is/webstudio/issues/2631

PR #2642, opened 2023-11-22
fix: Incorrectly quoted URLs in CSS

istarkov:
## Description

Even we have this
https://github.com/webstudio-is/webstudio-builder/blob/9af97656af2e06a789f416bd93e1e1d0e756f063/packages/asset-uploader/src/utils/sanitize-s3-key.ts#L26

Additionally seems like best practice is to add backslashes for `"` and `\` symbols, even this symbols are impossible having sanitize we are using. (In case of sanitizer change everything can break)

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/fonts/src/get-font-faces.ts
```diff
@@ -16,7 +16,8 @@ export type FontFace = {
   fontStretch?: string;
 };
 
-const sanitizeCssUrl = (str: string) => str.replace(/["\\]{1}/g, "\\$&");
+// Use JSON.stringify to escape double quotes and backslashes in strings as it automatically replaces " with \" and \ with \\.
+const sanitizeCssUrl = (str: string) => JSON.stringify(str);
```
kof, line 20: lol why bother with an extra function then

kof (approved):

packages/fonts/src/get-font-faces.ts
istarkov: JSON.stringify looks strange, and it looks like beautiful hack for me. Just to give name to this hack

PR #2654, opened 2023-11-24
fix: show active style states as remote

TrySound:
Succeeds https://github.com/webstudio-is/webstudio/pull/2136 in more simple way. Active states are preresolved for everything except local styles. And local styles shown as overwritten when defined state is active.

Fixes https://discord.com/channels/955905230107738152/1177028091227418654

## Steps for reproduction

1. create form component
2. choose error state
3. set background red
4. go to error state in style source menu
5. should be "remote" label
6. set background green
7. go to stateless style source
8. reset background color
9. should be "overwritten" label

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/property-name.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/property-name.tsx
@@ -37,11 +37,13 @@ import {
 import { humanizeString } from "~/shared/string-utils";
 import { getInstanceLabel } from "~/shared/instance-utils";
 import { StyleSourceBadge } from "../style-source";
+import type { WsComponentMeta } from "@webstudio-is/react-sdk";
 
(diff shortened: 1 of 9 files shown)


kof:
- [ ] seems like the local token color in the tooltip should be blue, not purple?

image

- [ ] when Error state is chosen, value comes from doesn't contain Local on base breakpoint, but if you switch to a different breakpoint it contains Local

image

kof:
Overwritten label is right, but we should probably also tell user which state the value is coming from in that tooltip?

image

TrySound:
a bit out of scope and design, fixed for remote label though
> Overwritten label is right, but we should probably also tell user which state the value is coming from in that tooltip?

kof (approved):

Issue #2615, opened 2023-11-15
CLI–exported project has images pointing to Vercel regardless of deploy target


sibsfinx:
When initializing via `npx webstudio` and choosing `netlify-edge-functions`, I still get Vercel image optimization path:
```
vctr-webstudio-test [main●] % ag vercel
app/constants.mjs
20:  // https://vercel.com/blog/build-your-own-web-framework#automatic-image-optimization
22:    "/_vercel/image?url=" +
```
This makes images missing when building static, e.g. with https://github.com/jacob-ebey/remix-ssg.
I'm going to use it on other services rather than Vercel, so It makes tricky to build and deploy with working images.

Is there any way to configure this? Anything I could help with a PR maybe?

JayaKrishnaNamburu:
I took a look, the cli is currently using the `vercel` as the default template. So, by default the exported projects are generating `.vercelignore` file and the image loaders for vercel. 
Should there be a bare-bones static template ?

JayaKrishnaNamburu:
@sibsfinx instead of creating a new template. There is a workaround too. Since the default template is `vercel` the default vercel files are generated everytime. But you can pass `--template` option so the cli uses that instead of using vercel. Here is a example of the custom template usage.

https://github.com/webstudio-is/webstudio/blob/main/fixtures/webstudio-custom-template/package.json#L12

```sh
webstudio build --template ./custom-template
```

JayaKrishnaNamburu:
@kof this can be closed now. Or should this be kept open till the updates are released ?

PR #2660, opened 2023-11-27
fix: Display error messages when a template is not passed for building a project

JayaKrishnaNamburu:
## Description

fixes #2538 and #2615

Don't use `vercel` as default template while exporting from cli

## Code Review

## Before requesting a review

- [x] made a self-review

## Before merging

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -20,14 +20,14 @@ export const buildOptions = (yargs: CommonYargsArgv) =>
     })
     .option("template", {
       type: "string",
-      describe: `Template to use for the build [choices: ${PROJECT_TEMPALTES.toString()}]`,
+      describe: `Template to use for the build [choices: ${PROJECT_TEMPALTES.join(
+        ", "
+      )}]`,
     });
 
 // @todo: use options.assets to define if we need to download assets
 export const build = async (
-  options: StrictYargsOptionsToInterface<typeof buildOptions> & {
-    template: string;
-  }
+  options: StrictYargsOptionsToInterface<typeof buildOptions>
 ) => {
   try {
     await access(LOCAL_DATA_FILE);
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -25,7 +25,9 @@ export const buildOptions = (yargs: CommonYargsArgv) =>
 
 // @todo: use options.assets to define if we need to download assets
 export const build = async (
-  options: StrictYargsOptionsToInterface<typeof buildOptions>
+  options: StrictYargsOptionsToInterface<typeof buildOptions> & {
+    template: string;
+  }
 ) => {
   try {
     await access(LOCAL_DATA_FILE);
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -20,7 +20,6 @@ export const buildOptions = (yargs: CommonYargsArgv) =>
     })
     .option("template", {
       type: "string",
-      default: "vercel",
       describe: `[Experimental] Template to use for the build [choices: ${PROJECT_TEMPALTES.toString()}]`,
     });
 
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -20,7 +20,7 @@ export const buildOptions = (yargs: CommonYargsArgv) =>
     })
     .option("template", {
       type: "string",
-      describe: `[Experimental] Template to use for the build [choices: ${PROJECT_TEMPALTES.toString()}]`,
+      describe: `Template to use for the build [choices: ${PROJECT_TEMPALTES.toString()}]`,
     });
 
 // @todo: use options.assets to define if we need to download assets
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -20,7 +20,6 @@ export const buildOptions = (yargs: CommonYargsArgv) =>
     })
     .option("template", {
       type: "string",
-      default: "vercel",
       describe: `[Experimental] Template to use for the build [choices: ${PROJECT_TEMPALTES.toString()}]`,
```
TrySound, line 23: Should it still be called experimental?

kof:
Is this going to throw an error when user didn't use the init and didn't call build with a template either? or are you going to try  to deploy vanilla template without errors and get other errors, similarly to the problem we got now with vercel being default?

kof:
Solution I think we discussed was NOT having a default template and throwing an error that says exactly that there is no template and user needs to run build with a template

JayaKrishnaNamburu:
> Is this going to throw an error when user didn't use the init and didn't call build with a template either? or are you going to try to deploy vanilla template without errors and get other errors, similarly to the problem we got now with vercel being default?

Yes, the current default is the remix starter. So it doesn't have any platform specific defaults. And they can be modified and deployed to any platforms.  I feel it's much better than throwing error when template is not passed right. Do we still need to throw error and not allow without `--template` flag ?

If the requires result is to throw error when users run
```sh
webstudio build
```

without any template. Then we need some modifications, because currently the cli doesn't store the template values in config.

JayaKrishnaNamburu:
Ok, so the `cli` can throw a error when users are running `build` without passing any template. 
Eg:

- While running
```sh
webstudio build 
```
we can show the error.
```sh
webstudio build
Error:
 Please choose a template from the list of supported templates.
 Please check webstudio --help for more details
    at prebuild (file:///Users/jk/Documents/webstudio/webstudio/packages/cli/lib/cli.js:351:11)
    at Object.build [as handler] (file:///Users/jk/Documents/webstudio/webstudio/packages/cli/lib/cli.js:681:
```

Which restricts users to run `build` without passing a template. But there is one more use-case to take care. This is from the `init` flow.

During the flow, there is a question like

```sh
✔ Would you like to create a project folder? (no to use current folder) … yes
✔ Please enter a project name … testing
✔ Please paste a link from the Share dialog in the builder … https://apps.webstudio.is/builder/2518abc9-9483-4092-bdce-745e27907de3?authToken=7253b71f-059c-4ad7-811b-3c1e33b198a0&mode=preview
Saved credentials for project 2518abc9-9483-4092-bdce-745e27907de3.
  You can find your config at /Users/jk/Library/Preferences/webstudio-nodejs/webstudio-config.json

? Would you like to setup a deploy target? › (y/N)
```

The flow breaks if the users select `N` for the option. So, should there be a re-consideration of the flow to avoid there too ?

kof:
@JayaKrishnaNamburu what is the problem init interrupts there? There was no build yet, right?

JayaKrishnaNamburu:
Yes, the flow get's interupted there. If the users respond with `N`. As there is no template choosen. When the cli runs the build step during the flow

```js
const { installDeps } = await prompt({
      type: "confirm",
      name: "installDeps",
      message: "Would you like to install dependencies? (recommended)",
      initial: true,
    });
    shouldInstallDeps = installDeps;
  }

  await sync({ buildId: undefined, origin: undefined, authToken: undefined });

  await build({
    ...options,
    ...(projectTemplate && { template: projectTemplate }),
  });
```

https://github.com/webstudio-is/webstudio/blob/main/packages/cli/src/commands/init-flow.ts#L94

It breaks to through the error that there should be some template selected. So, i am just curious how to handle this UX. Should be question be answered with some defaults. Or should the case is handled in some different way or something. Open for options 👍

kof:
Maybe worth pointing out that you are interrupting the init and will need to do the manual steps and print the --help doc where they can read about the steps?

JayaKrishnaNamburu:
ok sounds good, will add both the changes 👍

JayaKrishnaNamburu:
Hey i changed a little, and made the `--template` to be a mandatory requirement. And so the question of `Do you want to set a deploy` becomes obsolete. So removed it. Please suggest your changes, i am not sure what's right here so. Here is a quick recording of both the flows now. You can check it here instead of running it locally.

https://github.com/webstudio-is/webstudio/assets/11075561/90004c23-f5c1-430f-971b-a97f67cf2072



https://github.com/webstudio-is/webstudio/assets/11075561/82398f0e-4335-47a8-a064-34da14afb237

packages/cli/src/config.ts
```diff
@@ -53,6 +53,7 @@ export const jsonToGlobalConfig = (json: unknown) => {
 export type GlobalConfig = z.infer<typeof zGlobalConfig>;
 
 export const PROJECT_TEMPALTES = [
+  "vanilla",
```
JayaKrishnaNamburu, line 56: Added new `vanilla` option to export as plain remix projects without any deploy target defaults.

kof (approved):

PR #2663, opened 2023-11-28
refactor: reimplement duplicating instances

TrySound:
Part of https://github.com/webstudio-is/webstudio/issues/2516

Here added new copy implementation. There are 2 differences

- it does not change existing instances and not change their children, we will need this to copy pages
- it handles cross-project copy of portals (slots) properly, now each portal content is inserted separately if the root does not exist and skips otherwise
- much better covered with tests

## Steps for reproduction

1. try duplicating form and change form state prop
2. try duplicating form error message and change form state prop

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/instance-utils.ts
+++ b/apps/builder/app/shared/instance-utils.ts
@@ -1,4 +1,6 @@
+import { nanoid } from "nanoid";
 import { toast } from "@webstudio-is/design-system";
+import { equalMedia, type StyleValue } from "@webstudio-is/css-engine";
 import {
   type Instance,
   type Instances,
@@ -13,6 +15,8 @@ import {
   DataSources,
   Props,
   DataSource,
+  Prop,
+  Breakpoint,
 } from "@webstudio-is/sdk";
 import { findTreeInstanceIdsExcludingSlotDescendants } from "@webstudio-is/sdk";
 import {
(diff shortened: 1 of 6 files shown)


apps/builder/app/shared/instance-utils.ts
```diff
@@ -533,6 +547,17 @@ const traverseStyleValue = (
   value satisfies never;
 };
 
+type InstancesSlice = {
```
kof, line 550: as an idea: InstancesFragment in case you like it more

apps/builder/app/shared/instance-utils.ts
```diff
@@ -687,3 +712,452 @@ export const getInstancesSlice = (rootInstanceId: string) => {
     assets: slicedAssets,
   };
 };
+
+export const findAvailableDataSources = (
```
kof, line 716: shouldn't have test since exported?

packages/react-sdk/src/core-components.ts
```diff
@@ -1 +1,3 @@
+export const portalComponent = "Slot";
```
kof, line 1: I am fine with renaming it btw to Portal, whenever the time is right

kof (approved):

packages/react-sdk/src/core-components.ts
TrySound: Would be more like ws:portal

apps/builder/app/shared/instance-utils.ts
TrySound: More like slice of cheese with holes, fragment we use for component name already

PR #2671, opened 2023-12-06
refactor: support parameter data source for collection item

TrySound:

Here added dataSource.type = 'parameter' without initial value because value is always provided by context. It let us fix a few issues with variables without complicating logic.

- prevent deleting parameter variable (collection item) in variables panel
- allow changing only name of parameter variable
- show only name in the list of variables
- fix removing expression from collection data
- hide item parameter from collection variable list
- prevent editing parameter value from props
- make collection unstylable

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/controls/combined.tsx
+++ b/apps/builder/app/builder/features/settings-panel/controls/combined.tsx
@@ -30,6 +30,10 @@ export const renderControl = ({
     return;
   }
 
+  if (meta.control === "json" && (prop === undefined || prop.type === "json")) {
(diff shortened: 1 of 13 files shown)


istarkov (approved):

kof:
I don't think I understand what parameter data source is.

TrySound:
@kof See collection item cannot be a variable because in our case variables have value and updater (setValue) and item cannot have updater. To successfully distinct them in ui logic I introduced data source with "parameter" type similar to prop with "parameter" type.

PR #2674, opened 2023-12-08
feat: support page url pattern in builder

TrySound:

Here extended page "path" field to accept [url pattern](https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API) compatible syntax to parametarize url with id, date, slug or anything else. Such data can be passed to backend/cms requests and use the same webstudio page for dynamic content.

Preview fields automatically added when template fields appear in path.

- fixed json local value
- create "Page params" parameter data source when template is used in path
- added preview fields to compile path template and preview in ui

Screenshot 2023-12-08 at 11 35 30

## Steps for reproduction

1. create new page
2. type `/blog/:slug*` in path
3. bind expression to body id prop `$Page params<Enter>.id`
4. preview 'my-slug' in slug param

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Question: is design on the screenshot approved/designed by @taylornowotny?

kof:
- [ ] we need to explain (probably in a tooltip?) how all of this works, otherwise this is unusable

taylornowotny:
@kof No, these two new fields (ID and Slug) do not meet design standards. They aren't in figma but it should be apparent that they should be formatted like the other text fields in the page settings UI.

Fixes:
- [ ] These properties should use the same layout as the Path field above them
- [ ] Labels should be in title case like all other property labels

taylornowotny:
> * [ ]  we need to explain (probably in a tooltip?) how all of this works, otherwise this is unusable

I suggest showing the ? icon in the label for each of these properties and putting the tooltip there. I use the icon instead of putting the tt directly on the label when the tt content would be instructions for something rather than just a description of the property.
Alt: instead of a tooltip there can be a paragraph of gray text like you see elsewhere in the page settings.

TrySound:
This is a prototype hidden with flag. No design discussed yet.

@taylornowotny these two fields are auto generated from path template. Those `:id` and `:slug` params are the source of these fields. Urls may have many such params so I used horizontal layout to save space.

TrySound:
> Labels should be in title case like all other property labels

It is also important to have original case to not confuse users with accessing page params fields in expressions. For example `` `Page params`.slug ``

taylornowotny:
> > Labels should be in title case like all other property labels
> 
> It is also important to have original case to not confuse users with accessing page params fields in expressions. For example `` `Page params`.slug ``

I defer to @kof on this one

taylornowotny:
> This is a prototype hidden with flag. No design discussed yet.
> 
> @taylornowotny these two fields are auto generated from path template. Those `:id` and `:slug` params are the source of these fields. Urls may have many such params so I used horizontal layout to save space.

All good, I thought so. Only responded becuase asked. Please make the requested layout adjustment whenever you can.

istarkov (approved):

PR #2682, opened 2023-12-13
feat: support page path params in site compiler

TrySound:

Here remix route params are adapted to URLPattern format and passed to generate page component as prop.

For this added 2 remix specific utilities
- convert subset of url pattern to remix route
- generate utility to adapt remix params to URLPattern params for example for `/*` `params['*']` is replaced with `params[0]` and for `/:slug*` `params['*']` is replaced with `params['slug']`

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/remix.test.ts
+++ b/packages/react-sdk/src/remix.test.ts
@@ -1,32 +1,34 @@
 import { expect, test } from "@jest/globals";
-import { generateRemixParams, getRemixRoute } from "./remix";
+import { generateRemixParams, generateRemixRoute } from "./remix";
 
 test("convert home page to remix route", () => {
-  expect(getRemixRoute("")).toEqual("_index.tsx");
-  expect(getRemixRoute("/")).toEqual("_index.tsx");
+  expect(generateRemixRoute("")).toEqual("_index.tsx");
+  expect(generateRemixRoute("/")).toEqual("_index.tsx");
 });
 
 test("convert path to remix route", () => {
-  expect(getRemixRoute("/blog")).toEqual("[blog]._index.tsx");
-  expect(getRemixRoute("/blog/my-introduction")).toEqual(
+  expect(generateRemixRoute("/blog")).toEqual("[blog]._index.tsx");
+  expect(generateRemixRoute("/blog/my-introduction")).toEqual(
     "[blog].[my-introduction]._index.tsx"
   );
 });
 
 test("convert wildcard to remix route", () => {
-  expect(getRemixRoute("/blog/*")).toEqual("[blog].$.tsx");
+  expect(generateRemixRoute("/blog/*")).toEqual("[blog].$.tsx");
 });
 
 test("convert named group with * modifier to remix route", () => {
-  expect(getRemixRoute("/blog/:slug*")).toEqual("[blog].$.tsx");
+  expect(generateRemixRoute("/blog/:slug*")).toEqual("[blog].$.tsx");
 });
 
 test("convert named group with ? modifier to remix route", () => {
-  expect(getRemixRoute("/:id?/:slug?")).toEqual("($id).($slug)._index.tsx");
+  expect(generateRemixRoute("/:id?/:slug?")).toEqual(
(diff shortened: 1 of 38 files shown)


packages/react-sdk/src/remix.test.ts
```diff
@@ -0,0 +1,60 @@
+import { expect, test } from "@jest/globals";
+import { generateRemixParams, getRemixRoute } from "./remix";
+
+test("convert home page to remix route", () => {
+  expect(getRemixRoute("")).toEqual("_index.tsx");
```
istarkov, line 5: `getRemixRoute` imo `get` is bad name having that you transform one route format to another

packages/react-sdk/src/remix.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (modifier === "?") {
+      return `($${name})`;
+    }
+    return `$${name}`;
+  }
+  return `[${segment}]`;
+};
+
+export const getRemixRoute = (pathname: string) => {
```
istarkov, line 20: Plz use `/**` comments to give a short description from what format to wht format you are converting with links etc on originnal docs

istarkov: i.e.
```ts
/**
 * Transformrs [URL ROUTE](link) format to [Remix](file-route-link) format
**/
export const getRemixRoute = (pathname: string) => {
```

packages/react-sdk/src/remix.ts
```diff
@@ -0,0 +1,47 @@
+const getRemixSegment = (segment: string) => {
+  if (segment === "*") {
+    return "$";
+  }
+  const match = segment.match(/^:(?<name>\w+)(?<modifier>\*|\?)?$/);
```
istarkov, line 5: I would give a comment example what we are matching here
And probably that we don't support various original patterns
i.e.
`https://example.com/jo*` in our case should throw instead of `$` isn't it?

packages/react-sdk/src/remix.ts
```diff
@@ -0,0 +1,47 @@
...
+  const name = match?.groups?.name;
+  const modifier = match?.groups?.modifier;
+  if (name) {
+    if (modifier === "*") {
+      return "$";
```
istarkov, line 10: imo should throw here

packages/react-sdk/src/remix.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (modifier === "?") {
+      return `($${name})`;
```
istarkov, line 13: https://remix.run/docs/en/main/file-conventions/routes#escaping-special-characters
Interesting that in remix spec special chars should be inside `[`, `]` and nothing is written about `(` and `)`
Imo should be tested with various chars

packages/react-sdk/src/remix.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    return `$${name}`;
```
istarkov, line 15: same, what about special chars

kof (approved):

packages/react-sdk/src/remix.ts
istarkov: Probably work in remix `[partofroute]:param` ?

packages/react-sdk/src/remix.ts
TrySound: We validate this case out in builder

packages/react-sdk/src/remix.ts
TrySound: It does not. Why should it throw? This is "*" named equivalent. See getRemixParams set the name parsed from url pattern.

packages/react-sdk/src/remix.ts
TrySound: `($lang)` is optional segment

packages/react-sdk/src/remix.ts
TrySound: What chars?

PR #2684, opened 2023-12-14
fix: implement own hotkeys

TrySound:
Fixes https://discord.com/channels/955905230107738152/1184660413326053426/1184660413326053426

Primitives exposed by react-hotkeys-hook are quite useless and buggy for us. Here added own simple implementation which should fit our needs.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/commands-emitter.ts
+++ b/apps/builder/app/shared/commands-emitter.ts
@@ -1,4 +1,3 @@
-import { isHotkeyPressed } from "react-hotkeys-hook";
 import { atom } from "nanostores";
 import { $publisher, subscribe } from "~/shared/pubsub";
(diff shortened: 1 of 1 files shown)


kof:
I am ok with moving it to our codebase, but it feels like they just have a bug we could have contributed with a fix

kof (approved):

kof (approved):
Tested duplication bug - works

TrySound:
> I am ok with moving it to our codebase, but it feels like they just have a bug we could have contributed with a fix

It's not really a but but implementation very specific to their cases which does not fit us well. I had to add workarounds for it. Here it is more straightforward, we just check current event instead of relying on race conditions.

PR #2691, opened 2023-12-14
fix: pointer-events and tooltip contents for css-sections title

### LINKED ISSUE #2665: Fix the main tooltip for Transitions
The Transition tooltip is busted

To do:
- [ ] Show tooltip when the label is hovered even when no value is applied. Please update Box Shadows, Backgrounds, and Transitions
- [ ] Add tooltip description from the design
- [ ] Tooltips on properties have a fixed width of 208px

Current
Screenshot 2023-11-30 at 10 27 42 AM

Design
Screenshot 2023-11-30 at 11 13 14 AM

Figma
https://www.figma.com/file/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?type=design&node-id=8086%3A51795&mode=design&t=KA5xOTf6y2M5HKvO-1

JayaKrishnaNamburu:
## Description

fixes #2665 

The PR fixes the missing tooltips for `box-shadow`, `transition`, and `background` sections. Since the `label` is using pointer-events as `inherit`. And the parent `div` has pointer-events of `none`. It is not triggering the mouse events for the radix tooltip to show.

## Steps for reproduction

Hover over the headings of the above properties, should show the tooltips now.

## Code Review

   detailed review (read every line)

- [ ] hi @taylornowotny, i need your help
   -  please check the tooltip contents for these three proeprties.
  
## Before requesting a review

- [x] made a self-review


packages/design-system/src/components/section-title.tsx
```diff
@@ -197,7 +197,7 @@ export const SectionTitleLabel = forwardRef(
           // Instead, we need to render the label using a div that has position:absolute and pointer-events:none
```
istarkov, line 197: See comment it was intentionally set to inherit to inherit pointer-events none

istarkov: So your code and comments are wrong. I would check why it was intentionaly set as pointer-events:none and inherit and this is not breaking that case

JayaKrishnaNamburu: But that's disabling tooltips. And I couldn't understand what exactly need to be made from the comment.

istarkov: Having comment something has been done here with intent. You are breaking this intent but leaving old comment. The easiest way is try to find PR where it was implemented probably something would be clear

istarkov:
plz rename branch name to something short like 10-15 symbols. We have a bug preventing getting real links on PR apps

istarkov:
To be clear Im not getting the issue too

So for some unknown reason we decided to have tooltips for 3 sections, and to not have tooltips on the others.

From what I understand this can break behaviour of non tooltip sections.

istarkov:
Also seems like the whole section button must be tooltipable and not the label only as before.
was:
image

should:
image

taylornowotny:
Every CSS property must have a label with a tooltip. When we have a whole
section representing one CSS property, then the only place to put that
label + tooltip is in the section title.

taylornowotny:
> Also seems like the whole section button must be tooltipable and not the label only as before. was: image
> 
> should: image

No, the tooltip should be on the label as indicated by the design. The Add (+) button should have its own tooltip.

istarkov:
> Every CSS property must have a label with a tooltip. When we have a whole
section representing one CSS property, then the only place to put that
label + tooltip is in the section title.

background is not single css property but a pack of

image

istarkov:
Anyway @JayaKrishnaNamburu Label must be a button to have an own tooltip. So isLabelButton can be extended to have this logic. As I remember it must be a button because of keyboard events.

See tooltip on keyboard focus

image

So I would revert the  code above, and will think how to make label a button ;-)

taylornowotny:
I had the "background" shorthand property in mind. Even if the properties are separated out, this section follows the appropriate UI pattern for adding multiple values in a list.

taylornowotny:
Screenshot 2023-12-14 at 7 07 58 AM

Tried to test, getting this message now in chrome. I re-installed chrome and also tried Brave and issue persists. Something broken with this deployment?

kof:
@taylornowotny yes, I was updating the message and forgot to reenable the check, we fixed it now, hold on

taylornowotny:
@JayaKrishnaNamburu The tooltips look great! Nice work on this, thank you!

One more thing that is my fault: I didn't add this tooltip in the design. This tooltip should say "Add a transition"
Can you change that for me please? 

Screenshot 2023-12-14 at 11 43 36 AM

JayaKrishnaNamburu:
Please us the new #2699 PR for all the future discussions.

PR #2697, opened 2023-12-14
feat: New tokens for text links - text styles and hover colors

taylornowotny:
New color tokens:
foreground.link-hover.contrast
foreground.link-hover.text-more-subtle
foreground.link-hover.text-subtle
foreground.link-hover.main 

Description: The color of a text link in the hover state when the normal text is the color in the token name.

New text composite tokens:
Regular Link (renamed from prev: Link)
Label Link
 Mono Link
  Mono Bold Link

Description:
Same as the named text style with added underline to indicate a link.


packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -493,28 +535,31 @@ export const color = {
   foregroundOverwrittenFlexUi: "#dc3d43",
   foregroundInversePrimary: "#ff22ae",
   foregroundLinkNotVisited: "#006adc",
```
kof, line 537: @taylornowotny this already exists as foregroundLink

packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -493,28 +535,31 @@ export const color = {
...
+  foregroundLinkHoverMain: "#656869",
```
kof, line 538: why is this called foregroundLinkHoverMain and not just foregroundLinkHover?

packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -493,28 +535,31 @@ export const color = {
...
+  foregroundLinkHoverTextSubtle: "#656869",
```
kof, line 539: Why is it called HoverText ? if you are styling a link color its always the text, why is it not "foregroundLinkHoverSubtle"

packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -493,28 +535,31 @@ export const color = {
...
+  foregroundLinkHoverTextMoreSubtle: "#656869",
```
kof, line 540: same here

packages/design-system/src/__generated__/figma-design-tokens.ts
taylornowotny: When text color is foreground.main, text links should be foreground.link-hover.main

This is the pattern for all of these colors

packages/design-system/src/__generated__/figma-design-tokens.ts
taylornowotny: It was renamed

packages/design-system/src/__generated__/figma-design-tokens.ts
taylornowotny: I don't remember where this token came from. It's old, not part of this update.

taylornowotny: Not all links are text

taylornowotny: There's a color token called "foreground.text-subtle". If we have a text link that uses foreground.text-subtle, then the hover color would be foreground.link-hover.text-subtle

packages/design-system/src/__generated__/figma-design-tokens.ts
kof: is main the default? or when is it usable

taylornowotny: Main is the default foreground color against light backgrounds, yes. It's the near-black color.

packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -493,28 +535,31 @@ export const color = {
...
+  foregroundLinkHoverContrast: "#adb1b4",
```
kof, line 541: foregroundLinkHoverMain: "#656869",
foregroundLinkHoverTextSubtle: "#656869",
foregroundLinkHoverTextMoreSubtle: "#656869",

all have the same color

taylornowotny: Correct. A choice made for clarity, since there is not easy way to say that one token works for all 3 cases.

packages/design-system/src/__generated__/figma-design-tokens.ts
kof: there is no link with non-text, so its semantically confusing

packages/design-system/src/__generated__/figma-design-tokens.ts
kof: but foregroundLink exists too

packages/design-system/src/__generated__/figma-design-tokens.ts
taylornowotny: you can ask Mark to rename both text-subtle and the link hover to not use the word text

packages/design-system/src/__generated__/figma-design-tokens.ts
kof: also it would make sense to have foregroundLink as a default, meaning not visited and then a visited one

taylornowotny: I renamed foreground.link to foreground.link-hover.main  so it should not exist any more. If it does it is a duplicated.

I did not make a 'visited' color but you can ask Mark to make it if you want.

packages/design-system/src/__generated__/figma-design-tokens.ts
kof: if main is the default color, then there mist be also " foregroundLinkMain"

kof: honestly I don't understand the entire system here

taylornowotny: We need a color change when a text link is hovered. A text link is the same color as the rest of the text, so links do not need a different default color, only a different hover color.

So I identified the colors that are used for text in our UI, and made a hover color for each of them.

When text is foreground.main, links will also be foreground.main and links in the hover state will be foreground.link-hover.main

packages/design-system/src/__generated__/figma-design-tokens.ts
```diff
@@ -320,6 +320,48 @@ export const typography = {
...
+    fontSize: "12px",
+    lineHeight: "16px",
+    letterSpacing: "0.005em",
+    textTransform: "none",
+    textDecoration: "underline",
+    textIndent: "0px",
+  },
+  monoBoldLink: {
+    fontFamily: "Roboto Mono, RobotoMono, menlo, monospace",
+    fontWeight: 500,
```
kof, line 347: mono bold and mono are same font weight, is it intentional?

kof: Why do we need monoBold if we have the same value

taylornowotny: That was a mistake. Ijust fixed it in figma and pushed to github. LMK if I need to make a PR for that update

kof: yes need a pr but also need to clarify everything about the link colors, I tried to systematize and make them usable in code and it was a bad experience, nothing is clear:

seems like we need

Colors
- main (hover & visited)
- contrast (hover & visited)
- subtle (hover & visited)
- moreSubtle (hover & visited)

Variants
- regular
- label
- mono

kof: Unless we completely don't care about visited, then each color has to have default and hover color

kof: Actually I don't think we need visited colors at all

taylornowotny: Yea the color tokens could use a re-organization if you think its worth investing in that.

All those colors that you listed are tokens, they are just organized so as not to disrupt the greater system. Link hovers are not just called "hovers" becuase they are specifically for text links and nothing else. Regular text does change color on hover.

Yes you can group the link hovers with their respective normal text colors, but I made the decision to group them separately and together instead.

Current: foreground.link-hover.main   - groups link-hovers within foreground
Possible: foreground.main.link-hover - puts each link-hover within its reference color

taylornowotny:
It's not letting me create a new PR becuase I think the update that I pushed was added to this PR

packages/design-system/src/__generated__/figma-design-tokens.ts
kof: I wouldn't wawnt to toouch all colors right now but I want to understand the links and I don't

kof: This is what I see we have in total for links:

  foregroundLink: "#006adc",
  foregroundLinkVisited: "#793aaf",
  foregroundLinkNotVisited: "#006adc",
  foregroundLinkHoverMain: "#656869",
  foregroundLinkHoverTextSubtle: "#656869",
  foregroundLinkHoverTextMoreSubtle: "#656869",
  foregroundLinkHoverContrast: "#adb1b4",

kof:
> It's not letting me create a new PR becuase I think the update that I pushed was added to this PR

Update is perfect, I thought you can't update

packages/design-system/src/__generated__/figma-design-tokens.ts
taylornowotny: These I am unfamiliar with. I think @markxtang must have added them. I would say they should be deleted, but Mark should confirm first that it will not break a design that he made.

> foregroundLink: "#006adc",
> foregroundLinkVisited: "#793aaf",
> foregroundLinkNotVisited: "#006adc",

kof: is there no way to check if they are used anywhere in figma? 

I mean if there is "bus factor", what do we do?

kof: in code these 2 only used in 2 places: unsupported browser alert and publish dialog

foregroundLink: "#006adc",
foregroundLinkVisited: "#793aaf",

taylornowotny: There's not a native way, but I found a plugin that finds where styles are used. I confirmed none of these tokens are used in any designs, so I removed all 3.

I don't know why these are used in code but not in the design. You should check each of those instances to see if they need to be replaced with one of the new link-hover colors, or just leave it the same color but not connected to a token.

kof:
I still don't see the default and hover colors addressed, like I described there https://github.com/webstudio-is/webstudio/pull/2697#discussion_r1429199255

So I still don't understand the system

kof:
Also what is link vs regularLink, I found both in typography

taylornowotny:
Link was renamed to Regular Link to clarify that it is the Regular text
style + an underline

It is possible that tokens are not being removed automatically when they
are deleted or renamed in Figma. You may need to delete them manually.

Regarding the nomenclature, I’m sorry you’re having trouble understanding
it. I believe you can do it if you read my explanation carefully - I can’t
explain any better than I already did. This was the simplest organization
that was apparent to me, but if you want to organize it differently then
you are free to do so, just coordinate with Mark.

kof:
> It is possible that tokens are not being removed automatically when they
are deleted or renamed in Figma. You may need to delete them manually.

You managed to delete the other tokens, so it seems more like it wasn't deleted in figma at all?

taylornowotny:
It does not exist anymore in Figma because I renamed it. Maybe the plug-in
makes a new token when a token is renamed instead of deleting the old one.
I cant check right now but you’ll probably have to just delete it manually
regardless

kof (approved):
will send update in a separate PR

kof:
After analyzing what we need for a link component, I decided to not have special styles for it, it should just inherit from the text where it is in. We currently don't have places where we really need a "sticking out" link color.

PR #2698, opened 2023-12-15
feat: support resources in schema and compiler

TrySound:

Here added new schema type "Resources" to contain configuration of requests to services. Their first use case is referencing from variable to store loaded data.

For now added only client side generation. Later add support for server code generation as part of remix loaders.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk/src/schema/resources.ts
+++ b/packages/sdk/src/schema/resources.ts
@@ -1,0 +1,22 @@
+import { z } from "zod";
+
+const ResourceId = z.string();
+
+export const Resource = z.union([
+  // !!! this is prototype only resource type
+  // build other resources before exposing to production
+  z.object({
+    type: z.literal("getjson"),
+    id: ResourceId,
+    instanceId: z.string(),
+    // expression
+    url: z.string(),
+  }),
+  z.never(),
+]);
+
+export type Resource = z.infer<typeof Resource>;
+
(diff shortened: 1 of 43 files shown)


packages/sdk/src/schema/resources.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from "zod";
+
+const ResourceId = z.string();
+
+export const Resource = z.union([
+  // !!! this is prototype only resource type
+  // build other resources before exposing to production
+  z.object({
+    type: z.literal("getjson"),
```
kof, line 9: probably need a better type name, so if I see correctly you want different types for get, post etc and the type also is specific for a json response? 

So if you wanted to fetch text, you would create "gettext"?  as a ressource type? Almost feels like method and responseType should be separate fields there

kof (approved):
approving, but I don't understand the idea behind getjson type

packages/sdk/src/schema/resources.ts
TrySound: No, it's not about methods. It's just a configuration type. Could be grapnel, could be full blown rest api or just simple rest request.

Here type is for prototype only. We will come up with better resources when start integrating stuff.

PR #2701, opened 2023-12-16
feat: generate resource loaders

TrySound:

Here added generation of server modules with resource loaders. These loaders are invoked in remix loader and result is passed to page component.

Check fixtures and tests.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/resources-generator.test.ts
+++ b/packages/react-sdk/src/resources-generator.test.ts
@@ -49,7 +49,7 @@ test("generate getjson resource loader", () => {
       const [
       resourceName,
       ] = await Promise.all([
-      _getjson(\"https://my-json.com\"),
+      _getjson("https://my-json.com"),
       ])
       return {
       resourceName,
--- a/packages/react-sdk/src/resources-generator.test.ts
+++ b/packages/react-sdk/src/resources-generator.test.ts
@@ -1,0 +1,142 @@
+import { expect, test } from "@jest/globals";
+import stripIndent from "strip-indent";
+import { createScope, type Page } from "@webstudio-is/sdk";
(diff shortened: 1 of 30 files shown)


packages/react-sdk/src/resources-generator.test.ts
```diff
@@ -0,0 +1,142 @@
...
+          const data = await response.json()
+          return data;
+        }
+        return {
+          error: await response.text()
+        }
+      }
+      type Params = Record<string, string | undefined>
+      export const loadResources = async (_props: { params: Params }) => {
+      const [
```
kof, line 49: I know its only a test, but is there an easy way to make it look nicely formatted? Also why not use inline snapshots?

kof (approved):

packages/react-sdk/src/resources-generator.test.ts
TrySound: > I know its only a test, but is there an easy way to make it look nicely formatted?

Formatting makes code generation code more complex. Pieces of code are quite small so not completely unreadable.

> Also why not use inline snapshots?
Because jest doesn't work with prettier v3

kof: so we are not using inline snapshots at all now? I would wait for prettier support and not upgrade, inline snapshots are important we should be able to use them

TrySound: We use prettier v3 for a long time. Always better to bump it earlier to avoid massive git conflicts. Inline snapshots don't do much honestly. And output has escaped double quotes which are harder to read.

PR #2711, opened 2023-12-24
feat: add resource variables support in builder

TrySound:

Here added Value/Resource tabs to "new/edit variable" panel to support creating both value and http request based variables.

Url, headers and body are expressions with support to all available in scope value and page params variables.

Actual loading of resources in builder will be added later.

Screenshot 2023-12-24 at 13 42 34
Screenshot 2023-12-24 at 13 42 20

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/variables.tsx
+++ b/apps/builder/app/builder/features/settings-panel/variables.tsx
@@ -16,6 +16,10 @@ import {
   InputErrorsTooltip,
   InputField,
   Label,
+  PanelTabs,
+  PanelTabsContent,
+  PanelTabsList,
+  PanelTabsTrigger,
   ScrollArea,
   Separator,
   SmallIconButton,
@@ -42,6 +46,7 @@ import {
   $dataSources,
   $instances,
   $props,
+  $resources,
   $selectedInstanceSelector,
   $variableValuesByInstanceSelector,
 } from "~/shared/nano-states";
@@ -53,6 +58,7 @@ import {
   formatValue,
   formatValuePreview,
 } from "~/builder/shared/expression-editor";
+import { ResourcePanel } from "./resource-panel";
 
 /**
  * convert value expression to js value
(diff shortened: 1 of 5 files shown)


kof:
- The "(Expression)" in labels is suboptimal, we probably need some kind of icon or something minimalistic, its fine for the prototype though
- I would guess expression in body allows to access results and do something with them, but what does expression in url does?

kof (approved):

apps/builder/app/builder/features/settings-panel/variables.tsx
```diff
@@ -108,38 +115,46 @@ const saveVariable = (
...
+        value,
+      };
+      if (typeof value === "string") {
+        variableValue = { type: "string", value };
+      }
+      if (typeof value === "number") {
+        variableValue = { type: "number", value };
+      }
+      if (typeof value === "boolean") {
+        variableValue = { type: "boolean", value };
```
kof, line 137: this feels like it shold be generalized in a utility somewhere, I think I saw similar code already before

kof:
@TrySound do we want take expressions in any input in the future? in that case it could be just our standard, then we don't need to explain anywhere one can use expression

TrySound:
> I would guess expression in body allows to access results and do something with them, but what does expression in url does?

We want to parameterize with url params, some constants or access tokens. Those values are passed to generated resources in remix loader too

> do we want take expressions in any input in the future? in that case it could be just our standard, then we don't need to explain anywhere one can use expression

ideally we need a switcher like n8n has instead of popover which cary whole variable management with it. In this specific case we would prefer to have a text field by default and then switch to expression if necessary.

kof:
> We want to parameterize with url params, some constants or access tokens. Those values are passed to generated resources in remix loader too

so url could look like `"/path/" + token` ? 

> ideally we need a switcher like n8n has instead of popover which cary whole variable management with it. In this specific case we would prefer to have a text field by default and then switch to expression if necessary.

seems like we need to think this better through with Mark

TrySound:
> so url could look like "/path/" + token ?


Yes, or "/blog/" + slug

PR #2713, opened 2023-12-25
feat: CSS engine has now a Style Sheet rendering with atomic classes which will lead to a 10x smaller bundle size

kof:
## Description

CSS engine has now a Style Sheet rendering with atomic classes which will lead to a much smaller bundle size

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/css/css.ts
```diff
@@ -73,21 +81,25 @@ export const generateCssText = (data: Data, options: CssOptions) => {
...
+      breakpoint: breakpointId,
+      style,
+    };
+    // @todo add support for both regular and atomic style sheets
+    //sheet.addStyleRule(
+    //  `[${idAttribute}="${instanceId}"]${state ?? ""}`,
+    //  styleRule,
+    //  transformer
+    //);
+    const { classes } = atomicSheet.addStyleRule(styleRule, transformer);
```
TrySound, line 101: regular could have the same api and bind to classes instead of attributes.

kof: its now mostly the same api (rule, string, transformer) though in regular string is any selector, where in atomic its a selector suffix, since class selector is being generated

packages/css-engine/src/core/style-sheet-atomic.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  test("use default media rule when there is no matching one registered", () => {
+    const { classes } = sheet.addStyleRule(
+      {
+        style: style0,
+        breakpoint: "x",
+      },
+      ":hover"
+    );
+    expect(sheet.cssText).toMatchInlineSnapshot(`
```
kof, line 29: is there anything we can do to force jest to indent this properly?

packages/react-sdk/src/css/css.ts
kof: Just realized what you meant, the reason it was over the attribute so that there is no need to add a class name, since there was already an id there anyways and classname would correspond the  attribute, but I guess we could change that ....

kof: with the id its kindof nice to debug as you can know which css rule applies to which instance directly from css rule declaration, maybe we could keep the id and use it as a class: .{id} {}

kof: hmm, the same api will only work if we generate the class, not use the id, otherwise we need a different arguments interface to pass the ids

kof: we could use the same hashing as with atomic ofc

kof: hmm maybe its a good idea to extend the interface to always pass along instance id

kof: thought a bit more about it, I think its not worth it ... but we can do it any time later if we redeside

packages/css-engine/src/core/style-sheet-atomic.ts
```diff
@@ -0,0 +1,67 @@
...
+        // This property-value combination has already been added.
+        if (
+          "selectorText" in styleRule &&
+          newStyleRule.selectorText === styleRule.selectorText
+        ) {
+          break;
+        }
+        addRule = true;
+        break;
+      }
```
istarkov, line 52: This for always stops on the first rule, why you use for at all?

packages/css-engine/src/core/style-sheet-atomic.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      if (addRule) {
+        newStyleRule.onChange = this.#onChangeRule;
```
istarkov, line 55: Does it possible that newStyleRule doesn't have onChange at all. For me it looks like contructor parameter. No need to expose omChange property at all

packages/css-engine/src/core/style-sheet-regular.ts
```diff
@@ -0,0 +1,28 @@
...
+const defaultMediaRuleId = "__default-media-rule__";
+
+export class StyleSheetRegular extends StyleSheet {
+  addStyleRule(
+    rule: CssRule,
+    selectorText: string,
+    transformValue?: TransformValue
+  ) {
+    const mediaRule = this.addMediaRule(rule.breakpoint || defaultMediaRuleId);
+    if (mediaRule === undefined) {
```
istarkov, line 14: Probably better to throw in `addMediaRule` so in ts it would always return defined value

packages/css-engine/src/core/style-sheet.ts
```diff
@@ -1,15 +1,101 @@
...
+    this.#isDirty = true;
+  }
+  get cssText() {
+    if (this.#isDirty === false) {
+      return this.#cssText;
     }
+    this.#isDirty = false;
+    const css: Array<string> = [];
+
+    css.push(...this.#fontFaceRules.map((rule) => rule.cssText));
```
istarkov, line 65: Am I right we want to inline this into html in the near future?

istarkov (approved):
`for` looks like a bug, everything else is +-ok

packages/css-engine/src/core/style-sheet.ts
kof: I don't know it makes sense to completely stop caching css files, we need to do more benchmarking, I guess it would make sense for smaller sites that change very often, and make less sense for larger sites like our landing that changes less often

- The more it changes the less there is benefit in caching - so inlining makes more sense with frequent changes
- The bigger the file the less it makes sense to inline

istarkov: I mean #fontFaceRules

packages/css-engine/src/core/style-sheet-atomic.ts
kof: huh that's a good point, how does it work at all

packages/css-engine/src/core/style-sheet-atomic.ts
kof: yeah, can do, I meant the mental model more like rule.on("change") ... like adding an event handler, which is why I added it later

packages/css-engine/src/core/style-sheet.ts
kof: oh yeah, but you can use the engine, then get cssText and inline it

packages/css-engine/src/core/style-sheet-atomic.ts
kof: done

packages/css-engine/src/core/style-sheet-regular.ts
kof: done

packages/css-engine/src/core/style-sheet-atomic.ts
kof: upon rereading turned out it was fine, after we learned that we need to add the rule, there is no point in iterating any further, it just speeds things up

packages/css-engine/src/core/index.ts
```diff
@@ -1,12 +1,13 @@
-export { CssEngine } from "./css-engine";
 export type {
   AnyRule,
   StyleRule,
   MediaRule,
   PlaintextRule,
   FontFaceRule,
 } from "./rules";
-export * from "./create-css-engine";
+export { StyleSheetRegular } from "./style-sheet-regular";
```
TrySound, line 8: Why expose both classes and factories?

packages/css-engine/src/core/style-sheet-atomic.ts
```diff
@@ -0,0 +1,58 @@
...
+      const className = `c${hash(
+        newStyleRule.cssText + selectorSuffix + rule.breakpoint
+      )}`;
+      classes.push(className);
+      newStyleRule.selectorText = `.${className}${selectorSuffix}`;
+
+      const ruleExists = mediaRule.rules.some((styleRule) => {
+        return (
+          "selectorText" in styleRule &&
+          // This property-value combination has already been added.
```
TrySound, line 41: Is it necessary to check for selectorText existance? Seems like yiu pass empty string by default

packages/css-engine/src/core/style-sheet-atomic.ts
```diff
@@ -0,0 +1,58 @@
...
+      );
+      styleRules.push(newStyleRule);
+      // "c" makes sure hash always starts with a letter.
+      const className = `c${hash(
+        newStyleRule.cssText + selectorSuffix + rule.breakpoint
+      )}`;
+      classes.push(className);
+      newStyleRule.selectorText = `.${className}${selectorSuffix}`;
+
+      const ruleExists = mediaRule.rules.some((styleRule) => {
```
TrySound, line 38: Could mediaRule.rules be a map?

packages/react-sdk/src/component-generator.test.ts
```diff
@@ -687,6 +691,7 @@ test("avoid generating collection parameter variable as state", () => {
 test("generate params variable when present", () => {
   expect(
     generatePageComponent({
+      classesMap: new Map([["body", ["cls"]]]),
```
TrySound, line 694: Write separate test to check classes. These cases checks unrelated stufff.

packages/react-sdk/src/css/css.ts
```diff
@@ -8,6 +11,7 @@ import type {
   StyleDeclKey,
   StyleSourceSelection,
 } from "@webstudio-is/sdk";
+import { createAtomicStyleSheet } from "@webstudio-is/css-engine";
```
TrySound, line 14: Merge imports

packages/react-sdk/src/css/css.ts
```diff
@@ -49,21 +56,25 @@ export const createImageValueTransformer =
...
-export const generateCssText = (data: Data, options: CssOptions) => {
+export const generateCss = (
+  data: Data,
+  { assetBaseUrl, atomic = false }: CssOptions
+) => {
   const assets: Assets = new Map(data.assets.map((asset) => [asset.id, asset]));
   const breakpoints = new Map(data.breakpoints);
   const styles = new Map(data.styles);
   const styleSourceSelections = new Map(data.styleSourceSelections);
+  const classesMap: Map<string, Array<string>> = new Map();
```
TrySound, line 67: const classesMap= new Map<string, string[]>();

packages/css-engine/src/core/index.ts
kof: so far only for types in tests

kof: can be avoided or reduced to atype only

packages/css-engine/src/core/style-sheet-atomic.ts
kof: it could, but we won't be able to control the order, maybe that's alright

packages/react-sdk/src/component-generator.test.ts
kof: agree

packages/css-engine/src/core/style-sheet-atomic.ts
kof: I set it after that, its not beautiful, I am creating a rule to get cssText and then create hash, alternatively I have to move hash creation to the style rule

kof: actually I can just pass a function to a constructor so regular one can return a different one than atomic, so the logic is more clear

kof: rewrote

packages/css-engine/src/core/style-sheet-atomic.ts
kof: done

packages/react-sdk/src/component-generator.test.ts
kof: done

PR #2722, opened 2023-12-31
feat: add variables to settings panel

TrySound:

Here added variables section to settings panel similar to props section. UI is slightly different from previous implementation.

- the list shows only variables defined specifically on current instance
- selecting item starts editing instead of "menu" button
- resource variables can be deleted when unused

Screenshot 2023-12-31 at 18 01 35
Screenshot 2023-12-31 at 18 01 48
Screenshot 2023-12-31 at 18 01 35

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
@@ -219,10 +219,10 @@ const $selectedInstanceVariables = computed(
 
 export const ResourcePanel = ({
   variable,
-  onCancel,
+  onClose,
 }: {
   variable?: DataSource;
-  onCancel: () => void;
+  onClose: () => void;
 }) => {
   const resources = useStore($resources);
   const resource =
@@ -346,7 +346,7 @@ export const ResourcePanel = ({
       )}
 
       <Flex justify="end" css={{ gap: theme.spacing[5] }}>
-        <Button color="neutral" onClick={onCancel}>
+        <Button color="neutral" onClick={onClose}>
           Cancel
         </Button>
         <Button
@@ -379,7 +379,7 @@ export const ResourcePanel = ({
                 resources.set(newResource.id, newResource);
               }
             );
-            onCancel();
+            onClose();
           }}
         >
           Save
--- a/apps/builder/app/builder/features/settings-panel/settings-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/settings-panel.tsx
@@ -1,6 +1,8 @@
 import type { Instance } from "@webstudio-is/sdk";
 import { SettingsSection } from "./settings-section/settings-section";
 import { PropsSectionContainer } from "./props-section/props-section";
+import { VariablesSection } from "./variables-section";
+import { isFeatureEnabled } from "@webstudio-is/feature-flags";
 
 export const SettingsPanelContainer = ({
   selectedInstance,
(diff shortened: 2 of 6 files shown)


kof:
@taylornowotny any thoughts?

kof:
- [ ] sorting isn't working, if its too much work, just hide the drag handle

kof:
- [ ] when binding has not yet been added, we should probably disable "delete" button, otherwise its not clear what you might be deleting since there is not value set

image

kof:
- [ ] can't close the binding dialog when I deleted expression value

image

kof:
- [ ] either fix sorting or hide sort handle

image

kof:
- [ ] position of the binding dialog is a little strange, we need to think of something better

image

kof:
- [ ] lets get rid of cancel/save buttons

image

image

kof:
- [ ] seems like you are not using floating panel, it should position on the left from the settings panel, like with all other panels e.g. box shadow, fonts, transitions etc

image

kof:
- [ ] when value is bound to an expression, the input looks weird with the background color inside, probably it comes from the editor, what can we do about it? I know its not part of this PR

image

kof:
- [ ] I managed to bind tag and render box as an anchorby using value "a" , this should be impossible, since in this case only values provided by the select box should be possible (I know its not part of this PR)

image

kof (approved):
not everything need to be fixed in this pr, feel free to create separate once for each, but if you do please transfer those points to an issue

TrySound:
> can't close the binding dialog when I deleted expression value, seems like a natural way would be when user deleted expression and is closing the dialog to act same way as with clicking delete button

>  lets get rid of cancel/save buttons

The problem with lack of "save" buttons is validation. At what point it should show error? Users will close popover with invalid expression which is not saved and have not idea what is going on.

> when value is bound to an expression, the input looks weird with the background color inside, probably it comes from the editor, what can we do about it? I know its not part of this PR

It's just disabled input

kof:
> The problem with lack of "save" buttons is validation. At what point it should show error? Users will close popover with invalid expression which is not saved and have not idea what is going on.

I think we can do the following: 
1. while user is typing we don't show errors right away
2. when input gets unfocused - we validate and show error
3. when user starts typing while having error state, we remove the error and go from point 1
4. we also indicate the error with a red border (I afaik ds input has state error like that and code input should too)
5. if input is in error state and user closes - don't save the error state

kof:
> It's just disabled input

then it must be broken, lets check what DS says about disabled input state

TrySound:
> then it must be broken, lets check what DS says about disabled input state

I don't get it. It doesn't look broken. It looks disabled. Following DS.

image

kof:
> I don't get it. It doesn't look broken. It looks disabled. Following DS.

Right now what I see:

image

What it should be according to your screenshot

image



Do you see this gray background inside the input/textarea?

kof:
you may not see it depending on your screen, but on macbook I can see it, her its this gray background outlined

screen

kof:
- [ ] Just noticed that this is not using floating panel component

image

kof:
- [ ] noticed that user can have right now both Value and Ressource, but only one can be possible at the same time, so there is a UI issue, we are not telling the user this, we need to figure out the best possible UI for this, it might be that tabs is not a fit at all

PR #2727, opened 2024-01-03
feat: switch props to binding popover

TrySound:

Here finally replaced variables panel with binding popover in all prop controls.
Additionally did better support of expressions in file and url controls.

The tricky part is managing expressions on control components level instead of global props panel level which gives us more granular control of behavior though requires more boilerplate code in each control.

There are still some bugs for example with loosing focus when start editing expression which I will fix in following validation fix.

## Testing
- check id in Body (string)
- aria-max something in Body (number)
- check "show" in body child (boolean)
- check href in Link (url)
- check src in Image (file)
- check state in Form (radio)
- check aria-autocomplete in body (select)
- check remove binding in resource fields

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
- [ ] lets reduce the initial size of binding dots to 12px from 14 (looks a bit too massive right now), but add a 2px margin so that the hoverable surface is still quite big

kof:
- [ ] bound show/id controls for example don't show as "bound" right now, is this on a todo?

kof:
- [ ] I think its already reported in the main issue - binding select with a custom value shouldn't be posible

image

kof:
- [ ] binding on a radio needs a better look, I think we discusssed it with taylor not sure if it was designed, his idea was to have a border around the radio items so that the binding button doesn't look so weird there

image

kof (approved):

TrySound:
> bound show/id controls for example don't show as "bound" right now, is this on a todo?

Do you mean they are not disabled? They shouldn't because we want to control state with props to preview changes like error or success state in forms.

kof:
> Do you mean they are not disabled? They shouldn't because we want to control state with props to preview changes like error or success state in forms.

I am confused, wasn't the plan for bound value to have a special state that is non-editable?

TrySound:
We always supported editing single variable expression with props to preview features. So no it wasn't the plan to stop supporting it. We disabled only complex expression which cannot be reflected  in variable values.

PR #2729, opened 2024-01-04
feat: Link typography tokens and a new Link component in DS

kof:
## Description

- Decided to not have separate colors for link, it should just use regular text colors. 
- Added a Link component to the design system, modelled after MUI Link component

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/link.tsx
```diff
@@ -0,0 +1,65 @@
...
+      },
+      hover: {
+        textDecoration: "none",
+        "&:hover": { textDecoration: "underline" },
+      },
+      always: {
+        textDecoration: "underline",
+        "&:hover": { textDecoration: "underline" },
+      },
+    },
```
TrySound, line 58: This better be decided on project level

kof: I checked all our usages, this works, also I stole this api from material ui https://mui.com/material-ui/react-link/#underline

I think we should be fine, this gives us more of a standard api for links

kof: the 3 underline behavior is all we ever gonna need

TrySound (approved):

PR #2730, opened 2024-01-05
feat: improve binding validation

TrySound:

Here changed how bindings are validated

- closing with broken expression will show errors and prevent close
- the next close will not block and keep the old valid expression
- close with valid expression invoked saving
- props no longer can change the type of input
- binding button becomes red and gets tooltip when value has invalid type
- binding popover also shows non-blocking error about type
- each control provides own validation

image
Screenshot 2024-01-05 at 17 13 15

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/expression-editor.tsx
+++ b/apps/builder/app/builder/shared/expression-editor.tsx
@@ -212,9 +212,14 @@ const autocompletionStyle = css({
 const emptyScope: Scope = {};
 const emptyAliases: Aliases = new Map();
 
+const errorStyle = css({
+  borderColor: theme.colors.borderDestructiveMain,
+});
+
 export const ExpressionEditor = ({
   scope = emptyScope,
   aliases = emptyAliases,
+  error = false,
   autoFocus = false,
   readOnly = false,
   value,
@@ -229,6 +234,7 @@ export const ExpressionEditor = ({
    * variable aliases to show instead of $ws$dataSource$id
    */
   aliases?: Aliases;
+  error?: boolean;
   autoFocus?: boolean;
   readOnly?: boolean;
   value: string;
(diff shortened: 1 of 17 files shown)


kof:
- [ ] error state should probably also use red border, like we have on InnputField color="error"

image

kof:
- [ ] when I hit tab, blur doesn't happen, so there is no validation as well, I should be able to focus the next button
image

kof:
- [ ] needs better error message

image

"Data Ws Show" is a bunch of acronyms nobody will understand

kof:
- [ ] ESC should always work same way as the close button - this the the only quick way to close the dialog from the keyboard, it works this way in all floating panels, so it should here along with the tab

image

kof:
Oh I get it, tab inside the code editor has a different function since its not just an input .... I guess there is no way out in this case

kof:
seems like codemirror is using tab for autocompletion and esc to close the autocompletion, though when no autocompletion is shown, it shouldn't prevent ESC from propagating, we should be able to hit esc to close when there is no autocompletion

TrySound:
On which prop?

kof:
Any prop? I tried on id, but also link prefetch, link target

apps/builder/app/builder/shared/expression-editor.tsx
```diff
@@ -212,9 +212,14 @@ const autocompletionStyle = css({
...
 const emptyAliases: Aliases = new Map();
 
+const errorStyle = css({
+  borderColor: theme.colors.borderDestructiveMain,
+});
+
 export const ExpressionEditor = ({
   scope = emptyScope,
   aliases = emptyAliases,
+  error = false,
```
kof, line 222: why not choosing a similar interface we have in InputField and others:

prop color - defines colors e.g. color=error

kof (approved):

PR #2736, opened 2024-01-06
fix: Ensure user can't enter invalid canvas width value in breakpoints editor

kof:
## Description

User was able to delete the value and get a NaN error Ref https://github.com/webstudio-is/webstudio/issues/2734

## Steps for reproduction

1. delete the value
2. see the min width rendered

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/breakpoints/width-input.tsx
+++ b/apps/builder/app/builder/features/breakpoints/width-input.tsx
@@ -52,7 +52,9 @@ const useEnhancedInput = ({
 
   const getValue = () => {
     const value = inputRef.current?.valueAsNumber;
-    return typeof value === "number" && isNaN(value) === false ? value : min;
+    return typeof value === "number" && Number.isNaN(value) === false
+      ? value
+      : min;
   };
 
   return {
--- a/apps/builder/app/builder/features/breakpoints/width-input.tsx
+++ b/apps/builder/app/builder/features/breakpoints/width-input.tsx
@@ -52,7 +52,7 @@ const useEnhancedInput = ({
 
   const getValue = () => {
     const value = inputRef.current?.valueAsNumber;
-    return typeof value !== "number" || isNaN(value) ? min : value;
+    return typeof value === "number" && isNaN(value) === false ? value : min;
   };
 
   return {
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -55,15 +50,21 @@ const useEnhancedInput = ({
     onChangeComplete: handleChangeComplete,
   });
 
+  const getValue = () => {
+    const value = inputRef.current?.valueAsNumber;
+    return typeof value !== "number" || isNaN(value) ? min : value;
```
TrySound, line 55: typeof value !== "number" could be undefined check

isNaN -> Number.isNaN

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -55,15 +50,21 @@ const useEnhancedInput = ({
...
+  };
+
   return {
     ref: scrubRef,
     inputRef,
-    onChange(event: ChangeEvent<HTMLInputElement>) {
-      setIntermediateValue(event.target.valueAsNumber);
+    onChange() {
+      setIntermediateValue(getValue());
```
TrySound, line 62: How does it work? Does input get the state before it is updated by react?

kof: its a controlled input, so on change you can pretty much get the value from the element

kof: though I want to refactor this again, because currently there is no invalid state handling so you can't just delete the value to type a new one

apps/builder/app/builder/features/breakpoints/width-input.tsx
kof: good catch

TrySound (approved):

apps/builder/app/builder/features/breakpoints/width-input.tsx
```diff
@@ -52,7 +52,7 @@ const useEnhancedInput = ({
 
   const getValue = () => {
     const value = inputRef.current?.valueAsNumber;
-    return typeof value !== "number" || isNaN(value) ? min : value;
+    return typeof value === "number" && isNaN(value) === false ? value : min;
```
TrySound, line 55: Number.isNaN pls

kof: we should lint that, always forgetting

PR #2737, opened 2024-01-07
feat: improve variable popover

TrySound:

Here changed variable popover UI

- removed save/cancel buttons - save is happening on close
- replaced value/resource tabs with type select
- support additional string, number and boolean values
- add validation to value and resource forms

Also added form utilities to manage and compose field validation.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
@@ -1,6 +1,12 @@
 import { computed } from "nanostores";
 import { nanoid } from "nanoid";
-import { useId, useMemo, useState } from "react";
+import {
+  forwardRef,
+  useId,
+  useImperativeHandle,
+  useMemo,
+  useState,
+} from "react";
 import { useStore } from "@nanostores/react";
 import type { DataSource, Resource } from "@webstudio-is/sdk";
 import { encodeDataSourceVariable } from "@webstudio-is/react-sdk";
@@ -9,6 +15,7 @@ import {
   Button,
   Flex,
   Grid,
+  InputErrorsTooltip,
   InputField,
   Label,
   Select,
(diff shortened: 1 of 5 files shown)


kof:
- [ ] JSON needs to be rendered all capital
image

kof:
- [ ] I would expect it to save and close after typing value and hitting Enter, you?

image

kof:
- [ ] I would expect the input for json value doesn't need quotes any more

image

kof:
- [ ] in the list item says "object", but shouldn't it be also JSON? or both object? either way they should be the same, right?

image

kof:
- [ ] when creating a variable and going right after type selector, it shows error on the name, though I wasn't adding name yet
image

kof (changes requested):
ux improvements needed

TrySound:
> when creating a variable and going right after type selector, it shows error on the name, though I wasn't adding name yet

You wanted name to be first focused field. You want it to be type now?

TrySound:
> I would expect the input for json value doesn't need quotes any more

So empty object then?

TrySound:
> in the list item says "object", but shouldn't it be also JSON? or both object? either way they should be the same, right?

So no need to distinct "array" and "object" there? Those are not selected types but compact value preview.

kof:
> You wanted name to be first focused field. You want it to be type now?

No, name should be focused, but its weird when I click on type and get an error on name right away, seems like we need to check if user changed name or alternatively check if its empty and handle it separately

kof:
> So empty object then?

Ideally just empty and not show error until its time to save

kof:
> So no need to distinct "array" and "object" there? Those are not selected types but compact value preview.

Hmm, why do we have distinction between array and object there but not when creating the variable? I am surely missing something

kof:
> but compact value preview.

What is a compact value preview? like instead of "{a: 1}" you write "object"? In that case why not represent both by JSON?

kof:
Reg object/array vs JSON - its actually more correct to call it JSON, because its not a javascript object or array, its actually a JSON, right? meaning it can't have references or any other data structures, its a serializable object

TrySound:
> No, name should be focused, but its weird when I click on type and get an error on name right away, seems like we need to check if user changed name or alternatively check if its empty and handle it separately

This is how every validation library works. Blur always triggers validation of the field.

kof:
> This is how every validation library works. Blur always triggers validation of the field.

Alright, leave it for now as-is, it felt definitely weird to me, but maybe I am wrong

kof (approved):
looking great!

PR #2744, opened 2024-01-08
Fix: Determine CNAME properly when domain has multiple extensions

johnsicili:
## Description

When a domain like "www.example.co.za" is added, the UI shows "www.example" in the name field when it should show just "www".

## Steps for reproduction

1. Add new domain 
2. Add "www.example.co.za"
3. Observe the Name field (should now be www instead of www.example)

## Code Review

- [ ] hi @kof, I need you to review the heck out of this
  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/topbar/domains.tsx
```diff
@@ -74,12 +74,13 @@ const CopyToClipboard = (props: { text: string }) => {
 };
 
 const getCname = (domain: string) => {
```
kof, line 76: lets add some unit tests for this @istarkov ?

istarkov:
What if domain is real subdomain like `hello.world.example.com` above will not work. 
As the right cname will be `hello.world` and not `hello`

apps/builder/app/builder/features/topbar/domains.tsx
istarkov: See my comment below, this will not work. I don't know how to distinguish domains like `co.za` from ordinary domains like `example.com`

istarkov:
Based on this https://en.wikipedia.org/wiki/Country_code_second-level_domain some country code domains has 140 subdomains like co.za see brasil, some don't. 
Examples: `co.uk`, `org.uk`, `vet.br`,  `wiki.br` 
I haven't found the list of such domains, and not even sure that without that list any algorithm exists to extract cname properly in cases like above

johnsicili:
I concur. Here are the options as I see it:

1. Check against a list of second level domains (won't be perfect, but would catch almost 100% of use cases)
2. Make assumptions... like if we can assume a domain name is more than three characters I think logic would be easier (not so hot of a solution though)
**3. Don't support multi-subdomains/deal with it when somebody actually has a use case. I for one would advocate for dev-www.example.com instead of two subdomains, but then again I'm not familiar with the use case (my vote).**
4. Have the user explicitly enter the various parts of their domain (e.g. subdomain(s), domain + extensions)

I believe I'd need to update the code as I think I missed a use case.

kof:
https://handwiki.org/wiki/Country_code_top-level_domain

We need to get that list of cctlds and check against it, seems like there is no single place that maintains that list well, so we will have to hardcode it

kof:
Update we found a proper database of those suffixes https://publicsuffix.org/list/public_suffix_list.dat and Ivan is working on adding a proper check for those, closing this pr

PR #2752, opened 2024-01-11
fix: Add source and reset details in tooltips for flex gap

### LINKED ISSUE #2733: Flex gap doesn't have a tooltip with a full style source info
Flex gap icons should show the same style tootip with full info like other properties

image

JayaKrishnaNamburu:
## Description

fixes #2733 

Tooltips on FlexGap properties don't have details of `columnGap` or `rowGap`.  Now, on hover on these two properties will show the properties details and  able to reset the values from the tooltip

## Steps for reproduction

1. Hover on the flex-gap properties.
2. The property will show the details of the property and allows to reset.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

### DIFF
--- a/packages/design-system/src/components/enhanced-tooltip.tsx
+++ b/packages/design-system/src/components/enhanced-tooltip.tsx
@@ -71,6 +71,12 @@ export const EnhancedTooltip = forwardRef(
      */
     const allowOpenRef = useRef(true);
 
+    const handleDisableOpen = () => {
+      allowOpenRef.current = false;
+      setOpen(false);
+      showTooltipDelayed.cancel();
+    };
+
     const showTooltipDelayed = useDebouncedCallback(
       () => {
         if (allowOpenRef.current) {
@@ -83,12 +89,18 @@ export const EnhancedTooltip = forwardRef(
 
     // We debounce blur/focus events in a single function, to skip fast `blur/focus` which can occur during select menu hovers
     const handleFocusEventsDebounced = useDebouncedCallback(
-      (eventType: "focus" | "blur") => {
+      (eventType: "focus" | "blur" | "mouseLeave") => {
         if (eventType === "blur") {
           isFocusedRef.current = false;
           allowOpenRef.current = true;
           setOpen(false);
           showTooltipDelayed.cancel();
+          return;
+        }
+
+        if (eventType === "mouseLeave") {
+          allowOpenRef.current = true;
+          isFocusedRef.current = true;
           return;
         }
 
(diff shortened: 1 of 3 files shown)


kof (commented):
doesn't work for me, please check both row gap and column gap in both linked and not linked (button between them)

kof (changes requested):
doesn't work for me yet

JayaKrishnaNamburu:
Before
tooltips-before
After
tooltips-after

@kof  I think, i might have fixed something else then 😅. Or is the same not working on your side ?

JayaKrishnaNamburu:
There is a catch though, even in connected mode. The the builder never applies `flex-gap` directly. Instead it applies `rowGap` and `columnGap` with same values. 

So, we can't really reset the in a regular flow with a reset button in the tooltip.
But we can add logic to show the tooltip and everything on the `link-button` when they are in connected mode. By mimicking the behaviour combining `columnGap` and `rowGap`. Sounds good ?

kof:
Maybe it should apply rowGap and columnGap always and never just gap?

kof:
When I tried, in connected mode, tooltip was visible on the second one but not on the first one, when value was defined

JayaKrishnaNamburu:
> Maybe it should apply rowGap and columnGap always and never just gap?

Yup, that's the best way to go. As, the current UX gives users to define them individually or use in connected mode. In that case, the current implementation don't need any changes right. Can you check once again and confirm please

kof:
After disconnecting and changing one value, one of the tooltips doesn't show up.

1. change connected value
2. disconnect
3. change one value
4. expected both values can show a tooltip individually and reset individually

JayaKrishnaNamburu:
Found the issue, looks like. When users use `drag` operation to change the values. That is when the tooltips are not able to trigger. Might be something to do with the events there. If the inputs are just changed manually then the tooltips seems to work fine 👍. Will look into this

packages/design-system/src/components/enhanced-tooltip.tsx
```diff
@@ -117,12 +111,6 @@ export const EnhancedTooltip = forwardRef(
         handleFocusEventsDebounced("blur");
         event.preventDefault();
       },
-      onPointerDown: () => {
-        handleDisableOpen();
-      },
-      onKeyDown: () => {
-        handleDisableOpen();
-      },
```
JayaKrishnaNamburu, line 125: @kof @istarkov looks like the tooltips are intentionally disabled once the icon is interacted on inputs. Removed them and everything works as expected. Can you think of any specific case why this is added or it was just a bug ? 🤔

kof: is it meant to be only during interaction? to avoid having both tooltip and interaction at the same time?

JayaKrishnaNamburu:
Had the same hunch initially, but even without disabling it. It seems to work without any issues on my side testing.

packages/design-system/src/components/enhanced-tooltip.tsx
JayaKrishnaNamburu: Had the same hunch initially, but even without disabling it. It seems to work without any issues on my side testing.

kof: I bet you can hover and then click to drag on the icon and you get the tootlip open at the same time

JayaKrishnaNamburu: Screen Recording 2024-01-11 at 23 38 51
That's the strage part for me, it wasn't showing the tooltip when the keyboard click happens / drag is in progress. So we don't need to explicitly handle the case here.

kof: maybe its for some other case, @istarkov will know

kof: Oh I just read the code, its clear why - we enhanced tooltip implements a different delay logic, so it needed to prevent the other events to delay differently

JayaKrishnaNamburu:
Screen Recording 2024-01-12 at 18 40 19

Required behaviour.
- Users can hiver over the flex icons and can access the tooltip for the info.
- Users should be able to drag the values to update in the `columnGap` or `rowGap` inputs.
- When users visit the icon back, the tooltip should appear again.
- If the users just click on the icon, to start the drag behaviour. It shouldn't show the tooltip, as there is a drag operation that is in progress.

Current Behaviour
- The tooltips don't pop-up when the drag operation is in progress. As the the handers are manually disabling tooltip with `triggerProps` event handlers.
- But, when users completes the drag and come back to the icon, the tooltips never show up again. Since, the tooltip is manually disabled to show in the above step.

Fix
- Added a `onMouseLeave` listener. So, we can listen to actually when the mouse left the icon and reset it to `true`. So, when the operations are done and the cursor is again on the icon. The tooltips will behave as usual.

The above attached gif will show all the phases of the behaviour.

kof (approved):
works well!

PR #2753, opened 2024-01-11
fix: tooltip use full width when value section or reset button shows

### LINKED ISSUE #1892: Tooltips use full width when Value section or Reset button shows
Tooltips should expand to 100% of the max-width of 208px when the "Value comes from section" and the Reset button are included.

This is a design improvement that was not in the original design docs. It will help with consistent appearances and prevent tooltips from being awkwardly skinny and tall if there's not a description.

ditorodev:
## Description

Closes #1892 
Sets the `maxWidth` and `width` properties of the Tooltip in `PropertyTooltip` to `208px` (using the token) and removes the inner `maxWidth` of `TooltipContent` this is so we can ensure the same size (which after padding comes to 224px) across all Tooltips in the sidebar no matter their content

## Steps for reproduction

1. Hover over a CSS property in the `side-panel`
2. With or without content the tooltip that appears should always be 224px (with the inner content being 208px)

## Code Review

- [ ] hi @kof @taylornowotny , I need you to check that this suffices with the new design spec

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

Note: @kof How should I add tests for this? E2E checking that all `PropertyName` components render at 224px all the time? Or should we just not care to test this?


kof:
extremely sorry for ignoring this pr, seems like it got fixed in the meantime

PR #2754, opened 2024-01-12
feat: support editing text content in settings panel

TrySound:

Here added "Text Content" system prop to all elements which allows to edit plain text content though settings panel.

It will be used to support instance content binding instead of using HtmlEmbed with code binding.

Screenshot 2024-01-12 at 11 51 39

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/controls/text-content.tsx
+++ b/apps/builder/app/builder/features/settings-panel/controls/text-content.tsx
@@ -5,8 +5,6 @@ import {
   Flex,
   Label,
   theme,
-  Tooltip,
-  rawTheme,
 } from "@webstudio-is/design-system";
 import type { Instance } from "@webstudio-is/sdk";
 import {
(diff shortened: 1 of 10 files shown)


istarkov (approved):

apps/builder/app/builder/features/settings-panel/controls/text-content.tsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+export const TextContent = ({
+  instanceId,
+  meta,
+  propName,
+  computedValue,
+}: ControlProps<"textContent">) => {
+  const textContent = useTextContent(instanceId);
+  const updateTextContent = (type: "text", value: string) => {
```
kof, line 66: why inlining that function when it could be outside of the component?

kof:
- [ ] Text Content is now everywhere, while it should be only on those instances which should allow it. The question is how we define which instances allow it, I see 3 options:
1. explicitely enable/disable over meta
2. automatically based on html spec data
3. both, automatically choose the right default, but allow enabling/disabling over meta

kof:
- [ ] first time instance is added - binding dot isn't visible, even after reload, then when I type some text and save - binding dot appears, but I still can't bind the value

TrySound:
This PR does not support binding yet. Only shows button.

kof (approved):
I am not sure if the new text content will appear without a feature flag, if it won't - feel free to merge it if it will - we can't merge because of the ttodos I mentioned and we need it behind a flag

PR #2764, opened 2024-01-15
fix: Default link always underline

kof:
## Description

Failed to check after last changes that the default link used most of the time and it should have underline=always by default, only rare cases have underline  hover or none

## Steps for reproduction

1. export dialog
2. blocking alert

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):
❤️

PR #2769, opened 2024-01-16
feat: support bound and overwritten binding button variants

TrySound:

- added blue "bound" binding button
- added pink "overwritten" binding button
- added tooltip to reset overwriten binding

Screenshot 2024-01-16 at 22 20 11
Screenshot 2024-01-16 at 22 20 05

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
- [ ] found something strange, when I overwrite the value with a temp one, I see the value change in the variables panel

image

kof (approved):

TrySound:
> Its kindof like I mutated the variable instead of mutating the value of the property

Because we preview variable not just property

PR #2772, opened 2024-01-17
fix: support mouse select in expression autocomplete

TrySound:

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/expression-editor.tsx
+++ b/apps/builder/app/builder/shared/expression-editor.tsx
@@ -169,7 +169,7 @@ const autocompletionStyle = css({
     ...textVariants.mono,
     border: "none",
     backgroundColor: "transparent",
-    // override none set on body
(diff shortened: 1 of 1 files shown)


kof (approved):

apps/builder/app/builder/shared/expression-editor.tsx
```diff
@@ -169,6 +169,8 @@ const autocompletionStyle = css({
     ...textVariants.mono,
     border: "none",
     backgroundColor: "transparent",
+    // override none set on body
+    pointerEvents: "auto",
```
kof, line 173: when do we set it on body? why does it come up here, is it because of popover?

kof: worth mentioning the "why", not just the "what"

Issue #2650, opened 2023-11-22
♻️ Redirects Manager


itsNintu:
The redirects manager will help in simplifying redirecting traffic from any URL to a new one, ensuring that visitors and search engines are directed to the correct pages.

JayaKrishnaNamburu:
The initial iteration of Redirects management is added in
#2779 in the UI
#2824 for deployments

These mark all the redirects that users are adding under project settings as `301`. But we should still allow users to mark a specific redirect as `302` if needed. And it will be added in a new PR.

kof:
Done, releasing today

PR #2779, opened 2024-01-18
feat: Add redirects UI for `301` redirects to external URL's

JayaKrishnaNamburu:
related to #2650 

## Description

This PR adds only the UI for adding redirects menu under project settings. To keep the PR's small, the deployment side of things will be added in another PR.

## Steps for reproduction

1. Add `?features=redirects` at the end of the URL.
2. Open Project Settings.
3. Add a `path` and `url` at the end.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


JayaKrishnaNamburu:
redirects

kof:
Let me tweak CSS a bit and we are good to go for this one. We need to see later regarding validation.

apps/builder/app/builder/features/seo/project-redirects.tsx
```diff
@@ -0,0 +1,170 @@
...
+  paddingLeft: 0,
+  overflowY: "scroll",
+  listStyle: "none",
+});
+
+export const ProjectRedirectionSettings = (props: {
+  settings: ProjectSettings;
+  onSettingsChange: (settings: ProjectSettings) => void;
+}) => {
+  const [oldPath, setOldPath] = useState<string | null>(null);
```
kof, line 30: we use undefined and trying to avoid null as much as possible for consistency

apps/builder/app/builder/features/seo/project-redirects.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const redirectKeys = Object.keys(redirects);
+  const isValidRedirects =
+    oldPath !== null && isValidOldPath && newPath !== null && isValidNewPath;
+
+  useEffect(() => {
+    if (oldPath === null) {
+      return setIsValidOldPath(true);
+    }
+    return setIsValidOldPath(oldPath.charAt(0) === "/");
+  }, [oldPath]);
```
kof, line 45: I think the usage of useEffect is "wrong" here, you can call setIsValidOldPath from event handler, I think this is a classic example from react docs "how not to do it" because useEffect is meant for synchronization of states, not really to do effects

kof (changes requested):
some implementation issues

apps/builder/app/builder/features/seo/project-redirects.tsx
JayaKrishnaNamburu: Good one, i was just playing around with `input` validation colors. And missed removing those. Made the changes now 👍

apps/builder/app/builder/features/seo/project-redirects.tsx
```diff
@@ -0,0 +1,184 @@
...
+    isValidOldPath &&
+    newPath !== undefined &&
+    isValidNewPath;
+
+  const validateOldPath = () => {
+    if (oldPath === undefined) {
+      setIsValidOldPath(true);
+      return;
+    }
+    setIsValidOldPath(oldPath.charAt(0) === "/");
```
kof, line 48: ideally tihs validation is part of the schema, e.g. see how we refine Page path from validation, this way you can have both client-side and also server-side validation and its all in one place, because you will likely add more things for validations over time

kof:
Remaining 
- [ ] enter to submit
- [ ] validation def. broken in many ways: typing "a"  then adding slash is not revalidating, you are probably one stroke too early, you need to flush the state with flushSync
- [ ] we should use URL api to validate if the target is valid
- [ ] if path is used on the right, we should check if it matches with actually existing paths and show error to the user to prevent them from mistyping

JayaKrishnaNamburu:
> create a redirect automatically when user changes a path on page settings

I would like to handle these in a separate PR, when the feature from end-top-end is done. Since, i am sure there will be more cases to it that just the `path` change. Like page duplication, delete etc. And i feel, it's little confusing to handle all those before the deployment is also done 👍 

> undo-redo is actually working juged by the xhr tab, but its not reflecting in the UI, must be some bug

This seems to break for `Atomic CSS` too, let's add a separate ticket for the same.

apps/builder/app/shared/nano-states/pages.ts
```diff
@@ -15,3 +20,19 @@ export const $selectedPage = computed(
     return findPageByIdOrPath(selectedPageId, pages);
   }
 );
+
+export const $existingRoutePaths = computed([$pages], (pages) => {
```
kof, line 24: discussed to just turn this into utility function

apps/builder/app/shared/nano-states/pages.ts
```diff
@@ -15,3 +20,19 @@ export const $selectedPage = computed(
...
+  if (pages === undefined) {
+    return;
+  }
+
+  const paths = [];
```
kof, line 29: btw this should be a Map or a Set, Map has benefit for other use cases, where you get the page as a value

packages/design-system/src/components/tooltip.tsx
```diff
@@ -145,7 +148,8 @@ export const InputErrorsTooltip = ({
       {...rest}
       content={content ?? " "}
       open={errors !== undefined && errors.length !== 0}
-      side="right"
+      side={side ?? "right"}
```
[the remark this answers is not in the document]
JayaKrishnaNamburu: Probably because the tooltip component is using `top` as default here
https://github.com/webstudio-is/webstudio/blob/main/packages/design-system/src/components/tooltip.tsx#L103-L111
```jsx
<TooltipPrimitive.Portal>
          <Content
            ref={ref}
            side="top"
            align="center"
            sideOffset={2}
            collisionPadding={8}
            arrowPadding={8}
            {...props}
          >
```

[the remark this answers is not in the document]
kof: right, but then why changing the default we use system-wide instead of passing what you need in your case?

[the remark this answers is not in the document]
JayaKrishnaNamburu: Since the tooltips are side by side, it's kinda blocking what's on the right side. If someone is editing the content after adding initially. It's blocking to see what's on the right input.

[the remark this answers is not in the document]
kof: right, because its specifically the case with forms

kof (approved):
👍 from my side, lets get @TrySound's eye as well

packages/sdk/src/schema/pages.ts
```diff
@@ -102,11 +102,31 @@ const ProjectMeta = z.object({
   code: z.string().optional(),
 });
 
+export const ProjectNewRedirectPathSchema = PagePath.or(
+  z.string().refine((data) => {
+    try {
+      return Boolean(new URL(data));
```
TrySound, line 108: Casting here doesn't do anything so better explicitly return true
```
new URL(data)
return true
```

packages/sdk/src/schema/pages.ts
```diff
@@ -102,11 +102,31 @@ const ProjectMeta = z.object({
...
+      return false;
+    }
+  }, "Must be a valid URL")
+);
+
+export const PageRedirectSchema = z.object({
+  old: PagePath,
+  new: ProjectNewRedirectPathSchema,
+});
+export const ProjectRedirectsSchema = z.array(PageRedirectSchema);
```
TrySound, line 119: Let's inline this type. Build array is not big deal but easy to confuse these types.

apps/builder/app/builder/features/sidebar-left/panels/pages/page-utils.test.ts
```diff
@@ -461,3 +462,34 @@ describe("filterSelfAndChildren", () => {
...
+      id: "blogId",
+      meta: {},
+      name: "Blog",
+      path: "/blog/:id",
+      rootInstanceId: "rootInstanceId",
+      title: "Blog",
+    });
+
+    const result = getExistingRoutePaths(pages);
+    expect(result.size).toBe(2);
```
TrySound, line 492: no need to check size if you check whole thing after that

apps/builder/app/builder/features/seo/project-redirects.tsx
```diff
@@ -0,0 +1,224 @@
...
+            css={{ flexShrink: 0 }}
+          >
+            Add
+          </Button>
+        </Flex>
+      </Grid>
+
+      {redirectKeys.length > 0 ? (
+        <Grid css={{ px: theme.spacing[5] }}>
+          <ScrollArea css={{ maxHeight: theme.spacing[22] }}>
```
TrySound, line 181: We don't need nested scroll
image

apps/builder/app/builder/features/seo/project-redirects.tsx
```diff
@@ -0,0 +1,224 @@
...
+            <List asChild>
+              <Flex
+                direction="column"
+                gap="1"
+                css={{
+                  mx: theme.spacing[5],
```
TrySound, line 187: Let's put padding on parent instead

apps/builder/app/builder/features/seo/project-redirects.tsx
kof: I am assuming there could be dozens of redirects and this will be a problem if you want to get anything after that section, unless we keep this section always as a last one

With more sections added we probably are going to add a navigation on the left and each section will be just it's own page

kof: actually, ok lets kick out the inner scroll, and just keep that section as a last one for now

apps/builder/app/builder/features/seo/project-redirects.tsx
```diff
@@ -177,46 +176,42 @@ export const ProjectRedirectionSettings = (props: {
...
-                        </Flex>
-                        <SmallIconButton
-                          variant="destructive"
-                          icon={<TrashIcon />}
-                          onClick={() => handleDeleteRedirect(index)}
-                        />
+        <Grid
+          css={{
+            px: theme.spacing[5],
+            mx: theme.spacing[5],
```
TrySound, line 182: Could it be just px?

JayaKrishnaNamburu:
@TrySound can we merge it now. And start with the deployment side of things ?

TrySound (approved):

PR #2787, opened 2024-01-21
fix: insert copy on reparent

TrySound:

Here replaced reparent implementation with copy/paste logic to support changing scope of variables.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/shared/instance-utils.test.ts
```diff
@@ -515,32 +517,227 @@ test("insert template data with only new style sources", () => {
 });
 
 describe("reparent instance", () => {
-  test("from collectioo item", () => {
+  test("between instances", () => {
+    // body
+    //   box
+    //     text
+    //   button
```
kof, line 524: seems like this comment is missing last text node

TrySound: I usually write only hierarchy of instances. Same like we do in navigator.

kof (approved):

PR #2808, opened 2024-01-23
fix: reinitialize commands when builder is reopen

TrySound:
Here fixed the issue with builder commands stopped being available to canvas because initialized only once when app starts.

Here changed initialization to each time builder is open.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
are we removing those commands subscriptions when we are navigating away from the app?

Can we abstract away the whole "subscription game" when switching to or away between apps? 

We need a way to enforce those things, because its difficult to test, we will always forget this stuff. We need some form of "enforcement"

PR #2813, opened 2024-01-24
refactor: inline actions into props

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/2791

Here refactored actions generation and computing
to simplify rendering slots as components.

before

```ts
const onClick = () => {
variableName = 1
set$variableName(variableName)
}
...
<Box
data-ws-id="..."
data-ws-component="Box"
onClick={onClick}>
```

after

```ts
<Box
data-ws-id="..."
data-ws-component="Box"
onClick={() => {
variableName = 1
set$variableName(variableName)
}}>
```

Similar thing done in builder. Actions are generated and executed per each prop when compute prop values instead of precomputing globally along with variables.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):

kof:
generator is quite complex, we nede to consider having a proper abstraction in terms of how we generate code and how tests look like, something that enables:
1. tests that actually pass jsx/components so we understand the test better
2. generator that can have different framework targets that can be plugins

PR #2814, opened 2024-01-25
feat: support avif upload

kof:
## Description

We simply didn't have avif in the list for available formats for upload and I used the opportunity to reuse the mime list

@istarkov  CFs resizer input formats don't have avif, https://developers.cloudflare.com/images/image-resizing/format-limitations/ 
will it automatically ignore it? or do we need to do something to support avif upload?

## Steps for reproduction

1. upload avif image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/asset-uploader/src/utils/mime.ts
+++ b/packages/asset-uploader/src/utils/mime.ts
@@ -17,11 +17,11 @@ const mimeCategories = [
 
 type MimeCategory = (typeof mimeCategories)[number];
 
-const extensionToMime = new Map([
-  [".avif" as string, "image/avif"],
-  [".bmp", "image/bmp"],
+// https://developers.cloudflare.com/images/image-resizing/format-limitations/
+const imageExtensionsToMime = new Map([
+  [".avif", "image/avif"],
   [".gif", "image/gif"],
-  [".ico", "image/vnd.microsoft.icon"],
+  [".ico", "image/x-icon"],
   [".jpeg", "image/jpeg"],
   [".jpg", "image/jpeg"],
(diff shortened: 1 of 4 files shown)


packages/asset-uploader/src/utils/mime.ts
```diff
@@ -17,24 +17,38 @@ const mimeCategories = [
 
 type MimeCategory = (typeof mimeCategories)[number];
 
-const extensionToMime = new Map([
-  [".avif" as string, "image/avif"],
-  [".bmp", "image/bmp"],
```
kof, line 22: I have no idea why we had bmp in the list ...

packages/asset-uploader/src/utils/mime.ts
```diff
@@ -17,24 +17,38 @@ const mimeCategories = [
...
+// https://developers.cloudflare.com/images/image-resizing/format-limitations/
+const imageExtensionsToMime = new Map([
+  [".avif", "image/avif"],
   [".gif", "image/gif"],
-  [".ico", "image/vnd.microsoft.icon"],
+  [".ico", "image/x-icon"],
```
kof, line 24: correct mime type for ico
https://stackoverflow.com/questions/13827325/correct-mime-type-for-favicon-ico

kof:
Turns out we don't support avif because its not supported by all browsers yet, we use it as output only when supported.
CF resizer doesn't support it as input, so if we added upload, it wouldn't be converted to other formats when browser doesn't support it, closing for now.

PR #2824, opened 2024-01-25
experimental: Add support for redirects in deployment

JayaKrishnaNamburu:
## Description

WIP, still a lot of cases to cover. Please don't review yet

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


JayaKrishnaNamburu:
Currently we can add support for 5 different types of redirect patterns. When the destination is a static path. And not mixed with dynamic attributes in it. The source can have static paths though. Here are the pattern of the paths that can be supported.
```json
{
          "old": "/blog-post/:id",
          "new": "/blog/post"
        },
        {
          "old": "/blog-post/1234",
          "new": "/blog/post"
        },
        {
          "old": "/github",
          "new": "https://www.github.com"
        },
        {
          "old": "/blog-post",
          "new": "/blog/post"
        },
        {
          "old": "/about",
          "new": "/new-about"
        }
```
And a working example for the same

https://drive.google.com/file/d/15sq2WSioy0hspmsWttlRXHrr3YhenNG0/view?usp=sharing

cc @kof @istarkov @TrySound

packages/cli/templates/defaults/app/redirects.ts
```diff
@@ -0,0 +1,28 @@
+import { matchRoutes } from "@remix-run/router";
+import { type PageRedirect } from "@webstudio-is/sdk";
+
+export const matchPathFromRedirects = (
```
TrySound, line 4: I would put it into page-utils in sdk

packages/cli/templates/defaults/app/redirects.ts
```diff
@@ -0,0 +1,28 @@
...
+) => {
+  const redirectMap: Record<string, string> = {};
+  const routePaths: Array<{ path: string }> = [];
+
+  for (const redirect of redirects) {
+    redirectMap[redirect.old] = redirect.new;
+    routePaths.push({ path: redirect.old });
+  }
+
+  const matchedRoutes = matchRoutes(routePaths, new URL(url));
```
TrySound, line 16: Is remix able to match such routes syntax `/:id`?

packages/cli/src/prebuild.ts
```diff
@@ -638,6 +638,33 @@ ${utilsExport}
...
+      const redirectFileName = `${redirectPagePath}.ts`;
+
+      const content = `import { type LoaderArgs } from "@remix-run/server-runtime";
+import { redirect } from "@remix-run/server-runtime";
+import { matchPathFromRedirects } from "~/redirects";
+
+export const loader = (arg: LoaderArgs) => {
+  const redirectPath = matchPathFromRedirects(arg.request.url, ${JSON.stringify(
+    redirects
+  )});
```
TrySound, line 656: Why match at runtime if we already know new redirect here at build time?

JayaKrishnaNamburu: Thats a good catch, that got me thinking. We are not supporting dynamic destination at the moment. So, this logic can happen at cli time. And then just add the result string in the `redirect`. Will make the necessary changes 👍

packages/cli/templates/defaults/app/redirects.ts
JayaKrishnaNamburu: Yes, for the incoming routes. Let's say uses request with `/blog/1234`, `/blog/124234` etc. But while defining the redirects user might have just added `/blog/:id`. Then we need to reply on this. to match `/blog/1234` -> `/blog/:id` and then find the destination for the result.

packages/cli/src/prebuild.ts
```diff
@@ -638,6 +639,29 @@ ${utilsExport}
...
 
+  const redirects = siteData.build.pages?.redirects;
+  if (redirects !== undefined && redirects.length > 0) {
+    spinner.text = "Generating redirects";
+
+    for (const redirect of redirects) {
+      const redirectPagePath = generateRemixRoute(redirect.old);
+      const redirectFileName = `${redirectPagePath}.ts`;
+
+      const redirectPath = matchPathFromRedirects(redirect.old, redirects);
```
TrySound, line 650: I don't get what this does? Why we match old path with redirects when you already have an access to redirect.new? Remix shouldn't be involved for sure. Our dynamic path syntax is not 100% compatible with remix router.

packages/cli/src/prebuild.ts
```diff
@@ -638,6 +638,29 @@ ${utilsExport}
...
 
+  const redirects = siteData.build.pages?.redirects;
+  if (redirects !== undefined && redirects.length > 0) {
+    spinner.text = "Generating redirects";
+
+    for (const redirect of redirects) {
+      const redirectPagePath = generateRemixRoute(redirect.old);
+      const redirectFileName = `${redirectPagePath}.ts`;
+
+      const redirectPath = redirects.find((item) => redirect.old === item.old);
```
TrySound, line 649: Why search here? You already have current item in "redirect" variable, no?

packages/cli/src/prebuild.ts
```diff
@@ -638,6 +638,29 @@ ${utilsExport}
...
+      const content = `import { type LoaderArgs } from "@remix-run/server-runtime";
+import { redirect } from "@remix-run/server-runtime";
```
TrySound, line 651: You can combine imports

packages/cli/src/prebuild.ts
```diff
@@ -638,6 +638,25 @@ ${utilsExport}
...
+    spinner.text = "Generating redirects";
+
+    for (const redirect of redirects) {
+      const redirectPagePath = generateRemixRoute(redirect.old);
+      const redirectFileName = `${redirectPagePath}.ts`;
+
+      const content = `import { type LoaderArgs, redirect } from "@remix-run/server-runtime";
+
+export const loader = (arg: LoaderArgs) => {
+  return redirect("${redirect.new}", 301);
```
TrySound, line 652: 301 is not a good default, 302 is better and we may need a way to customize this.

JayaKrishnaNamburu: But in the UI, the builder is showing as `301` should that also be changed ?
https://github.com/webstudio-is/webstudio/blob/main/apps/builder/app/builder/features/project-settings/redirect-section.tsx#L134
```jsx
<Text variant="titles">301 Redirects</Text>
```

TrySound: Honestly not sure what the idea was here. Need to check with @kof

TrySound: Maybe leave more granular redirects to page settings redirect

kof: What's the question?

TrySound: Дo redirects UI supports only 301 permanent redirect?

kof: why would we want to limit that code to one specific instead of making it flexible?

kof (approved):
approving upfront

TrySound (approved):

PR #2827, opened 2024-01-26
feat: Allow change route template in cli custom templates

istarkov:
## Description

Allow change route template, so various tests in saas become easier. 
Also now each template can have own route template.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/cli/src/prebuild.ts
```diff
@@ -418,18 +431,14 @@ export const prebuild = async (options: {
...
-      )
-    ),
-    "utf8"
+  const routeTemplatePath = normalize(
+    join(cwd(), "__templates__", "route-template.tsx")
   );
 
+  const routeFileTemplate = await readFile(routeTemplatePath, "utf8");
+
+  await rmdir(dirname(routeTemplatePath), { recursive: true });
```
TrySound, line 440: maybe just rm?

TrySound (approved):

packages/cli/src/prebuild.ts
istarkov: directory (dirname)

TrySound: fs.rm
> Removes files and directories (modeled on the standard POSIX rm utility).

PR #2836, opened 2024-01-28
feat: Dashboard thumbnails

kof:
Ref https://github.com/webstudio-is/webstudio/issues/2826

## Description

Rendering image preview of the home page social image

https://github.com/webstudio-is/webstudio/assets/52824/df23b0e2-7844-4c08-807d-ab16c83cd218

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/prisma-client/prisma/migrations/20240127230238_project-preview/migration.sql
```diff
@@ -0,0 +1,26 @@
...
+ALTER COLUMN
+  "previewImageAssetId"
+SET
+  NOT NULL;
+
+-- Drop DashboardProject View
+DROP VIEW IF EXISTS "DashboardProject";
+
+-- Update DashboardProject View
+CREATE VIEW "DashboardProject" AS
```
istarkov, line 13: use ‘create or replace view’ instead of drop/create

packages/prisma-client/prisma/migrations/20240127230238_project-preview/migration.sql
```diff
@@ -0,0 +1,26 @@
+-- AlterTable
+ALTER TABLE
+  "Project"
+ALTER COLUMN
+  "previewImageAssetId"
+SET
```
istarkov, line 6: Why not null? If not null then what is default value.

packages/prisma-client/prisma/migrations/20240127230238_project-preview/migration.sql
istarkov: Left join assetId with assets table to not use all that logic with assets.findMany

packages/prisma-client/prisma/schema.prisma
```diff
@@ -135,18 +135,19 @@ view UserProduct {
...
+  domain              String                 @unique
+  user                User?                  @relation(fields: [userId], references: [id])
+  userId              String?
+  build               Build[]
+  isDeleted           Boolean                @default(false)
+  files               File[]
+  projectDomain       ProjectDomain[]
+  latestBuild         LatestBuildPerProject?
+  authorizationToken  AuthorizationToken[]
+  previewImageAssetId String
```
istarkov, line 150: its not string but relation to assets table, with relation you will not even need to left join view

packages/prisma-client/prisma/migrations/20240127230238_project-preview/migration.sql
istarkov: or see other comment, assetId is relation like user_id user etc

istarkov (commented):
Using prisma or sql will simplify code

istarkov (approved):

PR #2844, opened 2024-01-30
refactor: generate page meta in loader

TrySound:
Here refactored page meta with simpler type than full page data and moved it into factory which is invoked lazily in loader.

This will let us support variables and resources in page settings later.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/page-meta-generator.test.ts
+++ b/packages/react-sdk/src/page-meta-generator.test.ts
@@ -16,7 +16,7 @@ test("generate minimal static page meta factory", () => {
     })
   ).toMatchInlineSnapshot(`
 "export const getPageMeta = ({}: {
-  params: Record<string, any>;
+  params: Record<string, undefined | string>;
   resources: Record<string, any>;
 }): PageMeta => {
   return {
@@ -55,7 +55,7 @@ test("generate complete static page meta factory", () => {
     })
   ).toMatchInlineSnapshot(`
 "export const getPageMeta = ({}: {
-  params: Record<string, any>;
+  params: Record<string, undefined | string>;
   resources: Record<string, any>;
 }): PageMeta => {
   return {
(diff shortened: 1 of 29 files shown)


packages/react-sdk/src/page-meta-generator.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        path: "",
+        rootInstanceId: "",
+        title: "Page title",
+        meta: {},
+      },
+      dataSources: new Map(),
+    })
+  ).toMatchInlineSnapshot(`
+"export const getPageMeta = ({}: {
+  params: Record<string, any>;
```
kof, line 19: can it be a not any? just wondering

kof (approved):

packages/react-sdk/src/page-meta-generator.ts
```diff
@@ -0,0 +1,53 @@
...
+  const titleExpression = JSON.stringify(page.title);
+  const descriptionExpression = JSON.stringify(page.meta.description);
+  const excludePageFromSearchExpression = JSON.stringify(
+    page.meta.excludePageFromSearch
+  );
+  const socialImageAssetIdExpression = JSON.stringify(
+    page.meta.socialImageAssetId
+  );
+  let generated = "";
+  generated += `export const getPageMeta = ({}: {\n`;
```
kof, line 26: there really must be a better way to build that than this ... maybe fn.toString() could do this So we can write normal functions, then turn them to strings or use a parser/compiler ...

TrySound: toString would not work because code is dynamic based on data. Static functions are just imported from sdk.

Using any compiler is a hell of an overhead for node_modules. Own stuff would require maintaining.

PR #2853, opened 2024-01-31
refactor: convert page meta fields into expressions

TrySound:
Here added huge migration to convert page meta into expressions and adapted the code to compute those.

## Code Review

  - test it on preview (creating new project/page should work without migration)

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound:
Cli tests will be fixed after migration is complete.

packages/react-sdk/src/expression.ts
```diff
@@ -209,3 +209,12 @@ export const generateExpression = ({
     },
   });
 };
+
+export const executeExpression = (expression: undefined | string) => {
```
kof, line 213: how did we execute epressions before? weird we didn't have that eval so far

packages/sdk/src/schema/pages.ts
```diff
@@ -43,7 +43,7 @@ const commonPageFields = {
   meta: z.object({
     description: z.string().optional(),
     title: z.string().optional(),
-    excludePageFromSearch: z.boolean().optional(),
+    excludePageFromSearch: z.string().optional(),
```
kof, line 46: why is that a string now?

kof: because its an expression which is a string? basically we don't know the resulting value type

kof (approved):
tested that migration passes and page settings are still editing

packages/react-sdk/src/expression.ts
TrySound: Yes, it won't. This is cli only utility to extract value from excludePageFromSearch if possible. We never needed this in cli because all expressions were executed in generated runtime.

PR #2859, opened 2024-02-02
feat: support bindings in page meta

TrySound:
Here added support for bindings in page meta for following fields

- title
- description
- social image url
- exclude from search
- custom meta content

Screenshot 2024-02-02 at 17 40 37
Screenshot 2024-02-02 at 17 40 44

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Amazing, only one problem is bindin dot on a checkbox

kof:
maybe something like this would work?
image

also in this case it definitely feels too hard to access if only button/input hover is howing it and not the label, since those are tiny, I would totally miss it since I can use the label to check/uncheck

kof (commented):
2 things to improve regarding design/ux

kof (approved):

PR #2861, opened 2024-02-02
chore: Add DIRECT_URL to the .env.example file

JayaKrishnaNamburu:
## Description

Connecting to database now needs `DIRECT_URL` too, referene
https://github.com/webstudio-is/webstudio/blob/main/packages/prisma-client/prisma/schema.prisma#L13-L17

```sh
datasource db {
  provider  = "postgres"
  url       = env("DATABASE_URL")
  directUrl = env("DIRECT_URL")
}
```

## Code Review

## Before requesting a review

- [x] made a self-review

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "storybook": "pnpm run --parallel --recursive storybook:dev",
     "storybook:build": "storybook build",
-    "migrations": "cd apps/builder && pnpm migrations --dev",
+    "migrations": "cd apps/builder && pnpm migrations migrate --dev",
     "build-figma-tokens": "cd packages/design-system && pnpm build-figma-tokens",
     "prepare": "git config core.hooksPath .git/hooks/ && simple-git-hooks",
     "local:version-snapshot": "pnpm -r exec pnpm version prepatch --preid $(cat /dev/urandom | LC_ALL=C tr -dc 'a-z' | fold -w 8 | head -n 1)",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "storybook": "pnpm run --parallel --recursive storybook:dev",
     "storybook:build": "storybook build",
-    "migrations": "cd apps/builder && pnpm migrations migrate --dev",
+    "migrations": "cd apps/builder && pnpm migrations --dev",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "storybook": "pnpm run --parallel --recursive storybook:dev",
     "storybook:build": "storybook build",
-    "migrations": "cd apps/builder && pnpm migrations --dev",
+    "migrations": "cd apps/builder && pnpm migrations migrate --dev",
```
JayaKrishnaNamburu, line 17: When i run migrations in local, it needs either `--dev` or `--force` flag as mandatory. I am manually adding when i run migrations till now.

kof: @istarkov

kof:
also its weird that locally user needs both  DATABASE_URL and DIRECT_URL to work, it should be possible to use one variable not 2, I don't really understand how we ended up having both of them

package.json
istarkov: Prev command already had --dev. nothing has changed except you add `migration`

JayaKrishnaNamburu: Apologies, I mixed up. It was the missing _migrate_ command itself

istarkov:
It's 2 different variables on all envs except local. Why we need to do something special only for local development just to reduce one line in local environment.

package.json
istarkov: we use it not only with migrate ;)

istarkov: the only reason it exists is —dev flag ;)

istarkov:
Anyway some env providers, like infisical allows to interpolate env variables like
DIRECT=${DATABASE} 
that simplifies setup of secrets as all variables names exists in all envs (just some interpolated some are not)
infisical ui shows env diffs and absence of some variable will cause even more questions

kof:
> It's 2 different variables on all envs except local. Why we need to do something special only for local development just to reduce one line in local environment.

because its confusing when you just want to quickly get it up and running to contribute, we should fallback to one variable and only require one variable locally for simplicity

Do we need 2 of them in the cloud? or is only one needed depending on how we connect? is it possible to define the one that is being needed?

Can we name them semantically distinct so that I can understand from the naming which one is for which purpose?

kof:
as for more distinct naming, this should work:

url       = env("DATABASE_URL")
directUrl = env("DATABASE_DIRECT_URL")

kof:
fixing that naming in prisma https://github.com/prisma/docs/pull/5612

kof (approved):

PR #2862, opened 2024-02-03
feat: Use a number as a shortcut for breakpoints without modifiers

kof:
## Description

Both ctrl/cmd + number are used to in browsers to switch between tabs. Webflow is just using a number shortcut without modifier, lets do the same

## Steps for reproduction

1. hit numbers to switch breakpoints
2. can still type numbers in inputs

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/commands.ts
+++ b/apps/builder/app/builder/shared/commands.ts
@@ -30,7 +30,7 @@ const makeBreakpointCommand = <CommandName extends string>(
   number: number
 ): Command<CommandName> => ({
   name,
-  defaultHotkeys: [`meta+${number}`, `ctrl+${number}`],
+  defaultHotkeys: [`${number}`],
   handler: () => {
     selectBreakpointByOrder(number);
   },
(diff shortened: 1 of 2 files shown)


apps/builder/app/shared/commands-emitter.ts
```diff
@@ -57,7 +57,7 @@ const findCommandsMatchingHotkeys = (event: KeyboardEvent) => {
     pressedKeys.add("shift");
   }
   if (altKey) {
-    pressedKeys.add("altKey");
+    pressedKeys.add("alt");
```
kof, line 60: drive-by fix, I think we are not using alt yet, but it was inconsistent with others

apps/builder/app/builder/shared/commands.ts
```diff
@@ -30,7 +30,8 @@ const makeBreakpointCommand = <CommandName extends string>(
   number: number
 ): Command<CommandName> => ({
   name,
-  defaultHotkeys: [`meta+${number}`, `ctrl+${number}`],
+  defaultHotkeys: [`${number}`],
+  disableHotkeyOnFormTags: true,
```
TrySound, line 34: this too
disableHotkeyOnContentEditable: true,

apps/builder/app/shared/commands-emitter.ts
TrySound: good catch

TrySound (approved):

PR #2867, opened 2024-02-04
fix: Normalize border style

kof:
## Description

We don't support inset/outset in the style panel which is the initial for inputs/buttons

New defaults are:

- input, textarea, button - solid
- radio, checkbox - none (already was none, no change,but we didn't reflect that in sp)

## Steps for reproduction

1. add those components to canvas
2. see that  input, textarea, button - have solid border style in style panel
3. see that radio, checkbox - show none

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/css/normalize-type-check.ts
+++ b/packages/react-sdk/src/css/normalize-type-check.ts
@@ -7,7 +7,8 @@ import type { Style } from "@webstudio-is/css-engine";
 const normalizeWithKeyof = { ...normalize };
 
 type ExportedTags = keyof typeof normalizeWithKeyof;
+type ExpectedTags = HtmlTags | "checkbox" | "radio";
 
-type ValidTags = ExportedTags extends HtmlTags ? ExportedTags : false;
+type ValidTags = ExportedTags extends ExpectedTags ? ExportedTags : false;
 
 normalizeWithKeyof satisfies Record<ValidTags, Style>;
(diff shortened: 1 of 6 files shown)


TrySound (approved):

packages/react-sdk/src/css/normalize-type-check.ts
```diff
@@ -7,7 +7,8 @@ import type { Style } from "@webstudio-is/css-engine";
 const normalizeWithKeyof = { ...normalize };
 
 type ExportedTags = keyof typeof normalizeWithKeyof;
+type ExpectedTags = HtmlTags | "checkbox" | "radio";
```
TrySound, line 10: To not break this semantic you can just inline those presets into components.

kof: Sure the question is whats the rightful place for this stuff? what is this typecheck even doing?

Type checking this is hard because we only use tags and those are the same tag with different type combination, to get a type checker here we need {tag, type} object

TrySound (approved):

packages/react-sdk/src/css/normalize-type-check.ts
kof: discussed, we are keeping those as exceptions, added a comment to explain

PR #2871, opened 2024-02-05
feat: Clonable project

kof:
## Description

Every project is now clonable if you have a shared link with at least view permission or higher.

## Steps for reproduction

1. get a link with view permission from another owner
2. clone it
3. repeat the same from your own project

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk/src/schema/pages.ts
```diff
@@ -114,26 +115,24 @@ export const ProjectNewRedirectPathSchema = PagePath.or(
...
-export type ProjectMeta = z.infer<typeof ProjectMeta>;
+export type CompilerSettings = z.infer<typeof CompilerSettings>;
 
 export type Page = z.infer<typeof Page>;
 
 export const Pages = z.object({
   meta: ProjectMeta.optional(),
-  settings: ProjectSettings.optional(),
+  compiler: CompilerSettings.optional(),
+  redirects: z.array(PageRedirect).optional(),
```
TrySound, line 135: These seem like completely unrelated changes and will break atomic styles. Let's do renamings separately.

apps/builder/app/shared/router-utils/path-utils.ts
```diff
@@ -73,6 +73,8 @@ export const builderDomainsPath = (method: string) => {
   }`;
 };
 
+export const projectPath = (method: string) => `/project/${method}`;
```
TrySound, line 76: doesn't constrain anything, maybe just write path where it is used?

apps/builder/app/shared/clone-project/clone-project.tsx
```diff
@@ -0,0 +1,148 @@
+import { useState } from "react";
```
TrySound, line 1: apps/builder/app/shared/clone-project.tsx

apps/builder/app/builder/features/project-settings/meta-section.tsx
```diff
@@ -26,29 +28,41 @@ const imgStyle = css({
...
-      });
-    };
+const defaultMetaSettings: ProjectMeta = {
+  siteName: "",
+  faviconAssetId: "",
+  code: "",
+};
 
+export const MetaSection = () => {
+  const [meta, setMeta] = useState($pages.get()?.meta ?? defaultMetaSettings);
```
TrySound, line 38: please use function initializer, constant calling of .get in render deoptimizes stores

packages/sdk/src/schema/pages.ts
kof: I needed to clean up settings, could be in stages, but really would take much more time to make in separate prs. I think nobody uses that css setting, but to be sure we can check in prod and add a migration if needed

apps/builder/app/shared/router-utils/path-utils.ts
kof: we have all paths in this file, is spreading them across the codebase a good idea?

apps/builder/app/builder/features/project-settings/meta-section.tsx
kof: do you mean using useStore?

apps/builder/app/shared/router-utils/path-utils.ts
TrySound: What the different slightly refined path would make? "method" would spread the same way

apps/builder/app/builder/features/project-settings/meta-section.tsx
TrySound: I mean useState(() => here)

kof: actually since I changed how we save, we do it now right away without waiting for blur or close, we can as well use useStore, because there is no need to have a stale/static state any more, wdyt?

apps/builder/app/shared/router-utils/path-utils.ts
kof: yes, its just about keeping all paths in one place and if needed extend from here in case we need a change

packages/project/src/shared/schema.ts
```diff
@@ -11,6 +11,7 @@ export const Title = z
 
 export const ClonableSettings = z.object({
   isClonable: z.boolean().optional(),
+  isPublic: z.boolean().optional(),
```
TrySound, line 14: Aren't those exclusive?

kof: no, public is meant as in can be listed publicly on a directory, where clonable can be any project with even view link permission

TrySound: I'd replace isPublic with  visibility: private | public

PR #2897, opened 2024-02-11
feat: Autosave settings

kof:
## Description

This makes all changes in settings panel inside inputs, textareas and code input save automatically with slight debounce, without waiting for blur etc.

## Steps for reproduction

1. try to change any input/textarea/code
2. change should apply right away

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/settings-panel/shared.tsx
```diff
@@ -155,7 +156,8 @@ export const Label = ({
 
 export const useLocalValue = <Type,>(
   savedValue: Type,
-  onSave: (value: Type) => void
+  onSave: (value: Type) => void,
+  { autoSave = true } = {}
```
TrySound, line 160: why make it optional?

kof: to avoid passing it everywhere since so far I didn't see a need to customize it at all, maybe we should just remove it

TrySound (approved):

PR #2898, opened 2024-02-11
experimental: generate variables for page meta website

TrySound:
Here added support for variables in page meta in generated websites. Prevented creating unused variables. We can reuse the same way for other generated parts later.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/page-meta-generator.ts
+++ b/packages/react-sdk/src/page-meta-generator.ts
@@ -17,36 +17,34 @@ export type PageMeta = {
 };
 
 export const generatePageMeta = ({
-  scope: resourcesScope,
+  globalScope,
   page,
   dataSources,
 }: {
-  scope: Scope;
+  globalScope: Scope;
   page: Page;
   dataSources: DataSources;
 }) => {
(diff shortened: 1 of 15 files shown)


packages/react-sdk/src/page-meta-generator.ts
```diff
@@ -11,27 +17,33 @@ export type PageMeta = {
...
+  scope: resourcesScope,
   page,
   dataSources,
 }: {
   scope: Scope;
   page: Page;
   dataSources: DataSources;
 }) => {
+  // reserve initial parameters
+  const scope = createScope(["params", "resources"]);
```
kof, line 29: is this initialScope and the one passed the real one? I got a bit confused  with semantic difference between passed scope and this one and why the other one is maped as resourcesScope

kof: if the generatePageMeta always received resourcesScope maybe its better to call it that way and not map?

kof (approved):

PR #2899, opened 2024-02-11
fix: Fix shared link deletion crash

kof:
## Description

When deleting a link, blur event from link name triggers an update, which executes after link has been deleted

## Steps for reproduction

1. create a link
2. delete it
3. should not crash

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound:
Crasheд when put focus into name fielд anд then click дelete.

kof:
@TrySound that was the behaviour before this fix, I can't reproduce it any more

TrySound:
Looks like can't reproduce anymore

TrySound (approved):

PR #2901, opened 2024-02-12
feat: Add all input types

kof:
## Description

With this pr we are enabling all input types

https://github.com/webstudio-is/webstudio/assets/52824/0a09928d-68b6-4e0d-964e-7c19e92dc61a

## Steps for reproduction

1. add text input
2. try different types

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/text.tsx
```diff
@@ -9,7 +9,9 @@ type Props = ComponentProps<typeof defaultTag> & {
 };
 
 export const Text = forwardRef<ElementRef<typeof defaultTag>, Props>(
-  ({ tag: Tag = defaultTag, children, ...props }, ref) => {
+  ({ tag = defaultTag, children, ...props }, ref) => {
+    // Can't map it in the destricturing, default type won't be generated correctly
+    const Tag = tag;
```
TrySound, line 14: how does it work now without this change?

kof: it ignores default value if you add the mapping to the argument ...

TrySound (approved):

packages/sdk-components-react/src/text.tsx
kof: My guess is that someone chagned it to use destructuring/maping and didn't rebuild or we upgraded the compiler library and it broke

PR #2907, opened 2024-02-13
chore: Change tabs order

kof:
## Description

Just changed the order because it makes no sense to have assets before pages and navigator

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/sidebar-left/panels/index.ts
```diff
@@ -5,7 +5,7 @@ import * as pages from "./pages";
 
 export const panels = {
   components,
+  pages,
```
TrySound, line 8: object fields order is not guaranteed. Better use the other structure

kof: not officially but its actually guaranteed, but lets make it more clear with a change

TrySound (approved):

PR #2931, opened 2024-02-21
refactor: move insert instance children logic into utility

TrySound:
Here split and consolidated utility for adding children to instances. It can be universally reused almost everywhere.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/copy-paste/plugin-instance.ts
+++ b/apps/builder/app/shared/copy-paste/plugin-instance.ts
@@ -25,7 +25,7 @@ import {
   isInstanceDetachable,
   updateWebstudioData,
   getWebstudioData,
-  addInstanceChildrenMutable,
+  insertInstanceChildrenMutable,
 } from "../instance-utils";
 import { portalComponent } from "@webstudio-is/react-sdk";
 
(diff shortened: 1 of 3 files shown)


apps/builder/app/shared/copy-paste/plugin-instance.ts
```diff
@@ -204,32 +199,10 @@ export const onPaste = (clipboardData: string): boolean => {
...
-    const { position } = dropTarget;
-    if (position === "end") {
-      parentInstance.children.push(child);
-    } else {
-      parentInstance.children.splice(position, 0, child);
-    }
+    const children: Instance["children"] = [
+      { type: "id", value: newRootInstanceId },
+    ];
+    addInstanceChildrenMutable(data, children, pasteTarget);
```
kof, line 205: add is usually considered "at the end" semantically, I think "insert" is more correct

kof: if this is also inserting at any position

kof (approved):

PR #2939, opened 2024-02-23
refactor: render html for whole route

TrySound:
Here move root html to every route to be able to set html lang attribute. Split separately to test before whole language feature.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

istarkov (changes requested):
```ts
import { Links, Meta, Outlet as DefaultOutlet } from "@remix-run/react";
import { useEffect } from "react";

const Remount = () => {
  useEffect(() => {
    console.log("MOUNT");
    return () => console.log("UNMOUNT");
  }, []);
  return null;
};
/**
 * We are using Outlet prop from index layout when user renders project from a subdomain.
 */
export const Root = ({
  Outlet = DefaultOutlet,
}: {
  Outlet: typeof DefaultOutlet;
}) => {
  return (
    <html lang="en">
      <head>
        <meta charSet="utf-8" />
        <meta name="viewport" content="width=device-width,initial-scale=1" />
        <Meta />
        <Remount />
        <Links />
      </head>

      <Outlet />
    </html>
  );
};
```

In original implemetation doesn't remount during local routing

New implementation probably will do remount what can kill some 3rd parties etc

I would test same as above behavior as I think we need to preserve it

PR #2945, opened 2024-02-25
fix: replace unknown variables on paste

TrySound:
Here fixing the issue with saving expressions with variables not available in the scope which leads to user confusion and compilation errors.

image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/expression-editor.tsx
+++ b/apps/builder/app/builder/shared/expression-editor.tsx
@@ -424,7 +424,8 @@ export const ExpressionEditor = ({
         onChange={(value) => {
           try {
             let hasReplacements = false;
-            // replace unknown webstudio variables with null to prevent invalid compilation
+            // replace unknown webstudio variables with undefined
+            // to prevent invalid compilation
             const newExpression = validateExpression(value, {
               effectful: true,
               transformIdentifier: (identifier) => {
@@ -433,7 +434,7 @@ export const ExpressionEditor = ({
                   aliases.has(identifier) === false
                 ) {
                   hasReplacements = true;
-                  return `null`;
+                  return `undefined`;
                 }
                 return identifier;
               },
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/shared/expression-editor.tsx
```diff
@@ -417,7 +421,33 @@ export const ExpressionEditor = ({
...
+            // replace unknown webstudio variables with null to prevent invalid compilation
+            const newExpression = validateExpression(value, {
+              effectful: true,
+              transformIdentifier: (identifier) => {
+                if (
+                  decodeDataSourceVariable(identifier) &&
+                  aliases.has(identifier) === false
+                ) {
+                  hasReplacements = true;
+                  return `null`;
```
kof, line 436: I am not sure about `null`, its shorter, but null means a specific value, where undefined is more like it literally isn't defined, semantically much clearer

kof (approved):

apps/builder/app/builder/shared/expression-editor.tsx
kof: Also if you imagine that the variable user is trying to access hasn't been assigned a value, the initial value is undefined in js

kof: we can also consider using something completely different to not mix up with existing semantics, eg.. `unknown`

kof: If you think semantics of undefined are misleading in this case, I think they are great, null doesn't work for me.

kof: Btw do we store the id for that insertion if we can't find it immediately in case it appears later? Will it start working when that variable exists in the scope?

TrySound: No, this will not longer store the id. The variable is completely replaced from expression.

I can try to make undefined work. It seems not supported out of the box by jsep.

PR #2949, opened 2024-02-26
experimental: Marketplace settings

kof:
## Description

- split project settings
- add marketplace settings

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/project/src/db/project.ts
```diff
@@ -246,3 +246,21 @@ export const updateDomain = async (
...
+  {
+    projectId,
+    marketplaceApprovalStatus,
+  }: {
+    projectId: Project["id"];
+    marketplaceApprovalStatus: MarketplaceApprovalStatus;
+  },
+  context: AppContext
+) => {
+  await assertEditPermission(projectId, context);
```
istarkov, line 260: Looks wrong, as in this case anyone with edit permission can set any `marketplaceApprovalStatus`
but the idea that final status can be set only by predefined lists of moderators.

kof: done

apps/builder/app/builder/features/project-settings/section-general.tsx
```diff
@@ -129,6 +130,52 @@ export const MetaSection = () => {
...
+
+  return (
+    <Grid gap={2} css={{ mx: theme.spacing[5], px: theme.spacing[5] }}>
+      <Label htmlFor={ids.atomicStyles}>Compiler</Label>
+      <CheckboxAndLabel>
+        <Checkbox
+          checked={settings.atomicStyles ?? true}
+          id={ids.atomicStyles}
+          onCheckedChange={(atomicStyles) => {
+            if (typeof atomicStyles === "boolean") {
```
TrySound, line 168: we may want to exclude indeterminate from our component. Caz here I don't know at what point it will appear and block user input.

TrySound (approved):

apps/builder/app/builder/features/project-settings/section-general.tsx
kof: I think it can only be indeterminate if there is no default checked value

apps/builder/app/builder/features/project-settings/project-settings.tsx
```diff
@@ -20,27 +34,80 @@ const ProjectSettingsView = ({
...
+  >(settingNames[0]);
+
   return (
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
       <DialogContent
         // Left Aside panels (e.g., Pages, Components) use zIndex: theme.zIndices[1].
         // For a dialog to appear above these panels, both overlay and content should also have zIndex: theme.zIndices[1].
         css={{
-          width: theme.spacing[34],
+          width: "max-content",
```
istarkov, line 47: IMO this can lead cause serious width issues with Dialog having that max-width is set to none

istarkov: Also it's hard to detect what child gives width for the whole element

apps/builder/app/builder/features/project-settings/section-marketplace.tsx
```diff
@@ -0,0 +1,322 @@
...
+const useMarketplaceApprovalStatus = () => {
+  const { send, data, state } = trpc.setMarketplaceApprovalStatus.useMutation();
+  const project = useStore($project);
+
+  const status =
+    data?.marketplaceApprovalStatus ??
+    project?.marketplaceApprovalStatus ??
+    "UNLISTED";
+
+  useEffect(() => {
```
istarkov, line 71: We can use onSuccess send callback without using `useEffect`
Also every mutation can have errors, what if it fails?

apps/builder/app/builder/features/project-settings/section-redirects.tsx
```diff
@@ -118,16 +120,19 @@ export const RedirectSection = () => {
...
-      {
-        old: oldPath,
-        new: newPath,
-        status: httpStatus ?? "301",
-      },
-      ...redirects,
-    ]);
-    setOldPath("");
-    setNewPath("");
+    flushSync(() => {
```
istarkov, line 123: Imo flushSync must have comments why it's used here and what it solves.

packages/project/src/db/project.ts
```diff
@@ -246,3 +246,28 @@ export const updateDomain = async (
...
+    marketplaceApprovalStatus: MarketplaceApprovalStatus;
+  },
+  context: AppContext
+) => {
+  if (
+    marketplaceApprovalStatus === "APPROVED" ||
+    marketplaceApprovalStatus === "REJECTED"
+  ) {
+    // User can't approve or reject
+    throw new Error("Invalid status");
```
istarkov, line 265: Comment can be moved into Error

istarkov (approved):

apps/builder/app/builder/features/project-settings/section-marketplace.tsx
kof: OHH NICE, I forgot we have onSuccess and hated this effect while writing

apps/builder/app/builder/features/project-settings/project-settings.tsx
kof: done

PR #2955, opened 2024-02-29
feat: Add filter section in style panel

JayaKrishnaNamburu:
## Description

This PR adds the `filter` property in css. By default we add a `blur(0px)` when users click on `+`. And then users will be able to edit the style. The editing experience is same as editing and managing `transitions`, `box-shadows`.

User should be able to `sort`, `add`, `delete` layers. Right now the style value is parsed and separated into `tuple`. And in the next iteration we can work on UI for individual `filter-function`. Eg: `blur`, `sepia`, `hue-rotate` takes different type of props etc.

## Steps for reproduction
- Add a filter
- Edit a filter
- Add multiple filter by pasting multiple filter functions at once.

## Code Review

- [ ] hi @kof , I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/sections/box-shadows/box-shadow-layer.tsx
```diff
@@ -18,8 +18,12 @@ import { toValue } from "@webstudio-is/css-engine";
 import { ColorThumb } from "../../shared/color-thumb";
 import type { LayerProps } from "../../style-layers-list";
 
-const useLayer = (layer: TupleValue) => {
+const useLayer = (layer: LayerValueItem) => {
   return useMemo(() => {
+    if (layer.type !== "tuple") {
+      return undefined;
```
kof, line 24:
```suggestion
      return;
```

apps/builder/app/builder/features/style-panel/sections/box-shadows/box-shadow-layer.tsx
```diff
@@ -51,7 +55,13 @@ const useLayer = (layer: TupleValue) => {
 
 export const Layer = (props: LayerProps) => {
   const { index, id, layer, isHighlighted, onDeleteLayer, onLayerHide } = props;
-  const { name, shadow, color } = useLayer(layer);
+  const properties = useLayer(layer);
+
+  if (layer.type !== "tuple" || properties === undefined) {
+    return null;
```
kof, line 61:
```suggestion
    return;
```

apps/builder/app/builder/features/style-panel/sections/transitions/transition-layer.tsx
```diff
@@ -32,6 +39,17 @@ export const Layer = (props: LayerProps) => {
...
     return { transition: toValue(layer), label: toValue(label) };
   }, [layer]);
+};
+
+export const Layer = (props: LayerProps) => {
+  const { id, index, layer, isHighlighted, onDeleteLayer, disabled } = props;
+  const properties = useLayer(layer);
+
+  if (layer.type !== "tuple" || properties === undefined) {
+    return null;
```
kof, line 49:
```suggestion
    return;
```

TrySound:
How does it relate to backdrop-filter?

kof:
@TrySound we need to do the same for backdrop filter, building out both at the same time makes sense to me if they are the same, probably even reusing exactly the same code

TrySound:
My worry is it good for ui to implement so many properties as sections?

kof:
Looking at Webflow's effects section, really classify as effects only: transitions, filters, backdrop-filteres, transforms, blending

kof:
- [ ] @JayaKrishnaNamburu please make sure we don't need to migrate the data structure when we start adding the UI, the data structure is better be final upfront, because migrating would be lots of work

apps/builder/app/builder/features/style-panel/sections/filter/filter-content.tsx
```diff
@@ -0,0 +1,100 @@
...
+        rows={3}
+        name="description"
+        value={intermediateValue?.value ?? filter ?? ""}
+        css={{ minHeight: theme.spacing[14], ...textVariants.mono }}
+        state={intermediateValue?.type === "invalid" ? "invalid" : undefined}
+        onChange={handleChange}
+        onKeyDown={(event) => {
+          if (event.key === "Enter") {
+            handleComplete();
+            event.preventDefault();
```
istarkov, line 94: I would comment the reason why preventDefaut is used

apps/builder/app/builder/features/style-panel/sections/filter/filter-layer.tsx
```diff
@@ -0,0 +1,50 @@
...
+        index={index}
+        label={<Label truncate>{layer.value}</Label>}
+        buttons={
+          <SmallIconButton
+            variant="destructive"
+            tabIndex={-1}
+            icon={<SubtractIcon />}
+            onClick={(event) => {
+              onDeleteLayer(index);
+              event.preventDefault();
```
istarkov, line 43: Why prevent?

apps/builder/app/builder/features/style-panel/sections/filter/filter.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <SectionTitle
+          dots={getDots(currentStyle, [property])}
+          suffix={
+            <Tooltip content={"Add a filter"}>
+              <SectionTitleButton
+                prefix={<PlusIcon />}
+                onClick={() => {
+                  addLayer(
+                    property,
+                    parseFilter("blur(0px)"),
```
istarkov, line 45: Would be good to give a name for "blur(0px)" like INITIAL_FILTER = "blur(0px)"

apps/builder/app/builder/features/style-panel/sections/filter/filter.tsx
```diff
@@ -0,0 +1,82 @@
...
+              <SectionTitleLabel color={layerStyleSource}>
+                {label}
+              </SectionTitleLabel>
+            }
+            onReset={() => deleteProperty(property)}
+          />
+        </SectionTitle>
+      }
+    >
+      {filterValue?.type === "tuple" && filterValue.value.length > 0 && (
```
istarkov, line 70: Am I right that we use tuple type here just because it's joined with space in our style rendering https://github.com/webstudio-is/webstudio-builder/blob/de7f764932477e6702d481e9459c16f5bfd1d62a/packages/css-engine/src/core/to-value.ts#L112
Would be good to comment this internal knowledge. Or probably we need copy of Tuple type with some semantically correct name. Not sure

packages/css-data/src/parse-css-value.ts
```diff
@@ -11,7 +11,7 @@ import {
 import { keywordValues } from "./__generated__/keyword-values";
 import { units } from "./__generated__/units";
 
-const cssTryParseValue = (input: string) => {
+export const cssTreeTryParseValue = (input: string) => {
```
istarkov, line 14: Why we need to expose `cssTree` (like internal implementation detail that we use csstree) whats wrong with previous name

packages/css-data/src/property-parsers/filter.ts
```diff
@@ -0,0 +1,63 @@
...
+      value: input,
+    };
+  }
+
+  const layers: TupleValueItem[] = [];
+  csstree.walk(cssAst, (node) => {
+    if (node.type === "Value") {
+      for (const child of node.children) {
+        if (child.type === "Function") {
+          layers.push({ type: "keyword", value: csstree.generate(child) });
```
istarkov, line 53: In no case it's keyword we have UnparsedValue for such things from what I know

packages/css-data/src/property-parsers/filter.ts
```diff
@@ -0,0 +1,63 @@
+import * as csstree from "css-tree";
```
istarkov, line 1: Where is the test?

istarkov (changes requested):

packages/css-data/src/parse-css-value.ts
JayaKrishnaNamburu: I just added `export` there ivan didn't change any name.

packages/css-data/src/property-parsers/filter.ts
JayaKrishnaNamburu: Can't test it, as i added a comment in filter parser. The css-tree parser can't validate `filter`. So, i am relying on a validator which uses `CSSStyleSheet` which is available only in browser. Should we polyfill with `jsdom` and then add tests. Currently it's not available, so it keeps failing 👍

packages/css-data/src/property-parsers/filter.ts
kof: shouldn't we have type: "function"?

kof: so far the only function we had was rgb() and we added rgb as a type, not as a generic function, maybe it makes sense to just keep adding specific functions? since you want to make sure each function interface is strictly defined

packages/css-data/src/parse-css-value.ts
istarkov: But name has chaged 
cssTryParseValue 
cssTreeTryParseValue
image

packages/css-data/src/property-parsers/filter.ts
istarkov: Probably add throw in non browser env then?

JayaKrishnaNamburu:
We are adding a `FunctionValue` type in the https://github.com/webstudio-is/webstudio/pull/2986 PR. Will revisit and update the parsing according to it once the PR is merged.

packages/css-data/src/property-parsers/filter.ts
JayaKrishnaNamburu: Then i am not sure, if it's helping in writing tests. Basically, we are not polyfiling any browser env here. So, the tests always throws error. And we are using inbuilt browser `CSSStyleValue` function itself and there is not custom logic that tests actually run in local.

apps/builder/app/builder/features/style-panel/sections/filter/filter-content.tsx
```diff
@@ -0,0 +1,97 @@
...
+        gap: theme.spacing[3],
+        minWidth: theme.spacing[30],
+      }}
+    >
+      <Label>
+        <Flex align={"center"} gap={1}>
+          Code
+          <Tooltip
+            variant="wrapped"
+            content={<Text>TODO: Add content for the filter tooltip</Text>}
```
kof, line 71: @todo

apps/builder/app/builder/features/style-panel/sections/filter/filter-content.tsx
```diff
@@ -0,0 +1,97 @@
...
+          >
+            <InfoCircleIcon />
+          </Tooltip>
+        </Flex>
+      </Label>
+      <TextArea
```
kof, line 77: @todo use the new code editor

JayaKrishnaNamburu: We added a separate ticket for this #2977 to handle this for the rest of the properties too.

kof: right, but its better to comunicate that there too just in case

apps/builder/app/builder/features/style-panel/sections/filter/filter-content.tsx
JayaKrishnaNamburu: Should i leave it as a todo, or can you point me a place where i can get some content to fill in here that would be great.

TrySound (approved):

apps/builder/app/builder/features/style-panel/sections/box-shadows/box-shadow-layer.tsx
```diff
@@ -18,8 +18,12 @@ import { toValue } from "@webstudio-is/css-engine";
 import { ColorThumb } from "../../shared/color-thumb";
 import type { LayerProps } from "../../style-layers-list";
 
-const useLayer = (layer: TupleValue) => {
+const useLayer = (layer: TupleValue | FunctionValue) => {
```
istarkov, line 21: Why Both supported? Imo better to always have strong type for Layer even if there is just 1 function

apps/builder/app/builder/features/style-panel/sections/box-shadows/box-shadow-layer.tsx
```diff
@@ -49,9 +53,15 @@ const useLayer = (layer: TupleValue) => {
   }, [layer]);
 };
 
-export const Layer = (props: LayerProps) => {
+export const BoxShadowLayer = (props: LayerProps) => {
   const { index, id, layer, isHighlighted, onDeleteLayer, onLayerHide } = props;
-  const { name, shadow, color } = useLayer(layer);
+  const properties = useLayer(layer);
+
+  if (layer.type !== "tuple" || properties === undefined) {
```
istarkov, line 60: I think Box Shadow Layer should receive only layer of type Tuple so props can be changed to have a right type and do type check above. Also this looks like an error but not something what can be just `return ;`
You need to show something like `toast` that incorrect type has came to this component

istarkov (approved):

PR #2957, opened 2024-03-04
feat: CLI Mutiple template support

istarkov:
## Description

We have that depending on some external parameters we will need to have custom builds. Like adding analytics, web vitals etc on specific resources.

Instead of relying on env variables (has different support on different build systems vite, esbuild etc)

We can just use additional templates override. i.e.

```bash
# Without analytics
webstudio build --template ./custom-template

# With analytics
webstudio build --template ./custom-template --template ./custom-template-analytics
```

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

JayaKrishnaNamburu (approved):
Looks good 👍

PR #2965, opened 2024-03-04
feat: reset shortcut hint for all style panel style info tooltips

kof:
Currently user had no way to discover option/alt + click on label in style panel

image

## Steps for reproduction

1. click on any property value that is set in style panel
2. see the hint

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/separator.stories.tsx
+++ b/packages/design-system/src/components/separator.stories.tsx
@@ -1,9 +1,9 @@
-import { Separator } from "./separator";
+import { Shortcut } from "./shortcut";
 
 export default {
-  title: "Library/Separator",
+  title: "Library/Shortcut",
 };
 
-const SeparatorStory = () => <Separator />;
+const ShortcutStory = () => <Shortcut value={["cmd", "z"]} />;
 
-export { SeparatorStory as Separator };
+export { ShortcutStory as Shortcut };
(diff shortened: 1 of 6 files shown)


packages/design-system/src/components/separator.stories.tsx
```diff
@@ -1,9 +1,9 @@
-import { Separator } from "./separator";
```
TrySound, line 1: are you sure you wanna do this to separator?

packages/design-system/src/components/shortcut.tsx
```diff
@@ -9,6 +9,7 @@ const shortcutSymbolMap: Record<string, string> = {
   shift: "⇧",
   option: "⌥",
   backspace: "⌫",
+  click: "+click",
 };
```
TrySound, line 13: we might wanna switch to Sitnik's keyux for this later

kof: ofc not, edited the wrong file

packages/design-system/src/components/separator.stories.tsx
kof: ofc not, edited the wrong file

packages/design-system/src/components/shortcut.tsx
kof: just realized I had forgotten there is kbd element, switched to using it, I am not sure how keyux will work with radix and stuff

PR #2972, opened 2024-03-05
fix: fix code editor scrolling

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/2956

Here fixed the issue with scrolling code editor
by setting height on editor instead of content.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/shared/expression-editor.tsx
+++ b/apps/builder/app/builder/shared/expression-editor.tsx
@@ -341,7 +341,9 @@ const emptyScope: Scope = {};
 const emptyAliases: Aliases = new Map();
 
 const wrapperStyle = css({
-  "& .cm-content": {
+  // do not override .cm-content here
+  // it may break scrolling
+  "& .cm-editor": {
     // 1 line is 16px
     // set and max 20 lines
     maxHeight: 320,
--- a/apps/builder/app/builder/shared/expression-editor.tsx
+++ b/apps/builder/app/builder/shared/expression-editor.tsx
@@ -341,13 +341,9 @@ const emptyScope: Scope = {};
 const emptyAliases: Aliases = new Map();
 
 const wrapperStyle = css({
-  // avoid modifying height in .cm-content
-  // because it breaks scroll events and makes scrolling laggy
-  "& .cm-editor": {
-    // 1 line is 16px
-    // set and max 20 lines
-    maxHeight: 320,
-  },
+  // 1 line is 16px
+  // set and max 20 lines
+  "--ws-code-editor-max-height": "320px",
 });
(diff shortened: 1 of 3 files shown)


kof (approved):

apps/builder/app/builder/shared/expression-editor.tsx
```diff
@@ -341,7 +341,9 @@ const emptyScope: Scope = {};
 const emptyAliases: Aliases = new Map();
 
 const wrapperStyle = css({
-  "& .cm-content": {
+  // do not override .cm-content here
+  // it may break scrolling
```
kof, line 345: please describe what the problem is we discovered, we won't be able to recall it later

kof: ideally ofc we have something reusable that has only a single place  where these classes are used at all ... these 2 places will be 5 places later

kof: I would love for it to be a component that  we can control over props and all those classes are hidden inside of it, working with those global classes shouldn't be necessary in react

kof: e.g. you could expose the necessary option via css variable if prop doesn't work

apps/builder/app/builder/shared/expression-editor.tsx
```diff
@@ -341,11 +341,9 @@ const emptyScope: Scope = {};
...
 
 const wrapperStyle = css({
-  "& .cm-content": {
-    // 1 line is 16px
-    // set and max 20 lines
-    maxHeight: 320,
-  },
+  // 1 line is 16px
+  // set and max 20 lines
+  "--ws-code-editor-max-height": "320px",
```
kof, line 346: you just violated our RFC https://github.com/webstudio-is/webstudio-rfcs/blob/main/text/002-css-vars.md

"We never copy-paste a CSS variable name across modules."

apps/builder/app/builder/shared/html-editor.tsx
```diff
@@ -68,12 +68,10 @@ const autocompletionStyle = css({
...
   position: "relative",
-  "& .cm-content": {
-    // 1 line is 16px
-    // set min 10 lines and max 20 lines
-    minHeight: 160,
-    maxHeight: 320,
-  },
+  // 1 line is 16px
+  // set min 10 lines and max 20 lines
+  "--ws-code-editor-min-height": "160px",
```
kof, line 73: same here

PR #2986, opened 2024-03-10
feat: Add `FunctionValue` to the css-engine to describe css functions effeciently.

JayaKrishnaNamburu:
## Description

This PR adds a new `FunctionValue` type to the `css-engine`. This helps in defining `blur(5px)`, `drop-shadow(10px 10px 10px red)` and `translate3d(42px, -62px, -135px)` type of functions more clearly. There is a flag in the `FunctionValue` with the name `seperator` which defined that type of separator to be used  while passing the function arguments. 
In `css-tree` they use a `Operator` node. But we don't have something similar to that in the schema. So, using a new flag to define it.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/packages/css-engine/src/core/to-value.test.ts
+++ b/packages/css-engine/src/core/to-value.test.ts
@@ -132,6 +132,33 @@ describe("Convert WS CSS Values to native CSS strings", () => {
     expect(value).toBe("10px 20px 30px 40px");
   });
 
+  test("function", () => {
+    const translate3D = toValue({
+      type: "function",
+      name: "translate3d",
+      seperator: ",",
+      args: [
+        { type: "unit", value: 42, unit: "px" },
+        { type: "unit", value: -62, unit: "px" },
+        { type: "unit", value: -135, unit: "px" },
+      ],
+    });
+
+    const dropShadowValue = toValue({
+      type: "function",
+      name: "drop-shadow",
+      args: [
+        { type: "unit", value: 10, unit: "px" },
+        { type: "unit", value: 10, unit: "px" },
+        { type: "unit", value: 10, unit: "px" },
+        { type: "keyword", value: "red" },
+      ],
+    });
+
+    expect(translate3D).toBe("translate3d(42px, -62px, -135px)");
+    expect(dropShadowValue).toBe("drop-shadow(10px 10px 10px red)");
+  });
+
   test("sanitize url", () => {
     const assets = new Map<string, { path: string }>([
       ["1234567890", { path: `fo"o\\o.png` }],
--- a/packages/css-engine/src/core/to-value.test.ts
+++ b/packages/css-engine/src/core/to-value.test.ts
@@ -136,23 +136,24 @@ describe("Convert WS CSS Values to native CSS strings", () => {
     const translate3D = toValue({
       type: "function",
       name: "translate3d",
-      seperator: ",",
-      args: [
-        { type: "unit", value: 42, unit: "px" },
-        { type: "unit", value: -62, unit: "px" },
-        { type: "unit", value: -135, unit: "px" },
-      ],
+      args: {
+        type: "keyword",
+        value: "42px, -62px, -135px",
+      },
     });
 
     const dropShadowValue = toValue({
       type: "function",
       name: "drop-shadow",
-      args: [
-        { type: "unit", value: 10, unit: "px" },
-        { type: "unit", value: 10, unit: "px" },
-        { type: "unit", value: 10, unit: "px" },
-        { type: "keyword", value: "red" },
-      ],
+      args: {
+        type: "tuple",
+        value: [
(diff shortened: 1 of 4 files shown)


packages/css-engine/src/core/to-value.test.ts
```diff
@@ -132,6 +132,33 @@ describe("Convert WS CSS Values to native CSS strings", () => {
     expect(value).toBe("10px 20px 30px 40px");
   });
 
+  test("function", () => {
+    const translate3D = toValue({
+      type: "function",
+      name: "translate3d",
+      seperator: ",",
```
kof, line 139: do you think we should have separator as an option in the type? is it really like this in CSS? My understanding was that when you see space separated arguments in a function e.g. `drop-shadow(16px 16px 20px red);` `16px 16px 20px red` are not 4 arguments, its one css value just like you have one value in `box-shadow: 60px -16px teal`

kof: Its even using a different type: Dimension vs Number and uses Operator. I think this is that way because you are looking at the AST, I would rather learn from CSS spec way of thinking

kof: Generally I don't expect css-tree package to be the most acknowledgeable source, css spec should be one

TrySound: The idea was is that args is not an array but a StyleValue type and can be either tuple or layers and rendered based on this info.

kof: was just reading on mdn drop-shadow filter:

> Applies the parameter <shadow> as a drop shadow, following the contours of the image. The shadow syntax is similar to <box-shadow> (defined in the [CSS backgrounds and borders module](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_backgrounds_and_borders)), with the exception that the inset keyword and spread parameter are not allowed.

"shadow syntax is similar to <box-shadow>"

That means it is not literally reusing the same implementation, its a standalone one and so they have differences ...

kof: I would still look at this more from data perspective than from syntax, meaning the comma vs space thing is a "syntax" concern, though we are not building an AST, we are serializing the data.

kof: In terms of data I still think there is a function and there is arguments, in one case its multiple arguments and in another case its just one and this feels to me like one box shadow value

kof: > The idea was is that args is not an array but a StyleValue type and can be either tuple or layers and rendered based on this info.

@TrySound so there is always ONE "argument" - that makes sense to me too

JayaKrishnaNamburu: Sounds good to me, I am going to push an update now 👍🏽

JayaKrishnaNamburu: For reference, here is something

```js
{
      type: "function",
      name: "drop-shadow",
      args: {
        type: "tuple",
        value: [
          { type: "unit", value: 10, unit: "px" },
          { type: "unit", value: 10, unit: "px" },
          { type: "unit", value: 10, unit: "px" },
          { type: "keyword", value: "red" },
        ],
      },
 }
```

kof: args vs arg? if its always just one

kof (approved):

packages/css-engine/src/core/to-value.test.ts
kof: we can keep it args though and just use one, this is probably more natural for a funciton

TrySound: Args is used in that css parser api

packages/css-engine/src/schema.ts
```diff
@@ -54,6 +54,20 @@ const RgbValue = z.object({
...
+
+export const FunctionValue: z.ZodType<{
+  type: "function";
+  name: string;
+  seperator?: string | undefined;
+  args: StyleValue;
+}> = z.object({
+  type: z.literal("function"),
+  name: z.string(),
+  seperator: z.string().optional(),
```
kof, line 67: this is supposed to be removed

packages/css-engine/src/schema.ts
```diff
@@ -54,6 +54,20 @@ const RgbValue = z.object({
 });
 export type RgbValue = z.infer<typeof RgbValue>;
 
+export type FunctionValue = z.infer<typeof FunctionValue>;
+
+export const FunctionValue: z.ZodType<{
+  type: "function";
+  name: string;
+  seperator?: string | undefined;
```
TrySound, line 62: Not needed anymore

TrySound (approved):

PR #2987, opened 2024-03-10
feat: Make local style visually more distinct

kof:
Based on feedback, its not clear that Local style is special:
1. can't be reordered
2. can't be reused

To explain this visually better, we:

1. replaced Local with an icon
2. placed it after the input
3. added label in the menu
4. added hint in the menu

Screenshot 2024-03-10 at 15 32 20
Screenshot 2024-03-10 at 15 32 09
Screenshot 2024-03-10 at 15 32 04
Screenshot 2024-03-10 at 15 31 53

image

image

## Steps for reproduction

1. click button
5. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/style-source/style-source.tsx
```diff
@@ -110,21 +111,19 @@ export type ItemSource = "token" | "componentToken" | "tag" | "local";
...
 const EditableText = ({
   label,
-  isEditable,
   isEditing,
   onChangeEditing,
   onChangeValue,
 }: EditableTextProps) => {
   const { ref, handlers } = useContentEditable({
-    isEditable,
+    isEditable: false,
```
TrySound, line 126: did you mean true?

apps/builder/app/builder/features/style-panel/style-source/style-source-input.tsx
```diff
@@ -428,7 +449,20 @@ export const StyleSourceInput = (
     },
   });
 
-  const inputProps = getInputProps();
+  const inputProps = getInputProps({
+    onKeyDown(event) {
+      if (
+        event.key === "Backspace" &&
```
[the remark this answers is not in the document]
kof: it was removed because the local token was last and was not removable, so we had to remove this, but now I copied it back from the old implementation, because now its good again

apps/builder/app/builder/features/style-panel/style-source/style-source.tsx
kof: good catch, also weird why renaming still works

kof: ah it is for editing via double click, fixed now

TrySound (approved):
Honestly liked how it was before but what can I know, I don't use webstudio

kof:
> Honestly liked how it was before but what can I know, I don't use webstudio

From my research with the users it became clear that they are confused because Local is too similar to tokens ... so far I received positive feedback on the changes ...

PR #3020, opened 2024-03-15
fix: HTML Embed Code editor inside Dialog Content looses focus

### LINKED ISSUE #2998: HTML Embed Code editor inside Dialog Content looses focus
https://share.descript.com/view/rp6b9qkp1LU

## Repro
1. insert dialog
2. insert html embed into dialog content
3. focus code editor
4. see it looses focus after a few seconds

istarkov:
## Description

closes #2998

Also fixes name edits etc.
The same issue with radix as we had in the past, focus lock on any mutation inside dialog aquires focus. 
And we have no good way to prevent programmatic focus inside iframe.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/inspector-edits.ts
+++ b/apps/builder/app/canvas/inspector-edits.ts
@@ -1,0 +1,12 @@
+import { $inspectorLastInputTime, $props } from "~/shared/nano-states";
+import { setInert } from "./shared/inert";
+
+export const subscribeInspectorEdits = () => {
+  const unsubscribeInputTime = $inspectorLastInputTime.listen(setInert);
+  const unsubscribeProps = $props.listen(setInert);
+
+  return () => {
+    unsubscribeInputTime();
+    unsubscribeProps();
+  };
+};
--- a/apps/builder/app/canvas/shared/inert.ts
+++ b/apps/builder/app/canvas/shared/inert.ts
@@ -1,0 +1,32 @@
+let resetTimeoutHandle: number | undefined = undefined;
+
+const resetAutoDisposeInert = () => {
+  document.body.removeAttribute("inert");
+  clearTimeout(resetTimeoutHandle);
+  resetTimeoutHandle = undefined;
+};
+
+// 1000 ms is a reasonable time for the preview to reset.
+// Anyway should never happen after user has finished preview changes (can happen during preview changes)
+const AUTO_DISPOSE_INERT_TIMEOUT = 1000;
+
+// A brief delay to ensure mutation observers within the focus scope are activated by the preview changes.
+const DISPOSE_INERT_TIMEOUT = 300;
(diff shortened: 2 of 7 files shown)


apps/builder/app/canvas/inspector-edits.ts
```diff
@@ -0,0 +1,12 @@
+import { $inspectorLastInputTime, $props } from "~/shared/nano-states";
+import { setInert } from "./shared/inert";
+
+export const subscribeInspectorEdits = () => {
+  const unsubscribeInputTime = $inspectorLastInputTime.listen(setInert);
```
kof, line 5: why do we need a date vs counter? just a thought.

kof: or even just a true/false switch, seems like you are just using it as a way to trigger set setInert, so could be basically anything, date made me think its somehow depending on the date value

apps/builder/app/canvas/shared/inert.ts
```diff
@@ -0,0 +1,32 @@
...
+  document.body.removeAttribute("inert");
+  clearTimeout(resetTimeoutHandle);
+  resetTimeoutHandle = undefined;
+};
+
+// 1000 ms is a reasonable time for the preview to reset.
+// Anyway should never happen after user has finished preview changes (can happen during preview changes)
+const AUTO_DISPOSE_INERT_TIMEOUT = 1000;
+
+// A brief delay to ensure mutation observers within the focus scope are activated by the preview changes.
```
kof, line 13: this could use an example, to better understand what can happen, e.g. user does x, then something happens, then we wait then .etc ...  you know a practical example, its very hard to understand what is happening there without full knowledge

kof (approved):

apps/builder/app/canvas/shared/inert.ts
istarkov: Example added at setInert

apps/builder/app/canvas/inspector-edits.ts
istarkov: For me Date.now() is just simpliest form as there is no need to read state etc.
And with Date easier to give a name to variable. 
Have no strong opinion here, lastInputTime has some physical sense as it says that actions depending on it somehow depends on time (and they are), counter or switch have not.

istarkov: IMO the main issue in cases like this are using atoms and subscirptions instead of imperative calls into a canvas.

How it would be better to call `canvas.setInert()` instead of 3 atoms and 3 subscriptions we have to call setInert.

PR #3022, opened 2024-03-16
refactor: transpile expression without code generation

TrySound:

Our code generation with jsep has a big problem of not preserving parenthesis entered by user. And jsep is not able to provide the information.

As a solution I suggest to switch parser to acorn. It is 10 times bigger than jsep but provides more strict parsing with location info for every node. Here I use it to transpile simply with string manipulation which does not touch spaces or parenthesis entered by user.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
what about template strings? :)

istarkov (approved):

PR #3034, opened 2024-03-20
refactor: move forms properties into server module

TrySound:
Here moved forms properties, page meta and remix params into server module as used only in loader and never executed on client. Especially forms properties may contain user credentials and remix does not guarantee it to not appear in client side bundle.

Also moved those into sdk as universal generators.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
haven't tested

PR #3037, opened 2024-03-20
fix: Style updates on all instances upon change

istarkov:
Closes https://github.com/webstudio-is/webstudio/issues/2988

## Description

When a token value is modified, the change will be immediately visible across all N instances using the same token

## Steps for reproduction

Create token, use it on 2 instances, change token styles, local styles etc to see it work

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):
works!

PR #3042, opened 2024-03-21
experimental: setup style object model

TrySound:

Here added style object model basic functionality with inheritance and a few keywords like "initial", "inherit" and "unset".

The algorithm is following css spec and will get more features in next PRs.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/style-object-model.test.ts
+++ b/apps/builder/app/shared/style-object-model.test.ts
@@ -1,0 +1,213 @@
+import { expect, test } from "@jest/globals";
+import type { StyleDecl, StyleSource } from "@webstudio-is/sdk";
+import {
+  type StyleObjectModel,
+  type StyleSelector,
+  getComputedStyleDecl,
+} from "./style-object-model";
+
+const getStyleByStyleSourceId = (styles: StyleDecl[]) => {
+  const styleByStyleSourceId = new Map<
+    StyleSource["id"],
+    Map<StyleDecl["property"], StyleDecl>
+  >();
+  for (const styleDecl of styles) {
+    const styleSourceStyles =
+      styleByStyleSourceId.get(styleDecl.styleSourceId) ?? new Map();
+    styleByStyleSourceId.set(styleDecl.styleSourceId, styleSourceStyles);
+    styleSourceStyles.set(styleDecl.property, styleDecl);
+  }
+  return styleByStyleSourceId;
+};
+
+test("use cascaded style when specified and fallback to initial value", () => {
+  const styles: StyleDecl[] = [
+    {
+      breakpointId: "base",
+      styleSourceId: "bodyLocal",
+      property: "width",
+      value: { type: "unit", value: 10, unit: "px" },
+    },
+  ];
+  const model: StyleObjectModel = {
+    styleSourcesByInstanceId: new Map([["body", ["bodyLocal"]]]),
+    styleByStyleSourceId: getStyleByStyleSourceId(styles),
+  };
+  const styleSelector: StyleSelector = {
+    instanceSelector: ["body"],
+  };
+  // cascaded property
+  expect(
+    getComputedStyleDecl({ model, styleSelector, property: "width" })
+      .computedValue
+  ).toEqual({ type: "unit", unit: "px", value: 10 });
+  // initial for not inherited property
+  expect(
+    getComputedStyleDecl({ model, styleSelector, property: "height" })
+      .computedValue
+  ).toEqual({ type: "keyword", value: "auto" });
+  // initial for inherited property
+  expect(
+    getComputedStyleDecl({ model, styleSelector, property: "color" })
+      .computedValue
+  ).toEqual({ type: "keyword", value: "black" });
+});
+
+test("support initial keyword", () => {
+  const styles: StyleDecl[] = [
+    {
+      breakpointId: "base",
+      styleSourceId: "bodyLocal",
+      property: "width",
+      value: { type: "keyword", value: "initial" },
+    },
+  ];
+  const model: StyleObjectModel = {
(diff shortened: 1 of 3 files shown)


apps/builder/app/shared/style-object-model.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      ["level1", ["level1Local"]],
+    ]),
+    styleByStyleSourceId: getStyleByStyleSourceId(styles),
+  };
+  const styleSelector: StyleSelector = {
+    instanceSelector: ["level3", "level2", "level1"],
+  };
+  // should inherit declared value
+  expect(
+    getComputedStyleDecl({ model, styleSelector, property: "width" })
```
kof, line 118: nit:  getStyleDecl or computeStyleDecl would be shorter

TrySound: Mirroring browser api

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
...
+      const styleDecl = style?.get(property);
+      if (styleDecl) {
+        declaredValues.push(styleDecl.value);
+      }
+    }
+  }
+
+  // https://drafts.csswg.org/css-cascade-5/#cascaded
+  const cascadedValue = declaredValues.at(-1);
+  return { cascadedValue };
```
kof, line 82: I am a bit confused how this return value matches the type above at the expression declaration

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
...
+};
+
+const isKeyword = (styleValue: undefined | StyleValue, keyword: string) =>
```
kof, line 85: hmm, semantics of isKeyword might be a bit confusing here since you are passing a keyword string that needs to match

mby matchKeyword?

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
...
+      specifiedValue = inheritedValue;
+    } else {
+      specifiedValue = initialValue;
+    }
+
+    // https://drafts.csswg.org/css-cascade-5/#computed
+    computedValue = specifiedValue;
+  }
+
+  return { computedValue };
```
kof, line 154: same question here, I didn't run ts compiler but from the look of it I see that it expects:

```
{
  model: StyleObjectModel;
  styleSelector: StyleSelector;
  property: Property;
}
```

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
...
+  property: Property;
+}) => {
+  const { instanceSelector } = styleSelector;
+  const propertyData = properties[property];
+  const inherited = propertyData.inherited;
+  const initialValue = propertyData.initial;
+  let computedValue: StyleValue = initialValue;
+
+  // start computing from the root
+  for (const instanceId of instanceSelector.slice().reverse()) {
```
kof, line 119: .slice is fine, but I think you would want to use Array.from() in a modern js, just because you are not actually slicing, you want a new array

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
+import { properties } from "@webstudio-is/css-data";
+import { StyleValue } from "@webstudio-is/css-engine";
+import type { Instance, StyleDecl, StyleSource } from "@webstudio-is/sdk";
+
+/**
+ *
+ * use cases where computing is called
```
kof, line 7:
```suggestion
 * Use cases where computing is called
```

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * 1. compiler
+ *
+ * put the whole tree into the model once
+ * lazily compute each point in the tree reusing cached cascaded values
```
kof, line 12:
```suggestion
 * Put the whole tree into the model once,
 * lazily compute each point in the tree reusing cached cascaded values.
```

kof: etc, try to make sentences complete, its easier to read this way

apps/builder/app/shared/style-object-model.ts
```diff
@@ -0,0 +1,155 @@
...
+  instanceSelector: InstanceSelector;
+};
+
+type Style = Map<StyleDecl["property"], StyleDecl>;
+
+/**
+ * model contains all data and cache of computed styles
+ * and manages reactive subscriptions
+ */
+export type StyleObjectModel = {
```
kof, line 51: nit: I think you could call it StylesModel

kof (approved):

apps/builder/app/shared/style-object-model.ts
TrySound: this object is input, the output will get other features later

apps/builder/app/shared/style-object-model.ts
TrySound: argument type

apps/builder/app/shared/style-object-model.ts
TrySound: model is not the same as class
we have data and operations on it
looks like model
and again association with cssom

apps/builder/app/shared/style-object-model.ts
TrySound: those are not sentences but machine instructions like pseudo code

apps/builder/app/shared/style-object-model.ts
kof: cssom has methods on the model

PR #3046, opened 2024-03-23
fix: keep the execution order of scripts when executing HTML Embed on the client

kof:
https://share.descript.com/view/q8jpNCbmOxP

## Description

Originally all remote scripts in html embed were executed async, no matter if they have async attribute or not.
Now they respect the order and only execute async if there is async attribute.

## Steps for reproduction

This should fail on first load because first script is loaded async and second script depends on it
```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.12.3/gsap.min.js" async=""></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.12.3/ScrollTrigger.min.js"></script>
<script>
console.log('gsap', gsap)  
</script>
```

This should not fail
```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.12.3/gsap.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.12.3/ScrollTrigger.min.js"></script>
<script>
console.log('gsap', gsap)  
</script>
```

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/html-embed.tsx
```diff
@@ -8,53 +8,110 @@ import {
...
+      script.src = sourceScript.src;
+    } else {
+      script.textContent = sourceScript.innerText;
+    }
+
+    document.head.appendChild(script);
+    sourceScript.replaceWith(script);
+
+    // Run the callback immediately for inline scripts.
+    if (sourceScript.hasAttribute("src") === false) {
```
TrySound, line 37: why check property above and attribute here?

packages/sdk-components-react/src/html-embed.tsx
```diff
@@ -8,53 +8,110 @@ import {
...
+        resolve(script);
+      };
+      script.onerror = reject;
+      script.src = sourceScript.src;
+    } else {
+      script.textContent = sourceScript.innerText;
+    }
+
+    document.head.appendChild(script);
+    sourceScript.replaceWith(script);
```
TrySound, line 34: why replace script here? Wouldn't it execute it twice?

kof: you mean in case it was already executed from server-side render? we only should run this when client-only option is on or we are in the builder/preview

packages/sdk-components-react/src/html-embed.tsx
kof: where is the property check you are mentioning?

kof: oh you mean sourceScript.src ? no reason, fixing

packages/sdk-components-react/src/html-embed.tsx
```diff
@@ -8,53 +8,110 @@ import {
...
+        return insertScript(script);
+      });
+    }
+  });
+
+  // Insert the script tags in parallel.
+  Promise.all(
+    asyncTasks.map(async (task) => {
+      return await task();
+    })
```
TrySound, line 65: asyncTasks.map(task => task())

packages/sdk-components-react/src/html-embed.ws.ts
```diff
@@ -18,9 +18,16 @@ export const meta: WsComponentMeta = {
...
+      description:
+        "Activate it for any scripts that can mutate the DOM or introduce interactivity. This only affects the published site.",
+    },
     executeScriptOnCanvas: {
       ...props.executeScriptOnCanvas,
-      label: "Run script on canvas",
+      label: "Run scripts on canvas",
+      description:
+        "Allow script execution on canvas without switching to preview mode. This only affects build mode.",
     },
```
TrySound, line 31: We might want to support some jsdoc like params instead of customizing generated props meta

TrySound (approved):

PR #3047, opened 2024-03-23
experimental: support custom properties

TrySound:

Here added support for css custom properties in style object model. Added guaranteed-invalid value to schema to represent initial value of custom properties.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/instance-utils.ts
+++ b/apps/builder/app/shared/instance-utils.ts
@@ -697,7 +697,8 @@ const traverseStyleValue = (
     value.type === "invalid" ||
     value.type === "unset" ||
     value.type === "rgb" ||
-    value.type === "function"
+    value.type === "function" ||
+    value.type === "guaranteedInvalid"
   ) {
     callback(value);
     return;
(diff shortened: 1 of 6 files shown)


apps/builder/app/shared/instance-utils.ts
```diff
@@ -697,7 +697,8 @@ const traverseStyleValue = (
     value.type === "invalid" ||
     value.type === "unset" ||
     value.type === "rgb" ||
-    value.type === "function"
+    value.type === "function" ||
+    value.type === "guaranteedInvalid"
```
kof, line 701: What's the point of having this type? is it to allow an invalid value while not showing an error? This sounds like it could be called "initial"

TrySound: Guaranteed-invalid is the name of initial value of custom properties. "initial" should be always resolved into something. This value is what it resolves into. And it cannot be any other value because the logic use it to fallback into initial value of property which use custom property.

kof (approved):

PR #3055, opened 2024-03-26
experimental: support breakpoints and states in style object model

TrySound:

Here improved sorting declared values and computing cascade value. Added support for breakpoints and states. Introduced specificity rules to sort by State-Breakpoint-StyleSource without complicated logic.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/style-object-model.test.ts
+++ b/apps/builder/app/shared/style-object-model.test.ts
@@ -8,14 +8,14 @@ import {
 
 const getStyleByStyleSourceId = (styles: StyleDecl[]) => {
   const styleByStyleSourceId = new Map<
-    StyleSource["id"],
-    Map<StyleDecl["property"], StyleDecl>
+    `${StyleSource["id"]}:${string}`,
+    StyleDecl[]
   >();
   for (const styleDecl of styles) {
-    const styleSourceStyles =
-      styleByStyleSourceId.get(styleDecl.styleSourceId) ?? new Map();
-    styleByStyleSourceId.set(styleDecl.styleSourceId, styleSourceStyles);
-    styleSourceStyles.set(styleDecl.property, styleDecl);
+    const key = `${styleDecl.styleSourceId}:${styleDecl.property}` as const;
+    const styleSourceStyles = styleByStyleSourceId.get(key) ?? [];
+    styleByStyleSourceId.set(key, styleSourceStyles);
+    styleSourceStyles.push(styleDecl);
   }
   return styleByStyleSourceId;
 };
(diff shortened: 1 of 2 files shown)


kof (approved):

apps/builder/app/shared/style-object-model.test.ts
```diff
@@ -558,3 +588,252 @@ test("resolve non-cyclic references in custom properties", () => {
...
+    },
+  ];
+  const model: StyleObjectModel = {
+    styleSourcesByInstanceId: new Map([["body", ["bodyLocal"]]]),
+    styleByStyleSourceId: getStyleByStyleSourceId(styles),
+  };
+  const styleSelector: StyleSelector = {
+    instanceSelector: ["body"],
+    // large is not active breakpoint
+    activeBreakpoints: ["base", "small"],
```
kof, line 661: I think the right term is "applicable", e.g. applicableBreakpoints

apps/builder/app/shared/style-object-model.test.ts
```diff
@@ -558,3 +588,252 @@ test("resolve non-cyclic references in custom properties", () => {
...
+    activeStates: new Set(),
```
kof, line 662: same for states

apps/builder/app/shared/style-object-model.test.ts
TrySound: This is more about whether provided list of breakpoints is activated from ui and rendered in canvas

kof: checked with the spec, that's the term they use

kof: > This is more about whether provided list of breakpoints is activated from ui and rendered in canvas

From builder standpoint yes, but from the logic that's the list of breakpoints you need to consider when calculating the value, right? 

I mean these breakpoints conditions are satisfying the viewport potentially, right?

TrySound: > checked with the spec, that's the term they use

please attach a link

kof: https://www.w3.org/TR/mediaqueries-5/ you will find the word "apply"  and "applicable"

TrySound: Very vague. Styles are applicable with certain condition condition. Conditions are not applicable, they define when to apply

TrySound: The query already applied/activated/matched

kof: then seems it is applicable :)

TrySound: applicable - can be potentially applied

here the query already active/matching

kof: I think applicable is when they are actually machtching, same semantics, but feel free to use "matching" ... active is just overused and often has too many meanings

PR #3056, opened 2024-03-26
Getting build to pass

ankarnus:
Vercel was throwing a build error because a file in the `node_modules` directory of `emotion/hash` was missing. This probably isn't the correct fix but adding it to `devDependencies` using the latest version pulls in the file needed.

I'm also bumping the node version since I don't see it interfering with a production build and it's easier to work with locally.


ankarnus:
Meant to open in my own repo (just testing out the builder).

PR #3059, opened 2024-03-27
fix: Fix transitions for ephemeral changes

kof:
## Description

1. Stop updating the value when change is using scrub (ephemeral)
2. Stop using isEphemeral for arrow keys handling, it should set the vale right away

## Steps for reproduction

1. scrub the delay or duration in transition 
2. expect only delay/duration input changing as well as code area, but not the list in style panel 

## Code Review

- [ ] hi @JayaKrishnaNamburu , I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
```diff
@@ -52,6 +55,7 @@ export const TransitionContent = ({
 
   const { property, timing, delay, duration } =
     useMemo<ExtractedTransitionProperties>(() => {
+      // @todo remove the effect from useMemo
```
kof, line 58: @JayaKrishnaNamburu

JayaKrishnaNamburu (approved):
Looks good, working as expected too 👍

PR #3062, opened 2024-03-27
fix: Remove keywords for transition delay and duration, enable sorting layers

JayaKrishnaNamburu:
Update: this PR is not fixing it fully, there is still an issue when cleaning up field and pressing arrow down it goes into keywords mode.

@JayaKrishnaNamburu will fix it with the next PR

## Description

- Remove keyword values for `transition-delay` and `transition-duration`. Keywords are valid if they are directly applied as individual properties. But when they are used under `transition` shorthand. The `keyword` values actually make the property invalid. For more reference (https://www.w3.org/TR/css-transitions-1/#propdef-transition-duration)
-  Add `draggable` for the  `CSSValueListItem` to show the drag indicator for individual layers.
- Set intermediate values only when the layer is changed.

## Code Review

  - test it on preview

## Before requesting a review
- [x] made a self-review

## Before merging

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import {
   toValue,
   type InvalidValue,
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/features/style-panel/sections/transitions/transition-layer.tsx
```diff
@@ -54,6 +54,7 @@ export const TransitionLayer = <T extends TupleValue>(props: LayerProps<T>) => {
     >
       <CssValueListItem
         id={id}
+        draggable={true}
```
kof, line 57:
```suggestion
        draggable
```

apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
```diff
@@ -50,25 +50,18 @@ export const TransitionContent = ({
     IntermediateStyleValue | InvalidValue | undefined
   >({ type: "intermediate", value: transition });
 
+  useEffect(() => {
```
kof, line 53: wouldn't this effectively overwrite what user is typing if some value comes from outside?

kof:
- [ ] prevent from going into keyword mode in case of delay and duration

kof (approved):

apps/builder/app/builder/features/style-panel/sections/transitions/transition-content.tsx
kof: can you tell me what this code is for?

kof: I can't figure out when is this actually needed

kof: I feel stupid, I can't see why this code is needed, so I did the dumb thing - deleted it, so we don't merge stuff that isn't doing anything, if it did do something, I am sure you know what and we can bring it back

PR #3066, opened 2024-03-27
fix: add finalise job to lost pixel

d-ivashchuk:
## Description

This PR fixes monorepo mode in lost-pixel

### DIFF
--- a/.github/workflows/vis-reg-tests.yml
+++ b/.github/workflows/vis-reg-tests.yml
@@ -29,7 +29,7 @@ jobs:
         run: pnpm run --filter=@webstudio-is/design-system storybook:build
 
       - name: Lost Pixel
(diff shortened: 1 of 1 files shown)


kof (approved):

.github/workflows/vis-reg-tests.yml
```diff
@@ -50,7 +50,22 @@ jobs:
...
+      runs-on: ubuntu-latest
+
+      steps:
+        - name: Checkout
+          uses: actions/checkout@v2
+
+        - name: Lost Pixel Finalize
+          uses: lost-pixel/lost-pixel@v3.16.0
+          env:
+            LOST_PIXEL_CONFIG_DIR: apps/builder
```
TrySound, line 68: Does it need to do only for builder project and not for design-system?

PR #3123, opened 2024-04-09
build: remove unused build scripts

TrySound:
dev, build and dts are not necessary in not published packages and can be removed. Builder vite setup is able to access webstudio export conditions. Only prisma client need to be built before builder.

This will reduce CI time which runs build many times in various workflows.

## Code Review

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/ai/package.json
+++ b/packages/ai/package.json
@@ -10,10 +10,7 @@
     "typecheck": "tsc",
     "test": "NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
     "checks": "pnpm typecheck && pnpm test",
-    "dev": "pnpm build --watch",
-    "build": "rm -rf lib && pnpm build:prompts && esbuild 'src/**/*.ts' --outdir=lib",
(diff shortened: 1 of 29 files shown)


packages/ai/package.json
```diff
@@ -24,20 +21,12 @@
   },
   "exports": {
     ".": {
-      "webstudio": "./src/index.ts",
-      "types": "./lib/types/index.d.ts",
-      "import": "./lib/index.js"
+      "webstudio": "./src/index.ts"
```
kof, line 24: that looks a little weird, isn't this an entry point name? why is the name webstudio? is it like "main"?

TrySound: it is our custom condition, we use it to reference sources within monorepo

kof: in that case maybe there is a convetion like with "@" or just call it "monorepo"?

TrySound: convention is within project to not conflict with other projects

PR #3129, opened 2024-04-10
refactor: simplify toggle control

kof:
## Description

Better interface, simpler

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/layout/layout.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/layout/layout.tsx
@@ -356,7 +356,6 @@ const LayoutSectionFlex = ({
               icons={{
                 wrap: WrapIcon,
                 nowrap: NoWrapIcon,
-                default: WrapIcon,
               }}
               currentStyle={currentStyle}
               setProperty={setProperty}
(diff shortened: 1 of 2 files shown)


TrySound:
Looks like flex wrap this is the only use case for toggle control which makes its behavior unexpected. It also does not give info which values user choose. Maybe switch to menu control like all other layout controls?

kof:
I am switching menu control to the same interface as  toggle control in the next pr

apps/builder/app/builder/features/style-panel/sections/layout/layout.tsx
```diff
@@ -352,19 +352,12 @@ const LayoutSectionFlex = ({
...
-                  Icon: NoWrapIcon,
-                  value: "nowrap",
-                },
-              ]}
-              DefaultIcon={WrapIcon}
+              items={["wrap", "nowrap"]}
+              icons={{
+                wrap: WrapIcon,
+                nowrap: NoWrapIcon,
+                default: WrapIcon,
```
TrySound, line 359: what's the points of default if first item is default?

kof: well just to not have a logic inside the control that decides the first one is default?

kof: default is basically only ever needed when advanced value is used just to not have an empty one, maybe using the first would be fine?

kof: actually lets kill the default

TrySound (approved):

PR #3137, opened 2024-04-11
experimental: Advanced section improvements

kof:
## Description

A bunch of improvements for advanced section

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/nested-input-button.tsx
```diff
@@ -49,7 +48,16 @@ const style = css({
         },
       },
     },
+    size: {
+      1: {
```
TrySound, line 52: let's use names pls
numbers are cryptic
same reason theme.spacing is super unreadable and unwritable

kof: spacing is hard because there are too many values, in this case there are only few sizes and there is no difference if I callit 1,2 or small/large ... 

E.g. radix new theming uses numbers, here is an example: https://www.radix-ui.com/themes/docs/components/button#api-reference

The spacing we use are coming from tailwind https://tailwindcss.com/docs/theme#spacing

Example for tailwind size https://tailwindcss.com/docs/size

TrySound: idk all shit

>  there is no difference if I callit 1,2 or small/large ...

Well, I would not have an idea 1 is small and 2 is bigger
I read it more like size=1 is one size, size=2 is another size.

kof: no, its logical that 2 is bigger than 1, its a scale, [1,2,3,4,5]

TrySound (approved):

PR #3177, opened 2024-04-15
fix: Rewrite Vimeo component to fix a bug and make logic readable

kof:
Closes https://github.com/webstudio-is/webstudio/issues/3097

## Description

1. rewrite logic, hopefully now its readable
2. ensure spinner showing right after click
3. ensure player works even if play button and/or spinner are hidden

## Steps for reproduction

1. @istarkov is aware

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/vimeo.tsx
```diff
@@ -68,7 +62,7 @@ type VimeoPlayerOptions = {
   transparent?: boolean;
 };
 
-const getUrl = (options: VimeoPlayerOptions) => {
+const getVideoUrl = (options: VimeoOptions) => {
```
istarkov, line 65: Looks like a bug
In UI I see enter video URL, Im entering `https://player.vimeo.com/video/831343124`

The code below transforms it into

```
    const userUrl = new URL(options.url);
    url = new URL(IFRAME_CDN);
    url.pathname = `/video${userUrl.pathname}`;
```

causing double `/video/video` to appear

istarkov:
I think you can warmConnections on useEffect will work on mobile too and it's just `preconnect`

packages/sdk-components-react/src/vimeo.tsx
```diff
@@ -350,6 +224,115 @@ export type VimeoOptions = Omit<
...
+    }
+
+    if (showPreview === false) {
+      onPreviewImageUrlChange(undefined);
+      return;
+    }
+
+    if (previewImageUrl === undefined) {
+      loadPreviewImageUrl(videoUrl)
+        .then(onPreviewImageUrlChange)
```
istarkov, line 291: Usually this is bug in react as effect can be reexecuted and outdated `onPreviewImageUrlChange` will be called
Any async with callbacks in React are bugs

istarkov (approved):

packages/sdk-components-react/src/vimeo.tsx
kof: yeah, potentially, is there an easy way to do fetch from a component?

packages/sdk-components-react/src/vimeo.tsx
kof: In the UI the expection is to enter vimeo url that you get form the vimeo UI 
image

kof:
> I think you can warmConnections on useEffect will work on mobile too and it's just preconnect
Do you suggest to warm the connection on mobile in useEffect and on desktop on pointer over? Or do you suggest to always  warm in in useEffect

PR #3183, opened 2024-04-16
build: Move build time env vars in env.static

istarkov:
## Description

This variables are build time and not dependent on deploy. Moreover we don't know them during production deploy.
But they are used as versioning at saas level 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/env/vite-env.d.ts
+++ b/apps/builder/app/env/vite-env.d.ts
@@ -1,0 +1,4 @@
+interface ImportMetaEnv {
+  readonly GITHUB_REF_NAME: string | undefined;
+  readonly GITHUB_SHA: string | undefined;
+}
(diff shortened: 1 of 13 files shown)


apps/builder/app/env/vite-env.d.ts
```diff
@@ -0,0 +1,4 @@
+interface ImportMetaEnv {
```
TrySound, line 1: can this interface be defined at env.static.server?

istarkov: Error in non d.ts files
image

istarkov: image

PR #3190, opened 2024-04-17
build: Release 17.04.2024 (NEVER MERGED!!!!)

istarkov:
## 99% case

1. Create branch, name it `release-{dd}-{mm}-{yyyy}.staging` _(only postfix matters `.staging` release and date is just for devops debug purposes)_
2. Test on staging.
Find it here.
image
3. If everything is ok open Vercel Inspection and Promote to production
image

### Advanced Case:

Create same name branch on saas `release-{dd}-{mm}-{yyyy}.staging`. 
Now all saas services are used from that branch.


istarkov:
New Release is out

PR #3222, opened 2024-04-22
build: Fix timeout error

istarkov:
The only way I found to make it work is to deploy all functions again
Was:
image

Now:
image

Any ideas why rest/ai so much smaller than all other routes?


istarkov:
depending on `export config = {}` we have random function sizes.
1. If export config is on all routes then all functions except rest/ai


image

kof (approved):

PR #3227, opened 2024-04-23
refactor: add nesting rule to css engine

TrySound:
Nesting rule is universal representation for
instance style with support for nested state selectors and nested media queries. Css is generated css of course without css nesting and optimize media queries but in the future may be generated with nesting for simplicity.

This is essential to support mixins in css engine. Will support it in atomic css in other PRs.


packages/css-engine/src/core/rules.ts
```diff
@@ -2,6 +2,107 @@ import type { Style, StyleProperty, StyleValue } from "../schema";
...
+  return `${breakpoint}:${selector}:${property}`;
+};
+
+/**
+ * Universal style rule with nested selectors and media queries support
+ * Rules are generated by each media query
+ * and heavily cached to avoid complex computation
+ *
+ * selector {
+ *   \@media breakpoint {
```
kof, line 25:
```suggestion
 *   @media breakpoint {
```

TrySound: Nope, flattens indent caz jsdoc treat everything starting with @ as own directive

packages/css-engine/src/core/style-sheet-regular.test.ts
```diff
@@ -21,6 +22,203 @@ const style2 = {
...
+  test("generate rules for breakpoint", () => {
+    const sheet = createRegularStyleSheet();
+    const rule = sheet.addNestingRule(".instance");
+    rule.setDeclaration({
+      breakpoint: "base",
+      selector: "",
+      property: "width",
+      value: { type: "keyword", value: "auto" },
+    });
+    rule.setDeclaration({
```
kof, line 44: since you are adding multiple declarations, shouldn't this be `addDeclaration`?

kof: or are you making sure these are unique by property name? e.g. if I call setDeclaration with property height twice, will it add 2 declarations? if yes, then its an "add" if no, then its a set.

If there is deduplication and its a set, then maybe it would be clearer if the interface would be rule.setDeclaration('height', {rest})

kof: as I wrote this rule.setDeclaration('height', {rest}) I realized that this is basically setProperty

kof: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration

In CSSOM CSSStyleDeclaration is the entire style block,can you point me to the refence for terminology you used?

TrySound: declaration here is single property but also with nested selector (state) and breakpoint. So basically same as our StyleDecl.

kof: are you ensuring unique declaration? if not maybe call it addDeclaration?

kof (approved):

packages/css-engine/src/core/style-sheet-regular.test.ts
TrySound: Yes, ensure by unique nested selector, breakpoint and property within top level selector

PR #3228, opened 2024-04-23
feat: Sitemap resource

istarkov:
## First Part 

Here is only new local resource `/$resources/sitemap.xml` support

It can't be implemented as route for now as we have 2 issues:
1. Cloudflare does not allow server fetches from same domain
2. Remix does not have fetch like sveltekit allowing to bypass real http roundtrip when fetching local resources

The idea is to write/use it like any route in case of anything of above will be solved.

Most PR Size is generated new fixture page.
Move SiteData into the resource.

Here is example of rendering this resource using html.

Project to play:
https://sitemap.staging.webstudio.is/builder/cab934af-8615-411a-99c1-5822a98eac74 

Published site:
https://sitemap-saas-t6j12.wstd.work/

image

## Next steps:

- Add `origin` to the system, or even better add `system.hostUrl` with origin, hostname etc
- Add xml element and hardcoded sitemap.xml page

## Steps for reproduction

1. click button
4. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
@@ -443,7 +443,7 @@ export const ResourceForm = forwardRef<
         return "URL is required";
       }
       try {
-        new URL(evaluatedValue);
+        // new URL(evaluatedValue);
       } catch {
         return "URL is invalid";
       }
--- a/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
@@ -12,6 +12,7 @@ import type { DataSource, Resource } from "@webstudio-is/sdk";
 import {
   encodeDataSourceVariable,
   isLiteralExpression,
+  isLocalResource,
 } from "@webstudio-is/sdk";
 import {
   Box,
@@ -443,8 +444,12 @@ export const ResourceForm = forwardRef<
         return "URL is required";
       }
       try {
-        // new URL(evaluatedValue);
+        new URL(evaluatedValue);
       } catch {
+        if (isLocalResource(evaluatedValue, "sitemap.xml")) {
+          return;
+        }
+
         return "URL is invalid";
       }
     },
--- a/packages/cli/src/prebuild.ts
+++ b/packages/cli/src/prebuild.ts
@@ -43,10 +43,10 @@ import {
   findTreeInstanceIds,
   getPagePath,
   parseComponentName,
-  executeExpression,
   generateFormsProperties,
   generateResourcesLoader,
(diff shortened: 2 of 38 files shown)


packages/sdk/src/resources-generator.ts
```diff
@@ -86,10 +86,31 @@ export const generateResourcesLoader = ({
...
+
+  if (hasResources) {
+    generated += `import { sitemap } from "./[sitemap.xml]";\n`;
+  }
+
   generated += `export const loadResources = async (_props: { system: System }) => {\n`;
   generated += generatedVariables;
   if (hasResources) {
+    generated += `
+    const customFetch: typeof fetch = (input, init) => {
```
TrySound, line 99: Seems like static function, maybe put it into sdk?

istarkov: yes

istarkov: but probably here will be better as sitedata etc will be arguments

apps/builder/app/builder/features/settings-panel/resource-panel.tsx
```diff
@@ -445,6 +446,10 @@ export const ResourceForm = forwardRef<
       try {
         new URL(evaluatedValue);
       } catch {
+        if (isLocalResource(evaluatedValue, "sitemap.xml")) {
+          return;
+        }
```
TrySound, line 451: could be checked before url parser

packages/cli/src/prebuild.ts
```diff
@@ -682,19 +682,7 @@ export const projectMeta: ProjectMeta =
...
-            .filter(
-              (page) =>
-                executeExpression(page.meta.excludePageFromSearch) !== true
-            )
-            .map((page) => ({
-              path: page.path,
-              lastModified: siteData.build.updatedAt,
-            })),
-        },
+        getStaticSiteMapXml(siteData.pages, siteData.build.updatedAt),
```
TrySound, line 685: xml? but it is json 😁

packages/sdk/src/resources-generator.test.ts
```diff
@@ -35,30 +31,45 @@ test("generate resources loader", () => {
...
+import { sitemap } from "./[sitemap.xml]";
+export const loadResources = async (_props: { system: System }) => {
+
+    const customFetch: typeof fetch = (input, init) => {
+      if (typeof input !== "string") {
+        return fetch(input, init);
+      }
+
+      if (isLocalResource(input, "sitemap.xml")) {
+        // @todo: dynamic import sitemap ???
```
TrySound, line 45: what this about?

PR #3232, opened 2024-04-23
feat: Resizable and draggable inspector dialog

kof:
Make an option for resizable and draggable dialog. Inspector and marketplace will benefit from it

https://github.com/webstudio-is/webstudio/assets/52824/53c1c9fa-bb81-4fcd-b7a3-80b508537fdd

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound:
maximize no longer works

packages/design-system/src/components/dialog.tsx
```diff
@@ -21,23 +24,121 @@ export const DialogClose = Primitive.Close;
...
+    ) {
+      return;
+    }
+    const { rect, point } = initialDataRef.current;
+    const movementX = point.x - event.pageX;
+    const movementY = point.y - event.pageY;
+    let left = Math.max(rect.x - movementX, 0);
+    left = Math.min(left, window.innerWidth - rect.width);
+    let top = Math.max(rect.y - movementY, 0);
+    // We want some part of the dialog to be visible but otherwise let it go off screen.
```
istarkov, line 86: Strange comment starting from 'but', Im not getting it. 
IMO this `  // We want some part of the dialog to be visible` is enough

istarkov: Also why top only?

packages/design-system/src/components/dialog.tsx
```diff
@@ -107,16 +208,32 @@ const contentShow = keyframes({
   from: { opacity: 0, transform: "translate(-50%, -48%) scale(0.96)" },
   to: { opacity: 1, transform: "translate(-50%, -50%) scale(1)" },
 });
-const contentStyle = css(floatingPanelStyle, {
-  position: "fixed",
+
+const centeredContent = {
   top: "50%",
   left: "50%",
   transform: "translate(-50%, -50%)",
```
istarkov, line 215: transforms can cause count of bug, of wrong position detection to various visual glitches. In case of fixed elements I would suggest to use { left: 0, top: 0,  bottom: 0, right: 0, margin: auto}

istarkov (approved):

packages/design-system/src/components/dialog.tsx
kof: intentional behavior so that the header of the dialog is always fully visible and so that you can make the dialog mostly offscreen and see whats behind

packages/design-system/src/components/dialog.tsx
```diff
@@ -21,23 +24,121 @@ export const DialogClose = Primitive.Close;
...
+    target.style.top = `${rect.top}px`;
+    target.style.transform = "none";
+    initialDataRef.current = {
+      point: { x: event.pageX, y: event.pageY },
+      rect,
+    };
+  };
+
+  const handleDrag: DragEventHandler = (event) => {
+    event.preventDefault();
```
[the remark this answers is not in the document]
kof: deleted, can't even remember what I was trying to fix

PR #3235, opened 2024-04-24
fix: save resources on reload

TrySound:
Here fixed the issue with resources not refreshed while editing and clicking on refresh button. Also fixed cached value not restored after closing resource panel.


kof (approved):
This works, but honestly what if we saved changes like we do in properties panel as user types? 

Risk is the resource is really large and takes time to fetch and slows things down?

PR #3245, opened 2024-04-26
feat: Canvas drawing of xml element

istarkov:
## Description

This is only for Canvas rendering.

image

image

Editable
image

For generating it will change a little.

## Steps for reproduction

[Play Ref https://xml-element.development.webstudio.is/builder/cab934af-8615-411a-99c1-5822a98eac74](https://xml-element.development.webstudio.is/builder/cab934af-8615-411a-99c1-5822a98eac74)

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

packages/sdk-components-react/src/xml-element.tsx
```diff
@@ -0,0 +1,32 @@
...
+  children: ReactNode;
+} & Record<typeof componentAttribute, string>;
+
+export const XmlElement = forwardRef<ElementRef<"div">, Props>(
+  ({ tag = "element", children, ...props }, ref) => {
+    const isTextChild = Children.toArray(children).every(
+      (child) => typeof child === "string"
+    );
+
+    const elementName = tag.replaceAll(/[^\p{L}]+/gu, "");
```
kof, line 18: can't we use browsers own xml parser?

packages/react-sdk/src/components/component-meta.ts
```diff
@@ -24,6 +24,7 @@ export const componentCategories = [
   "media",
   "forms",
   "radix",
+  "xml",
```
kof, line 27: Hmm do we need to show that component in the components panel at all? If yes we need to think where, custom category for it is probably a bit too much?

packages/sdk-components-react/src/components.ts
```diff
@@ -30,3 +30,4 @@ export { Vimeo } from "./vimeo";
 export { VimeoPreviewImage } from "./vimeo-preview-image";
 export { VimeoPlayButton } from "./vimeo-play-button";
 export { VimeoSpinner } from "./vimeo-spinner";
+export { XmlElement } from "./xml-element";
```
kof, line 33: I am a bit torn about this name XmlElement, it becomes a bit confusing because its a component that's called element, can't we call it just XML?

packages/sdk-components-react/src/xml-element.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const defaultTag = "div";
+
+// We don't want to enable all tags because Box is usually a container and we have specific components for many tags.
+type Props = {
+  tag: string;
+  children: ReactNode;
+} & Record<typeof componentAttribute, string>;
+
+export const XmlElement = forwardRef<ElementRef<"div">, Props>(
+  ({ tag = "element", children, ...props }, ref) => {
```
kof, line 13: what's "element" as a tag name?

packages/sdk-components-react/src/xml-element.ws.ts
```diff
@@ -0,0 +1,22 @@
...
+  type WsComponentPropsMeta,
+} from "@webstudio-is/react-sdk";
+
+import { props } from "./__generated__/xml-element.props";
+
+export const meta: WsComponentMeta = {
+  category: "xml",
+  order: 6,
+  type: "container",
+  icon: XIcon,
```
kof, line 14: I think this is twitter icon, I will get you a proper one

packages/sdk-components-react/src/xml-element.tsx
kof: for example aren't -_ or . allowed?

kof: probably can't sanitize with the dom, only parse ...

kof: found something https://github.com/MylesBorins/xml-sanitizer/blob/main/index.js

apps/builder/app/builder/features/sidebar-left/panels/components/components.tsx
```diff
@@ -50,6 +50,14 @@ export const TabContent = ({ publish, onSetActiveTab }: TabContentProps) => {
             if (category === "hidden") {
               return false;
             }
+
+            if (
```
kof, line 54: just discussed with Bogdan, we should for now keep it hidden, so this can be removed

kof (approved):

packages/sdk-components-react/src/xml-element.ws.ts
kof: Will provide an icon, because we need one for the navigator even if the component is hidden from components panel

packages/sdk-components-react/src/xml-element.tsx
istarkov: Improved

packages/sdk-components-react/src/xml-element.tsx
istarkov: Don't get, here can be any default, why not element

packages/react-sdk/src/components/component-meta.ts
istarkov: Now hidden under the flag, we will decide later what to do

apps/builder/app/builder/features/sidebar-left/panels/components/components.tsx
istarkov: Then we decided to leave under flag for now

packages/sdk-components-react/src/components.ts
istarkov: Something that draw xml element, not an xml imo

packages/sdk-components-react/src/xml-element.tsx
istarkov: We will deciede later for now will merge as is

packages/sdk-components-react/src/components.ts
istarkov: Let's decide later will merge as is

packages/sdk-components-react/src/xml-element.tsx
kof: what is element doing in case the default is applied?

kof: when is it possible to not get a tag and land on default?
what's the intention of having a default here, who will see it?

kof: seems like it should be possible to not have a tag provided and right now I can only imagine an empty string there

PR #3246, opened 2024-04-26
fix: Prevent editing Code Text component as text

kof:
Code Text component is designed to render code and in the future also to add syntax highlighting. With this PR it won't be possible to treat it like a Box any more and code will be entered using code property.

Benefit is: better UX for editing code and no unintended side effects later with highlighting.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/code-text.ws.ts
```diff
@@ -49,6 +49,13 @@ export const meta: WsComponentMeta = {
 };
 
 export const propsMeta: WsComponentPropsMeta = {
-  props,
-  initialProps: ["id", "className", "lang"],
+  props: {
+    ...props,
+    code: {
+      required: true,
+      control: "code",
```
TrySound, line 56: This will highlight it only as html

packages/sdk-components-react/src/code-text.tsx
```diff
@@ -1,14 +1,35 @@
-import { type ElementRef, type ComponentProps, forwardRef } from "react";
+import {
+  type ElementRef,
+  type ComponentProps,
+  forwardRef,
```
TrySound, line 4: 😠

packages/sdk-components-react/src/code-text.ws.ts
kof: what should we do about it?

packages/sdk-components-react/src/code-text.tsx
kof: what about it?

packages/sdk-components-react/src/code-text.ws.ts
kof: I see 2 options:
1. use no syntax highlighting at all in this case
2. allow user to select language in the editor

kof: technically we already have a lang prop there, we could pick a default and use the same prop to highlight in the editor as well as on the canvas later

The questions there:
1. how do we provide to the code editor the selected langage on the other prop, currently each prop doens't have access to other props
2. we will need to lazy load language packages to avoid loading all of them at once, right?

kof: or we keep it as a follow-up issue and for now just have a way to define code editor without syntax highlighting, this will be more appropriate for this issue, which is only trying to prevent user from addding children from this component

packages/sdk-components-react/src/code-text.tsx
TrySound: Wanna get rid of forwardRef already 😄

kof: I totally didn't realize we already are on the version that doesn't require it lol

apps/builder/app/builder/shared/code-editor.tsx
```diff
@@ -1,386 +1,122 @@
+import { forwardRef, useMemo, type ComponentProps, useEffect } from "react";
```
TrySound, line 1: If put base code editor here the diff would not blow up

kof: I have no idea what you mean

TrySound: splitting code editor stuff into code-editor-base made diff unreadable

TrySound (approved):

apps/builder/app/builder/shared/code-editor.tsx
TrySound: I would just keep both components in single module

kof: I didn't split it now, it was before

kof: I think git didn't recognize renaming because it also had changes

TrySound: because after remaining you created a module with the same name

PR #3247, opened 2024-04-27
fix: Move css filter properties to be used as tuples instead of layers in the data-layer

### LINKED ISSUE #3243: Filters don't work when using more than one layer
Turns out filters are not working at all when using more than one layer because they are compiled using a comma instead of a space.

JayaKrishnaNamburu:
fixes #3243 

## Description
 
We should use tuples for handling filters. Don't know when i switched to layers, but the initial implementation of the filters started as tuples itself. 

- [ ] We might probably need a migration for the existing projects to change it to tuple if needed. Let me know, I can help with that as well. 

## Steps for reproduction
1. Add a css filter in the sections panel.
2. Add another css filter and edit it using the code view.
3.  Swap filters using the drag and drop action.
4. Deploy the project and the multiple filters applied in the same element should work as expected.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging


apps/builder/app/builder/features/style-panel/style-layers-list.tsx
```diff
@@ -26,27 +30,27 @@ export type LayerProps<T> = {
...
   onLayerHide: (index: number) => void;
   onDeleteLayer: (index: number) => void;
-  onEditLayer: (index: number, layers: LayersValue) => void;
+  onEditLayer: (index: number, layers: LayersValue | TupleValue) => void;
   createBatchUpdate: CreateBatchUpdate;
   deleteProperty: DeleteProperty;
 };
 
-type LayerListProperties<T> = SectionProps & {
+type LayerListProperties<T, K> = SectionProps & {
```
kof, line 38: Please try to not use T,K when possible, e.g. in this case K is LayersValue  and T is TupleValue

kof: or maybe I am wrong, haven't looked exactly and T K isn't makeit it obvious

kof (approved):
Tested, works!

JayaKrishnaNamburu:
Updated the typings, should we write some migrations or should we merge as it is. As it a fairly new feature.

kof:
merge witout migrations, fixing it would potentially breaksomething because when they tested it didn't have affect and now it will

PR #3248, opened 2024-04-27
refactor: reimplement atomic styles with nesting rules

TrySound:
Here used new nesting rules to generate atomic styles. Replaced separate atomic style sheet class with just a method to generate styles atomically. Inheritance doesn't work well with private fields caz require a lot of internals to be public.


packages/css-engine/src/core/style-sheet-atomic.test.ts
```diff
@@ -1,130 +1,176 @@
-import { describe, beforeEach, test, expect } from "@jest/globals";
```
kof, line 1: I am very confused about what's going on with file naming here, before this PR, there was style-sheet-atomic.ts and style-sheet-atomic.test.ts and the same thing for regular now tests for atomic are using regular style sheet?

very confusing everryhting you did with renaming

packages/css-engine/src/core/style-sheet.ts
```diff
@@ -119,6 +120,39 @@ export class StyleSheet {
     this.#cssText = css.join("\n");
     return this.#cssText;
   }
+  generateAtomic(options: { getKey: (rule: NestingRule) => string }) {
```
kof, line 123: so you killed the separation between atomic style sheet and regular style sheet and now there is only one style sheet that does both?

kof: I feel like this is a road towards a more dirty abstraction

packages/css-engine/src/core/style-sheet.ts
```diff
@@ -88,7 +88,13 @@ export class StyleSheet {
   markAsDirty() {
     this.#isDirty = true;
   }
-  get cssText() {
+  generateWith({
```
kof, line 91: regarding naming, I used toString in all rules to get a css string from a model with a .cssText shorthand when you don't need options

Would be great to keep it consistent here and in the nesting rule the same

kof (approved):
- [ ] a bit of naming cleanup

packages/css-engine/src/core/style-sheet.ts
TrySound: toString in js is not suppose to accept any input data. So it may be confusing. It is confusing for me at some point. It's like `''.toString({ content: 'content' })`

kof: then lets rename all of those toString in all rules to toCss or something

TrySound: in other PR

PR #3251, opened 2024-04-27
chore: don't fix boolean attributes to empty string values in html embed

kof:
## Description

When validating html embed, dom innerHTML always turns boolean attributes into empty string, e.g. <div hidden></div> will always become <div hidden=""></div> and this is not necessary, we can ignore those mismatches.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/settings-panel/controls/code.tsx
```diff
@@ -70,8 +70,17 @@ const validateHtml = (value: string): Error | undefined => {
   const div = document.createElement("div");
   div.innerHTML = value;
   const expected = div.innerHTML;
-  // We don't need to show error for unnecessary whitespace.
-  if (value.replace(/\s/g, "") !== expected.replace(/\s/g, "")) {
+  const clean = (value: string) => {
+    return (
```
istarkov, line 74: What's wrong with

```
function isValidHTML(htmlString) {
    const parser = new DOMParser();
    const doc = parser.parseFromString(htmlString, 'text/html');
    return !doc.querySelector('parsererror');
}
```

What the case of `clean(blabla) === clean(otherblabla)`

kof: I tried it, nothing was triggering a parseerror, it may be not implemented in chrome, so it's kinda useless, then you get the same issue I am fixing above

kof: The problem is to identify broken HTML and because browser is flexible and always tries to fix it we never know if something is really broken, so what I am trying to do there is to minimize the amount of false positives:
1. different amount of whitespace
2. `boolean=""` is the same as `boolean`

istarkov: lets add a comment then

istarkov (approved):

PR #3253, opened 2024-04-29
refactor: reimplement canvas style renderer

TrySound:
Here refactored canvas styles with nesting rules which allow to update each declaration incrementally and closer to our data format.

Things to test
- initial rendering
- updating values
- ephemeral values
- selecting state, rendered as separate stylesheet

Added shim.ts to contain simplified polyfills for future methods like Set.prototype.difference which I use here to find added and deleted styles.

Removed warning from parseCssValue which pollutes log with outdated messages.

Additionally fixed old bugs
- support ephemeral styles with selector state, for example hover was not updated on on canvas while updating ephemeral styles
- remove ephemeral custom property from body (bug appeared when added support ephemeral styles in tokens)

### DIFF
--- a/apps/builder/app/shared/shim.ts
+++ b/apps/builder/app/shared/shim.ts
@@ -1,0 +1,17 @@
+export const setDifference = <Item>(current: Set<Item>, other: Set<Item>) => {
+  const result = new Set<Item>(current);
+  if (current.size <= other.size) {
+    for (const item of current) {
+      if (other.has(item)) {
+        result.delete(item);
+      }
+    }
+  } else {
+    for (const item of other) {
+      if (current.has(item)) {
+        result.delete(item);
+      }
+    }
+  }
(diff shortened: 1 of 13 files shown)


apps/builder/app/shared/shim.ts
```diff
@@ -0,0 +1,17 @@
+export const setDifference = <Item>(current: Set<Item>, other: Set<Item>) => {
```
kof, line 1: setDifference reads like set method like setX etc ... I was starring at it for a while until I realized you mean a difference for Set

TrySound: yeah, naming is hard

kof:
- [ ] after ephemeral updates e.g. scrubbing width, you are now leaving the ephemeral style property in the dom, previously we were clearing it up

PR #3256, opened 2024-04-29
refactor: drop tsconfig paths

TrySound:
We use tsconfig paths option only to resolve single constants module. It requires additional plugin in vite and all other tools. At least in generated published sites it can be just dropped.

Moved __templates__/route-template.tsx to app/routes/template.tsx


TrySound:
Works in staging

istarkov (approved):

PR #3276, opened 2024-05-01
feat: Fullscreen preview when on base breakpoint

kof:
Closes https://github.com/webstudio-is/webstudio/issues/2324

Explanation https://share.descript.com/view/8NayOadrxo8

Biggest problem it addresses is that users on base breakpoint want to see the site in a full width without having to resize each time.

Additionally user doesn't want it to resize when switching between preview and canvas on a custom selected size.

1. removed remaining hook wrappers from nanostores
2. moved the resize handles to the outside of the canvas and gave them some space - fixes the problem with dragging to the edge and then not being able to get them back because mouse grabs the browser itself
3.  On base breakpoint and higher when there is no min-width breakpoint and user switches to preview, we use the entire workspace, then going back, we calculate based on base breakpoint start.
4. On all other breakpoints we keep the size user has chosen between the preview and canvas, we don't reset it.

## Steps for reproduction

1. without min breakpoints on canvas
  - on base, switch to preview, see maximum canvas size, switch back to canvas, see base canvas width
  - on every breakpoint and in between see that it keeps the size
1.  add min-width breakpoint, on base, switch to preview, see the same size

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
```diff
@@ -0,0 +1,129 @@
...
+  // Note: it is still possible to get intersecting min and max breakpoints.
+  if (isBaseBreakpoint(selectedBreakpoint)) {
+    // Base is the only breakpoint
+    if (breakpoints.length === 1) {
+      return workspaceWidth;
+    }
+
+    const grouped = groupBreakpoints(breakpoints).filter(
+      ({ minWidth, maxWidth }) => {
+        if (minWidth && minWidth < workspaceWidth) {
```
istarkov, line 77: Is it possible that we are on base breakpoint and minWidth < workspaceWidth?

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
```diff
@@ -0,0 +1,129 @@
...
+          return true;
+        }
+        if (maxWidth && maxWidth > minCanvasWidth) {
```
istarkov, line 80: Not sure why we are filtering this too

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
```diff
@@ -0,0 +1,129 @@
...
+          return true;
+        }
+      }
+    );
+    let lowestMinWidth = grouped
+      .filter(({ minWidth }) => {
+        if (minWidth && minWidth < workspaceWidth) {
```
istarkov, line 87: already filtered above

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
```diff
@@ -0,0 +1,129 @@
...
+          return true;
+        }
+      })
+      .at(-1)?.minWidth;
+    lowestMinWidth =
+      lowestMinWidth === undefined ? workspaceWidth : lowestMinWidth - 1;
+
+    let highestMaxWidth = grouped
+      .filter(({ maxWidth }) => {
+        if (maxWidth && maxWidth > minCanvasWidth) {
```
istarkov, line 97: already filtered

istarkov (approved):

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
kof: fixed, now only checking for undefined

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
kof: fixed, now only checking for undefined

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
kof: added comments

apps/builder/app/builder/features/breakpoints/calc-canvas-width.ts
istarkov: I mean greater ;)

kof: yes its possible, why not, you can have a tiny workspace and a large min-width, this is why we have scaling , so you can test on a large breakpoint without having a large screen

PR #3283, opened 2024-05-02
feat: allow non-latin characters in old-path redirects

### LINKED ISSUE #3225: Support non latin characters in page path
The issue was brought by this discussion https://github.com/webstudio-is/webstudio-community/discussions/138

It turns out our validation of a path is super basic and doesn't support any non latin characters in pat and redirects. I think this could be easily supported

JayaKrishnaNamburu:
## Description

fixes #3225 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk/src/schema/pages.ts
+++ b/packages/sdk/src/schema/pages.ts
@@ -129,8 +129,20 @@ export const ProjectNewRedirectPath = PagePath.or(
   }, "Must be a valid URL")
 );
 
+export const ProjectOldRedirectPath = PagePath.or(
+  z
+    .string()
+    .refine(
+      (path) =>
+        new RegExp(/^(?:\/[-_a-zA-Z0-9*:?\\\p{L}\\p{N}\\p{M}\\\/]+)+$/gmu).test(
+          path
+        ),
+      "Only a-z, 0-9, -, _, /, :, ?, * and non-latin characters are allowed"
+    )
+);
+
 export const PageRedirect = z.object({
-  old: PagePath,
+  old: ProjectOldRedirectPath,
   new: ProjectNewRedirectPath,
   status: z.enum(["301", "302"]).optional(),
 });
--- a/packages/sdk/src/schema/pages.ts
+++ b/packages/sdk/src/schema/pages.ts
@@ -134,7 +134,7 @@ export const ProjectOldRedirectPath = PagePath.or(
     .string()
     .refine(
       (path) =>
-        new RegExp(/^(?:\/[-_a-zA-Z0-9*:?\\\p{L}\\p{N}\\p{M}\\\/]+)+$/gmu).test(
+        new RegExp(/^(?:\/[-_a-zA-Z0-9*:?\\\p{L}\\p{N}\\p{M}\\/]+)+$/gmu).test(
           path
         ),
       "Only a-z, 0-9, -, _, /, :, ?, * and non-latin characters are allowed"
--- a/apps/builder/app/builder/features/project-settings/section-redirects.tsx
+++ b/apps/builder/app/builder/features/project-settings/section-redirects.tsx
@@ -17,9 +17,9 @@ import {
 import { ArrowRightIcon, TrashIcon } from "@webstudio-is/icons";
 import { useState, type ChangeEvent, useRef } from "react";
 import {
-  PagePath,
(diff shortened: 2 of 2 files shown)


kof:
Why only redirects? I think all page paths can support non-latin characters

packages/sdk/src/schema/pages.ts
```diff
@@ -129,8 +129,20 @@ export const ProjectNewRedirectPath = PagePath.or(
   }, "Must be a valid URL")
 );
 
+export const ProjectOldRedirectPath = PagePath.or(
+  z
+    .string()
+    .refine(
+      (path) =>
+        new RegExp(/^(?:\/[-_a-zA-Z0-9*:?\\\p{L}\\p{N}\\p{M}\\\/]+)+$/gmu).test(
```
kof, line 137: is there an original implementation of this? I think its a non-trivial thing to write and we should be able to use something battle-tested

JayaKrishnaNamburu:
Yup, thought of it. But didn't want to change without cross checking with the team once. 
So just added as a safe option for old redirects. As, urls from old projects can still be used. Will test a little bit, and change it to all page paths across the project 👍

packages/sdk/src/schema/pages.ts
JayaKrishnaNamburu: I just extended `\\p{L}\\p{N}\\p{M}\\` to the original one to allow non-latin characters though

kof: Btw is not possible to use any of the standard api for this? e.g. URL api

kof:
I just checked webflow and when I put "日本語" in there it gets transformed to "ri-ben-yu", do you know why?

packages/sdk/src/schema/pages.ts
JayaKrishnaNamburu: Will see, ideally should work though. Will check about the `+/-`'s of both approaches.

kof:
Lets see if twitter knows why Webflow converts this way

https://x.com/oleg008/status/1786101402869063997

packages/sdk/src/schema/pages.ts
JayaKrishnaNamburu: > where is the original coming from? is it worth linking to?

Original regex is from `PagesPath`, just extending with latin characters. Until there is a decision to support non-latin across all the paths inside webstudio projects. Maybe checking only for old paths (projects that are designed in another platforms)  would be safe way to go.

kof:
- [ ] investigate how mature projects are treating page urls and what they use for validation

kof:
As an idea, check what wordpress is doing, I think they had 20 years to figure this out

JayaKrishnaNamburu:
Here are my observations.
- Wordpress don't have a first hand support for redirects. But there is a plugin that brings in support for the same.  While pasting, the non-latin characters are not handled. So any old url's with non-latin characters in them can't be used.
Screenshot 2024-05-03 at 18 29 53

- Webflow seems to change them to translate them in the page urls. So, ideally looks like they don't support non-latin characters for the pages in the project. Can't test the `301` in webflow, as it needs a premium account.
Screenshot 2024-05-03 at 18 33 26



And one more thing, we can't directly use the `URL` to validate this because, there is a domain that is missing. So, the `URL` will always throw error unless we prepend with something like `www.example.com/product-page/商品名-7` for validation purposes. We can just stick with regex as it is now.


The decent middle ground can be, just allow support for non-latin characters in the redirects manager and that too only for **old-paths**. So existing users like the one from GitHub discussions can still use the manager to move their projects. And later the support for non-latin on project wide routes can be added.

TrySound:
> And one more thing, we can't directly use the URL to validate this because, there is a domain that is missing. So, the URL will always throw error unless we prepend with something like www.example.com/product-page/商品名-7 for validation purposes. We can just stick with regex as it is now.

there is nothing bad with prepending for new URL, we do this all the time. You can even pass fake origin as second argument.

JayaKrishnaNamburu:
Closing this PR, as the support for non-latin characters is a little bit bigger topic than just enabling in regex. Follow #3225 for further updates.

PR #3307, opened 2024-05-06
feat: Refactor existing `parseBoxShadow` to parse both box and text shadows

JayaKrishnaNamburu:
## Description

`box-shadow` and `text-shadow` both share the same syntax signature.
https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#syntax
https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow#syntax

The only difference is `text-shadow` don't have `inset` and `spread` values. And added a extractor which can be used in future for text-shadow UI.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


packages/css-data/src/property-parsers/__snapshots__/shadows.test.ts.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
kof, line 1: @TrySound what's your opinion on shapshots, I wonder if we had switched to inline snapshots

packages/css-data/src/property-parsers/box-shadow-property-extractor.test.ts
```diff
@@ -1,6 +1,6 @@
 import { describe, expect, test } from "@jest/globals";
 import { extractBoxShadowProperties } from "./box-shadow-property-extractor";
-import { parseBoxShadow } from ".";
+import { parseShadow } from ".";
```
kof, line 3: I suggest to import specifically the file instead of the index

packages/css-data/src/property-parsers/text-shadow-property-extractor.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  To validate the text-shadow string value, should be used
+  in combination with parseShadow function.
+
+  Eg:
+  extractBoxShadowProperties(parseShadow('boxShadow', "5em 5em 5em 5em #ffff"))
+*/
+
+export type ExtractedTextShadowProperties = {
+  color: RgbValue | KeywordValue | null;
```
kof, line 22: is there any way to not use null? we are generally trying to use undefined all the time and null only where some external api forces us to use null

kof (approved):
I am approving with a few minor comments, but lets get an approval from @TrySound

packages/css-data/src/property-parsers/text-shadow-property-extractor.ts
JayaKrishnaNamburu: Will switch to `undefined` then 👍

packages/css-data/src/property-parsers/__snapshots__/shadows.test.ts.snap
TrySound: yep, we use only inline snapshots

packages/css-data/src/property-parsers/text-shadow-property-extractor.ts
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+export type ExtractedTextShadowProperties = {
+  color?: RgbValue | KeywordValue;
+  offsetX: UnitValue;
+  offsetY: UnitValue;
+  blur: UnitValue;
+};
+
+export const extractTextShadowProperties = (
```
TrySound, line 28: is it different from box shadow extractor?

JayaKrishnaNamburu: Yes, `text-shadow` don't have `blur` and `spread`. And really didn't want to mix the extractors just in case if we need to modify them if needed too.

JayaKrishnaNamburu: @TrySound merged both the extractors now. Let's see if we really need to be separated down the line then.

TrySound (approved):

PR #3313, opened 2024-05-06
experimental: add descendent component

TrySound:

Here added support for new `<ws:descendent />` component. It allows to style any children or descendents of components. Primary usage is for embed content which renders html and need to style tags.

Descendent component accepts "selector" prop which can be suffix like ` img`, ` .image` or `::first-child` pseudo-element.

Potentially it could be used as tag styling for any element.

Can be tested in any element by dropping "Descendent" component from "internal" category.

### DIFF
--- a/apps/builder/app/canvas/shared/styles.ts
+++ b/apps/builder/app/canvas/shared/styles.ts
@@ -2,6 +2,7 @@ import { useEffect, useLayoutEffect } from "react";
 import { computed } from "nanostores";
 import { useStore } from "@nanostores/react";
 import {
+  Instance,
   getStyleDeclKey,
   type StyleDecl,
   type StyleSourceSelection,
@@ -12,6 +13,7 @@ import {
   addGlobalRules,
   createImageValueTransformer,
   getPresetStyleRules,
+  descendentComponent,
   type Params,
 } from "@webstudio-is/react-sdk";
 import {
@@ -23,7 +25,9 @@ import {
 import {
   $assets,
   $breakpoints,
+  $instances,
   $isPreviewMode,
+  $props,
   $registeredComponentMetas,
   $selectedInstanceSelector,
   $selectedStyleState,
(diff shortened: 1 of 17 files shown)


kof (approved):

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -159,6 +163,42 @@ const toVarValue = (styleDecl: StyleDecl): undefined | VarValue => {
   }
 };
 
+const $descendentSelectors = computed(
```
kof, line 166: this def. needs tests

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -227,10 +267,31 @@ export const subscribeStyles = () => {
     }
   );
 
+  const unsubscribeDescendentSelectors = $descendentSelectors.subscribe(
```
kof, line 270: def. needs tests

packages/css-engine/src/core/atomic.test.ts
```diff
@@ -175,3 +175,33 @@ test("distinct similar declarations from different breakpoints", () => {
...
+  const sheet = createRegularStyleSheet();
+  sheet.addMediaRule("x");
+  const rule1 = sheet.addNestingRule("instance");
+  rule1.setDeclaration({
+    breakpoint: "x",
+    selector: ":hover",
+    property: "display",
+    value: { type: "keyword", value: "block" },
+  });
+  const rule2 = sheet.addNestingRule("instance", " img");
```
kof, line 189: are you sure you want to have space and such as part of the selector string and not as something completely separated?

packages/css-engine/src/core/style-sheet-regular.test.ts
```diff
@@ -932,6 +932,35 @@ describe("Style Sheet Regular", () => {
     color: blue
   }
 }"
+`);
+  });
+
+  test("support descendent suffix", () => {
+    const sheet = createRegularStyleSheet();
+    sheet.addMediaRule("base", {});
+    const rule1 = sheet.addNestingRule(".instance");
```
kof, line 941: .instance or .element? technically you are talking about an element that is backing the instance, but in this case you are talking specifically about the element, instance is basically the component instance that contains that element, but you are not styling the instance

packages/css-engine/src/core/rules.ts
```diff
@@ -80,13 +81,25 @@ export class NestingRule {
...
     this.#mixinRules = mixinRules;
   }
   getSelector() {
     return this.#selector;
   }
+  setSelector(selector: string) {
+    this.#selector = selector;
+    this.#cache.clear();
+  }
+  getDescendentSuffix() {
```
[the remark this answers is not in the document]
TrySound: Here no combinators `::placeholder`

packages/css-engine/src/core/style-sheet-regular.test.ts
TrySound: Just a reference `[data-ws-id=instanceId]` -> `.instance`

packages/css-engine/src/core/atomic.test.ts
TrySound: Yep, string is enough

PR #3315, opened 2024-05-06
experimental: add content embed aka rich text component

TrySound:

Embed Content is a component with a way to bind html from cms with ability to style itself and all descendants generated from cms rich text.

Screenshot 2024-05-06 at 23 36 04
Screenshot 2024-05-06 at 23 35 50


johnsicili (commented):
Love it!

1. May need a different icon/add a tooltip to differentiate between HTML embed
2. I found it confusing why I can change the selector, but maybe there's a reason

Well done this is huge

kof:
- [ ] descendant should only be possible as a child of content embed, right? currently it can be moved around freely to any other instance

TrySound:
alright, changed a few things
- replaced Content Embed component with combination of html embed + descendants,
  Content Embed is builtin template now, just like Sheet is a template around Dialog
- moved Content Embed into "data" category
- disabled all props in descendant component because we currently want prevent any customization

kof:
> @kof descendant is inside of "internal" category which is not accessible from prod

@TrySound yes but once they are on canvas they can be moved to any instance, seems like they lack ancestor constraints?

kof (approved):

Issue #3308, opened 2024-05-06
SiteMap.xml whats left


istarkov:
#### Tasks:
- [x] **Integrate Document Type Selection**
  - Add a "Document Type" dropdown menu on the new/edit page immediately after the "Redirect" option.
  - Options for the dropdown:
    - `HTML`
    - `XML`
  - Hide all subsequent sections on the page when "XML" is selected.

- [x] **Adapt Script for XML Handling**
  - Modify the script generation logic to properly recognize and process files designated as XML document types.
  - Check behaviors for files with a `/sitemap.xml` path to differ from current processing rules.

- [x] **UI Adjustments for XML Pages**
  - For pages set to the XML document type, only display the "Collection" and "XMLNode" components.

PR #3316, opened 2024-05-06
build: Ability to publish Pages with document type XML

istarkov:
## Description

ref #3308 

Ability to render Pages with document type equal to `xml` 

If page with document type xml is created and has path `/sitemap.xml` this page is rendered as sitemap
otherwise implicit sitemap.xml is generated.

On publish we use following logic for rendering.

With doctype === xml
- we are rendering XmlNode as is
- `Body` rendered as `(props) => props.children`, 
- all other elements as `() => null`

- [x] - sitemap.xml must be excluded from sitemap

### Example Saas Sitemap with $resource and custom section and headers mixed with XmlNode

- App builder https://xml-gen.staging.webstudio.is/builder/d6bcc915-6634-4e2d-9ef1-90e8e3836003?pageId=KrgWBi6ANVxH2BfRbar19
- Published https://saas-sitemapxml-x1g2n.wstd.work/sitemap.xml

### Implicit sitemap
- App builder https://xml-gen.staging.webstudio.is/builder/96168940-4fbd-454a-9a03-ba4f2dffbbec
- Published https://saas-january-th6m1.wstd.work/sitemap.xml

Example with custom section and non XmlNode elements
image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/cli/src/prebuild.ts
```diff
@@ -489,15 +489,30 @@ export const prebuild = async (options: {
...
 
   spinner.text = "Generating routes and pages";
 
-  const routeTemplatePath = normalize(join(cwd(), "app/routes/template.tsx"));
+  // MARK: - Route templates read
+  const routeTemplatesDir = join(cwd(), "app/route-templates");
 
-  const routeFileTemplate = await readFile(routeTemplatePath, "utf8");
+  const routeTemplatePath = normalize(
+    join(routeTemplatesDir, "html-doc-type.tsx")
```
TrySound, line 500: html-template?

packages/cli/src/prebuild.ts
```diff
@@ -589,62 +639,67 @@ export const prebuild = async (options: {
     const pageMeta = pageData.page.meta;
     const favIconAsset = assets.get(projectMeta?.faviconAssetId ?? "");
     const socialImageAsset = assets.get(pageMeta.socialImageAssetId ?? "");
+
+    // MARK: - TODO: XML GENERATION
     const pageExports = `/* eslint-disable */
-/* This is a auto generated file for building the project */ \n
+      /* This is a auto generated file for building the project */ \n
```
TrySound, line 645: we don't run prettier for user code, would be weird indent

packages/cli/src/prebuild.ts
TrySound: or just html.tsx

packages/sdk/src/resources-generator.ts
```diff
@@ -89,7 +89,7 @@ export const generateResourcesLoader = ({
   generated += `import { loadResource, isLocalResource, type System } from "@webstudio-is/sdk";\n`;
 
   if (hasResources) {
-    generated += `import { sitemap } from "./[sitemap.xml]";\n`;
+    generated += `import { sitemap } from "./$resources.sitemap.xml";\n`;
```
TrySound, line 92: why this name?

packages/cli/src/prebuild.ts
istarkov: Fully unreadable for me otherwise. Also additional cognitive load to think about writing string code being looking good.

packages/sdk/src/resources-generator.ts
istarkov: just same as `$resources/sitemap.xml` in resource url

packages/cli/src/prebuild.ts
istarkov: Moved to `route-template` dir. So removed `template` keyword.
Renamed to just html, xml

packages/cli/templates/defaults/app/route-templates/implicit-sitemap.tsx
```diff
@@ -0,0 +1,34 @@
+import type { LoaderFunctionArgs } from "@remix-run/server-runtime";
```
kof, line 1: suggestion: default-sitemap.tsx

packages/cli/templates/defaults/app/route-templates/xml.tsx
```diff
@@ -0,0 +1,61 @@
...
+  url.protocol = "https";
+
+  const params = getRemixParams(arg.params);
+
+  const system = {
+    params,
+    search: Object.fromEntries(url.searchParams),
+    origin: url.origin,
+  };
+
```
kof, line 30: so every new system variable we will add will have to be added in all of these places and probably more? I think we already now know there will be many more system variables and we need to use a function

Additionally would be nice if system had a ts type System

packages/sdk-components-react/src/xml-node.tsx
```diff
@@ -1,15 +1,46 @@
...
+} from "react";
 
 export const defaultTag = "div";
 
 // We don't want to enable all tags because Box is usually a container and we have specific components for many tags.
 type Props = {
   tag: string;
+  xmlns?: string;
   children: ReactNode;
+  // xxxAttribute is used for typings only
```
kof, line 22: typo?

packages/sdk/src/page-utils.ts
```diff
@@ -79,6 +79,7 @@ export const getStaticSiteMapXml = (pages: Pages, updatedAt: string) => {
   const allPages = [pages.homePage, ...pages.pages];
   return (
     allPages
+      .filter((page) => (page.meta.documentType ?? "html") === "html")
```
kof, line 82: I read this twice, more straight forward
```suggestion
      .filter((page) => page.meta.documentType  === "html" ||  page.meta.documentType === undefined)
```

kof (changes requested):
Lets get the System Resource in? or did you want to merge as-is and make a second PR?

packages/sdk-components-react/src/xml-node.tsx
istarkov: No see code below. 
```
[idAttribute]: string;
[componentAttribute]: string;
```

packages/sdk/src/page-utils.ts
istarkov: why? any metrics why 2nd is better?
if you don't like brakets it can be `"html" === page.meta.documentType ?? "html"` anyway for me current is much more readable vs `||`

packages/cli/templates/defaults/app/route-templates/xml.tsx
kof: 1. turns out type System already exists in the builder
2. Reason why I think you can't just add another system like this without adding a type is because now we can't be sure where are all the places where we need to also add origin, it can fail without a warning

kof (approved):
- [ ] missing UI for System Resource (next pr)
- [x] please rename implicite-sitemap to default-sitemap (this is more accurate semantics)

packages/cli/templates/defaults/app/route-templates/xml.tsx
TrySound: all good here, loadResources and getPageMeta validate it

kof: according to @TrySound its being checked in the function we pass it to

PR #3319, opened 2024-05-07
feat: parse complex values with available parsers for sections

### LINKED ISSUE #3294: Box shadow is set but isn't listed in the layers
Here is a repro https://apps.webstudio.is/builder/7bdea1b5-202f-43dd-8891-579b7e9803e4?authToken=bed1507d-8217-4844-b998-728e598beb6c&mode=preview

You can copy it into local  env to work on it locally

1. Click on Use Case container
2. Select Neon Card token
3. See box shadow is there but its not being listed in the layers

image

JayaKrishnaNamburu:
## Description

fixes #3294 

This PR does two things,
- When users add complex values like `box-shadow`, `filter`, `transition`, `text-shadow`. They are parsed and added to the data layer accordingly. This will help the sections to properly handle them using the style-panel. 
- For projects that already have `unparsed` complex values in them. They are just showing a blue label as local state in the `ProeprtyName` component. We need to display default style, because even though the property is unparsed and valid. The sections can't take advantage of it to help it edit with UI.
Screenshot 2024-05-07 at 21 21 33

## Steps for reproduction

1. Add a property like `filter`, `box-shadow` in the advanced tab.
2. Now, start editing the value in the style-panel.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/collapsible-section.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/collapsible-section.tsx
@@ -26,7 +26,7 @@ export const getDots = (
       style?.value.type === "unparsed" ||
       style?.value.type === "guaranteedInvalid"
     ) {
-      return undefined;
+      return;
     }
 
     const source = getStyleSource(style);
--- a/apps/builder/app/builder/features/style-panel/shared/collapsible-section.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/collapsible-section.tsx
@@ -19,7 +19,17 @@ export const getDots = (
   const dots = new Set<"local" | "overwritten" | "remote">();
 
   for (const property of properties) {
-    const source = getStyleSource(currentStyle[property]);
+    const style = currentStyle[property];
+
+    // Unparsed values are not editable directly in the section, so we don't show the dot
+    if (
+      style?.value.type === "unparsed" ||
+      style?.value.type === "guaranteedInvalid"
+    ) {
+      return undefined;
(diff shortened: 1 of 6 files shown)


kof:
Just to clearify: what will UI look like when there is already existing unparsed value in transition for example?

apps/builder/app/builder/features/style-panel/shared/collapsible-section.tsx
```diff
@@ -19,7 +19,17 @@ export const getDots = (
...
   for (const property of properties) {
-    const source = getStyleSource(currentStyle[property]);
+    const style = currentStyle[property];
+
+    // Unparsed values are not editable directly in the section, so we don't show the dot
+    if (
+      style?.value.type === "unparsed" ||
+      style?.value.type === "guaranteedInvalid"
+    ) {
+      return undefined;
```
kof, line 29:
```suggestion
      return;
```

kof (approved):
looking good to me

JayaKrishnaNamburu:
> Just to clarify: what will UI look like when there is already existing unparsed value in transition for example?

If there is any existing `unparsed` value that has a style panel as well. The style-panel doesn't show any UI stuff like, `blue-chip` or `blue-dots`. But the value can still be edited using the `Advanced` tab. So sections only respond to structured and parsed data and advanced tab can take care of `unparsed` stuff.

istarkov (approved):

PR #3336, opened 2024-05-09
refactor: bump asset uploader dependencies

TrySound:
fontkit types added missing pieces. Figured otf is interpreted as ttf so no need to have it as separate type.

Testing
- upload font
- upload image


kof (approved):
haven't tested, should I?

PR #3342, opened 2024-05-11
fix: support multiline graphql in curl

TrySound:
Here fixed the issue with parsing curl by switching argv parser.


istarkov:
Doesn't work: 

Create resource

image

Cupy as curl, then paste in url field of the other resource and get different result

image

kof (approved):

PR #3346, opened 2024-05-12
experimental: autocomplete path params

TrySound:
Here added support for autocompleting inferred path params where their values are not specified by user.

Screenshot 2024-05-12 at 14 33 23


kof:
Do we need to show undefined as a value? I would show nothing  in the value inside the autocomplete

kof (approved):

TrySound:
Screenshot 2024-05-13 at 00 13 42

PR #3360, opened 2024-05-15
refactor: define resource control explicitly

TrySound:
Here simplified detecting system resource control with explicit resource field. Without it ui falls back to http ui.

Will be useful for graphql ui later.

Also disabled system resource option for non-paid users

## Testing

- create sitemap resource
- close variable popover
- open again and see system control instead of default http control


kof (approved):
didn't test

PR #3365, opened 2024-05-15
refactor: align text area error with input field

TrySound:
Here replaced state=invalid with color=error in textarea to align with input field and fixed focused input with error style.

image


kof (approved):
didn't test, would be easier if a visual change comes with a screenshot attached

TrySound:
added

PR #3366, opened 2024-05-15
refactor: use native validation for resource url

TrySound:
Here integrated native fields validation into our form utilities. It will let us have more isolated validation complexity within each field.

The trick was to detect if any error are not shown. Solved by iterating form.elements and checking its input.validity.valid and data-color=error attribute.

Screenshot 2024-05-16 at 01 08 37

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
@@ -2,9 +2,12 @@ import { computed } from "nanostores";
 import { nanoid } from "nanoid";
 import {
   forwardRef,
+  useCallback,
+  useEffect,
   useId,
   useImperativeHandle,
   useMemo,
+  useRef,
   useState,
 } from "react";
 import { useStore } from "@nanostores/react";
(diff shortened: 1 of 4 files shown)


kof (approved):

apps/builder/app/builder/features/settings-panel/resource-panel.tsx
```diff
@@ -51,7 +54,104 @@ import {
...
+}: {
+  aliases: Map<string, string>;
+  scope: Record<string, unknown>;
+  value: string;
+  onChange: (value: string) => void;
+  onCurlPaste: (curl: CurlRequest) => void;
+}) => {
+  const urlId = useId();
+  const ref = useRef<HTMLTextAreaElement>(null);
+  const [error, setError] = useState("");
```
kof, line 74: not sure but I think you could use html as well as a way to set error message

TrySound: of course I could, but I need to render it in tooltip, not only style when it exists

PR #3369, opened 2024-05-16
feat: Add Time Component

istarkov:
## Description

Mimics https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time 
The main difference as we use it to convert datetime field into Human readable format

Play here:
https://time.development.webstudio.is/builder/dd5668ed-76c3-4c82-aacd-4fa21411d5ae

TODO:
- [ ] - We need to think about init text see image below

## Steps for reproduction

We need to think about init text

image

image

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/time.tsx
```diff
@@ -0,0 +1,371 @@
...
+    const locale = `${languageOrDefault(language)}-${countryOrDefault(
+      country
+    )}`;
+
+    const options: Intl.DateTimeFormatOptions = {
+      dateStyle: dateStyleOrUndefined(dateStyle),
+      timeStyle: timeStyleOrUndefined(timeStyle),
+    };
+
+    const datetimeString = datetime === null ? "" : datetime.toString();
```
TrySound, line 350: why null?

TrySound: I would better do String(datetime) as input can be undefined as well

packages/sdk-components-react/src/time.ws.ts
```diff
@@ -0,0 +1,33 @@
...
+  type WsComponentMeta,
+  type WsComponentPropsMeta,
+} from "@webstudio-is/react-sdk";
+
+import { time } from "@webstudio-is/react-sdk/css-normalize";
+
+import { props } from "./__generated__/time.props";
+
+const presetStyle = {
+  time: [...time],
```
TrySound, line 15: why spread?

packages/sdk-components-react/src/time.tsx
istarkov: Null is not overwritten with = DEFAULT

packages/sdk-components-react/src/time.ws.ts
```diff
@@ -0,0 +1,33 @@
...
+} satisfies PresetStyle<"time">;
+
+export const meta: WsComponentMeta = {
+  category: "data",
+  type: "container",
+  label: "Time",
```
TrySound, line 21: Optional, can be inferred

packages/sdk-components-react/src/time.ws.ts
istarkov: copypasted, in case of any other styles will be added

packages/sdk-components-react/src/time.tsx
TrySound: 🤷 render "null" then

packages/icons/src/__generated__/svg/time.ts
```diff
@@ -0,0 +1 @@
...a3.432 3.432 0 0 0-.038.5l-.011.414H4.998l-.011-.414a3.432 3.432 0 0 0-.038-.5c-.036-.115-.207-.3-.332-.361a.689.689 0 0 0-.414-.053m-.516 3.026c0 .386.004.422.055.532a.734.734 0 0 0 .314.328.777.777 0 0 0 .523.014.699.699 0 0 0 .346-.338c.048-.109.055-.167.064-.536l.01-.415h4.002l.01.415c.009.369.016.427.064.536a.699.699 0 0 0 .346.338c.129.057.391.05.523-.014a.734.734 0 0 0 .314-.328c.051-.11.055-.146.055-.53v-.412l.333.01c.407.013.692.045.846.095.333.107.672.458.765.792.059.212.086.598.086 1.227v.578H1.657v-.578c0-.629.027-1.015.086-1.227.093-.335.434-.688.765-.791.182-.057.277-.07.665-.087l.438-.02.076-.004v.415m8.656 6.183c-.001 2.471-.01 2.897-.072 3.167-.065.288-.345.632-.63.777-.324.165-.086.157-4.641.157s-4.317.008-4.641-.157c-.285-.145-.565-.489-.63-.777-.062-.27-.071-.696-.072-3.167V7.327h10.686v2.584m-5.547-1.86c-.098.03-1.16.736-1.264.841-.233.233-.236.626-.008.886a.666.666 0 0 0 .7.167l.123-.047v1.459l-.195.011c-.211.011-.337.064-.463.197-.231.244-.215.685.033.909.194.175.168.172 1.278.172s1.084.003 1.278-.172c.248-.224.264-.665.033-.909-.126-.133-.252-.186-.463-.197l-.195-.011V9.976c0-1.555.002-1.534-.166-1.725a.562.562 0 0 0-.452-.216.98.98 0 0 0-.239.016"/></svg>`;
```
TrySound, line 1: looks like original icon was smaller

packages/sdk-components-react/src/time.tsx
istarkov: new Date(null).getTime() Valid ;-) so need to be excluded

istarkov: new Date("").getTime() gives NaN

packages/icons/src/__generated__/svg/time.ts
istarkov: We use fill for every icon we have.

packages/sdk-components-react/src/time.tsx
TrySound: shite

istarkov: Renamed "" to INVALID_DATE_STRING

TrySound (approved):

kof:
Questions I have

1. Using Datetime property vs Text Content similar to how html <time>1978-08-18T10:00:00Z</time> would do it

no strong opinion just wonder if you considered 

2. Language and country are separate fields, what happens if data source have this locale thing en_US that has both in one, how will we get that bound?

kof (approved):

packages/sdk-components-react/src/time.tsx
```diff
@@ -0,0 +1,373 @@
...
+    const options: Intl.DateTimeFormatOptions = {
+      dateStyle: dateStyleOrUndefined(dateStyle),
+      timeStyle: timeStyleOrUndefined(timeStyle),
+    };
+
+    const datetimeString =
+      datetime === null ? INVALID_DATE_STRING : datetime.toString();
+
+    const date = new Date(datetimeString);
+    const isValidDate = !Number.isNaN(date.getTime());
```
kof, line 355:
```suggestion
    const isValidDate = Number.isNaN(date.getTime()) === false;
```

packages/sdk-components-react/src/time.tsx
```diff
@@ -0,0 +1,373 @@
...
+    : DEFAULT_COUNTRY;
+};
+
+const dateStyleOrUndefined = (
+  value: unknown
+): Intl.DateTimeFormatOptions["dateStyle"] => {
+  if (["full", "long", "medium", "short"].includes(value as string)) {
+    return value as Intl.DateTimeFormatOptions["dateStyle"];
+  }
+  return undefined;
```
kof, line 318:
```suggestion
```

packages/sdk-components-react/src/time.tsx
```diff
@@ -0,0 +1,373 @@
...
+    const date = new Date(datetimeString);
+    const isValidDate = !Number.isNaN(date.getTime());
+
+    let formattedDate = datetimeString;
+
+    if (isValidDate) {
+      try {
+        formattedDate = new Intl.DateTimeFormat(locale, options).format(date);
+      } catch {
+        /* Do Nothing */
```
kof, line 363: shouldn't we at least warn about invalid date?

packages/sdk-components-react/src/time.ws.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const meta: WsComponentMeta = {
+  category: "data",
+  type: "container",
+  description:
+    "Converts machine-readable date and time to a human-readable format.",
+  icon: TimeIcon,
+  states: defaultStates,
+  presetStyle,
+  order: 0,
```
kof, line 26: as a last item please

packages/sdk-components-react/src/time.tsx
istarkov: It can be invalid. We just passthrough it then. Have no strong opinion should we be silent or expose some error. 
Probably in the future we can somehow mark such components as invalid with red border or like

PR #3377, opened 2024-05-18
experimental: support history in address bar

TrySound:
Address bar need to fill every time user want to test the page. Here added history suggestions which let users reuse peviously entered params.

Had to implement own combobox-like behavior to support syggestions universally on all inputs.

Screenshot 2024-05-18 at 17 35 54

### DIFF
--- a/apps/builder/app/builder/features/address-bar.tsx
+++ b/apps/builder/app/builder/features/address-bar.tsx
@@ -261,7 +261,7 @@ const Suggestions = ({
           type="button"
           role="option"
           tabIndex={-1}
-          css={{ textTransform: "none" }}
+          css={{ textTransform: "none", whiteSpace: "nowrap" }}
           onMouseEnter={(event) => {
             // select option on hover
             const options =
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/address-bar.tsx
```diff
@@ -101,9 +133,166 @@ const useCopyUrl = (pageUrl: string) => {
...
+          "[role=option][aria-selected=true]"
+        );
+        if (selected instanceof HTMLElement) {
+          // avoid submitting form when item is selected
+          event.preventDefault();
+          selected.click();
+        }
+      }
+    };
+    container.addEventListener("input", handleInput);
```
kof, line 230: was it impossible to use react for the handlers?

TrySound: yes, but would require using more refs instead of variable inside of effect

kof (approved):

apps/builder/app/builder/features/address-bar.tsx
kof: feels like the ritght way to use react tbh, since it does its events delegation stuff and fixes things for some events

PR #3388, opened 2024-05-20
feat: restore system params from history first entry

TrySound:
We want users to get good first impression and show already working cms page when clone from marketplace. Here improved bindings logic to always use the first history entry as default.

Screenshot 2024-05-20 at 13 41 17


kof (approved):
Didn't test!

PR #3393, opened 2024-05-21
refactor: switch variable values to native validation

TrySound:
Here migrated all left fields to native navigation. Got rid from legacy useField hook.
Replaced number input with just text input as more reliable. And reused expression linter for json value.

Check following forms
- string
- number
- boolean
- json

image


kof (approved):
tested

apps/builder/app/builder/features/settings-panel/variable-popover.tsx
```diff
@@ -704,6 +693,7 @@ export const VariablePopoverTrigger = forwardRef<
             }}
           >
             <form
+              noValidate={true}
```
kof, line 696:
```suggestion
              noValidate
```

PR #3403, opened 2024-05-22
feat: match unambiguous variables by name

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3401

Simple logic here, one variable matched by name then can be replaced with actual variable id otherwise user has to select variable explicitly.

Screenshot 2024-05-23 at 00 37 07


johnsicili:
Beautiful!! Will make sharing expressions much easier. Works well. Thanks!

kof (approved):

PR #3410, opened 2024-05-23
refactor: Move text placeholders to meta templates

kof:
## Description

Part of https://github.com/webstudio-is/webstudio/issues/3149 this allows us to understand if an instance has text child.
With this knowledge we can place instances into components that don't have text children and after them if they do.

## Steps for reproduction

1. insert text or paragraph
2. see there is expected placeholder text which is editable
3. select it, tell AI to change color or something
4. see it works and see in payload it didn't send placeholder text
5. try text editing on empty body - shouldn't allow it

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/react-sdk/src/component-generator.ts
+++ b/packages/react-sdk/src/component-generator.ts
@@ -275,6 +275,7 @@ export const generateJsxChildren = ({
   usedDataSources,
   indexesWithinAncestors,
   classesMap,
+  excludePlaceholders,
 }: {
   scope: Scope;
   children: Instance["children"];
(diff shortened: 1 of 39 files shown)


packages/react-sdk/src/component-generator.ts
```diff
@@ -288,6 +288,9 @@ export const generateJsxChildren = ({
   let generatedChildren = "";
   for (const child of children) {
     if (child.type === "text") {
+      if (child.placeholder) {
```
TrySound, line 291: I would keep it rendered so user see consistent content in canvas and builder

kof: so would need to filter out placeholders before passing to this function?

TrySound: no need to filter at all, the flag is for ai only

TrySound: ah yes, you are right. need to filter before passing, forgot it is used in jsx utils

TrySound (approved):

PR #3412, opened 2024-05-24
feat: Add postgresql patches

istarkov:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/patches/apply-patches.js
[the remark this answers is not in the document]
istarkov: FOrgot to delete Thanks!

TrySound (approved):

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
```diff
@@ -0,0 +1,200 @@
...
+};
+
+/**
+ *
+ * @param {string} dataStr
+ * @param {string} primaryKeyCommaSeparated - if empty then it's js array or plain object, otherwise is a map
+ * @param {string} patchesStr
+ * @returns {string}
+ */
+const patch = (dataStr, primaryKeyCommaSeparated, patchesStr) => {
```
kof, line 147: If I understood our discussion correctly, we found a way to move this function outside of plv8

TrySound: We found a way to move computing keys outside of plv8

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * @param {*} item
+   * @returns
+   */
+
+  const primaryKeyArray = primaryKeyCommaSeparated
+    .split(",")
+    .map((item) => item.trim());
+
+  const isPlainObjectOrArray =
```
kof, line 157: why do we need both?

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
```diff
@@ -0,0 +1,200 @@
...
+        if (primaryKeyArray.length === 1) {
+          return item[primaryKeyArray[0]];
+        }
+
+        return primaryKeyArray
+          .map((key) => (item[key] == null ? "" : item[key]))
+          .join(":");
+      };
+
+  let data = JSON.parse(dataStr);
```
kof, line 175: shoudln't this be in try/catch?

istarkov: why? let it fail

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
istarkov: what means both?

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
kof: do we have ways to handle that case?

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
kof: plain object and array

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
```diff
@@ -0,0 +1,200 @@
...
+    }
+  });
+
+  return draft;
+};
+
+/**
+ *
+ * @param {string} dataStr
+ * @param {string} primaryKeyCommaSeparated - if empty then it's js array or plain object, otherwise is a map
```
kof, line 143: if we can avoid havinng multiple types here we should, btw also @param types are wrong and we are not using them anywhere else

istarkov: What means wrong, this is fully typecheked js file. We used JSDoc typescript types before in middleware.js
There are no other ways to have types in js

istarkov: And js file as described in comments copy pasted here

packages/prisma-client/prisma/migrations/20240524155926_patch_map/migration.sql
istarkov: What the reason? Its perfectly ok if this function will fail on json parse. Without any wrapper.

kof: so what happens it it fails, did you try? where does error land? are you catching it somewhere in prisma then as a usual postgres error? or is it blowing up and doing some weird shit)

istarkov: This doesn't change any behavioir with SQL statements we did before. What if any previous SQL fail. It's not the refactor of how errors from SQL are shown on UI or console.  

Will be shown this
```
"errors": "\nInvalid `prisma.$executeRaw()` invocation:\n\n\nRaw query failed. Code: `XX000`. Message: `ERROR: SyntaxError: Unexpected number in JSON at position 1`"
```
As like as if any SQL statement before would fail.

And it's ok as there is 0 chance to get error like this without data corruption on db level what is not we are expecting when writing the code.

istarkov:
Closed in favor of postgrest

PR #3413, opened 2024-05-24
feat: add marketplace section in page settings

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3405

Reusing existing fields for marketplace does not work well when fields are dynamically bound to cms data. Here we replaced existing fields used in marketplace with static ones.

- ws:title ?? page title -> page name
- social image url ?? social image asset -> marketplace thumbnail ?? social image asset
- exclude page from searcg -> marketplace include
- ws:category -> marketplace category

Screenshot 2024-05-24 at 17 15 29

### DIFF
--- a/apps/builder/app/builder/features/project-settings/section-marketplace.tsx
+++ b/apps/builder/app/builder/features/project-settings/section-marketplace.tsx
@@ -300,10 +300,6 @@ export const SectionMarketplace = () => {
 
       <Grid gap={2} css={sectionSpacing}>
         <PanelBanner>
-          <Text>
-            {`You can add a "ws:category" meta tag in the page settings for pages being grouped by category. You can also define
-            "ws:title" for each page; otherwise, the page title will be used.`}
-          </Text>
           <Text color="destructive">
             {`Don't forget to publish your project after every change to make your
             changes available in the marketplace!`}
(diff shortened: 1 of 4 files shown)


kof:
- [ ] when not included we should probably hide the entire block to reduce the visual noise
- [ ] "Include this item in the marketplace" > "Include in the marketplace"
- [ ] "Marketplace item preview " - "Marketplace Preview"

kof:
Just tested, if we are not showing the section in the settings before user hits "start review" then we need to do something about the fact that user needs to change those settings before they hit start review, I see 2 options:

- show the section in page settings always but at least reduce it to the sectino title and first checkbox when not active
- have a checkbox or button that activates marketplace in project settings first, this shows page setting section, so one can change it before submitting for review

kof:
image
- [ ] lets remove the hint for ws: meta

TrySound:
>  when not included we should probably hide the entire block to reduce the visual noise

don't agree, this flag just allow to setup everything before "publishing", no way to setup if flag hides essential fields

TrySound:
maybe publish should trigger review?

kof:
> maybe publish should trigger review?

maybe, technically there is 3 steps:

1. activate marketplace settings in project settings
2. add marketplace page settings
3. publish

without publish, the 2 steps don't matter

kof:
anyways, if we don't have a good idea now, we should just keep thinking

kof (approved):

PR #3414, opened 2024-05-24
fix: avoid interfering with user input in color picker

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3408

Here fixed the issue with auto converting hex while user typing and improved performance while typing or dragging color.


kof (approved):
didn't test

PR #3421, opened 2024-05-26
refactor: host hyphenate-style-name

TrySound:
hyphenate-style-name utility is basically oneliner but requires to install two packages. Considering we will switch to hyphenate property naming by default, no reason to rely on this package.

Also switched to unprefixed property in css preview and property name tooltips.
Just shows background-clip instead of -webkit-background-clip.


kof:
> Also switched to unprefixed property in css preview and property name tooltips.
> Just shows background-clip instead of -webkit-background-clip.

Why is that a good idea? I think user wants to know in the css output what the actual CSS is

kof (approved):

TrySound:
Prefixes are hacks and no longer used by browsers for new features.
Ideally we should not show them too considering background-clip is widely supported.

PR #3428, opened 2024-05-27
fix: exclude experimental css properties

TrySound:
We included a lot of stuff which is not well supported and may change in the future.


packages/css-data/src/__generated__/properties.ts
```diff
@@ -163,36 +143,6 @@ export const properties = {
...
-    unitGroups: [],
-    inherited: false,
-    initial: {
-      type: "keyword",
-      value: "none",
-    },
-    popularity: 0.24468974,
-    appliesTo: "allElements",
-  },
-  aspectRatio: {
```
kof, line 186: we can't delete this :)

PR #3437, opened 2024-05-28
fix: try to avoid race conditions while loading resources

TrySound:
Our stores updates does not support transactions
which leads to a lot of updates with empty state.
Empty and valid states are both used to load resources and override each other. Here workarounded by scheduling loading to the next frame.


kof (approved):
approving it but it looks like a pretty bad hack that will bite us very randomely

PR #3439, opened 2024-05-28
fix: switch to native form in preview

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3435

Remix just does not work well on canvas/preview. Both links and forms are rendered natively without remix.

### DIFF
--- a/packages/sdk-components-react-remix/src/remix-form.tsx
+++ b/packages/sdk-components-react-remix/src/remix-form.tsx
@@ -1,5 +1,11 @@
-import { type ElementRef, type ComponentProps, forwardRef } from "react";
+import {
+  type ElementRef,
+  type ComponentProps,
+  forwardRef,
+  useContext,
+} from "react";
 import { Form, type FormProps } from "@remix-run/react";
+import { ReactSdkContext } from "@webstudio-is/react-sdk";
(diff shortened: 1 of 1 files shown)


packages/sdk-components-react-remix/src/remix-form.tsx
```diff
@@ -12,13 +18,19 @@ export const RemixForm = forwardRef<
...
 >((props, ref) => {
+  const { renderer } = useContext(ReactSdkContext);
   // remix casts action to relative url
   if (
     props.action === undefined ||
     props.action === "" ||
     props.action?.startsWith("/")
   ) {
-    return <Form {...props} ref={ref} />;
+    // remix forms specifies own action instead of provided one
```
kof, line 28: Is the form supposed to work on preview, like fully functional? If yes how would it work without remix?

kof (approved):

packages/sdk-components-react-remix/src/remix-form.tsx
TrySound: We do interception to simulate what remix does. But remix does it with all custom workarounds to support server which does not work with our interceptor.

Issue #2967, opened 2024-03-05
Unify code editing experience in style panel for complex values


kof:
- [x] Put code editor at the bottom of background gradient panel (to be consistent with box shadow, transitions) #3445 
- [x] Add code editor to background image #3445  
- [ ] ~~Use the new code editor component everywhere: backgrounds, transitions, box shadows, filter~~
- [x] When there is only one layer, drag handle shouldn't be visible, this is for all layered properties #3445
- [x] Save text values when closing/blur content box.
- [x] updating values on scrub instead of using ephemeral flag, which makes it slow https://github.com/webstudio-is/webstudio/pull/3318#pullrequestreview-2046045009

TrySound:
Let's not use autocompletion here at all, there is not much to suggest here

JayaKrishnaNamburu:
> Let's not use autocompletion here at all, there is not much to suggest here

I second this, because we are ading the UI controls to manage the same anyways. I will handle the rest and move to transforms.

kof:
Yes, lets not have autocompletion there

PR #3445, opened 2024-05-29
refactor: code editor improvements and editor for background-image

JayaKrishnaNamburu:
## Description

Fixes some of the things from the #2967 
- [x] Put code editor at the bottom of background gradient panel (to be consistent with box shadow, transitions)
- [ ] Add code editor to background image
- [x] When there is only one layer, drag handle shouldn't be visible, this is for all layered properties

Added the code-editor for the `background-image` too. But made it `readOnly`. Because, i am not sure the UX for that flow. The only thing i could think of allowing users to edit is. 
- We can allow users to set `url(https://......)` remote images by editing in the text area. In which case, we can remove the current `readOnly` textarea mode.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging


kof:
- [ ] label should be "Code" with a similar (i) icon and tooltip to others
image

JayaKrishnaNamburu:
bg-image

Here is the flow now after making the `background-image` to be editable.
- Users can choose a image from the `Choose Source` button from the UI and select any uploaded images.
- Users can just `copy/paste` the url from the text-area and paste to any other node. And the area parses it and links the asset instead.
- Users can set any remote `url's` using the text-area. 

Here is the current state of the `background-images` editor now. Please feel to chip in some ideas. 

PS: Right now the values save when users press `Enter` in the text-area. And need to handle the save on `blur` for the textarea.

cc @kof @TrySound

kof:
We can't remove the button on the image like that 
image

In Image component settings we have both, image picker button and an input
image

In this case we don't really need an input if we have the code input below, but the button should always stay:

When there is no source chosen or there is custom url:

image

When the image picker was used
image

JayaKrishnaNamburu:
Screen Recording 2024-06-04 at 4 05 01 PM
 cc @kof  updated the section now to show the `url` input filed all the time. let me know, if the UX and the layout is looking good now.

kof:
Looking good, I think "Image" label should be top-aligned, its currently in the middle between input and button

JayaKrishnaNamburu:
Screenshot 2024-06-04 at 5 05 34 PM

Moved the lalbel to the top now 👍

kof (approved):
seems like all done🎉

apps/builder/app/builder/features/style-panel/sections/backgrounds/background-gradient.tsx
```diff
@@ -86,32 +92,59 @@ export const BackgroundGradient = (
...
+        style={props.currentStyle}
+        description={
+          <>
+            Paste a CSS gradient, for example:
+            <br />
+            <br />
+            linear-gradient(...)
+            <br />
+            <br />
+            If pasting from figma, remove the “background” property name.
```
kof, line 115:
```suggestion
            If pasting from Figma, remove the "background" property name.
```

PR #3449, opened 2024-05-29
feat: improve CLI link command

TrySound:
Switched from readline to @clack/prompts package
with more rich ui docorations and better validation.

It should improve UX and @clack/prompts actually can replace both prompts and ora packages, in fact it is smaller than each of these and has builtin types.

https://github.com/webstudio-is/webstudio/assets/5635476/b03174f5-33ef-47a5-90b8-90ad0697c442


kof (approved):
looking great, please test it well :)

PR #3452, opened 2024-05-30
fix: drop marketplace legacy fallbacks

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3450

Got rid of ws:cetegory meta and excludePageFromSearch fallbacks in marketplace. Social image is the only fallback we agreed to left.


kof:
lets merge

kof (approved):

PR #3453, opened 2024-05-30
feat: improve CLI sync command

TrySound:
@clack/prompts supports spinner out of the box
so we can replace ora with it too.

Here switched sync command to this spinner
and added handling server errors.

https://github.com/webstudio-is/webstudio/assets/5635476/f6984b7c-c782-4859-a58d-335a39340119


kof (approved):
haven't tested

PR #3461, opened 2024-05-31
refactor: remove canvas max width detection

TrySound:
This seem does not affect wide screens anymore. Maybe caz we moved builder to client side rendering. build.breakpoints are no longer used so we can move data loading to client side as well.


kof (approved):
tested, don't see any jumps either, I think the problem didn't exist since we removed SSR from canvas

PR #3462, opened 2024-05-31
fix: Fix open state indication of the tabs in the left side bar

kof:
## Description

Additionally refactored the logic a bit and moved AI to the bottom. This way its more consistent with upper part all handling sidebar panels, being tabs

## Steps for reproduction

1. focus and hover should show a tooltip
2. open tab should have open state

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/sidebar-left/sidebar-left.tsx
+++ b/apps/builder/app/builder/features/sidebar-left/sidebar-left.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { Box, Tooltip, rawTheme } from "@webstudio-is/design-system";
+import { Box, rawTheme } from "@webstudio-is/design-system";
 import { useSubscribe, type Publish } from "~/shared/pubsub";
 import { $dragAndDropState, $isPreviewMode } from "~/shared/nano-states";
 import { panels } from "./panels";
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/features/sidebar-left/sidebar-left.tsx
```diff
@@ -46,59 +47,49 @@ const useHideActiveTabOnPreview = () => {
...
-        <AiIcon size={rawTheme.spacing[10]} />
-      </SidebarTabsTrigger>
-    </Tooltip>
+    <SidebarButton
+      label="AI"
+      data-state={clientSettings.isAiCommandBarVisible ? "active" : undefined}
+      onClick={() => {
+        setClientSetting(
+          "isAiCommandBarVisible",
+          clientSettings.isAiCommandBarVisible === true ? false : true
```
istarkov, line 56: I know your love for `!`
But really in what universe `clientSettings.isAiCommandBarVisible === true ? false : true` is more readable vs
`!clientSettings.isAiCommandBarVisible` 

Even more when I see this:
` clientSettings.isAiCommandBarVisible === true ` I assume that this is done for real reason and the only real reason to compare boolean with `=== true` if it can be undefined or null or even other type
But it's not, it's just boolean.

kof: this one was unnecessary actually

kof, line 56:
```suggestion
          clientSettings.isAiCommandBarVisible ? false : true
```

PR #3464, opened 2024-05-31
refactor: move builder data loading to client side

TrySound:
To synchronize data between tabs it should be consistent and be easily accessible. Loading from server creates huge latency and data can be changed by loading from another tab.

Here moved data loading to client side so it can be shared through IndexedDB.

Added new `/rest/data/:buildId` endpoint and defined global setter and getter for current state.


kof:
Just thinking if /data is the right name or /build if its only for fetching the build

TrySound:
Not only build, but also assets

istarkov (approved):

PR #3469, opened 2024-06-01
feat: insert a variable at the caret position from the list

TrySound:

Here improved variables list in bindings to insert variable at cursor instead of full expression override.

Also bumped codemirror and disabled autocomplete in variable "name".

https://github.com/webstudio-is/webstudio/assets/5635476/2a5482e2-b880-4bcd-abdf-10ce6db68d31


kof (approved):
nice!

PR #3471, opened 2024-06-01
feat: show usage count when delete variable

TrySound:

Now user can see how many usages left. It can give some hint where to look. We will explore in the future how to show the list of usages.

Also fixed usage detection on pages.

Screenshot 2024-06-01 at 18 45 53

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/variables-section.tsx
+++ b/apps/builder/app/builder/features/settings-panel/variables-section.tsx
@@ -31,6 +31,7 @@ import {
   $props,
   $resources,
   $selectedInstanceSelector,
+  $selectedPage,
   $variableValuesByInstanceSelector,
 } from "~/shared/nano-states";
 import { serverSyncStore } from "~/shared/sync";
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/settings-panel/variables-section.tsx
```diff
@@ -124,6 +126,20 @@ const $usedVariables = computed(
         }
       }
     }
+    if (page) {
```
kof, line 129: that list looks extremely suspicious: how are we supposed to know we need to update it if we add more things to page settings?

kof (approved):
- Great first step in this story
- I really don't see that logic with a hardcoded list of settings getting merged without us having issues later

apps/builder/app/builder/features/settings-panel/variables-section.tsx
kof: Please don't merge it like this.

TrySound: no way, we have it in pages ui, in compiler, in utils and in variables.

TrySound: Will try to generalize with incremental updates in the future

PR #3476, opened 2024-06-02
fix: Create a new page and mark it as home in the same step

### LINKED ISSUE #3470: Create a new page and mark it as home in the same step
When creating a page and marking it as home at the same step - it will fail to create a page

## Steps
1. create a page
2. check "make it home page"
3. hit create
4. nothing happens

istarkov:
## Description

closes #3470

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/sidebar-left/panels/pages/page-settings.tsx
+++ b/apps/builder/app/builder/features/sidebar-left/panels/pages/page-settings.tsx
@@ -1508,10 +1508,10 @@ const updatePage = (pageId: Page["id"], values: Partial<Values>) => {
       updatePageMutable(pages.homePage, values, pages.folders);
     }
 
-    for (const page of pages.pages) {
-      if (page.id === pageId) {
-        updatePageMutable(page, values, pages.folders);
-      }
+    const pageToUpdate = pages.pages.find((page) => page.id === pageId);
+
+    if (pageToUpdate !== undefined) {
+      updatePageMutable(pageToUpdate, values, pages.folders);
     }
 
     // swap home page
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/sidebar-left/panels/pages/page-settings.tsx
```diff
@@ -1505,11 +1504,22 @@ const updatePage = (pageId: Page["id"], values: Partial<Values>) => {
       return;
     }
 
+    if (pages.homePage.id === pageId) {
+      updatePageMutable(pages.homePage, values, pages.folders);
+    }
+
+    for (const page of pages.pages) {
```
kof, line 1511: seems like you would rather want to use .find or at least a break; once the page is found, .find is kinda making more semantical sense

kof (approved):

apps/builder/app/builder/features/sidebar-left/panels/pages/page-settings.tsx
istarkov: I just moved that code lines, there are no new code, just order fix

kof: iknow)

TrySound: honestly it's not better, for "semantics" we need .get(id)

kof: .find() returns one element, its semantically clearly better, perf doesn't really matter here, since you won't have thouthands pages

PR #3484, opened 2024-06-03
fix: treat link urls without pathname as self reference

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3478

This fixes the issue with reloading pages when href="?" and href="#" are specified without any pathname.


kof (approved):

kof:
Would have been nice to have tests for these

PR #3486, opened 2024-06-04
fix: show error on canvas when collection data is invalid

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3481 https://github.com/webstudio-is/webstudio/issues/3473

Screenshot 2024-06-04 at 12 54 22

### DIFF
--- a/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
+++ b/apps/builder/app/canvas/features/webstudio-component/webstudio-component.tsx
@@ -139,7 +139,18 @@ const InvalidCollectionDataStub = forwardRef<
 >((props, ref) => {
   return (
     <ErrorStub ref={ref} {...props}>
-      Collection component requires array in data property
+      The Collection component requires an array in the data property. When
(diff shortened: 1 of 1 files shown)


kof:
Lets think of a message that will tell them what to do or even link them to a document that explains it, this is going to be a major point of friction moving forward, lets try and help it as much as possible, @johnsicili please help us out here framing the message and potentially linking to something that's showing how to do it

johnsicili:
Collection doc is live. 

Error suggestion:

Collection component requires an array in the data property. If binding external data, the array is nested somewhere within. [Learn more](https://docs.webstudio.is/university/core-components/collection.md#whats-an-array)

kof:
@johnsicili great job on the doc, how about this text:

The Collection component requires an array in the data property. When binding external data, it is likely that the array is nested somewhere within, and you need to provide the correct path in the binding.

johnsicili:
Works for me

TrySound:
image

kof (approved):

PR #3491, opened 2024-06-04
experimental: Paste from webflow 4

kof:
## Description

- webflow css preset

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/shared/copy-paste/plugin-webflow/__generated__/style-presets.ts
```diff
@@ -0,0 +1,7212 @@
...
+    },
+    {
+      property: "display",
+      value: {
+        type: "keyword",
+        value: "block",
+      },
+    },
+  ],
+  "w-file-upload-uploading-btn": [
```
TrySound, line 2657: what are these? Do not look like tags

kof: the keys are either tags or classes, I will start using them in the next prs

kof: its how webflow defines styles on their custom abstractions like a link button etc

TrySound: they are not copied to clipboard too?

TrySound:
I think it makes sense to not parse values prematurely only on demand. This will reduce the size of generated module a lot and reduce loading time. Values is a more straightforward piece to parse.

kof:
> I think it makes sense to not parse values prematurely only on demand. This will reduce the size of generated module a lot and reduce loading time. Values is a more straightforward piece to parse.

We will see in the later PRs how much is actually unsused in the end, seems premature right now

kof:
Also if doing it lazy, then we need to load the CSS file instead of prebuilding and that's what I mentioned earlier the difference between 40 vs 120kb

kof:
Additional downside of runtime parsing is if there is something wrong being generated its not immediately obvious

TrySound:
I mean specifically leave values unparsed. Not whole structure. So it is still easy to access them.

TrySound (approved):

PR #3492, opened 2024-06-04
build: automatically create release

TrySound:
Here added workflow to automatically create release when tag is pushed to the repo.

All unconfigured types will go into "Other changes" section

Example of changelogs
https://github.com/TrySound/test-auto-release/releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,10 +15,10 @@ jobs:
       id: changelog
       with:
         commit-types: |
-          feat: New Features
-          fix: Bug Fixes
+          feat: Features
+          fix: Fixes
           docs: Documentation
-          experimental: Experimental features
+          experimental: Experimental
         semver: false
 
     - uses: ncipollo/release-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      contents: write
+    steps:
+    - uses: ardalanamini/auto-changelog@v4
+      id: changelog
+      with:
+        commit-types: |
+          feat: New Features
+          fix: Bug Fixes
+          docs: Documentation
+          experimental: Experimental features
```
istarkov, line 21: Should be matched ?
https://github.com/webstudio-is/webstudio/blob/3fa4afa2691227a77ded5248a6b3cdfb05c18630/.github/workflows/lint-pull-request.yaml#L29-L41

istarkov (approved):

.github/workflows/release.yml
TrySound: Most of these are not important to user and many we just don't use

istarkov: What it will do if chore etc will be in our PRs

TrySound: Put it into "Other changes" section

PR #3493, opened 2024-06-04
fix: close variable after creating

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3480

Here always close popover when variable is created to avoid duplicating with a few "Enter" presses on "name" field.


kof (approved):
haven't tested

PR #3502, opened 2024-06-06
refactor: allow show/hide options for filters and backdrop-filters

### LINKED ISSUE #3451: Allow `show/hide` for layers in filters, backdrop-filters
Properties like `background-image`, `box-shadow` and `transitions` are handled using `layers` as data-structure. Which allows individual layers to be hide-able by default. 
But `filters` can't be handled like `layers` because individual filter properties can't be appended using `,` at the time of generating the css value. And we use `tuple` so the individual filters can be appended without `,`. Which means all the properties should be able to have `show/hide` flags and so each value can be handled.

JayaKrishnaNamburu:
## Description

fixes #3451 
Allow `show/hide` for individual layers in `filters` and `backdrop-filters`. 

## Steps for reproduction

1. Add some layers for filters and backdrop-filters
2. Play with `show/hide` icon on each layer.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


apps/builder/app/builder/features/style-panel/style-layer-utils.ts
```diff
@@ -41,28 +41,34 @@ export const deleteLayer = <T extends TupleValue | LayersValue>(
   batch.publish();
 };
 
-export const hideLayer = (
+export const hideLayer = <T extends LayersValue | TupleValue>(
   property: StyleProperty,
   index: number,
-  layers: LayersValue,
+  layers: T,
```
kof, line 47: Please use a full word always when possible, here clearly it could be 
```suggestion
  layers: Layers,
```

packages/css-engine/src/core/to-value.ts
```diff
@@ -113,6 +113,11 @@ export const toValue = (
   }
 
   if (value.type === "function") {
+    // Right now, we are using function-value only for filter and backdrop-filter functions
+    if (value.hidden === true) {
```
kof, line 117: I would be ok with this too
```suggestion
    if (value.hidden) {
```

PR #3506, opened 2024-06-06
fix: generate conditional collection

TrySound:
Fixes https://github.com/webstudio-is/webstudio-community/discussions/151

Here fixed the issue with generating conditional collection code for published sites. Added fragment around collection when rendered inside of condition.


kof:
What I don't like is the fact that things like these break a build. Its not the first time and certainly not the last one. Broken generation should not result in a broken build entirely.

kof (approved):

PR #3513, opened 2024-06-07
build: run webstudio cli without rebuilding

TrySound:
Finally figured out how to leverage node conditions to run webstudio cli in fixtures without rebuilding it on every change.

Import conditions is another thing which allows
to constrain not exposed modules but internal ones.

Now to run cli locally it should be have this env
which is basically same we do with vite.
Works for both import and export conditions.

```
NODE_OPTIONS='--conditions=webstudio --import=tsx' webstudio
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,6 +9,12 @@
     "webstudio-cli": "./bin.js",
     "webstudio": "./bin.js"
   },
+  "imports": {
+    "#cli": {
+      "webstudio": "./src/cli.ts",
+      "default": "./lib/cli.js"
+    }
+  },
   "files": [
     "lib/*",
     "templates/*",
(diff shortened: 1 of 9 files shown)


packages/cli/package.json
```diff
@@ -19,7 +25,6 @@
     "typecheck": "tsc",
     "checks": "pnpm typecheck",
     "build": "rm -rf lib && esbuild src/cli.ts --outdir=lib --bundle --format=esm --packages=external",
-    "local-run": "tsx --no-warnings ./src/bin.ts",
```
kof, line 22: what's the replacement for this? `NODE_OPTIONS='--conditions=webstudio --import=tsx' webstudio` is this something we need to memorize to run it localy?

kof: or did you want to set NODE_OPTIONS into the .env by default?

TrySound: > something we need to memorize to run it localy?

not memoize, copy paste
local-run script is useless it is in cli package

istarkov (approved):

PR #3516, opened 2024-06-07
refactor: merge icon modules

TrySound:
Our vite setup in builder is getting slower
because of huge amount of modules.

Here I merged more than 400 icon modules into 2 modules string and components.

This should improved vite experience and slightly reduce bundling time.


kof (approved):
are we relying on treeshaking to avoid loading them all?

PR #3526, opened 2024-06-09
feat: Builder loading delight

kof:
Closes https://github.com/webstudio-is/webstudio/issues/3520

## Description

Before: 

https://github.com/webstudio-is/webstudio/assets/52824/4ba347e6-8994-45c6-bdee-cdd0738539fa

After:

https://github.com/webstudio-is/webstudio/assets/52824/416b4614-cf8d-4b2e-bf34-439fb65d3195

## Steps for reproduction

1. reload builder with large amount of data
2. click from dashboard to a project

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -17,7 +17,6 @@ import {
   Flex,
   Grid,
   Progress,
-  rawTheme,
 } from "@webstudio-is/design-system";
 import type { AuthPermit } from "@webstudio-is/trpc-interface/index.server";
 import { createImageLoader } from "@webstudio-is/image";
@@ -65,7 +64,7 @@ import type { TokenPermissions } from "@webstudio-is/authorization-token";
 import { useToastErrors } from "~/shared/error/toast-error";
 import { canvasApi } from "~/shared/canvas-api";
 import { loadBuilderData, setBuilderData } from "~/shared/builder-data";
-import { Webstudio1cIcon, WebstudioIcon } from "@webstudio-is/icons";
+import { WebstudioIcon } from "@webstudio-is/icons";
 
 registerContainers();
 
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -255,12 +255,11 @@ const Loading = ({ value }: { value: number }) => {
   useEffect(() => {
     const timerId = setInterval(() => {
       setProgress((progress) => {
-        if (progress >= 100) {
-          clearInterval(timerId);
-          return progress;
-        }
-        return progress + 1;
+        return progress >= 100 ? progress : progress + 1;
       });
+      if (progress >= 100) {
+        clearInterval(timerId);
+      }
     }, 50);
     return () => {
       clearInterval(timerId);
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -60,9 +60,11 @@ import { useToastErrors } from "~/shared/error/toast-error";
 import { canvasApi } from "~/shared/canvas-api";
 import { loadBuilderData, setBuilderData } from "~/shared/builder-data";
 import { WebstudioIcon } from "@webstudio-is/icons";
-import { atom } from "nanostores";
+import { atom, computed } from "nanostores";
 
 registerContainers();
+
+const $dataLoadingState = atom<"idle" | "loading" | "loaded">("idle");
 
 const useSetWindowTitle = () => {
   const project = useStore($project);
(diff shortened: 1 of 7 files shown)


istarkov:
When I open Many Nodes project on 1-3mb internet I have now, after loading animation I see this for 5s+

image

apps/builder/app/builder/builder.tsx
```diff
@@ -208,6 +227,72 @@ const NavigatorPanel = ({
...
+
+const Loading = ({ value }: { value: number }) => {
+  const [progress, setProgress] = useState(value);
+
+  // This is an approximation of a real progress that should come from "value".
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      setProgress((progress) => {
+        if (progress >= 100) {
+          clearInterval(timerId);
```
istarkov, line 259: side effects in set? plz fix

istarkov (approved):
Loading big projects on  +- ordinary internat has issue with no loading indicator at all after it reaches 100

Sideeffects inside set

apps/builder/app/builder/builder.tsx
```diff
@@ -208,6 +227,72 @@ const NavigatorPanel = ({
...
+      clearInterval(timerId);
+    };
+  }, []);
+
+  // When real value from progress comes in, we reset our estimate.
+  useEffect(() => {
+    setProgress(value);
+  }, [value]);
+
+  if (progress === 100) {
```
istarkov, line 275: See my screenshot. Better to do progress non linear
something like https://easings.net/#easeOutExpo 
so it reaches to 80% like linear and then has long period to reach 100%

apps/builder/app/builder/builder.tsx
kof: I don't like this entire effect, how can it be better written?

istarkov: css keyframes, transitions, some react animate library
as imo any animation should rely on time (RAF https://github.com/streamich/react-use/blob/ade8d3905f544305515d010737b4ae604cc51024/src/useRaf.ts#L4) or css

kof: Yes, though in this case, its not controlling each keyframe, that's done by the transition, its just setting rough steps every X ms, with a raf would be calling more state updates than this, I have now made it a bit better, but generally seems fine now

apps/builder/app/builder/builder.tsx
```diff
@@ -208,6 +223,108 @@ const NavigatorPanel = ({
...
+  const readyCount = Array.from(states.values()).filter(
+    (state) => state === "ready"
+  ).length;
+  const state = readyCount === states.size ? "ready" : "loading";
+  const progress = Math.round((readyCount / states.size) * 100);
+
+  return { state, progress, states };
+};
+
+const ProgressIndicator = ({ value }: { value: number }) => {
```
istarkov, line 284: This is something what starts from `value` so value is `initialValue` here

kof: I have 2 values, real value, which is value and a fakeValue. Its not just initial, because its updating, semantically in useState initial value is not updating

apps/builder/app/builder/builder.tsx
```diff
@@ -208,6 +223,108 @@ const NavigatorPanel = ({
...
+
+  // This is an approximation of a real progress that should come from "value".
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      setFakeValue((fakeValue) => {
+        fakeValue++;
+        return Math.min(fakeValue, 100);
+      });
+    }, 50);
+    if (value >= 100) {
```
istarkov, line 295: place it first in effect

apps/builder/app/builder/builder.tsx
```diff
@@ -208,6 +223,108 @@ const NavigatorPanel = ({
...
+    new Map([
+      ["dataLoadingState", "initial"],
+      ["selectedInstanceRenderState", "initial"],
+      ["canvasIframeState", "initial"],
+    ])
+  );
+
+  const dataLoadingState = useStore($dataLoadingState);
+  const selectedInstanceRenderState = useStore($selectedInstanceRenderState);
+  const canvasIframeState = useStore($canvasIframeState);
```
istarkov, line 262: Why this 3 states are selected as something what have 1/3 of progress. Let's be simple leave `canvasIframeState`

istarkov: IMO this has so low value and so many code, that should be simplified.

istarkov (approved):

PR #3532, opened 2024-06-10
refactor: display show/hide for layers in the style-panel only on hover

JayaKrishnaNamburu:
## Description

Show full label for the `CSSValueListItem` and display `show/hide` icons only when users hover over the layer. Or when the users focuses the item with keys.

## Steps for reproduction
1. Create a new layer under any of the properties which has layers
2. add some values so that the labels are big enough to test it. 
3. Hover over the layer with mouse or focus it using keyboard.
4. The label should display the full name in normal and truncate the label and show the `show/hide` icons.

Screen Recording 2024-06-10 at 5 52 25 PM

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


kof (approved):
- [ ] when item is focused, using arrow left/right was moving focus to the buttons visible-invisible, now its not

packages/design-system/src/components/css-value-list-item.tsx
```diff
@@ -66,6 +68,13 @@ const ItemButton = styled("button", {
   position: "relative",
 
   "&:focus-visible, &[data-focused=true], &[data-state=open]": {
+    [`& ${FakeIconButtonsWrapper}`]: {
```
kof, line 71: you need focus-within

PR #3533, opened 2024-06-11
experimental: set collisionPadding for background panel to reduce panel jumps.

### LINKED ISSUE #3511: Backgrounds panel issues
- [ ] jump when switching between image and gradient https://share.descript.com/view/gRPo8AORr7N
- [x] when choosing an image after having a url, url is normally being reset, so you don't have both, now I managed to have both 
    image
- [ ] shouldn't all properties land in the Code area? not just the url()?
- [x] Code area label should be Code

JayaKrishnaNamburu:
fixes #3511 

## Description

This is an experiment to see if setting `collisionPadding` can reduce the background panel jumps when switching tabs. `Popover` content always opens close to trigger, but there is no defined way to know when it repositions when the content is dynamically changed.
So, when the panel is opened close to the bottom of the screen, the panel moves away from arrows under the cursor. 

## Steps for reproduction

- Add a background
- Now, open the background panel and switch between the tabs.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/backgrounds/backgrounds.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/backgrounds/backgrounds.tsx
@@ -88,6 +88,12 @@ const Layer = (props: {
   return (
     <FloatingPanel
       title="Background"
+      // Background Panel is big, and the size differs when the tabs are changed.
+      // This results in the panel moving around when the tabs are changed.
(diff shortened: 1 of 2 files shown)


kof (approved):
Still jumps in case item is at the very bottom, but its much better now, so I think we should at least go with this if there is no better option

https://share.descript.com/view/Yv94kYnCGKH

PR #3543, opened 2024-06-12
fix: Properties like animation-name were receiving unneeded px suffix

kof:
Closes https://github.com/webstudio-is/webstudio/issues/3529

## Description

Values that can have custom identifiers (user defined string) were tested to accept a unit value and they did, because anything is valid.

Now they are not tested with units any more

## Steps for reproduction

1. add animation-name in advanced panel
2. use any name it should stay as-is
3. previously we would add a "px"

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
+++ b/apps/builder/app/builder/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
@@ -3,9 +3,9 @@ import type {
   StyleValue,
   InvalidValue,
 } from "@webstudio-is/css-engine";
+import { units, properties, parseCssValue } from "@webstudio-is/css-data";
 import type { IntermediateStyleValue } from "./css-value-input";
 import { evaluateMath } from "./evaluate-math";
-import { units, parseCssValue } from "@webstudio-is/css-data";
 import { toKebabCase } from "../keyword-utils";
 
 const unitsList = Object.values(units).flat();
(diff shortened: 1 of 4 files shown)


apps/builder/app/builder/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
```diff
@@ -23,22 +23,32 @@ export const parseIntermediateOrInvalidValue = (
...
   }
 
-  // Try value with existing or fallback unit
-  const unit = "unit" in styleValue ? styleValue.unit ?? "px" : "px";
-  let styleInput = parseCssValue(property, `${value}${unit}`);
+  const valueInfo = properties[property as keyof typeof properties];
 
-  if (styleInput.type !== "invalid") {
-    return styleInput;
+  // - When user enters a number, we don't know if its a valid unit value,
```
TrySound, line 28: can we leverage unitGroups for this? Seem equivalent to types.

kof: they are different, they just have some overlap

TrySound: Isn't the point of this change to detect unit?

kof: No the point of this pr is to stop guessing the unit when value can be anything, because with custom identifiers we can't guess it, anything is valid

istarkov (approved):

PR #3563, opened 2024-06-14
fix: Unsubscribe loading states after its ready

kof:
## Description

Some of the states we are using for progress UI can be changed like selected instance, and we need to stop listening to those events.

## Steps for reproduction

1. unselect an instance by pressing esc or clicking outside of canvas
2. see that loading indicator didn't appear again

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/builder.tsx
```diff
@@ -432,8 +432,18 @@ export const Builder = ({
 
   const navigatorLayout = useNavigatorLayout();
   const dataLoadingState = useStore($dataLoadingState);
+  const [loadingState, setLoadingState] = useState($loadingState.get());
```
TrySound, line 435: better use useState(() => $loadingState.get()) to avoid one-time subscribe on each render

TrySound (approved):

PR #3566, opened 2024-06-14
fix: validate using long-hand property for transition while saving

### LINKED ISSUE #3069: Transition delay and  duration don't support keywords
emove keyword values for transition-delay and transition-duration. Keywords are valid if they are directly applied as individual properties. But when they are used under transition shorthand. The keyword values actually make the property invalid. For more reference (https://www.w3.org/TR/css-transitions-1/#propdef-transition-duration)

JayaKrishnaNamburu:
## Description

fixes #3069 
The properties `transition-delay` and `transition-duration` are being saved directly considering they are being valued using the `CSSValueItemProperty`. But not all values in long hand are valid in `short-hand` property. So, we need to validate the property value before saving them

## Steps for reproduction

- Add a transition layer.
- Now change the values in the transition-delay and transition-duration and try to save them.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging


kof (approved):
didn't test

PR #3578, opened 2024-06-17
refactor: upgrade mdn-data

TrySound:

Here upgraded to the latest mdn data and improved properties detection. Excluded experimental and undocumented css properties.

- font-synthesis got longhands in mdn data
- background-position-x/y got excluded as experimental (though support is good) but we anyway write as shorthand
- excluded white-space-collapse, text-wrap and text-wrap-* until decide how to maintain them in the future
- field-sizing and zoom are new additions


kof (approved):

packages/css-data/src/custom-data.ts
```diff
@@ -18,9 +18,16 @@ export const valueTypes = [
...
+  "dashed-ident",
   "single-animation-composition",
   "palette-identifier",
   "flex",
+  "inset-area",
+  "offset-path",
+  "coord-box",
+  "anchor-element",
+  "try-tactic",
+  "try-size",
```
TrySound, line 30: I'm not sure if there is any use in these types

kof: I think the question is rather: would it hurt to support as much as possible?

kof: We are saying we support full css on the site

TrySound: Those are not properties but css types

kof: values, properties, all part of css, right?

TrySound: I mean specifically this list is used only to detect custom idents

PR #3584, opened 2024-06-18
refactor: use longhand properties to save transition values in style-panel

### LINKED ISSUE #3567: Save only transition longhands in style panel
We rely on longhand style properties to generate atomic styles without collisions.
When save both shorthand and longhand it's not deterministic what overrides the other
especially in visual builder where you can't just move code.

So we need to switch "transitions" in style panel to longhand properties.
- accept both shorthand and longhands in transitions section
- always write longhands into data
- give longhands more priority

Longhands is either single common value or multiple values per animation name. When amount of values not matching animation name list it is considered invalid.

This way we gracefully migrate to longhands without huge migration.

JayaKrishnaNamburu:
## Description

fixes #3567 

Use long-hand properties to save `transition` values in the style-panel.

## Steps for reproduction
- Add some layers in transition panel in UI.
- Delete some layers
- Swap some layers

## TODO

- [ ] Allow `show/hide` for each layer.
- [ ] Add tests for all the layer handlers for transition.

## Code Review

  - test it on preview

@TrySound i made some refactor to use individual properties as long-hands instead of short-hand property. I am still yet to add tests for the new handlers, since they need to be handled in a different way now for transitions. But you can still play around with the current setup.

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/style-panel/sections/sections.ts
```diff
@@ -25,6 +25,7 @@ export const sections = new Map<
     Section: (props: SectionProps) => ReactNode;
   }
 >([
+  ["transitions", transitions],
```
JayaKrishnaNamburu, line 28: Will remove this change, once adding tests. Just added for dev time convenience.

JayaKrishnaNamburu:
@TrySound and @kof the PR is ready for review now. I kept the UX exact as before, but saving the layer result as individual transitions properties instead of short-hand like we are doing before.

Once this is merged, there are some improvements that need to be made like #3593 will be doing all those in smaller PR's.

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
```diff
@@ -63,3 +100,262 @@ export const findTimingFunctionFromValue = (
...
+    if (
+      isValidTransitionValue(property) === true &&
+      isValidTransitionValue(duration) === true &&
+      isValidTransitionValue(timingFunction) === true &&
+      isValidTransitionValue(delay) === true
+    ) {
+      const layer: TupleValue = {
+        type: "tuple",
+        value: [property, duration, timingFunction, delay],
+        hidden: property.hidden ? true : false,
```
TrySound, line 167: `property.hidden ?? false`

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
```diff
@@ -63,3 +100,262 @@ export const findTimingFunctionFromValue = (
...
+  };
+  const {
+    transitionProperty,
+    transitionDuration,
+    transitionDelay,
+    transitionTimingFunction,
+  } = properties;
+
+  for (const [index] of transitionProperty.value.entries()) {
+    const property = transitionProperty.value[index];
```
TrySound, line 153: There is an edge case. Some longhands can have only single value. And this value is used to each transition property.

packages/css-data/src/property-parsers/transition-property-extractor.ts
```diff
@@ -12,19 +11,19 @@ export const isTimingFunction = (timing: string) => {
...
 };
 
 export const extractTransitionProperties = (
   transition: TupleValue
 ): ExtractedTransitionProperties => {
-  let property: KeywordValue | null = null;
-  let timing: KeywordValue | null = null;
-  let duration: UnitValue | null = null;
-  let delay: UnitValue | null = null;
+  let property: KeywordValue | undefined = undefined;
```
TrySound, line 23: `= undefined` is not required when type explicitly specifies it

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
JayaKrishnaNamburu: Yes, but here is the case why i am maintaining all the defaults even if there is just one value.
Let's say user is adding these three layers
```
opacity 0ms ease 0ms
width
background 10ms ease-in 10ms 
```

If we saw them as individual properties.

```
transition-property: opacity, width, background
transition-duration: 0ms, 10ms
transition-timing-function: ease, ease-in
transition-delay: 0ms, 10ms
```

Once we read them back to display as individual layers. This will end up coming ass

```
opacity 0ms ease 0ms
width 10ms ease-in 10ms 
background 
```

So, just adding the default values even if there is just one property added by user

packages/css-data/src/property-parsers/transition-property-extractor.ts
JayaKrishnaNamburu: `ts` was complaining for me with `Used before delaration type error`. So just initialised with `undefined`

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
TrySound: See what I mean
```
transition-property: opacity, width, background;
transition-duration: 10ms;
```
Here transition should expand into `10ms, 10ms, 10ms`

JayaKrishnaNamburu: I refactored a little bit, but looks like it's going to be a real tricky one. When users add `transition-property` or any of the long hands using the advanced tab. They get added as `unparsed`. So, we need to handle all these individual cases to make sure they are accessible in the style-panel again.

Will ping once the updates are done 👍

JayaKrishnaNamburu:
@TrySound and @kof you can play around and test the PR now. 
- You can add `long-hand` properties individually. And they are still displayed as layers in the style-panel UI. 
- You can add layers directly in the style-panel, and they are added as long-hand properties.
- You can swap, show/hide and add/delete layers for each transition layer too

Now, users can add transitions from the `Transitions` section in style-panel. And using long-hand properties in the advanced panel.

I have a doubt now, users can still add `transition` as a property in the advanced tab. It gets parsed and added to the project. But should that be handled in style panel ?
I think it will overcomplicate things, like we can't support both `long-hand` and `short-hand` using the same UI panels and same layers. If we convert them to long hands, then it will overwrite the values in advanced tab.

IMO, we can just leave short-hand value of transition in the advanced tab itself and care for only long hands in the style-panel UI editor.

packages/css-data/src/property-parsers/transition.ts
```diff
@@ -1,20 +1,63 @@
...
+  "filter",
+  "flex",
+  "background-color",
+];
+
 export const isAnimatableProperty = (
   property: string
 ): property is AnimatableProperty => {
-  return animatableProperties.some((item) => item === property);
+  if (property === "all") {
```
JayaKrishnaNamburu, line 42: common properties already include all, no need for explicit check here.

TrySound (approved):
Working great!

kof (approved):
Tested, seems to work fine

JayaKrishnaNamburu:
Will wrap up client migration in a new PR and merge both at once, to keep it consistent and no delays between these two for existing projects 👍

PR #3588, opened 2024-06-18
feat: switch to overflow longhands

TrySound:

Here refactored overflow UI to update "overflow-x" and "overflow-y" longhands instead of "overflow" shorthand.

Removed "overflow" from advanced panel suggestions.

Added support for expanding overflow shorthand for webflow copy paste.

And implemented very basic migrations utility
which always run after data is loaded from server
and sends patch for expanding overflow longhands.
Patch is not sent when no properties are modified. Though eventually we will need to support versioning to avoid executing old migrations.

Screenshot 2024-06-18 at 19 33 21

### DIFF
--- a/apps/builder/app/shared/webstudio-data-migrator.ts
+++ b/apps/builder/app/shared/webstudio-data-migrator.ts
@@ -1,0 +1,43 @@
+import { camelCase } from "change-case";
+import {
+  getStyleDeclKey,
+  type StyleDecl,
+  type WebstudioData,
+} from "@webstudio-is/sdk";
+import { toValue, type StyleProperty } from "@webstudio-is/css-engine";
+import { expandShorthands, parseCssValue } from "@webstudio-is/css-data";
+
+/**
+ *
+ * Transform loaded data and sync into server
+ * before passing into application
+ * should be idempotent because can be executed multiple times
+ *
+ * Very basic version of client migrations
+ * Should be versioned eventually to avoid running every time
+ *
+ * For now patch is prevented by excluding empty transactions from sync queue
+ *
+ */
+export const migrateWebstudioDataMutable = (data: WebstudioData) => {
+  for (const [styleDeclKey, styleDecl] of data.styles) {
+    const property = styleDecl.property as string;
+
+    // expand overflow shorthand into overflow-x and overflow-y longhands
+    if (property === "overflow") {
+      data.styles.delete(styleDeclKey);
+      const longhands = expandShorthands([
+        [property, toValue(styleDecl.value)],
+      ]);
+      for (const [hyphenedProperty, value] of longhands) {
+        const longhandProperty = camelCase(hyphenedProperty) as StyleProperty;
+        const longhandStyleDecl: StyleDecl = {
+          ...styleDecl,
+          property: longhandProperty,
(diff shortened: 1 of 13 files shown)


apps/builder/app/shared/webstudio-data-migrator.ts
```diff
@@ -0,0 +1,43 @@
...
+ * before passing into application
+ * should be idempotent because can be executed multiple times
+ *
+ * Very basic version of client migrations
+ * Should be versioned eventually to avoid running every time
+ *
+ * For now patch is prevented by excluding empty transactions from sync queue
+ *
+ */
+export const migrateWebstudioDataMutable = (data: WebstudioData) => {
```
kof, line 22: 1. migrations should be a separate folder with a file per migration type of structure imo
2. we should avoid having to run them all all the time, there should be some way to say that a mgiration isn't needed without iterating over data. Imaging having dozens or hundreds of them

kof: You would probably want to have a flag in data for each migration and once there is a flag, e.g. migrations.overflowShorthand === true then its not executing next time

TrySound: Yes, in the future we will have it all. For now the we iterate once over all styles and when nothing is mutated just do not send patch.

Whole migrations stuff would take time to implemented. Here's very fast solution. It does the job.

kof: I think adding flags is relatively easy to do and also better organizing migrations on a file system, its really just a flag

TrySound: just a flag will require migration in the future to support dozens
we also need to support sdk dependencies
so not just a flag

kof: Why not set the flag as part of json data? so it can be sent along with patches and no migration needed?

kof: You are basically making this migration functionality for everyone to use, so it should be good enough, otherwise its just yet another hack added to the system

kof: I wouldn't feel comfortable using it knowing it can lead to execution every time. I would also like to have a nicer interface for migrations

kof: Some things aren't meant to be incrementally added iwh some naive version first

kof: I understand we are under pressure with the release, but that can be postponed if anything

kof: I don't know if this is part of our requirement for the current release

TrySound: This flag you want to add will have to be set in data forever. And then we will bring another solution for migrations which will also be set in data forever. So we will have to support both solutions instead of just one which will emerge from what I implemented.

kof (approved):
approving because you keep insisting ... would rather not have this at all than another half -backed thing

istarkov (approved):

PR #3607, opened 2024-06-23
refactor: drop parseTransition shorthand and add migration for transition shorthand

### LINKED ISSUE #3593: Use `expandShorthands` and move away from `parseTransitions` and `extractTransitions`
After the latest #3584 is merged. We should replace using `parseTransitions` and `extractTransitions` to one step of `expandShorthands` which can take care of this in a single step.

JayaKrishnaNamburu:
fixes #3593 
## Description

This PR
- drops the `parseTransition` for parsing shorthand transition property.
- allows migration of existing transition property to long-hands.
- drops parsing `transition` when added from advanced tab. Because if we parse and add it as proper layers. It will get remigrated again every time the app is loaded.
- Adds tests for parsing individual transition properties and even parses timing functions as timing functions now.

- [x] there is a bug associated with expanding transitions from shorthands #3608 . Once fixed we need to test the new `FunctionValue` parsing behaviour.

## Steps for reproduction

- Any previous values of `transitions` should we migrated into new approach and should be available in the style-panel for using them.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/shared/webstudio-data-migrator.ts
```diff
@@ -24,7 +24,7 @@ export const migrateWebstudioDataMutable = (data: WebstudioData) => {
     const property = styleDecl.property as string;
 
     // expand overflow shorthand into overflow-x and overflow-y longhands
-    if (property === "overflow") {
+    if (property === "overflow" || property === "transition") {
```
JayaKrishnaNamburu, line 27: I just have one doubt in this process of data migration.  This runs everytime when the app is loaded. So, instead of just the old data structure. This will migrate all the `unparsed` values too in the app if the users add them now. 
PS: The values for `transition` becomes unparsed now when added from advanced tab. Should we add a check to migrate only if the old value is `layers` type for transitions proeprty.

TrySound: This will migrate only if these property names are appeared. No patches sent if these properties are already longhands.

TrySound: Ah, right. We need to remove transition from mdn data

JayaKrishnaNamburu: Done, dropped transition shorthand from property aggregator too 👍

TrySound (approved):

PR #3609, opened 2024-06-23
feat: Add native select control

istarkov:
## New

- Native Select

## Changes

- Improved detection algorithm to include invisible elements (like `option` or `display:none`).
- Added interceptor to handle pointer down events, preventing select from opening.
- Enhanced element selection on canvas to support `display:none` and option selection.
- Fixed prop types generation (previously, the value property was broken).
- Memory props has effect on canvas only

## Steps for reproduction

Add `Select` see it works

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/components.ts
+++ b/packages/sdk-components-react/src/components.ts
@@ -32,3 +32,5 @@ export { VimeoPlayButton } from "./vimeo-play-button";
 export { VimeoSpinner } from "./vimeo-spinner";
 export { XmlNode } from "./xml-node";
 export { Time } from "./time";
+export { Select } from "./select";
+export { Option } from "./option";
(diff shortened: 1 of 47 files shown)


packages/sdk-components-react/src/components.ts
```diff
@@ -32,3 +32,5 @@ export { VimeoPlayButton } from "./vimeo-play-button";
 export { VimeoSpinner } from "./vimeo-spinner";
 export { XmlNode } from "./xml-node";
 export { Time } from "./time";
+export { Select } from "./select";
+export { Option } from "./option";
```
TrySound, line 36: Do we need them separate? Or could be grouped like radix components?

istarkov: Props are not generated in a single file

kof:
- [ ] Prop value needs to be in initial props after name
- [ ] Prop multiple needs to be in initialProps
- [ ] Style preset on option don't match rendering when prop multiple is used, e.g. background color is transparent in style panel but is dark gray on canvas
- [ ] when using json value array with prop multiple, validation fails, because it thinks value can only be a string, but it can take an array too

packages/sdk-components-react/src/components.ts
TrySound: How so?
image

PR #3611, opened 2024-06-23
experimental: correctly match unordered shorthands values

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3608


JayaKrishnaNamburu (approved):
tested and works good now 👍 💯

PR #3640, opened 2024-06-28
feat: support -webkit-line-clamp

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/2647 Closes https://github.com/webstudio-is/webstudio/issues/3427

Here added -webkit-line-clamp and -webkit-box-orient properties support which is the only way to add multiline truncation at the moment.

All properties with prefixes are stored without camel casing. With font smoothing as the only exception because added very early.

Screenshot 2024-06-28 at 17 48 08

### DIFF
--- a/packages/css-data/bin/mdn-data.ts
+++ b/packages/css-data/bin/mdn-data.ts
@@ -17,6 +17,18 @@ import type {
 } from "@webstudio-is/css-engine";
 import * as customData from "../src/custom-data";
 
+/**
+ * Store prefixed properties without change
+ * and convert to camel case only unprefixed properties
+ * @todo stop converting to camel case and use hyphenated format
+ */
+const normalizePropertyName = (property: string) => {
+  if (property.startsWith("-")) {
+    return property;
+  }
+  return camelCase(property);
+};
+
 const units: Record<customData.UnitGroup, Array<string>> = {
   number: [],
   // consider % as unit
@@ -228,15 +240,11 @@ type FilteredProperties = { [property in Property]: Value };
 const experimentalProperties = [
   "appearance",
   "aspect-ratio",
-  // not standard and not implemented without prefix
-  // @todo get rid once the issue with types in radix sdk is resolved
-  "line-clamp",
-  // used in normalize
   "text-size-adjust",
+  "-webkit-line-clamp",
 ];
 
 const unsupportedProperties = [
-  "-webkit-line-clamp",
   "--*",
   // shorthand properties
   "all",
@@ -278,7 +286,9 @@ const filteredProperties: FilteredProperties = (() => {
   return result;
 })();
 
-const propertiesData = { ...customData.propertiesData };
+const propertiesData = {
+  ...customData.propertiesData,
+};
 
 let property: Property;
 for (property in filteredProperties) {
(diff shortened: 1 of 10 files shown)


packages/css-data/bin/mdn-data.ts
```diff
@@ -17,6 +17,18 @@ import type {
 } from "@webstudio-is/css-engine";
 import * as customData from "../src/custom-data";
 
+/**
+ * Store prefixed properties without change
+ * and convert to camel case only unprefixed properties
+ * @todo stop converting to camel case and use hyphenated format
+ */
+const normalizePropertyName = (property: string) => {
+  if (property.startsWith("-")) {
```
kof, line 26: hmm, so you don't camel-case the name if it starts with dash? is it for vendor prefixes or is it for variables? I am guessing for vendors, but then you will return lower case name and in other cases camel case ... I don't get the logic

packages/css-data/src/parse-css.ts
```diff
@@ -12,6 +12,25 @@ import { camelCase } from "change-case";
...
+ */
+export const normalizePropertyName = (property: string) => {
+  // these are manually added with pascal case
+  if (property === "-webkit-font-smoothing") {
+    return "WebkitFontSmoothing";
+  }
+  if (property === "-moz-osx-font-smoothing") {
+    return "MozOsxFontSmoothing";
+  }
+  if (property.startsWith("-")) {
```
kof, line 28: same question

packages/css-data/bin/mdn-data.ts
TrySound: Both custom properties and prefixed properties are not camel cased. Ideally convert all other properties to hyphenated as well.

packages/css-data/src/parse-css.ts
```diff
@@ -31,22 +50,14 @@ const unprefixProperty = (property: string) => {
   return property.replace(prefixRegex, "");
 };
 
-const camelCaseProperty = (property: string) => {
-  let camelcased = camelCase(property);
-  if (property[0] === "-") {
-    camelcased = camelcased[0].toLocaleUpperCase() + camelcased.slice(1);
```
kof, line 37: this used to turn
-webkit-something into WebkitSomething

TrySound: Now it doesn't

kof: well unless there was a bug, at least this was the intention

packages/css-data/bin/mdn-data.ts
kof: but then this function returns camel cased for everything else and hyphenated for variables and vendor prefixed once? this seems inconsistent and werid

TrySound: We wanted to move away from camel cased anyway. Why add more?

packages/css-data/src/parse-css.ts
TrySound: Well, prefixed properties parsed incorrectly with it 🤷

packages/css-data/bin/mdn-data.ts
kof: by creating inconsistency in the meantime it will create more confusion, lets do the switch at once

TrySound: Custom property is inconsistent and should not be camel cased

kof: actually you are right regarding custom property, it could be inconsistently named by the user, so it can't be reliably converted to camel case and then restored back to hyphenated

kof (approved):

PR #3648, opened 2024-06-29
experimental: parse non-standard properties

TrySound:

Fixed parsing non-standard properties with prefixes

- font-smoothing
- -webkit-font-smoothing
- -moz-osx-font-smoothing
- tap-highlight-color
- -webkit-tap-highlight-color
- -webkit-overflow-scrolling


kof (approved):
🔥

PR #3656, opened 2024-07-01
experimental: expand shorthands with css-wide keywords

TrySound:

According to spec css-wide keywords on shorthands should be passed to all their longhands as is.

### DIFF
--- a/packages/css-data/src/shorthands.test.ts
+++ b/packages/css-data/src/shorthands.test.ts
@@ -35,6 +35,23 @@ test("expand border", () => {
   ]);
 });
 
+test("expand border with css-wide keywords", () => {
+  expect(expandShorthands([["border", "INHERIT"]])).toEqual([
+    ["border-top-width", "INHERIT"],
+    ["border-right-width", "INHERIT"],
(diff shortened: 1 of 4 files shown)


packages/css-data/src/shorthands.test.ts
```diff
@@ -35,6 +35,23 @@ test("expand border", () => {
   ]);
 });
 
+test("expand border with css-wide keywords", () => {
+  expect(expandShorthands([["border", "INHERIT"]])).toEqual([
+    ["border-top-width", "INHERIT"],
```
kof, line 40: why is this one upper case?

kof (approved):

packages/css-data/src/shorthands.test.ts
TrySound: css is case-insensitive in most cases. Just a check it handles it.

kof: would be better to cheeck as a separate test, I wouldn't have guessed the test is about case sensitivity

PR #3661, opened 2024-07-02
build: run all tests and typechecks in CI

TrySound:
Here improved our checks. No longet rely on "checks" script but run "test" and "typescript" scripts from each package in monorepo.

Would also make each package less boilerplaty.


packages/cli/package.json
```diff
@@ -23,9 +23,7 @@
   ],
   "scripts": {
     "typecheck": "tsc",
-    "checks": "pnpm typecheck",
     "build": "rm -rf lib && esbuild src/cli.ts --outdir=lib --bundle --format=esm --packages=external",
-    "dev": "esbuild src/cli.ts --watch --bundle --format=esm --packages=external --outdir=./lib",
```
kof, line 28: sure we don't need this?

kof (approved):

PR #3676, opened 2024-07-05
fix: Fix transition

istarkov:
Fixes https://github.com/webstudio-is/webstudio/issues/3675

## Description

Already on production

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/style-layer-utils.ts
+++ b/apps/builder/app/builder/features/style-panel/style-layer-utils.ts
@@ -199,9 +199,11 @@ export const getHumanizedTextFromLayer = (
     case "transitionProperty":
       if (layer.type === "tuple") {
         const properties = [...layer.value];
-        const transitionProperty = properties.find(
+
+        let transitionProperty = properties.find(
           (item): item is KeywordValue =>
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/style-panel/style-layer-utils.ts
```diff
@@ -211,7 +213,10 @@ export const getHumanizedTextFromLayer = (
         );
 
         if (transitionProperty === undefined) {
-          throw `Transition property is missing from the layer ${JSON.stringify(layer)}`;
+          transitionProperty = {
+            type: "keyword",
+            value: "unknown",
+          };
```
JayaKrishnaNamburu, line 219: @istarkov even if the values are added from advanced tab. It is parsing them, so the only other thing i can think og fall under this is for older projects. But for older projects there was a migration that is ran when a project is loaded. Any specific case where it is failing ?

istarkov: Ruined prod today ;) Just add new transition

TrySound (approved):

Issue #3411, opened 2024-05-23
CSS Transforms


kof:
As a user I want to use CSS transforms

https://www.figma.com/design/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=10712-4063&m=dev

JayaKrishnaNamburu:
I have some questions w.r.t to each section in the transform section. The UI is 100% achievable if only there is one-way direction of adding styles in the application. But in the whole application, there is always a space where users can add values using the `code-editor` text-area. And the values are parsed back into the editor and populated in style-panel. Let's see on by one.

## Parsing the property value
`transform` is a very similar to filter which means all the possible functions are added to the same property without any `,` separated values like layers. So, first i thought we can just do exactly what we are doing in filters. But then after checking the UI, i realised we need to parse the value and group the functions into individual tuples again. We need to build a hierarchy on top of each filter function so we can show them as independnet layers. 

Eg:
The typical way of doing this, just parsing everything and converting into data layer functions.
```js
const transition: TupleValue = {
  type: "tuple",
  hidden: false,
  value: [
    {
      type: "function",
      name: "translate",
      args: {
        type: "layers",
        value: [{ type: "unit", unit: "px", value: 50 }],
      },
    },
    {
      type: "function",
      name: "rotate",
      args: {
        type: "layers",
        value: [{ type: "unit", unit: "deg", value: 50 }],
      },
    },
    {
      type: "function",
      name: "scale",
      args: {
        type: "layers",
        value: [
          { type: "keyword", value: "2" },
          { type: "keyword", value: "5" },
        ],
      },
    },
    {
      type: "function",
      name: "translateX",
      args: {
        type: "layers",
        value: [{ type: "unit", unit: "px", value: 50 }],
      },
    },
  ],
};
```
But this way, each function goes as an individual layer in the current layer list implementation. Like
```
- translate(50px)
- rotate(50deg)
- scale(2, 5)
- translateX(50px)
```

But the parser should parse them into

```js
const transition: TupleValue = {
  type: "tuple",
  hidden: false,
  value: [
    {
      type: "tuple",
      value: [
        {
          type: "function",
          name: "translate",
          args: {
            type: "layers",
            value: [{ type: "unit", unit: "px", value: 50 }],
          },
        },
        {
          type: "function",
          name: "translateX",
          args: {
            type: "layers",
            value: [{ type: "unit", unit: "px", value: 50 }],
[...]

PR #3696, opened 2024-07-08
experimental: parse `translate` css property into tuples

JayaKrishnaNamburu:
## Description

Parses `translate` css property and converts them into tuples. Related to the series of PR that enables transform property #3411 

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


packages/css-data/src/property-parsers/translate.test.ts
```diff
@@ -0,0 +1,81 @@
+import { test, expect } from "@jest/globals";
+import { parseTranslate } from "./translate";
+
+test("parses a valid translate value", () => {
```
TrySound, line 4: I would just write all tests in parse-css-value.test.ts
Btw I'm trying to avoid snapshots for json structures to make formatting more compact.

packages/css-data/src/property-parsers/translate.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (parsed.error) {
+    return {
+      type: "invalid",
+      value: input,
+    };
+  }
+
+  const translateValues: TupleValueItem[] = [];
+
+  try {
```
TrySound, line 42: is walking can result in runtime error?

packages/css-data/src/property-parsers/translate.test.ts
JayaKrishnaNamburu: sure will update it 👍

packages/css-data/src/property-parsers/translate.ts
JayaKrishnaNamburu: Nope, just comes in handy if we want to throw an error or semething while parsing. Just following from `filters`. Will remove, not actually needed here.

TrySound (approved):

PR #3702, opened 2024-07-09
fix: expandShorthands fails and prevent builder to load

istarkov:
## Description

Cherry-picked here https://github.com/webstudio-is/webstudio/pull/3674

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-data/src/shorthands.test.ts
```diff
@@ -1448,3 +1448,7 @@ test("expand position-try", () => {
     ["position-try-options", "--dashed-ident"],
   ]);
 });
+
+test("Does not fail on empty transition", () => {
+  expect(expandShorthands([["transition", ""]])).toEqual([]);
```
TrySound, line 1453: weird one

TrySound (approved):

packages/css-data/src/shorthands.test.ts
kof: I think any value can be empty, we need to test it with empty string for all properties

packages/css-data/src/shorthands.test.ts
```diff
@@ -1448,3 +1448,77 @@ test("expand position-try", () => {
     ["position-try-options", "--dashed-ident"],
   ]);
 });
+
+test("Does not fail on empty value", () => {
```
kof, line 1452: is there any way we could get a full list of all properties  as an array and get a loop over it, then add a bunch of values beyond empty string to get generated tests for all of them?

kof: I don't think its feasible to maintan a list of properties here

PR #3705, opened 2024-07-10
experimental: parse `scale` css property into tuples

JayaKrishnaNamburu:
## Description

Adds support for parsing `scale` css property. Will be used in bringing support for transforms ection in style panel.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ x added tests


packages/css-data/src/parse-css-value.test.ts
```diff
@@ -374,3 +374,57 @@ test("parse unknown properties as unparsed", () => {
     }
   );
 });
+
+test("parses a valid translate value", () => {
+  expect(parseCssValue("scale", "1.5")).toEqual({
+    type: "tuple",
+    value: [
+      {
+        type: "keyword",
```
TrySound, line 383: why do you use keyword everywhere for numbers? We have unit: 'number' for such values.

TrySound: `none` is the only keyword here

JayaKrishnaNamburu: Done, i remember seeing it somewhere and just continued it everytime. Will use `number` from now on 👍

TrySound (approved):

PR #3712, opened 2024-07-11
refactor: add universal css fragment parser

TrySound:
Here added new utility for builder to support parsing css fragments in style controls like background, transition, shadows etc.

The idea is to use regular parseCss where possible and fallback to specified property parsing.

This way we do not need to strip property name manually and can always have shorthands expanded in result.

Also user can paste multiple properties at once like background-color and background-image when describe gradient.

Screenshot 2024-07-11 at 16 38 06


JayaKrishnaNamburu (approved):
Looks good 👍

PR #3717, opened 2024-07-12
refactor: add repeated style control

TrySound:
Our repeated sections like background or transitions are too repetetive and require logic to store multiple properties as single style value.

Here added new control which covers more use cases out of the box and allows to avoid repetition for shorthand properties.

Switched only transitions for now.


apps/builder/app/builder/features/style-panel/sections/transitions/transitions.tsx
```diff
@@ -14,44 +13,66 @@ import { CollapsibleSectionRoot } from "~/builder/shared/collapsible-section";
...
+}: {
+  style: StyleInfo;
+  index: number;
+}) => {
+  // show label without hidden replacement
+  const propertyLayer = getLayer(style.transitionProperty?.value, index);
+  const property = humanizeString(
+    toValue(propertyLayer ? { ...propertyLayer, hidden: false } : undefined)
+  );
+  const duration = toValue(getLayer(style.transitionDuration?.value, index));
```
JayaKrishnaNamburu, line 52: Since, it's going to be `getLayer` approach using index now. And i see the transition extractor is not being used anymore I think we can delete the extractor too in this PR with it's tests.

And use the same approach here too
https://github.com/webstudio-is/webstudio/blob/main/apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts#L291

JayaKrishnaNamburu (approved):
Looks good, except one more refactor at adding default transition step 👍 🔥

apps/builder/app/builder/features/style-panel/sections/transitions/transitions.tsx
TrySound: Still used in transition popover. Would require more refactoring. Here focused on the list specifically.

PR #3721, opened 2024-07-14
refactor: split css literal parser

TrySound:
Here refactored value parser with literal parser
which can be reused in all functions, tuples and layers.

Color parser now relies on specific ast types.

Tuples are generated for all values with multiple nodes.

This all will let us build more universal parsers with less effort.


kof:
This a dangerous one, needs to be well tested

TrySound:
Plenty of tests seem work as expected

kof (approved):

PR #3723, opened 2024-07-14
refactor: move shadow parser into value parser

TrySound:
We don't need separate parsing logic for box-shadow and text-shadow properties. Generic css value parser is able to produce the same ast.


JayaKrishnaNamburu (approved):
Looks good, and working as expected too 👍

PR #3726, opened 2024-07-15
refactor: move transform parser into css value parser

TrySound:
Here dropped custom transform parsers in favor of generic value parser. Also switched to layers when single value is specified in scale and translate.


packages/css-data/src/parse-css-value.test.ts
```diff
@@ -385,23 +385,23 @@ test("parse transform property as tuple", () => {
         type: "function",
         name: "rotateX",
         args: {
-          type: "tuple",
+          type: "layers",
```
JayaKrishnaNamburu, line 388: Looks pretty good. My only doubt here is.
For some functions the values are separated using `,` and some are not. Does the new way of parsing takes care of those too.

Eg: `drop-shadow` function doesn't take `,` separated arguments.  But some other transform functions take `,` in their arguments.

In this specific case it is find, since `rotateX`, `rotateY` and `roateZ` takes only one argument. So, it doesn't really effect how the arguments are separated. Can we handle those complex cases too

JayaKrishnaNamburu (approved):

packages/css-data/src/parse-css-value.test.ts
TrySound: It does not handle drop-shadow etc. I think we would have a separate branch for them.
All listed functions have comma separated arguments. As for rotateXYZ, tuple doesn't make sense here either. Layers just makes it consistent with all other functions in transform.

JayaKrishnaNamburu: Yeah got that, was just curious how to handle those kind of cases by default.

PR #3729, opened 2024-07-15
refactor: move filter parsers to generic value parser

TrySound:
Here reuse generic css value parser for filter and backdrop-filter properties. Added section to parse filter functions with tuple arguments.


JayaKrishnaNamburu (approved):
Looks good 👍

PR #3730, opened 2024-07-15
fix: transition bugs

TrySound:
- fixed default duration
- fixed repeating logic when view, add or delete transitions
- prevent showing transition-behavior property when use transitions UI


kof (approved):
didn't test

PR #3732, opened 2024-07-15
build: forbid unused parameters

TrySound:
Enabled flag in tsconfig to forbid unused parameters.


kof (approved):
huh, this used to work some time in the past

istarkov (approved):

PR #3735, opened 2024-07-15
experimental: fix local styles when copy from webflow

TrySound:
Fixed the issue with pasting components with local styles for exmaple quick stack.


kof (approved):

kof:
fixes https://github.com/webstudio-is/webstudio/issues/3734

PR #3753, opened 2024-07-17
refactor: isolate remix specifics in cli

TrySound:
We wanna support other frameworks than remix.
At least to support SSG we need to additionally deploy vike.dev.

Here split all remix specifics into separate module. It provides the list of components (including overrides) and templates getter.

Templates allow to provide multiple module per route which is essential for many frameworks like svelte and vike which do not rely on treeshaking for client/server splitting.


kof (approved):
great start

PR #3758, opened 2024-07-19
experimental: Enhance Performance by Using Postgrest to Fetch Data from Build Table

istarkov:
## Description

[Big Project Example](https://postgrest.development.webstudio.is/builder/be939bf2-9095-4e17-b175-69de2bbf9597)

Reduces time of big projects loading:

   ✓ { scenario:new_o }...........: avg=2.58s    min=1.99s  med=2.54s    max=3.53s    p(90)=2.95s    p(95)=3.15s    p(99)=3.44s    count=31
     ✓ { scenario:old_o }...........: avg=5.5s     min=4.55s  med=5.12s    max=15.58s   p(90)=5.44s    p(95)=6.02s    p(99)=12.87s   count=31

2 times faster. (but highly depends on test times, project size etc sometimes it's 6x difference)

Reduces patch time too.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof (approved):

pnpm-lock.yaml
```diff
@@ -1572,6 +1575,28 @@ importers:
         specifier: 5.5.2
         version: 5.5.2
 
+  packages/postgrest:
```
TrySound, line 1578: maybe call it db?

packages/postgrest/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "generate-types": "pnpx supabase gen types --lang=typescript --db-url postgresql://postgres:pass@localhost/webstudio > ./src/__generated__/db-types.ts"
+  },
+  "dependencies": {
+    "@supabase/postgrest-js": "^1.15.8"
+  },
+  "devDependencies": {
+    "@jest/globals": "^29.7.0",
+    "@webstudio-is/jest-config": "workspace:*",
+    "@webstudio-is/tsconfig": "workspace:*",
+    "type-fest": "^4.3.1",
```
TrySound, line 28: are jest and type-fest used?

istarkov: type-fest highly probably will be used based on their docs.
jest probably

pnpm-lock.yaml
istarkov: If we will start move onto then yes. Not a pig deal as it's a   "private": true package

packages/postgrest/package.json
istarkov: Will remove jest

PR #3765, opened 2024-07-19
refactor: merge exclude from search flags

TrySound:
Write environment flag to page meta in load instead of exposing to client.


istarkov (approved):

istarkov:
test on staging plz

PR #3794, opened 2024-07-25
feat: use classes for preset styles

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/3787

Replaced [data-ws-component=...] selector with classes in generated styles. This will unlock cleaner html output in the future.

### DIFF
--- a/apps/builder/app/canvas/shared/styles.ts
+++ b/apps/builder/app/canvas/shared/styles.ts
@@ -11,7 +11,6 @@ import {
   collapsedAttribute,
   idAttribute,
   addGlobalRules,
-  addPresetRules,
   createImageValueTransformer,
   descendantComponent,
   type Params,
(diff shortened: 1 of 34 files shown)


packages/react-sdk/src/css/css.ts
```diff
@@ -69,25 +72,31 @@ export const generateCss = ({
...
-  for (const prop of props.values()) {
-    if (prop.name === "selector" && prop.type === "string") {
-      descendantSelectorByInstanceId.set(prop.instanceId, prop.value);
+  addGlobalRules(globalSheet, { assets, assetBaseUrl });
+  globalSheet.addMediaRule("presets");
+  const presetClasses = new Map<Instance["component"], string>();
+  const scope = createScope([], (name) => name);
+  for (const [component, meta] of componentMetas) {
+    const [_namespace, componentName] = parseComponentName(component);
+    const className = `ws-p-${scope.getName(component, kebabCase(componentName))}`;
```
kof, line 82: do we really need the "-p" or could it be just ws-name?

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -324,7 +323,22 @@ export const GlobalStyles = ({ params }: { params: Params }) => {
...
-    addPresetRules(presetSheet, metas);
+    presetSheet.addMediaRule("presets");
+    for (const [component, meta] of metas) {
+      for (const [tag, styles] of Object.entries(meta.presetStyle ?? {})) {
+        const rule = presetSheet.addNestingRule(
+          `${tag}:where([data-ws-component="${component}"])`
+        );
+        for (const declaration of styles) {
+          rule.setDeclaration({
+            breakpoint: "presets",
```
kof, line 334: weird to see breakpoint name presets, since they are usually  different

kof (approved):

packages/react-sdk/src/css/css.ts
TrySound: Can conflict with instance styles. I am planning to use `ws-i-*`. Alternatively would be deduplicated like this `ws-link ws-link_1` where first is preset and the second is local.

apps/builder/app/canvas/shared/styles.ts
TrySound: It's just static identifier to place all styles within single media rule.

PR #3795, opened 2024-07-26
refactor: disable atomic styles in generate stories

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3788

Here switched back to regular (non-atomic) styles
in generated stories in radix sdk.

### DIFF
--- a/packages/sdk-components-react-radix/src/__generated__/accordion.stories.tsx
+++ b/packages/sdk-components-react-radix/src/__generated__/accordion.stories.tsx
@@ -1,4 +1,3 @@
-import { useState } from "react";
 import {
   Box as Box,
   Text as Text,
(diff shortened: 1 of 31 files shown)


packages/sdk-components-react-radix/src/__generated__/accordion.stories.tsx
```diff
@@ -187,105 +169,106 @@ html {margin: 0; display: grid; min-height: 100%}
...
+  :where(address.w-box) {
     box-sizing: border-box;
     border-top-width: 1px;
     border-right-width: 1px;
     border-bottom-width: 1px;
     border-left-width: 1px;
     outline-width: 1px
   }
-  article:where([data-ws-component="Box"]) {
+  :where(article.w-box) {
```
kof, line 191: do you think it makes sense to just have a simple .w-box here?

kof (approved):

packages/sdk-components-react-radix/src/__generated__/accordion.stories.tsx
TrySound: Nope. We use :where to reset specificity and let stateless styles override hover. Same like we do with a:visited {} styles.

TrySound: Same effect would give be using `@layer preset {}`

kof: right, forgot about it

kof: ok but why do we need the tag selector as well?

PR #3805, opened 2024-07-27
refactor: migrate asset uploader to postgrest

TrySound:
Got some experience with using postgrest api.
Tested locally.

### DIFF
--- a/packages/asset-uploader/src/db/load.ts
+++ b/packages/asset-uploader/src/db/load.ts
@@ -26,6 +26,8 @@ export const loadAssetsByProject = async (
 
   const assets = await context.postgrest.client
     .from("Asset")
+    // use inner to filter out assets without file
+    // when file is not uploaded
     .select(
       `
         assetId:id,
--- a/packages/asset-uploader/src/db/load.ts
+++ b/packages/asset-uploader/src/db/load.ts
@@ -1,4 +1,3 @@
-import { prisma, type Project } from "@webstudio-is/prisma-client";
 import {
   authorizeProject,
   type AppContext,
@@ -8,7 +7,7 @@ import type { Asset } from "@webstudio-is/sdk";
 import { formatAsset } from "../utils/format-asset";
 
 export const loadAssetsByProject = async (
-  projectId: Project["id"],
+  projectId: string,
   context: AppContext,
   { skipPermissionsCheck = false }: { skipPermissionsCheck?: boolean } = {}
 ): Promise<Asset[]> => {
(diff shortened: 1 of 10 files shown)


packages/asset-uploader/src/db/load.ts
```diff
@@ -25,26 +24,27 @@ export const loadAssetsByProject = async (
...
-      file: { createdAt: "desc" },
-    },
-  });
+  const assets = await context.postgrest.client
+    .from("Asset")
+    .select(
+      `
+        assetId:id,
+        projectId,
+        file:File!inner (*)
```
istarkov, line 33: plz comment why inner is used and in what case it will give the wrong result

packages/asset-uploader/src/db/load.ts
```diff
@@ -25,26 +24,27 @@ export const loadAssetsByProject = async (
...
-  );
+  const result: Asset[] = [];
+  for (const { assetId, projectId, file } of assets.data ?? []) {
+    if (file) {
+      result.push(formatAsset({ assetId, projectId, file }));
+    }
+  }
+
+  // order by createdAt desc
+  result.sort((left, right) => right.createdAt.localeCompare(left.createdAt));
```
istarkov, line 47: What the need of sorting? If stability then `id` must be used too

packages/asset-uploader/src/upload.ts
```diff
@@ -39,23 +38,24 @@ export const createUploadName = async (
...
+  const uploadedCount = await context.postgrest.client
+    .from("File")
+    .select("*", { count: "exact", head: true })
+    .eq("isDeleted", false)
+    .eq("uploaderProjectId", projectId)
+    .eq("status", "UPLOADED");
+
+  const uploadingCount = await context.postgrest.client
+    .from("File")
+    .select("*", { count: "exact", head: true })
```
istarkov, line 50: what means head: true? does it needed?

TrySound: > When set to `true`, `data` will not be returned. Useful if you only need the count.

packages/asset-uploader/src/db/load.ts
TrySound: Looks like we do this on client anyway, maybe not needed

packages/asset-uploader/src/upload.ts
istarkov: `.select("*", { count: "exact", head: true })` * looks suspicious I have a feeling that it generates something bad on sql side

istarkov: Better to see generated sql in postgres logs

istarkov: also where  .eq("uploaderProjectId", projectId)

TrySound: Looks like without head: true it loads data along with count

istarkov (approved):

PR #3806, opened 2024-07-27
feat: drop data-ws-* attributes

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3763

We no longer rely on data-ws-id and data-ws-component in generated apps. Now we get a few benefits

- smaller workers to deploy
- cleaner html to inspect
- generated ssg can be modified directly by user

See generated fixtures

image


kof (approved):
🔥

PR #3812, opened 2024-07-29
experimental: add scaling `x` and `y` inputs to be changed together using locked behaviour

JayaKrishnaNamburu:
## Description

Adds `lock` behaviour to change `scaleX` and `scaleY` to change values at once. 

Screenshot 2024-07-29 at 11 53 16 AM

Screenshot 2024-07-29 at 1 37 27 PM

## Steps for reproduction

- Set a `scaleX` value
- the value `scaleY` also changes at the same time.
- You can press the lock and change it to `unclock`.
- Now you can change the `scaleX` and `scaleY` independently.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


kof:
- [ ]  wouldn't it be more practical to use % unit for scale? dealing with 0.1 steps instead of whole numbers seems much less intuitive, but there should be a unit selector anyways

JayaKrishnaNamburu:
@kof updated the icon to use lock and the default values to be `%` now for scale. The bug with `transition-delay` and `transition-duration` might not be a direct ones. Looking into it and fixing if needed in a new PR 👍

kof:
What's the "none"?

image

kof:
Why is % gone after switching to unitless?

image

kof:
After I switched to unitless, I tried to type a number and a "%" and got NaN

image

JayaKrishnaNamburu:
> Why is % gone after switching to unitless?
> 
> image

It ended up being a upstream bug for this one. Where `mdn` is sending only `number` as value for `unitGroups`. I am going to switch this to use a proxy proeprty to avoid this. They don't seem to merge it anytime soo. So, will drop a reference for it too in the code.

JayaKrishnaNamburu:
Ok, i replaced `scale` to use a proxy value for validation now. It shouldn't show `none` anymore in the keywords list. And should work with all cases like scrub, key strokes, onChange.

kof:
- [ ] typing in scale input results in immediate rerender, where if you compare to width input, typing doesn't rerender until you hit enter or unfocus

kof:
- [ ] Link and unlink tooltip text always says the same, if you compare to flex, it says Link or Unlink depending on the state

JayaKrishnaNamburu:
> * [ ]  Link and unlink tooltip text always says the same, if you compare to flex, it says Link or Unlink depending on the state

link-unlink-tooltip

Can you give it a try again please.

kof (approved):

PR #3824, opened 2024-07-30
fix: clone projects with preview images properly

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/3823

This fixes the issue with preview images set before assets are copied to the new project.

### DIFF
--- a/packages/prisma-client/prisma/migrations/20240730131207_clone_project_preview_imagea/migration.sql
+++ b/packages/prisma-client/prisma/migrations/20240730131207_clone_project_preview_imagea/migration.sql
@@ -1,0 +1,71 @@
+DROP FUNCTION IF EXISTS clone_project;
+CREATE FUNCTION clone_project(
+  project_id text,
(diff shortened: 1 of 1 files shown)


packages/prisma-client/prisma/migrations/20240730131207_clone_project_preview_imagea/migration.sql
```diff
@@ -0,0 +1,71 @@
...
+  FROM "Asset" AS asset, "File" AS file
+  WHERE
+    asset.name = file.name AND
+    file.status = 'UPLOADED' AND
+    asset."projectId" = old_project.id;
+
+  -- set preview image asset after copying assets to the project
+  UPDATE "Project"
+  SET "previewImageAssetId"= old_project."previewImageAssetId"
+  WHERE id=new_project.id;
```
TrySound, line 39: These are new

istarkov (approved):

PR #3878, opened 2024-08-05
refactor: switch authorization-token to postgrest

TrySound:
Upgraded postgrest client and migrated authorization-token package to it.

### DIFF
--- a/packages/authorization-token/src/db/authorization-token.ts
+++ b/packages/authorization-token/src/db/authorization-token.ts
@@ -1,16 +1,15 @@
+import type { Database } from "@webstudio-is/postrest/index.server";
 import {
-  prisma,
-  type AuthorizationToken,
-  type Project,
-} from "@webstudio-is/prisma-client";
-import {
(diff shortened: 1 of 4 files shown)


kof (approved):

istarkov (approved):

packages/authorization-token/src/db/authorization-token.ts
```diff
@@ -179,14 +181,16 @@ export const remove = async (
...
-        token: props.token,
-      },
-    },
-  });
+  const dbToken = await context.postgrest.client
+    .from("AuthorizationToken")
+    .delete()
+    .eq("projectId", props.projectId)
+    .eq("token", props.token)
+    .select()
```
istarkov, line 189: Not an issue, just found that we don't use result of remove anywhere so can skip select too

istarkov: Anyway it will throw if result is not "single" so it's ok

PR #3883, opened 2024-08-06
fix: parse transform-origin property values as tuples

JayaKrishnaNamburu:
## Description

The current parser is not able to parse all the possible values for `transform-origin` property. The property takes `single`, `double` and `triple` values. So these need to be parsed as `tuples` to support `triple` values now.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/packages/css-data/src/parse-css-value.test.ts
+++ b/packages/css-data/src/parse-css-value.test.ts
@@ -1,6 +1,7 @@
 import { describe, test, expect } from "@jest/globals";
 import { parseCssValue } from "./parse-css-value";
 import { toValue, type StyleProperty } from "@webstudio-is/css-engine";
+import { parseCss } from "./parse-css";
 
 describe("Parse CSS value", () => {
   describe("number value", () => {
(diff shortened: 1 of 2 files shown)


packages/css-data/src/parse-css-value.test.ts
```diff
@@ -920,3 +921,45 @@ describe("font-family", () => {
...
 });
+
+test("parse transform-origin", () => {
+  expect(parseCssValue("transformOrigin", "bottom")).toEqual({
+    type: "tuple",
+    value: [
+      {
+        type: "keyword",
+        value: "bottom",
+      },
```
TrySound, line 932: format in one line pls

TrySound (approved):

packages/css-data/src/parse-css-value.test.ts
```diff
@@ -1,6 +1,7 @@
 import { describe, test, expect } from "@jest/globals";
 import { parseCssValue } from "./parse-css-value";
 import { toValue, type StyleProperty } from "@webstudio-is/css-engine";
+import { parseCss } from "./parse-css";
```
TrySound, line 4: why import parseCss here?

PR #3897, opened 2024-08-08
feat: allow small screen size in preview

kof:
## Description

A user wants to be able to test in preview mode using dev tools screen simulation. 

## Steps for reproduction

1. resize the window to mobile size
2. see there is no. "too small alert"
3. see under 640px topbar elements got reduced
4. see that canvas is almost full width

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov:
image

On mobile view trying to move sidebars cause error

istarkov:
Why not just hide that draggable bars on mobile view. 
Detection like `@media (max-width: 767px) and (hover: none) {` should work well and allows css only solution

kof:
> Why not just hide that draggable bars on mobile view.

You mean the 5px handles on the left and right side of the canvas? Thought about it too, didn't want to invest more time into this issue.

kof:
But I agree, this way it would be more realistic to test

kof:
1. made canvas take full width, no space between, handles are now on top of canvas, I think this is ok since they are only visible when you hover them, the worst thing that can happen is you can't click 5 pixel in the corner. I think its unlikely that a user would want to make specifically the 5px on the side interactive, but I don't know, we will see.

2. made it work on touch because that's what chrome dev tools were emulating and there were several bugs with that

apps/builder/app/builder/features/workspace/workspace.tsx
```diff
@@ -56,7 +53,7 @@ const getCanvasStyle = (
 ) => {
   let canvasHeight;
 
-  if (workspaceRect?.height) {
+  if (workspaceRect?.height && scale !== 0) {
```
istarkov, line 56: In what case scale is 0?

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -69,6 +69,16 @@ const getValueDefault = (
   return clamp(value, minValue, maxValue);
 };
 
+const preventContextMenu = () => {
+  const handler = (event: MouseEvent) => {
+    event.preventDefault();
+  };
+  globalThis.addEventListener("contextmenu", handler);
```
istarkov, line 76: window no reason to use globalThis

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -231,11 +236,30 @@ export const numericScrubControl = (
   };
 };
 
+const setRootStyle = (
```
istarkov, line 239: Imo such code should have guard protection against parallel exection i.e.
It must not be called from 2 different places with the same property.
```
setRooltStyle // A component
setRootStyle // B component
dispose // A
dispose // B
```
broke root style

2 ways to avoid.
Use per property counter and save/restore original value into module level variable only if counter is 0

Or just fail if counter[prop] > 1

apps/builder/app/builder/features/workspace/workspace.tsx
kof: adding a comment

packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -165,9 +171,8 @@ export const numericScrubControl = (
         }, 150);
 
         targetNode.addEventListener("pointermove", handleEvent);
-        originalUserSelect =
-          targetNode.ownerDocument.documentElement.style.userSelect;
-        targetNode.ownerDocument.documentElement.style.userSelect = "none";
+        restoreUserSelect = setRootStyle(targetNode, "user-select", "none");
+        restoreContextMenu = preventContextMenu();
```
istarkov, line 175: Better to describe why we are preventing menu and how to reproduce. Same for select and other touches of global  things

istarkov (approved):

packages/design-system/src/components/primitives/numeric-gesture-control.ts
kof: agree, its technically impossible to have twice as this is used based on a single user interaction, so if we look at it from how its used, it should be fine, but from purely API perspective - yeah in theory its possible. 

Do you see any way this is a real concern or are we bike shedding here?

packages/design-system/src/components/primitives/numeric-gesture-control.ts
kof: added comments

PR #3900, opened 2024-08-08
refactor: migrate project to postgrest

TrySound:
Switched to postgrest in "project" package.
Do 3 requests now when load project to get
latest build and latest static build since
they are not inferred as related in postgrest.

### DIFF
--- a/packages/project/src/db/project.ts
+++ b/packages/project/src/db/project.ts
@@ -1,4 +1,4 @@
-import { prisma, Prisma } from "@webstudio-is/prisma-client";
+import { nanoid } from "nanoid";
 import {
   authorizeProject,
   type AppContext,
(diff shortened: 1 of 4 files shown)


packages/project/src/db/project.ts
```diff
@@ -22,20 +22,45 @@ export const loadById = async (
...
+    .from("Project")
+    .select(`*, previewImageAsset:Asset (*)`)
+    .eq("id", projectId)
+    .eq("isDeleted", false)
+    .single();
+  if (data.error) {
+    throw data.error;
+  }
+
+  const latestBuild = await context.postgrest.client
```
istarkov, line 35: Why not `lb:LatestBuildPerProject (*)` in select?

packages/project/src/db/project.ts
```diff
@@ -22,20 +22,45 @@ export const loadById = async (
...
+    .eq("projectId", projectId)
+    .maybeSingle();
+  if (latestBuild.error) {
+    throw latestBuild.error;
+  }
 
-  if (data === null) {
-    throw new Error(`Project ${projectId} not found`);
+  const latestStaticBuild = await context.postgrest.client
+    .from("LatestStaticBuildPerProject")
```
istarkov, line 45: same question as above

istarkov (approved):

PR #3908, opened 2024-08-09
fix: support html form in ssg

TrySound:
Here added RemixForm alias to native form in base sdk components. It should allow users to write at least native forms. Naming was not chosen well though.

### DIFF
--- a/packages/sdk-components-react/src/components.ts
+++ b/packages/sdk-components-react/src/components.ts
@@ -17,6 +17,7 @@ export { Subscript } from "./subscript";
 export { Button } from "./button";
 export { Input } from "./input";
 export { Form } from "./form";
+export { Form as RemixForm } from "./form";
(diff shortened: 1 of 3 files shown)


packages/sdk-components-react/src/components.ts
```diff
@@ -17,6 +17,7 @@ export { Subscript } from "./subscript";
 export { Button } from "./button";
 export { Input } from "./input";
 export { Form } from "./form";
+export { Form as RemixForm } from "./form";
```
kof, line 20: can we rename that form to remix form?

kof (approved):

packages/sdk-components-react/src/components.ts
TrySound: It's not remix form 😄
We just alias to it for compatibility

PR #3922, opened 2024-08-11
refactor: load compact data into builder

TrySound:
Load only arrays instead of maps from server
and convert data to map on client.

This reduces download size and server cpu usage.

340 kB -> 269 kB
218 kB -> 170 kB
22%


istarkov (approved):

kof (approved):
🔥

PR #3936, opened 2024-08-15
feat: style panel and settings panel keyboard shortcuts

kof:
## Description

- style panel - S
- settings panel - D
- ESC inside focused input - unfocuses input, enabling shortcuts
- ESC in left sidebar panels closes the panels
- improved contrast for some text elements

My best guess for D is that its next to S like on keyboard

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili (changes requested):
When search is active, Esc should blur the input so user can continue navigating with keyboard.

johnsicili:
The same is true for Style and Settings, as the style source and name (in Settings) are auto-focused.

johnsicili:
Contrast ratio on hints is too low

image

Can't wait for this!

kof:
> When search is active, Esc should blur the input so user can continue navigating with keyboard.

Yes, I think potentially we need to make any focused input loose focus when pressing esc, need to read up on the stadnards for this. We have this issue everywhere in general.

kof:
> Contrast ratio on hints is too low

Yes this is our standard subtle color from design system, I guess it was never tested on a tooltip, need to either make tooltip darker or the subtle color less subtle, needs to be tested on all usages.

kof:
Or maybe we need to finally start using relative colors, because subtle on light and subtle on dark need different brightness and adding a separate color for each like we started before is going to create too many colors. With relative colors we could just use one base color and choose 2 different opacity or brightness levels for dark and light.

kof:
@johnsicili contrast should be fixed now

johnsicili:
Nice! Another thing is when loading the project, Style Sources is focused by default. That's something new and don't think that should be the case

kof:
> Nice! Another thing is when loading the project, Style Sources is focused by default. That's something new and don't think that should be the case

Yes, I thought if we are making search in components autoFocus why not making the same for styles, what's the problem? Isn't this the most useful focused element when you open the style panel?

johnsicili:
Because it renders all keyboard shortcuts useless until you click out. Plus, the first action somebody takes when opening a project is almost always not adding a token to the body.

kof:
@johnsicili agreed, removing autofocus

johnsicili:
Use shortcut S for style, then press Tab – it focuses on the row at the bottom that shows the instance tree/nesting. However, if you click Style and then press tab, it focuses Style Sources which is the correct behavior

johnsicili (approved):
FIRE 😍

kof:
@johnsicili also added handling ESC in left sidebar panels

TrySound (approved):

apps/builder/app/builder/shared/commands.ts
```diff
@@ -134,6 +134,22 @@ export const { emitCommand, subscribeCommands } = createCommandsEmitter({
       },
       disableHotkeyOnFormTags: true,
     },
+    {
+      name: "openStylePanel",
+      defaultHotkeys: ["s"],
+      handler: () => {
+        $activeInspectorPanel.set("style");
+      },
+      disableHotkeyOnFormTags: true,
```
TrySound, line 143: need to forbid in content editable as well
now I cannot enter s or d when rename instances

PR #3938, opened 2024-08-15
refactor: rewrite html styles with css

TrySound:

We have a proper parser which can convert raw css into webstudio styles. Rewrite should simplify maintaining and allow to change final format without big rewrites.


kof (approved):
Nice, but we need to compare each and every property and value we generated to make sure there are no regressions, I didn't do it, did you?

TrySound:
Yep, I did compare previous styles with jest before deleting them.

PR #3939, opened 2024-08-15
refactor: optimize html styles access

TrySound:

Styles are accessed for specific fields and never iterated. Here output html styles as single map with composite key tag:property.


kof (approved):
didn't test

PR #3959, opened 2024-08-18
refactor: drop sorting by specificity

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3399 https://github.com/webstudio-is/webstudio/issues/1536

Even less work needed now to compute style.

Turns out since we check each possible style declaration the order of keys is determenistic and already can provide the right order.

Here removed computing specificity and ordering by it. Instead rearranged style declarations lookup.

### DIFF
--- a/apps/builder/app/shared/style-object-model.ts
+++ b/apps/builder/app/shared/style-object-model.ts
@@ -93,44 +93,12 @@ export const getPresetStyleDeclKey = ({
  * excluding everything after selected STYLESOURCE
  *
  */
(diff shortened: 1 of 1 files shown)


kof (approved):
uhh, is there tests making sure specificity is correct? this is veeery important

TrySound:
There are a lot of tests checking exactly this. Some broke, rearranging for loops fixed.

kof:
@TrySound as long as its impossible to break the order without breaking tests, all good

PR #3961, opened 2024-08-18
feat: added navigator panel keyboard shortcut and refactored navigator docking behavior

kof:
## Description

With shortcuts, it became very clear that we need to make navigator closable even  when it is undocked, because a shortcut should't just stop working.

I had to refactor loading animation because to cover with overlay the panel it now need a different way of rendering the overlay. Instead of passing some CSS, I am now rendering an overlay above everything else.

## Steps for reproduction

1. Navigator should close/open using
  a) tab trigger
  b) close button in the panel
  c) keyboard shortcut (z)
2. It should behave the same  way in docked and undocked state, with only one difference: when undocked - it takes space in the layout, makes canvas smaller. Docked - it goes above.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili (changes requested):
Issue: Add box, then add a heading (either via shortcuts or clicking). As soon as you add the child (heading), box is collapsed in the navigator.

On live, the box stays open so you can see what you're doing.

kof:
@johnsicili fixed it, it was collapsing it without checking if undocked

kof:
It should close when docked, but not touch it when undocked

johnsicili:
@kof I'm still seeing it. Was this the issue you fixed?

https://github.com/user-attachments/assets/d979ae2a-3830-4e41-baba-c69f66b21caf

kof:
@johnsicili great catch, didn't realize this one

johnsicili (approved):

TrySound (approved):

PR #3964, opened 2024-08-19
feat: Add tooltips for individual properties in style panels by the descriptions generated by ai

JayaKrishnaNamburu:
## Description

This PR adds tooltips for all the layer based sections in the transform property. The rest of the properties already have their tooltips managed by the `PropertyName` component. I have taken our previous tooltips as based and used a AI suggested one and proof read those. So, please let me know if there are any suggestions to change. 

## Steps for reproduction

- verify tooltips for `translate`, `scale`, `rotate` and `skew`, `box-shadow`, `text-shadow`, `drop-shadow` properties. 

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


kof (changes requested):
discussed to move the text to css-data

JayaKrishnaNamburu:
Hey I was writing a new prompt and new script to generate these descriptions for properties which we are building UIs. But need some more clarification on this script

Like let's take example of translate property. We can write a prompt to generate descriptions for translateX, translateY, translates.

But what happens when it is for skew or rotate. They basically functions under the transform property. So I am assuming we are going to keep a custom list of these properties which needs descriptions and we use that specific list, like this
-translateX
-translateY
-translateZ 
-rotateX 
-rotateY 
-rotateZ
-skewX 
-skewY 

with a prompt to generate the descriptions and not just property names like how it's currently happening in 'property-value-descriptions' file. Because skew is not an actual css property and it's just a function for transform property right.

And secondary, when I am using my own API key and ORG key form openAI. I get just 429 too many requests response and nothing useful. Is it something to do with personal account. Or should there be a paid account for generating such big number. I don't know, anyone faced the same.

kof:
why a new script? we already have a script that does all that, it must be missing these properties or something, that's it

kof:
>  Because skew is not an actual css property and it's just a function for transform property right.

I would just rename the script and change the key to whatever we need, so that it can be anything css as a key and description as a value

JayaKrishnaNamburu:
> why a new script? we already have a script that does all that, it must be missing these properties or something, that's it

I mean the prompt might need a little tweaking right. We just don't need the description for translate property. But we need descriptions for how property value of the translate value too right. Like x,y,z of translate etc.

kof:
yeah, tweaking prompt should be able to get you there

JayaKrishnaNamburu:
I have updated the script for generating properties a little bit, and we can generate the description for the syntaxes of properties that we managed individually using the UI controls. Please share your feedback to improve it, or the descriptions are looking fine now.

@kof @TrySound

apps/builder/app/builder/features/style-panel/property-label.tsx
```diff
@@ -323,10 +323,12 @@ export const PropertyInlineLabel = ({
   description,
   properties,
   disabled,
+  syntax,
 }: {
   label: string;
   description: string;
-  properties: [StyleProperty, ...StyleProperty[]];
+  properties?: [StyleProperty, ...StyleProperty[]];
+  syntax?: string;
```
TrySound, line 331: maybe for this specific component we could relax properties type while keeping autocomplete.
image

JayaKrishnaNamburu: Done, update the type and all the places where it is used 👍

TrySound (approved):

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -72,12 +107,87 @@ const newPropertiesNames = Object.keys(keywordValues)
...
+const newPropertySyntaxes = propertiesAndFunctionsSyntax.filter(
+  (property) =>
+    forceRegenerate ||
+    (typeof propertySyntaxesGenerated[property] !== "string" &&
+      typeof propertySyntaxesGenerated[property] !== "string")
+);
+
 let retries = 0;
 
 const backoff = (num: number) => {
```
kof, line 119: inlining retry with incremental backoff is a bit too much, there are good libs for that with battle tested logic, but also maybe you don't need incremental backoff, you can just repeat fixed amount of times after fixed amount of seconds and that's it, our use case is simple here

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -59,6 +61,39 @@ try {
 
 const batchSize = 16;
 
+const propertiesAndFunctionsSyntax = [
```
kof, line 64: there must be a better place for this list ....

packages/css-data/bin/property-value-descriptions.ts
kof: I used this small lib a years ago https://github.com/tim-kos/node-retry its tiny, but maybe there is something better @TrySound might know

kof: def. don't love the retry/backoff logic in the script here

kof: would love to organize this script a little better overall, its a bit hard to read, I just tried

packages/css-data/src/__generated__/property-value-descriptions.ts
```diff
@@ -494,10 +494,54 @@ export const propertiesGenerated = {
   viewTransitionName: "Sets a custom name for a transition effect",
   whiteSpaceCollapse:
     "Specifies how white space inside an element is collapsed",
+  "-webkit-box-orient":
+    '"Controls the direction of flex items in a flex container."',
```
kof, line 498: The prompt was trying to get rid of the additional quotes inside of single quotes, but seems like its not working any more. We need to either tweak the prompt again or get rid of the quotes manuallly

apps/builder/app/builder/features/style-panel/property-label.tsx
```diff
@@ -326,7 +326,7 @@ export const PropertyInlineLabel = ({
 }: {
   label: string;
   description: string;
-  properties: [StyleProperty, ...StyleProperty[]];
+  properties: [StyleProperty, ...StyleProperty[]] | string;
```
kof, line 329: its confusing, properties is plural, that should be an array ...  even if inside its only one element

JayaKrishnaNamburu: If i tried to keep `[StyleProeprty | string, ...StyleProperty[] | string]` it's breaking auto-complete. Found this might be a better way to differentiate those both.

TrySound (approved):

kof:
- [ ] text-shadow-offset-x and similar are internal representations, they don't exist in css, we need to make sure tooltip doesn't make it look like they do, lets chat what can be displayed instead in such cases

image

kof:
- [ ] backface visibility and perspective initial values now visible? what changed?

image

kof:
- [ ] something is wrong with spacing

image
vs

image

JayaKrishnaNamburu:
> * [ ]  backface visibility and perspective initial values now visible? what changed?
> 
> image

This started to appear after the migration to text-control. Taken care now 👍 
https://github.com/webstudio-is/webstudio/commit/ac26e8f7f7c3fdc796b18d9453157e9f5ed32398

JayaKrishnaNamburu:
Updated the tooltips now with a `label` and a `title`. 

> * [ ]  text-shadow-offset-x and similar are internal representations, they don't exist in css, we need to make sure tooltip doesn't make it look like they do, lets chat what can be displayed instead in such cases
> 

Screenshot 2024-09-02 at 10 42 57 PM
Screenshot 2024-09-02 at 10 42 50 PM

kof (approved):
Looking great

packages/css-data/bin/property-value-descriptions.ts
```diff
@@ -72,11 +76,63 @@ const newPropertiesNames = Object.keys(keywordValues)
...
 
-let retries = 0;
+const newPropertySyntaxes = propertiesAndFunctionsSyntax.filter(
+  (property) =>
+    forceRegenerate ||
+    (typeof propertySyntaxesGenerated[property] !== "string" &&
+      typeof propertySyntaxesGenerated[property] !== "string")
+);
+
+for (let i = 0; i < newPropertySyntaxes.length; ) {
```
kof, line 86: please use full words whenever possible, e.g. here 
```suggestion
for (let index = 0; index < newPropertySyntaxes.length; ) {
```

packages/css-data/bin/syntaxes.ts
```diff
@@ -0,0 +1,32 @@
+export const propertiesAndFunctionsSyntax = [
```
kof, line 1: @TrySound where can we put this file? I feel like bin is a bit bad for this type of data, wdyt?

TrySound: 🤷

apps/builder/app/builder/features/style-panel/property-label.tsx
```diff
@@ -320,13 +320,15 @@ export const PropertySectionLabel = ({
  */
 export const PropertyInlineLabel = ({
   label,
+  title,
   description,
   properties,
   disabled,
 }: {
   label: string;
+  title?: string;
```
TrySound, line 329: what this for?

packages/css-data/bin/syntaxes.ts
kof: I think it belongs to src, not bin, just not sure if we should have a separate file specifically for this or put it into custom-data.ts

kof: propertiesAndFunctionsSyntax is not descriptive enough imho, in 2 month I wouldn't know what that is and it needs a comment

kof: effectively this is a list of long-hands that we wish existed but they don't, right?

so its something like: customLonghandPropertyNames

apps/builder/app/builder/features/style-panel/property-label.tsx
JayaKrishnaNamburu: Screenshot 2024-09-03 at 10 47 33 AM
Screenshot 2024-09-03 at 10 47 19 AM

We changed the design here to hide the syntaxes when it not actually a valid css property. So, instead we show it as label. So, we have a label and a title. Not necessarily be the same so.

kof: we don't want to show box-shadow-offset-x because its not css standard, so we need something that is more descriptive in the title than what we have in the label, because X is not going to be clear, Offset X is clear

kof: TBH in case of Offset X and Offset Y we could say Offset X and Offset Y

kof: But we need to make it consistent everywhere, not sure how about other properties

apps/builder/app/builder/features/style-panel/sections/transforms/transform-and-perspective-origin.tsx
```diff
@@ -141,14 +135,28 @@ export const TransformAndPerspectiveOrigin = (
...
-              <Label>X</Label>
+            <Grid
+              gap="2"
+              align="center"
+              css={{ gridTemplateColumns: "auto 1fr" }}
+            >
+              <PropertyInlineLabel
+                label="X"
+                title={
+                  property === "transformOrigin" ? "Offset X" : "X Position"
```
kof, line 152: btw why is it not both Offset X Position X or X Offset and X Position ? I wonder what's the right way

JayaKrishnaNamburu: In MDN, they seem to use different wordings to represent these two
https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin#values

https://developer.mozilla.org/en-US/docs/Web/CSS/perspective-origin#values

kof: checked mdn, it uses x-offset and x-position, so why changing it and why not using humanizeString to get that value from data?

packages/css-data/bin/syntaxes.ts
JayaKrishnaNamburu: I think it's better to move it under the `custom-data.ts` file. Instead of a new way, And the name `customLonghandPropertyNames` is self explanatory too 👍

apps/builder/app/builder/features/style-panel/shared/shadow-content.tsx
```diff
@@ -93,6 +94,29 @@ const boxShadowInsetValues = [
   { value: "inset", Icon: ShadowInsetIcon },
 ] as const;
 
+const shadowPropertySyntaxes = {
```
kof, line 97: it feels like this mapping doens't make the code above simpler, maybe I am wrong, but have you checked if using `propertySyntaxes` directly everywhere leads to better readability? ignore this message if this is still better

apps/builder/app/builder/features/style-panel/sections/transforms/transform-and-perspective-origin.tsx
JayaKrishnaNamburu: Btw, we are not storing these things from the `mdn` right. To use it directly via `config`. Am i missing something, from where to use it 🤔. Except to keep them as strings like above.

kof: I thought you already had a config, but maybe I misremembered, you prob only have the long-form that combines property and position name

apps/builder/app/builder/features/style-panel/shared/shadow-content.tsx
JayaKrishnaNamburu: I personally believe this is much easier to read. Or else, we need to go through if checks every time we use these. As we are managing three properties `box-shadow`, `drop-shadow` and `text-shadow` using the same component.

apps/builder/app/builder/features/style-panel/sections/transforms/transform-and-perspective-origin.tsx
JayaKrishnaNamburu: Done 👍

JayaKrishnaNamburu:
Done, updated everything in the PR now. Give me a 👍, and i will merge it.

kof:
👍

PR #3968, opened 2024-08-20
fix: better sorting of transitions and allow `enter` key to save

### LINKED ISSUE #3958: Changing transition property requires clicking it in dropdown, even if it shows a check
When changing the transition property, for example, from the default opacity to height, height isn't selected until you click on it – even though height has a check before it, and I hit enter.

1. Add transition
2. Click into it and edit the property
3. Type `height`
4. See it checked, also try hitting enter
5. Click out of the transition and see opacity still there

https://github.com/user-attachments/assets/c155543e-7ed8-461d-a565-4a746d8321e8

JayaKrishnaNamburu:
fixes #3958 

## Description
This PR fixes the `enter` key behaviour for the transition-property combo box. And we updated the sort mechanism at the time of searching to show the most related property at the top. So, users don't need to scroll all the way to see the matched property. 

## Steps for reproduction

- Add a transition property layer.
- Now, click on the layer and the UI panel to edit the properties opens up.
- Type `height` or any other property in the combo box.
- Now, press `Enter`. The action should save the value of the property and update the layer.
- Next, search for some random property with a possibility for more results. Eg: `width`. 
- `width` or any other matching value that you searched for should always show up at the top of the list.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
@@ -50,7 +50,7 @@ export const TransitionProperty = ({
   onPropertySelection,
 }: TransitionPropertyProps) => {
   const valueString = toValue(property);
-  const [inputValue, setInputValue] = useState(valueString);
+  const [inputValue, setInputValue] = useState<string>(valueString);
   useEffect(() => setInputValue(valueString), [valueString]);
 
   const {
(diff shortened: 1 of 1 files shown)


kof:
- [ ] just found out that there is supposed to be some separation between the common section and the rest, now I don't seem to find any, what do I miss?

JayaKrishnaNamburu:
@kof looks like there might be some bug in the primitive itself. Because when the UI was initially built it was working good. Just cross check from old deployment. Will check once on the bug.
https://webstudio-builder-git-fix-transition-filtering-getwebstudio.vercel.app/builder/264a609f-ce63-4197-a364-68d239c2f56a
Screenshot 2024-08-20 at 6 56 31 PM

johnsicili:
@JayaKrishnaNamburu Enter works great! Thanks for the fix!

What's the point of having the checkmark if it's not selected by default, i.e., I can't click out without hitting enter or clicking it?

https://github.com/user-attachments/assets/54cae874-748f-482c-9b5c-bc5c7e1c3da9

I think auto-highlighting it would be more accurate. Not a big deal, though, and if complex, don't worry about it.

JayaKrishnaNamburu:
> https://webstudio-builder-git-fix-transition-filtering-getwebstudio.vercel.app/builder/264a609f-ce63-4197-a364-68d239c2f56a

I am not sure about the UX @johnsicili maybe we can open a new issue for the same to change on all places combo box is used.

JayaKrishnaNamburu:
Merging this, and opened a new issue to track the combobox issue with seperator
https://github.com/webstudio-is/webstudio/issues/3977

PR #3970, opened 2024-08-20
feat: add offset-position property under advanced section

### LINKED ISSUE #3963: CSS property missing: offset-position
Hi, 

I am trying to enter a css property in advance that doesn't exist:

offset-position

https://developer.mozilla.org/en-US/docs/Web/CSS/offset-position

Can you please add it to be supported by webstudio?

JayaKrishnaNamburu:
fixes #3963 

## Description
Adds `offset-position` property to be available under the advanced section. Of all the `offset-*` properties. Looks like only `position` is marked as experimental in `mdn-data`. But, according to can-i-use. It seems it has a fair amount of support among browsers.
https://caniuse.com/mdn-css_properties_offset-position
Screenshot 2024-08-20 at 11 44 59 AM

## Steps for reproduction
- Add `offset-position` property under the advanced tab.
- check for description of the property.
- Ability to reset the property from the advanced tab.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


packages/css-data/src/__generated__/properties.ts
```diff
@@ -2000,6 +2000,15 @@ export const properties = {
     },
     types: ["offset-path", "coord-box"],
   },
+  offsetPosition: {
```
TrySound, line 2003: offset-anchor seems missing as well
https://developer.mozilla.org/en-US/docs/Web/CSS/offset-anchor

JayaKrishnaNamburu: Good catch, but i see `offset-anchor` is marked as `standard` in the `mdn-data`. But inside `mdn-data` it is missing `mdn_url` and so the generator is skipping it. Kinda might need to add a `wild-card` check for this. What's you opinion ?

Any change i make to MDN, they are not merging at the moment.


```json
"offset-anchor": {
    "syntax": "auto | <position>",
    "media": "visual",
    "inherited": false,
    "animationType": "position",
    "percentages": "relativeToWidthAndHeight",
    "groups": [
      "CSS Motion Path"
    ],
    "initial": "auto",
    "appliesto": "transformableElements",
    "computed": "forLengthAbsoluteValueOtherwisePercentage",
    "order": "perGrammar",
    "status": "standard"
  },
```

TrySound: Some day 😄 
for now just add to experimental list

TrySound (approved):

PR #3987, opened 2024-08-22
fix: allow inserting instances into the body when it has text content

kof:
closes https://github.com/webstudio-is/webstudio/issues/3941

## Description

Allow inserting instances into the body when it has text content

## Steps for reproduction

1. select body, add text content over props
2. one-click insert any component
3. should insert and wrap text

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
+++ b/apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
@@ -187,8 +187,8 @@ export const usePropsLogic = ({
   });
   const canHaveTextContent =
     instanceMeta?.type === "container" &&
-    instance.component !== collectionComponent &&
-    instance.component !== "Body";
+    instance.component !== collectionComponent;
+
   const hasNoChildren = instance.children.length === 0;
   const hasOnlyTextChild =
     instance.children.length === 1 && instance.children[0].type === "text";
--- a/apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
+++ b/apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
@@ -187,7 +187,8 @@ export const usePropsLogic = ({
   });
   const canHaveTextContent =
     instanceMeta?.type === "container" &&
-    instance.component !== collectionComponent;
+    instance.component !== collectionComponent &&
+    instance.component !== "Body";
   const hasNoChildren = instance.children.length === 0;
   const hasOnlyTextChild =
     instance.children.length === 1 && instance.children[0].type === "text";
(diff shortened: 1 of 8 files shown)


apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
```diff
@@ -187,7 +187,8 @@ export const usePropsLogic = ({
   });
   const canHaveTextContent =
     instanceMeta?.type === "container" &&
-    instance.component !== collectionComponent;
+    instance.component !== collectionComponent &&
+    instance.component !== "Body";
```
TrySound, line 191: I would either check the instance is root
or allow text content on body but wrap text with span on insert

kof: I think its kinda useless to have text on body. We should have an isRoot function to be used everywhere, I have seen other places checking for component name, which I don't like but did the same thing

kof: I lmooked up and we have code like this in a lot of places:

    // 1 means root
    if (instanceSelector.length === 1) {
    
   everywhere there is a comment "1 means root" - honestly this is a bit of a smell, its telling this is not guessable and this logic only works when you actually have instanceSelector.
   
   I propose to rewrite all of these to isRootInstance(instance) function that takes an instance and what's inside - I don't really care, easierst right now is to keep component === 'Body' check but maybe you have a better idea.

kof: One thing that I already know will break is when we need to render a subtree, where root isn't body

kof: Btw in this case editing text should be even allowed, since its not a body.

kof: It seems that canHaveTextContent needs to be moved to sdk and should be reused across canvas and props editing and the check for having text has to actually depend on being body, or alternatively we allow having text there but we also allow inserting instance with 1 click when body has text just like we do with boxes, so that there is no error at all, it would just wrap and  insert

kof: oh more so there could be multiple roots in theory :)

kof: basically we can only tell what instance is root by looking at the available tree, not instance

TrySound (approved):

PR #3989, opened 2024-08-22
fix: transition-timing function update from the dropdown ui

### LINKED ISSUE #3957: Not able to change transition timing function
Try changing the default `ease` to something else, and it doesn't work.

Workaround is selecting custom, then selecting something else, and then it accepts it.

1. Add transition
2. Change timing
3. See red and it doesn't apply

https://github.com/user-attachments/assets/926765e3-a112-47bb-8ef7-4c929716c1e8

JayaKrishnaNamburu:
## Description

Fixes #3957 
Looks like there is a props mis-match during the refactoring from shorthand to using long hand properties. Now changed it to use the same convention every where for timing function. 

## Steps for reproduction

- Add a new transition layer.
- Update the `timing` from the transition-ui floating panel.
- The updated function from the select should reflect in UI, floating panel content box and in the style-panel layer name.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


kof (approved):
works great, merging it to release

PR #3990, opened 2024-08-22
chore: Project login

istarkov:
## Description

New Project login without any automation or link changes.

corresponding PR
https://github.com/webstudio-is/webstudio-saas/pull/319

## Steps for reproduction

- Goto [Tmp/Login manul page](https://oauth.staging.webstudio.is/tmp/login)
- Manually login/logout to project domains

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/services/token.server.ts
```diff
@@ -0,0 +1,178 @@
+import jwt from "@tsndr/cloudflare-worker-jwt";
```
TrySound, line 1: Not `jose`?

apps/builder/package.json
```diff
@@ -116,13 +120,15 @@
...
-    "zod": "^3.22.4"
+    "zod": "^3.22.4",
+    "zod-validation-error": "^3.3.1"
   },
   "devDependencies": {
     "@jest/globals": "^29.7.0",
     "@remix-run/dev": "^2.11.0",
     "@testing-library/react-hooks": "^8.0.1",
     "@types/cookie": "^0.6.0",
+    "@types/debug": "^4.1.12",
```
TrySound, line 131: 😢

TrySound (approved):

apps/builder/app/routes/auth.ws.ts
```diff
@@ -0,0 +1,37 @@
+import type { LoaderFunction } from "@remix-run/node";
+import createDebug from "debug";
+import { builderAuthenticator } from "~/services/builder-auth.server";
+import { isBuilderUrl } from "~/shared/router-utils/origins";
+
+const debug = createDebug("Auth").extend("Ws");
```
kof, line 6: we need to decide on convention we use for naming, I would prefer using exact module name
```suggestion
const debug = createDebug("auth").extend("ws");
```

apps/builder/app/routes/auth.ws_.callback.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      throw redirect("/error", {
+        headers: {
+          "Set-Cookie": await builderSessionStorage.commitSession(session),
+        },
+      });
+    }
+
+    debug("error", error);
+    console.error("error", error);
```
kof, line 67: why do you need both?

istarkov: console shown independent of debug, debug is colored in cli using color groups and easier to debug for me

apps/builder/app/routes/auth.ws.ts
kof: why do we extend"ws" here  but not in other places?

apps/builder/app/routes/logout.tsx
```diff
@@ -1,11 +1,49 @@
...
+import { builderAuthenticator } from "~/services/builder-auth.server";
+import {
+  getAuthorizationServerOrigin,
+  isBuilderUrl,
+} from "~/shared/router-utils/origins";
 
-export default function Logout() {
-  return null;
-}
+const debug = createDebug("Auth").extend("Logout");
```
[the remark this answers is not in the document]
kof: I haven't tried but I think here you won't get "ws:logout:auth", you will get just "logout:auth"

I think you need double extend here: createDebug("ws").extend('auth').extend("Logout")

and I also think we could create one debug function:
const debug = createDebug("ws")

and then reuse it everywhere, so that there is always "ws" namespace defined

apps/builder/app/routes/logout.tsx
```diff
@@ -1,11 +1,49 @@
...
+        );
+      }
+
+      return error;
+    }
+
+    throw error;
+  }
+
+  return json({});
```
kof, line 48: shouldn't this contain at least something? like the success state or so

apps/builder/app/routes/oauth.ws.authorize.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * Based RFC 6749 and RFC 7636
+ *
+ * https://datatracker.ietf.org/doc/html/rfc6749
+ *
+ * https://datatracker.ietf.org/doc/html/rfc7636
+ */
+export const loader: LoaderFunction = async ({ request }) => {
+  try {
```
kof, line 71: not in love with the giant try block, because its hard to see where it ends

apps/builder/app/routes/oauth.ws.authorize.tsx
```diff
@@ -0,0 +1,229 @@
...
+    debug("Authorize request received", request.url);
+
+    const url = new URL(request.url);
+    const searchParams = Object.fromEntries(url.searchParams);
+
+    const parsedRedirect = OAuthRedirectUriSchema.safeParse(searchParams);
+
+    if (!parsedRedirect.success) {
```
kof, line 79:
```suggestion
    if (parsedRedirect.success === false) {
```

apps/builder/app/routes/oauth.ws.authorize.tsx
```diff
@@ -0,0 +1,229 @@
...
+        },
+        { status: 400 }
+      );
+    }
+
+    let oauthError = createOauthError(redirect_uri, searchParams.state);
+
+    const parsedOAuthParams = OAuthParamsSchema.safeParse(searchParams);
+
+    if (!parsedOAuthParams.success) {
```
kof, line 118:
```suggestion
    if (parsedOAuthParams.success === false) {
```

apps/builder/app/routes/oauth.ws.authorize.tsx
```diff
@@ -0,0 +1,229 @@
...
+      redirectUri.searchParams.set("state", oAuthParams.state);
+
+      debug(
+        `Code ${code} created, redirecting to redirect_uri: ${redirectUri.href}`
+      );
+
+      return redirect(redirectUri.href);
+    }
+
+    if (!user) {
```
kof, line 190:
```suggestion
    if (user === ?) {
```

apps/builder/app/routes/oauth.ws.authorize.tsx
```diff
@@ -0,0 +1,229 @@
...
+      debug("Client is not registered");
+
+      return oauthError("unauthorized_client", "Client is not registered");
+    }
+
+    const oAuthParams = parsedOAuthParams.data;
+
+    const user = await authenticator.isAuthenticated(request);
+
+    if (user) {
```
kof, line 141: as an idea we could split this in a few functions, e.g. handleWithUser, handleWithoutUser 
this would make try/catch much smaller and still readable generally

apps/builder/app/routes/oauth.ws.token.ts
```diff
@@ -0,0 +1,156 @@
...
+ * OAuth 2.0 Token Request
+ *
+ * https://datatracker.ietf.org/doc/html/rfc7636
+ *
+ * https://datatracker.ietf.org/doc/html/rfc6749
+ */
+export const action = async ({ request }: ActionFunctionArgs) => {
+  debug("Token request received");
+  const jsonBody = Object.fromEntries((await request.formData()).entries());
+  debug("Token request received", jsonBody);
```
kof, line 44: why do you need 2 debug statements

apps/builder/app/routes/oauth.ws.token.ts
```diff
@@ -0,0 +1,156 @@
...
+        error_uri: "https://tools.ietf.org/html/rfc6749#section-5.2",
+      },
+      { status: 400 }
+    );
+  }
+
+  // verify the code_verifier against the stored code_challenge
+  if (
+    false ===
+    (await verifyChallenge(body.code_verifier, codeToken.codeChallenge))
```
kof, line 97: would be easier to read it you put it above over a variable

apps/builder/app/routes/oauth.ws.token.ts
istarkov: formData can fail

apps/builder/app/routes/tmp.login.tsx
```diff
@@ -0,0 +1,244 @@
...
+  ) => {
+    return [
+      {
+        id: data.projectId,
+      },
+    ];
+  };
+
+  return (
+    <div className="font-sans p-4">
```
kof, line 107: how did you get tailwind in here :)

apps/builder/app/routes/tmp.login.tsx
```diff
@@ -0,0 +1,244 @@
...
+      <h1 className="text-3xl">Welcome {data.user?.email ?? "Anon"}</h1>
+      <ul className="list-disc mt-4 pl-6 space-y-2">
+        {!data.isBuilderUrl && (
```
kof, line 110:
```suggestion
        {data.isBuilderUrl === ? && (
```

apps/builder/app/routes/tmp.login.tsx
```diff
@@ -0,0 +1,244 @@
...
+                className="text-blue-700 underline visited:text-purple-900"
+                href={noAccessProjectUrl.href}
+                rel="noreferrer"
+              >
+                Project link you have no access to
+              </a>
+            </li>
+          </>
+        )}
+        {!data.isBuilderUrl && data.user == null && (
```
kof, line 179:
```suggestion
        {data.isBuilderUrl === false && data.user == null && (
```

apps/builder/app/routes/tmp.login.tsx
```diff
@@ -0,0 +1,244 @@
...
+              className="text-blue-700 underline visited:text-purple-900"
+              href={"/login"}
+              rel="noreferrer"
+            >
+              Login at authorizer app
+            </a>
+          </li>
+        )}
+
+        {!data.isBuilderUrl && data.user != null && (
```
kof, line 191: same

apps/builder/app/services/auth-strategy/ws.server.ts
```diff
@@ -0,0 +1,75 @@
...
+>();
+
+/**
+ * The main issue with OAuth2Strategy is that it forces us to define authorizationEndpoint, tokenEndpoint, and redirectURI
+ * at the time of constructing the instance.
+ * However, in our case, these values are dynamic because we use multiple domains, such as development, staging, and production.
+ * The solution is to wrap the options object with a proxy, retrieving the necessary options from localAsyncStorage and proxying
+ * all requests through asyncLocalStorage.run
+ */
+export class WSStrategy<
```
kof, line 26:
```suggestion
export class WsStrategy<
```

apps/builder/app/routes/tmp.login.tsx
istarkov: see file name, its created to test current implementation

apps/builder/app/services/auth-strategy/ws.server.ts
```diff
@@ -0,0 +1,75 @@
...
+    >,
+    overrideOptions: (request: Request) => OAuth2StrategyOptionsOverrides
+  ) {
+    super(options, verify);
+
+    this.options = new Proxy<OAuth2StrategyOptions>(this.options, {
+      get: (target, property, receiver) => {
+        const store = asyncLocalStorage.getStore();
+        return (
+          store?.[property as keyof OAuth2StrategyOptions] ??
```
kof, line 45: shouldn't we check. property in OAuth2StrategyOptions and throw otherwise?

apps/builder/app/services/builder-auth.server.ts
```diff
@@ -0,0 +1,93 @@
...
+    throwOnError: true,
+  }
+);
+
+builderAuthenticator.use(
+  new WSStrategy<User>(
+    {
+      clientId: env.AUTH_WS_CLIENT_ID,
+      clientSecret: env.AUTH_WS_CLIENT_SECRET,
+      authorizationEndpoint: "https://OVERRIDED_ENDPOINT/oauth2/authorize",
```
kof, line 33:
```suggestion
      authorizationEndpoint: "https://OVERRIDEN_ENDPOINT/oauth2/authorize",
```

apps/builder/app/services/builder-auth.server.ts
```diff
@@ -0,0 +1,93 @@
...
+      tokenEndpoint: "https://OVERRIDED_ENDPOINT/oauth2/token",
+      redirectURI: "https://OVERRIDED_ENDPOINT/auth/callback",
```
kof, line 35: same here

apps/builder/app/services/token.server.ts
```diff
@@ -0,0 +1,178 @@
...
+  const keyBuffer = encoder.encode(key);
+
+  if (keyBuffer.length === length) {
+    return keyBuffer;
+  }
+
+  if (keyBuffer.length > length) {
+    // Truncate the key if it's too long
+    return keyBuffer.slice(0, length);
+  } else {
```
kof, line 16: no need for else here

apps/builder/app/services/token.server.ts
```diff
@@ -0,0 +1,178 @@
...
+export const encrypt = async (text: string, key: string) => {
+  const enc = new TextEncoder();
+  const keyBuffer = await crypto.subtle.importKey(
+    "raw",
+    normalizeKey(key),
+    { name: "AES-GCM" },
+    false,
+    ["encrypt"]
+  );
+  const iv = crypto.getRandomValues(new Uint8Array(12));
```
kof, line 33: whats iv?

apps/builder/app/services/token.server.ts
```diff
@@ -0,0 +1,178 @@
...
+): Promise<boolean> => {
+  const encoder = new TextEncoder();
+  const data = encoder.encode(codeVerifier);
+  const hashBuffer = await crypto.subtle.digest("SHA-256", data);
+  const base64Hash = Buffer.from(hashBuffer).toString("base64url");
+
+  return base64Hash === codeChallenge;
+};
+
+const CodeTokenPayloadEncSchema = z.object({
```
kof, line 75: we really didn't change the overall convention for zod schema naming, so far we used without suffix

kof: searching through codebase I can see now 2 notations ... really not liking this fact

kof: if we decided to use Schema prefix then lets rename the other places too

apps/builder/app/shared/router-utils/path-utils.ts
```diff
@@ -55,6 +56,38 @@ export const builderUrl = (props: {
   return url.href;
 };
 
+export const builderUrlNew = (props: {
```
kof, line 59: I would rather call this builderUrl and rename the other one to builderUrlDeprecated so that its absolutely clear what to use

packages/trpc-interface/src/authorize/project.server.ts
```diff
@@ -157,6 +172,43 @@ export const hasProjectPermit = async (
...
+    return checkProjectPermit(
+      props.projectId,
+      props.permit,
+      authInfo,
+      context.postgrest.client
+    );
+  } finally {
+    const diff = Date.now() - start;
+
+    console.info(`hasProjectPermit execution ${diff}ms`);
```
kof, line 208: why should this be logging all the time?

kof (changes requested):
Do I understand correctly that this is not yet being used? otherwise is the steps in the description all there is to test?  I feel its a bit suspicious and there should be many more things to test with this amount of code changes

apps/builder/app/routes/oauth.ws.token.ts
kof: is it happening often? I thought it would be only the case when data is entirely broken ... 
if its an absolute edge case I would still not add 2 debug statements permanentlt

apps/builder/app/routes/auth.ws_.callback.ts
kof: why do we need to see that error all the time in console if you are not debugging? seems like you are going to spam the console

packages/trpc-interface/src/authorize/project.server.ts
istarkov: Not the change of current PR

kof:
Might be just my test, but login took for me much longer, we might need a loading indicator

apps/builder/app/services/auth-strategy/ws.server.ts
istarkov: const x = {}
x.blabla <- does it throw?

same here

kof: yes, but I think you might want a good error message if wrong property is used? just to be nice

apps/builder/app/routes/oauth.ws.authorize.tsx
istarkov: It’s linear code just a lot of linear logic IMO adding functions for this makes it less readable

apps/builder/app/services/auth-strategy/ws.server.ts
istarkov: No just want to be ordinary proxy, some options can be not defined but accessed it’s ok

packages/trpc-interface/src/authorize/project.server.ts
kof: its ok, just don't get this type of stuff in general, we shouldn't be spamming console on a regular basis without a reason or control, right?

kof: this could be also a debug statement I guess

apps/builder/app/shared/router-utils/path-utils.ts
kof: or is the plan to replace the old one entirely in the next step?

apps/builder/app/routes/auth.ws_.callback.ts
istarkov: If we are here this error does not somehow tracked above. 
If we see it it's a real error need to be fixed ASAP. 
The debug library doesn’t natively support log levels like “info”, “warn”, or “error”. It’s designed to provide debugging output and uses a simple on/off mechanism based on namespaces.
Having it's coloring and ability to use search in console there is no way to leave here debug or console only message.

console and debug are for different purposes. console should work always, debug should work only if namespace is enabled, and allows to filter output by that namespace.

apps/builder/app/routes/logout.tsx
istarkov: changed on `throw new Error("Should not reach this point");`

apps/builder/app/routes/auth.ws_.callback.ts
kof: got it, I am wondering if we could have used one for both, because this seems like going to be everywhere and its not very nice, even though I understand the reasoning.

As an idea debug with error could be always logged by using error namespace

apps/builder/app/routes/oauth.ws.authorize.tsx
istarkov: Simplified removed another if

apps/builder/app/routes/oauth.ws.authorize.tsx
istarkov: Most routers allows to intercept errors after the main logic without using try catch but not remix loader.

apps/builder/app/shared/router-utils/path-utils.ts
istarkov: Plan to replace entirely

apps/builder/app/services/token.server.ts
istarkov: Yep, jose is a little bit harder. This is small and simple, anyway in case of using assymetric keys signing we will switch on jose.

packages/trpc-interface/src/authorize/project.server.ts
istarkov: We used 3rd party proovider for this. Since that time it's not needed as no issues with permissions checks we have

istarkov: Removed

.github/actions/vercel/action.yaml
```diff
@@ -79,6 +79,27 @@ runs:
         vercel build
       shell: bash
 
+    - name: Patch
+      run: |
+        # When we deploy on Vercel, it generates a URL like webstudio-saas-mahqcavgo-getwebstudio.vercel.app.
+        # We use the alias oauth-wstd-00-staging.vercel.app for routing, which maps to oauth.staging.webstudio.is on the worker.
+        # Issue: Vercel proxies also set x-forwarded-host, which overrides our header.
+        # We are adding x-forwarded-ws-host on the worker as a workaround, but issues with request.url persist.
+        # Remix and the Vercel adapter lack support for header selection https://github.com/vercel/vercel/blob/9d4d4b6deb6294506016106f78e71f1984adcc7f/packages/remix/defaults/server-node.mjs#L44
```
kof, line 88: pinging @ryanflorence and @TooTallNate just in case you guys see this notification

Issue #3977, opened 2024-08-21
Combobox separator is not rendering at the right place for dynamically rendered list items.


JayaKrishnaNamburu:
We found the bug while fixing #3968 for transition-property.
The `combobox` is built using he separator to  differentiate common transition properties all rest of the other properties. But looks like the separator is not rendering in the correct place now.

Here is a screenshot from the old rendering. And here is the link for a old deployment where you can cross confirm.
https://webstudio-builder-git-fix-transition-filtering-getwebstudio.vercel.app/builder/264a609f-ce63-4197-a364-68d239c2f56a

![transition-property](https://github.com/user-attachments/assets/954f9989-13ce-43ee-9d57-784082db0a70)

<img width="269" alt="Screenshot 2024-08-20 at 6 56 31 PM" src="https://github.com/user-attachments/assets/029c77bd-e9c5-4450-8864-42c483e86adf">

PR #3992, opened 2024-08-22
feat: Show the properties that are already applied on the node at the top of the transition property list.

### LINKED ISSUE #3982: Make it easy to find the right property in transition properties list
We are adding transforms section under #3411 

Among the transform properties `translate`, `scale` are applied on elements using the individual properties [translate](https://developer.mozilla.org/en-US/docs/Web/CSS/translate) and [scale](https://developer.mozilla.org/en-US/docs/Web/CSS/scale). But `skew` and `rotate` and managed using the [transform](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) property as functions. So, when users want to apply css transitions using the `transform` as `transition-property` like this

Screenshot 2024-08-21 at 9 14 14 PM

It applies only for `rotate` and `skew`. Because `translate` and `scale` are not added under the `transform` property. This behaviour is because we are managing `translate` and `scale` using the individual property behind the scenes when applied under the Transform section.

Here is a detailed explanation of the same from @johnsicili 
https://www.tella.tv/video/troubleshooting-transform-ui-4anv

## Proposal

After some discussion with @kof. The most sensible way to avoid such things seemed to show the list of properties that are already applied on the element that are effected by transition property.

 Which makes this list to be from this.

image
 
 to this
 
image

cc @kof @TrySound

JayaKrishnaNamburu:
fixes #3982  #3977 

## Description

This PR aims to improve the experience for selecting `transition-property` values while selecting from the dropdown.
And now, we show the `Rotate` and `Skew` next to the `Transform` layer name to exactly show what properties are being effected when something is applied on transform.

Screenshot 2024-08-22 at 8 18 29 PM

And the properties on the node are show at the top under the `Defined` sections.
Screenshot 2024-08-22 at 8 19 17 PM

## Steps for reproduction
- Add a transition layer and set the property to transform. 
- This will show the layer name as `Transform (Rotate, Skew)`
- Change the node to multiple breakpoints and it should change the defined properties list items.
- When there are no properties set on the node, the `Derived` section should not show on the list items.
- Should be able to select the item from dropdown using the arrow keys when in focus.
- Select the list item using the mouse.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
```diff
@@ -86,6 +89,21 @@ export const findTimingFunctionFromValue = (
...
+  currentStyle: StyleInfo
+): Set<string> => {
+  const properties: Set<string> = new Set();
+  let property: StyleProperty;
+  for (property in currentStyle) {
+    const value = currentStyle[property];
+    const prop = hyphenateProperty(property);
+    if (isAnimatableProperty(prop) === true && value?.local !== undefined) {
+      properties.add(prop);
+    }
```
JayaKrishnaNamburu, line 102: Right now, it is doing a very basic check if the values are set in local to the node. And the currentStyle style info seems to respect the breakpoints too when changed.

Opened the PR to give a brief experience of the UX for this. Please let me know the other cases that we can improve here.

kof:
- [ ] like I mentioned earlier, state styles like hover should be part of the list

JayaKrishnaNamburu:
transition-property

Heya @kof i have updated the PR now, to check for `preset`, `nextSource` and `previousSource` too to show the properties that are applied on the properties. Please le me know if there are any other style related effects that needs to be covered. I covered the type of effects i got they can be applied to.

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
```diff
@@ -29,29 +29,40 @@ import {
...
 export const TransitionProperty = ({
   property,
+  currentStyle,
   onPropertySelection,
 }: TransitionPropertyProps) => {
   const valueString = toValue(property);
   const [inputValue, setInputValue] = useState<string>(valueString);
   useEffect(() => setInputValue(valueString), [valueString]);
+  const propertiesOnInstanceSet = useMemo(
+    () => getAnimatablePropertiesOnInstance(currentStyle),
```
kof, line 56: maybe getTransitionableProperties ? to avoid confusion with keyframes animation

kof: though if they are the same, then the name is great

kof: ignore this comment, seems this is how its called in mdn

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
```diff
@@ -86,6 +89,30 @@ export const findTimingFunctionFromValue = (
   );
 };
 
+export const getAnimatablePropertiesOnInstance = (
```
kof, line 92: @TrySound due to your recent changes, does this need any change?

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
```diff
@@ -183,19 +215,29 @@ export const TransitionProperty = ({
               <ComboboxScrollArea>
                 {isOpen && (
                   <>
-                    <ComboboxLabel>Common</ComboboxLabel>
-                    {commonProperties.map(renderItem)}
+                    {propertiesOnInstance.length > 0 && (
```
kof, line 218: @TrySound what do we call propertied defined on an instance, including local, token and states?

kof:
A question: when we have border defined on all sides, shouldn't we shorten the property name for transition? Otherwise its not very practical

image

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
TrySound: cascaded values

kof:
Looks like the default value also went into the list, but I am not sure it should, wdyt?

image

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
TrySound: defined is fine too

kof (changes requested):
lets clarify a few things

apps/builder/app/builder/features/style-panel/sections/transitions/transition-property.tsx
kof: defined sounds good

apps/builder/app/builder/features/style-panel/sections/transitions/transition-utils.ts
TrySound: I'll do it later myself

JayaKrishnaNamburu:
> Looks like the preset value also went into the list, but I am not sure it should, wdyt? My assumption was that the properties user actively defined should go in there to help choose the right property, but presets are there anyways, if you take them this list will almost never be empty.
> 
> 
> 
> image
> 
> 

Most of the presets are resetting the values to defaults. So, probably be a good idea not to show them 👍🏽

JayaKrishnaNamburu:
> A question: when we have border defined on all sides, shouldn't we shorten the property name for transition? Otherwise its not very practical
> 
> 
> 
> image
> 
> 

Sounds good, but these are coming from presets right now. But in any case that could become a new story of its own. As I see it's not just the border right. We need some long hand to short hand style of parsing with existing properties by figuring out the short hand property they make up something like that.

kof:
@TrySound how hard is it to convert long to short? I have a feeling we already have this logic

johnsicili:
JFYI there is a border property in there too 

image

TrySound:
> @TrySound how hard is it to convert long to short? I have a feeling we already have this logic

Not for this, but nobody forbids to just write "border" in transition-property.

JayaKrishnaNamburu:
I just updated the PR to exclude presets stuff. And regarding the conversion of property names from long hand to shorthand to show under `Defined` column. I think we have a separate story for it. And we don't need to hold the transforms release imo.

cc @kof @TrySound

JayaKrishnaNamburu:
I just updated the PR to exclude presets stuff. And regarding the conversion of property names from long hand to shorthand to show under `Defined` column. I think we have a separate story for it. And we don't need to hold the transforms release imo.

cc @kof @TrySound

kof:
> Not for this, but nobody forbids to just write "border" in transition-property.

I know one can write it, the point here is to make it easy. Not for this means we don't have a converter?

kof:
We can merge it but I think the suggestions we make for properties around borders and such are going to be useless 50% of the time because each one individually and the user won't know they can use a shorthand, so it would be misleading them.

kof (approved):

PR #3993, opened 2024-08-22
refactor: switch to style object model in color control

TrySound:

Here switched to computed style declarations in color control. Used only to get value (used value) and currentcolor (computed value).

### DIFF
--- a/apps/builder/app/builder/features/style-panel/controls/color/color-control.tsx
+++ b/apps/builder/app/builder/features/style-panel/controls/color/color-control.tsx
@@ -1,4 +1,5 @@
-import { useState } from "react";
+import { useStore } from "@nanostores/react";
+import { useMemo, useState } from "react";
 import { Flex } from "@webstudio-is/design-system";
 import { toValue } from "@webstudio-is/css-engine";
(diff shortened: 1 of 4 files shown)


apps/builder/app/builder/features/style-panel/controls/color/color-control.tsx
```diff
@@ -45,25 +47,26 @@ export const ColorControl = ({
   deleteProperty,
   isAdvanced,
 }: ControlProps) => {
+  const $computedStyleDecl = useMemo(
+    () => createComputedStyleDeclStore(property),
+    [property]
+  );
+  const computedStyleDecl = useStore($computedStyleDecl);
```
kof, line 54: seems very unusual and mentally complex to generate a store in a hook and then memoize it in the component as well

is there anything we can do to simplify? If not, I think this is the case where it would be better in a custom hook, so we don't have to write it in every control.

kof (approved):
If you want me to test, I need a more precise tests. Approving but haven't tested.

TrySound:
You can test color and border color properties. How current color works. But anyway it heavily covered with tests so much more reliable than old implementation.

PR #4011, opened 2024-08-26
fix: hide outline when ephemeral update is in progress

kof:
Ref https://github.com/webstudio-is/webstudio/issues/3914
Closes https://github.com/webstudio-is/webstudio/issues/3800

## Description
This allows to see borders/outlines when changing ephimeral styles e.g. scrubbing border-outline width or color

This is not 100% fixing the problem, as you still can't see the border/outline when you change non ephimeral value, but there is not much we can actually do, because outline has to be blue for clarity, making it look like the user color, would be potentially very confusing.

## Steps for reproduction

1. change any ephimeral styles like padding drag/drop or scrub any input
2. see outline is hidden during that change

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/use-style-data.ts
+++ b/apps/builder/app/builder/features/style-panel/shared/use-style-data.ts
@@ -1,4 +1,4 @@
-import { useCallback, useRef } from "react";
+import { useCallback } from "react";
 import { useStore } from "@nanostores/react";
 import {
   type Breakpoint,
@@ -18,7 +18,6 @@ import {
 import { useStyleInfo } from "./style-info";
 import { serverSyncStore } from "~/shared/sync";
 import { $ephemeralStyles } from "~/canvas/stores";
-import { $isEphemeralUpdateInProgress } from "~/builder/shared/nano-states";
 
 export type StyleUpdate =
   | {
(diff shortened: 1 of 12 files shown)


johnsicili:
Click into border color field and click away. No border appears again until refresh

https://github.com/user-attachments/assets/2be3d636-be18-41eb-a51e-8bb4869738a2

johnsicili:
It seems to only disappear on scrubbing. But when I click in and arrow, it doesn't. This is probably needed because when you're fine-tuning to the point where you don't want to see the border, you need fine control, which isn't obtained easily via scrubbing.

https://github.com/user-attachments/assets/4dbf9ca9-93b0-4f18-84fb-efd9fcf217a0

kof:
@johnsicili same in wf, I also don't see a good way to solve this, because outline happens outisde and you can't generally let it get covered by something else. The only thing  I could imagine is some button or shortcut that would hide the outline until you turn it on again

kof:
@johnsicili I can add a view menu point "Show canvas outlines" and add a shortcut for it

apps/builder/app/builder/features/style-panel/shared/use-style-data.ts
```diff
@@ -18,6 +18,7 @@ import {
 import { useStyleInfo } from "./style-info";
 import { serverSyncStore } from "~/shared/sync";
 import { $ephemeralStyles } from "~/canvas/stores";
+import { $isEphemeralUpdateInProgress } from "~/builder/shared/nano-states";
```
TrySound, line 21: would checking $ephemeralStyles work for you?

johnsicili:
The outline doesn't always come back. Click the position dropdown, and click out. No outline

apps/builder/app/builder/features/style-panel/shared/use-style-data.ts
kof: actually it might

TrySound (approved):

PR #4017, opened 2024-08-27
refactor: migrate to property section label

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3914

New property label is used for section titles.

Screenshot 2024-08-27 at 21 18 25

### DIFF
--- a/apps/builder/app/builder/features/style-panel/property-label.tsx
+++ b/apps/builder/app/builder/features/style-panel/property-label.tsx
@@ -11,6 +11,7 @@ import {
   Flex,
   Kbd,
   Label,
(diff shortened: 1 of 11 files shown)


kof:
We also have a bug, where section butttons don't do anything but still show a button-label

image


Can we get that fix along this?

TrySound:
Yep, finally figured

JayaKrishnaNamburu (approved):
Looks good and working as expected, one thing i noticed is the tooltip descriptions for some sections need to be moved to be used from `property-value-descriptions` instead of the current static texts. I will take care of it in #3964

PR #4019, opened 2024-08-27
feat: Support timestamps in the Time Component

johnsicili:
## Description

Working with a blogging system that provides dates as 10-digit timestamps. This enables converting a unix timestamp to human readable date.

I have a 10 length check in code, but let everything else through. I was going to add a condition for ` else if (datetime.length !== 13)`, but it seems like we just pass through any characters anyway, like `asdf` in the datetime field.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/time.tsx
+++ b/packages/sdk-components-react/src/time.tsx
@@ -331,22 +331,15 @@ export const parseDate = (datetimeString: string) => {
   if (datetimeString === "") {
     return;
   }
-  const timestamp = Number(datetimeString);
-  if (
-    isNaN(timestamp) === false &&
-    Number.isNaN(new Date(datetimeString).getTime())
-  ) {
-    // 10 length timestamp is in seconds, convert to milliseconds.
-    datetimeString = new Date(
-      datetimeString.length === 10 ? timestamp * 1000 : timestamp
-    ).toISOString();
-  } else {
-    datetimeString = datetimeString.toString();
+  let timestamp = Number(datetimeString);
+  const isNumber = Number.isNaN(timestamp) === false;
+  if (isNumber && datetimeString.length === 10) {
+    timestamp = timestamp * 1000;
   }
+  const date = new Date(timestamp);
+  const isValid = false === Number.isNaN(date.getTime());
 
-  const date = new Date(datetimeString);
-  const isValidDate = false === Number.isNaN(date.getTime());
-  if (isValidDate) {
+  if (isValid) {
     return date;
   }
 };
--- a/packages/sdk-components-react/src/time.tsx
+++ b/packages/sdk-components-react/src/time.tsx
@@ -348,8 +348,18 @@ export const Time = React.forwardRef<React.ElementRef<"time">, TimeProps>(
       timeStyle: timeStyleOrUndefined(timeStyle),
     };
 
-    const datetimeString =
-      datetime === null ? INVALID_DATE_STRING : datetime.toString();
+    let datetimeString;
+    if (datetime === null) {
+      datetimeString = INVALID_DATE_STRING;
+    } else if (isNaN(Number(datetime)) === false) {
+      let timestamp = Number(datetime);
+      if (datetime.length === 10) {
+        timestamp = timestamp * 1000;
+      }
+      datetimeString = new Date(timestamp).toISOString();
+    } else {
+      datetimeString = datetime.toString();
+    }
 
     const date = new Date(datetimeString);
     const isValidDate = false === Number.isNaN(date.getTime());
--- a/packages/sdk-components-react/src/time.tsx
+++ b/packages/sdk-components-react/src/time.tsx
@@ -348,17 +348,20 @@ export const Time = React.forwardRef<React.ElementRef<"time">, TimeProps>(
       timeStyle: timeStyleOrUndefined(timeStyle),
     };
 
-    let datetimeString;
-    if (datetime === null) {
-      datetimeString = INVALID_DATE_STRING;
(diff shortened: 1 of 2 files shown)


packages/sdk-components-react/src/time.tsx
```diff
@@ -348,8 +348,18 @@ export const Time = React.forwardRef<React.ElementRef<"time">, TimeProps>(
...
     };
 
-    const datetimeString =
-      datetime === null ? INVALID_DATE_STRING : datetime.toString();
+    let datetimeString;
+    if (datetime === null) {
+      datetimeString = INVALID_DATE_STRING;
+    } else if (isNaN(Number(datetime)) === false) {
+      let timestamp = Number(datetime);
+      if (datetime.length === 10) {
```
istarkov, line 356: In `Sat Nov 20 2286` this logic will die ;-)

packages/sdk-components-react/src/time.tsx
```diff
@@ -348,8 +348,18 @@ export const Time = React.forwardRef<React.ElementRef<"time">, TimeProps>(
       timeStyle: timeStyleOrUndefined(timeStyle),
     };
 
-    const datetimeString =
-      datetime === null ? INVALID_DATE_STRING : datetime.toString();
+    let datetimeString;
+    if (datetime === null) {
+      datetimeString = INVALID_DATE_STRING;
+    } else if (isNaN(Number(datetime)) === false) {
```
istarkov, line 354: datetime="2024.10" is valid Date but also a valid number

istarkov: Mb additionally check that 
`Number.isNaN(new Date().getTime(datetime)) && isNaN(Number(datetime)) === false`

Number.isNaN(new Date().getTime(datetime)) <-- means that Date is invalid

istarkov: Or check that parsed number is integer `Number.isInteger` and that it's above some maximum possible year like > 2100

istarkov (approved):
I would check that datetime is not parsed as Date 
But should not be a big issue

packages/sdk-components-react/src/time.tsx
istarkov: Also what about negative numbers

packages/sdk-components-react/src/time.tsx
```diff
@@ -348,8 +348,21 @@ export const Time = React.forwardRef<React.ElementRef<"time">, TimeProps>(
       timeStyle: timeStyleOrUndefined(timeStyle),
     };
 
-    const datetimeString =
-      datetime === null ? INVALID_DATE_STRING : datetime.toString();
+    let datetimeString = INVALID_DATE_STRING;
+
+    if (datetime !== null) {
```
kof, line 353: Can't we normalize the timestamp first? seems much simpler
```js
let timestamp = Number(datetime);
if ( Number.isNaN(timestamp) === false && datetime.length === 10) {
  timestamp = timestamp * 1000
}

kof: turns out that 10 digit timestamp is the same unix timestamp but in seconds instead of milliseconds

kof: please also add a comment, I would have never guessed php is using this timestamp

johnsicili: Following up from DM @kof 

> Can't we normalize the timestamp first? seems much simpler

If we only needed logic on the 10 digit yes, but we need logic for all timestamps. Converting to an iso string lets them get parsed again by the date parser. If we leave them as a timestamp string, they'll fail. So we'd need to cast all timestamps to numbers which the condition would grow again and have to change the logic elsewhere too.

My understanding is you wanted to do something like this:

```js
if (datetime) {
      let timestamp = Number(datetime);
      if (Number.isNaN(timestamp) === false && datetime.length === 10) {
        timestamp = timestamp * 1000
      } else {
        datetimeString = datetime.toString();
      }
    }
    ```
    
    But that leaves all other timestamps untouched
    
    Open to suggestions.

kof: at this point I am lost, lets get this into a function and cover it with unit tests, this is too confusing

kof: might be too much for you, so someone could take over, but need a list of inputs you are trying to support

johnsicili: 1698377386 - A valid 10-digit Unix timestamp (October 26, 2023).
2147483647 - The maximum 32-bit signed integer timestamp (January 19, 2038).
1698377386000 - A valid 13-digit Unix timestamp (October 26, 2023).
9999999999999 - A far future date (November 20, 2286).

kof: cool, that's def. not clear from reading code, needs unit tests, I can add them if you want me to take over

johnsicili: that would be great, thanks

kof: Added tests, made sure they work with your implemeentation, then did what I suggested and tests passed, not sure what i am missing but all tests are passing

johnsicili:
2024.10 doesn't parse correctly now

kof:
branch name too long, tbc https://github.com/webstudio-is/webstudio/pull/4028

kof:
> 2024.10 doesn't parse correctly now

great, fixing in the new one

PR #4021, opened 2024-08-28
refactor: decouple style updates from react

TrySound:
Removed setProperty, deleteProperty and createBatchUpdates from react hook. All parameters can be accessed from global stores.

Note: review with hidden spaces.


kof (approved):
didn't test

PR #4023, opened 2024-08-28
refactor: switch menu and toggle controls to style object model

TrySound:
Here migrated menu control and toggle control from style panel
to new style engine and improve UX a bit.

Instead of disabling button, it gets "disabled" style and tooltip with warning. Though user still can use control to override advanced value in place.

Screenshot 2024-08-28 at 18 37 09

image


kof (approved):
Nice!

PR #4025, opened 2024-08-29
refactor: migrate select control to style object model

TrySound:

No visual or functional changes, just changed the source of computed value.


apps/builder/app/builder/features/style-panel/controls/select/select-control.tsx
```diff
@@ -29,7 +28,6 @@ export const SelectControl = ({
 
   return (
     <Select
-      disabled={isAdvanced}
```
kof, line 32: if select has advanced value, do we show it in that select?

kof (approved):

apps/builder/app/builder/features/style-panel/controls/select/select-control.tsx
TrySound: Yep, the flag was never passed. If we decide to indicate, it can be done inside control.

PR #4028, opened 2024-08-29
feat: Support multiple types of timestamps in the Time Component

kof:
## Description

Date can be in various forms, lets support them all:

1698377386 - A valid 10-digit Unix timestamp (October 26, 2023).
2147483647 - The maximum 32-bit signed integer timestamp (January 19, 2038).
1698377386000 - A valid 13-digit Unix timestamp (October 26, 2023).
9999999999999 - A far future date (November 20, 2286).
2024.10

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/time.test.ts
+++ b/packages/sdk-components-react/src/time.test.ts
@@ -21,6 +21,12 @@ test("far future date", () => {
   );
 });
 
+test("parse ISO date string", () => {
+  expect(parseDate("2024-08-29T13:36:17.000Z")).toEqual(
+    new Date("2024-08-29T13:36:17.000Z")
+  );
+});
+
 test("empty string", () => {
   expect(parseDate("")).toEqual(undefined);
 });
--- a/packages/sdk-components-react/src/time.test.ts
+++ b/packages/sdk-components-react/src/time.test.ts
@@ -29,6 +29,7 @@ test("parse ISO date string", () => {
 
 test("parse short date", () => {
   expect(parseDate("2024.10")).toEqual(new Date("2024-10-01T00:00:00.000Z"));
+  expect(parseDate("2024/10")).toEqual(new Date("2024-10-01T00:00:00.000Z"));
 });
 
 test("empty string", () => {
--- a/packages/sdk-components-react/src/time.test.ts
+++ b/packages/sdk-components-react/src/time.test.ts
@@ -20,3 +20,7 @@ test("far future date", () => {
     new Date("2286-11-20T17:46:39.999Z")
   );
 });
+
+test("empty string", () => {
+  expect(parseDate("")).toEqual(undefined);
+});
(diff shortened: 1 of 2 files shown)


johnsicili (approved):
Works. Thanks for the assist!

packages/sdk-components-react/src/time.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  expect(parseDate("1724938577")).toEqual(new Date("2024-08-29T13:36:17.000Z"));
+});
+
+test("maximum 32-bit signed integer timestamp", () => {
+  expect(parseDate("2147483647")).toEqual(new Date("2038-01-19T03:14:07.000Z"));
+});
+
+test("far future date", () => {
+  expect(parseDate("9999999999999")).toEqual(
+    new Date("2286-11-20T17:46:39.999Z")
```
TrySound, line 20: Huh, crazy year

packages/sdk-components-react/src/time.tsx
```diff
@@ -327,6 +327,23 @@ const timeStyleOrUndefined = (
...
 };
 
+export const parseDate = (datetimeString: string) => {
+  if (datetimeString === "") {
+    return;
+  }
+  const isNumber = /^\d+$/.test(datetimeString);
+  let timestamp = Number(datetimeString);
+  if (isNumber && datetimeString.length === 10) {
+    timestamp = timestamp * 1000;
```
TrySound, line 337: *=

packages/sdk-components-react/src/time.tsx
```diff
@@ -327,6 +327,23 @@ const timeStyleOrUndefined = (
   return undefined;
 };
 
+export const parseDate = (datetimeString: string) => {
+  if (datetimeString === "") {
+    return;
+  }
+  const isNumber = /^\d+$/.test(datetimeString);
```
istarkov, line 334: 01 is bad number
moreover it's check on positive Int can be done using JS i.e 
ordinary check isNumber +   Number.isInteger(value) && value >= 0;

packages/sdk-components-react/src/time.tsx
```diff
@@ -327,6 +327,23 @@ const timeStyleOrUndefined = (
...
+    timestamp = timestamp * 1000;
+  }
+  const date = new Date(isNumber ? timestamp : datetimeString);
+  const isValid = false === Number.isNaN(date.getTime());
+
+  if (isValid) {
+    return date;
+  }
+};
+
```
istarkov, line 346: IMO better to check that Date isValid first and only then try any other formats.
Or explain why we check non std date formats first

istarkov (approved):
anyway with tests it's ok

packages/sdk-components-react/src/time.tsx
istarkov: Why as Date with 10 digits string is NaN

istarkov: `new Date("1000000000")` <- INVALID

PR #4029, opened 2024-08-29
refactor: improve color picker

TrySound:
Color control can be a thin wrapper around color picker

- fixed "inherit" keyword support
- moved intermediate value into color picker
- merge highlight and change low level events


TrySound:
Not necessary

PR #4037, opened 2024-08-31
fix: Default form control values

kof:
## Description

- always use defaultValue and defaultChecked props when passing to react
- expose value and checked as props, remove default*
- reorder initial props to make more sense (name, value, checked - in this order)

Affected components:

Forms:
- checkbox
- radio
- input
- select
- textarea

Radix:
- checkbox
- radio group
- select
- switch

## Steps for reproduction

1. test each component named above how bound and unbound value/checked works
2. use in regular form and bind to url params
3. use in webhook form 

## Code Review

- [ ] hi @john, I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
tbc https://github.com/webstudio-is/webstudio/pull/4039

PR #4045, opened 2024-08-31
fix: Default form control values

kof:
## Description

- always use defaultValue and defaultChecked props when passing to react
- expose value and checked as props, remove default*
- reorder initial props to make more sense (name, value, checked - in this order)

Affected components:

Forms:
- checkbox
- radio
- input
- select
- textarea

Radix:
- checkbox
- radio group
- select
- switch

## Steps for reproduction

1. test each component named above how bound and unbound value/checked works
2. use in regular form and bind to url params
3. use in webhook form 

## Code Review

- [ ] hi @john, I need you to do
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react-radix/src/checkbox.ws.ts
```diff
@@ -146,7 +146,7 @@ export const metaCheckboxIndicator: WsComponentMeta = {
 
 export const propsMetaCheckbox: WsComponentPropsMeta = {
   props: propsCheckbox,
-  initialProps: ["id", "className", "checked", "name", "required"],
+  initialProps: ["id", "className", "name", "value", "required", "checked"],
```
TrySound, line 149: What does "value" mean in checkbox?

kof: same as in native checkbox https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value

johnsicili:
Adding my notes from DM:

Works well on checkbox, but not on a 2nd component I tested with: Switch.

Once I bind the value of `checked` to `system.search.hello`, the value can't be changed by the user. However, if I bind that same expression to `checked` on a checkbox, the default state is set and the user can change it.

The expression I'm using on `checked` is: `system.search.hello ? true : <the checked variable that comes with the component such as checkboxChecked>`. I also tried `system.search.hello ? true : false`


https://github.com/user-attachments/assets/1d40f028-4b91-4f82-9b7d-511005be895c

kof:
@johnsicili good catch, I didn't fix Switch

kof:
@johnsicili updated select and switch

johnsicili:
Checkboxes and switches work now. Radix radios already work and so do selects.

The same logic needs to be applied to Webstudio Components like Checkbox and Radio

kof:
html radio should be already working, checkbox - just found that I mistyped, should be also fixed now

johnsicili (approved):
Looks good thanks!

TrySound (approved):

packages/sdk-components-react-radix/src/select.tsx
```diff
@@ -24,13 +24,11 @@ import {
...
 
-// wrap in forwardRef because Root is functional component without ref
-export const Select = forwardRef<
-  HTMLDivElement,
-  ComponentPropsWithoutRef<typeof Root>
->((props, _ref) => {
-  return <Root {...props} />;
-});
+export const Select = forwardRef<HTMLDivElement, ComponentProps<typeof Root>>(
+  ({ value, defaultValue, ...props }) => {
```
istarkov, line 28: ref is needed even if not used
select.tsx:27 Warning: forwardRef render functions accept exactly two parameters: props and ref. Did you forget to use the ref parameter?

﻿So please don't delete _args they are for a reason

kof:
Ohh

PR #4046, opened 2024-08-31
feat: Hide the label of the instance when mouse is not hovering it

kof:
## Description

When instance is selected the label often gets in the way of seeing the element behind the label, we think its better to only show the label when user is hovering the instance

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili (approved):
Works well as far as I can tell. Will take some getting used to during an actual build but I think it has more upside potential than any loss.

kof:
Need to be vigilant if it makes the UX worse than. before

TrySound (approved):

PR #4058, opened 2024-09-03
feat: show generated css in property label

TrySound:
Here improved property label to show css declaration with computed (used) value.

image


kof:
we really should use a shorthand notation when all sides are the same, for ease of reading, but if too hard later ...

PR #4063, opened 2024-09-04
experimental: migrate toggle group to style object model

TrySound:

Switched to new style engine for toggle group controls. Those are tricky caz have property tooltip internally.

Improved tooltip to not show reset button and source when highlight not selected toggle button.

Can be tested
- background "repeat"
- flex child "sizing"
- flex child "order"
- size "overflow"


kof (approved):
tested

kof:
Are you aware that background repeat can not be resetted any more?

TrySound:
> Are you aware that background repeat can not be resetted any more?

None other controls in background content can

PR #4084, opened 2024-09-07
experimental: migrate transitions to style object model

TrySound:

Here migrated transitions section to new style engine along with "repeated style" control.

Changed behavior of transition code blue to write non-ephemeral changes caz ephemeral do not have enough visual feedback.

Note: review with hidden spaces


kof (changes requested):
- [ ] some properties don't get accepted, e.g. boxShadow

   https://share.descript.com/view/HyA2dFpDl1U
 - [ ] defined properties are camel cased instead of dash 
image

kof (approved):

PR #4086, opened 2024-09-08
refactor: add repeated style utils

TrySound:
Here replaced utils for adding and editing transitions with generic ones which should cover cases for most other sections.

Can be tested adding and editing transitions and code


kof:
Something is wrong, didn't compare to prod:
1. Defined section in transition doesn't show the transform from hover
2. Can't see the transform in the transforms list nor the label button when hover is not active

https://share.descript.com/view/fgeOeBln6Fw

TrySound:
Hover not selected nor active in dom. Why would it be shown in style panel? We do not show styles for inactive breakpoints. It always worked like this.

kof:
Yeah, I wasn't sure, but the defined in transition was showing properties on any states

kof (approved):

PR #4092, opened 2024-09-09
experimental: migrate shadows sections to style object model

TrySound:

Refactoed box shadow and text shadow sections with "repeated style" control and switched to new style engine. Simplified layer label logic.


kof (approved):
tested

PR #4097, opened 2024-09-09
experimental: migrate filter sections to style object model

TrySound:

Refactored filter and backdrop filter sections with repeated styles. Added tests for repeated style operations and added support for tuple. Got rid of legacy "style layers list" as no longer used anywhere.


kof (approved):
tested

PR #4107, opened 2024-09-11
fix: check for expanded templates when a template is selected from the cli init flow.

### LINKED ISSUE #4105: Template vanilla is not available
When I try to use the CLI to export a Vanilla js app I get this error: 

■  Template vanilla is not available
│  Please check webstudio --help for more details

It seems to work fine for a Static Site however.

JayaKrishnaNamburu:
fixes #4105 
## Description

This issue fixes picking the default vanilla template in the cli init-flow. Right now, when a template is selected from the `PROJECT_TEMPLATES` we are directly sending the value to `build`. But we need to check for the `expanded` value of the template in order to pick all the series of templates needed for the build. 

## Steps for reproduction
- create a project in the editor.
- Now, create a share-link for the project and copy it.
- Run `webstudio` in your local terminal and then the cli prompts for the URL, parse the url from the above step.
- Now, pick `Vanilla` as a template from the list of the templates.
- The project should build a vanilla template without any errors.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging


istarkov:
Doesnt look right, every template before had no issues with just config.

istarkov:
Or code used for template config is not reused

JayaKrishnaNamburu:
When it is being used as `webstudio build --template vanilla` the cli is actually picking the template and then expanding it. But when it is loading from the flow, it is just passing vanilla directly. Where `vanilla` folder doesn't exist. Looks like the logic is not being reused there.

istarkov:
See this code should be in a single place
https://github.com/webstudio-is/webstudio/blob/54e686a4e304855574c845ffef7352985b990400/packages/cli/src/commands/build.ts#L33-L52

Otherwise any changes in the build logic will ruin above cli part again and again.
Seems like cli must somehow reuse underlying utils

istarkov:
Probably every util buildOptions should be reused in the main cli

TrySound:
Or maybe just move the logic from yargs inside of "build" function so it automatically reused in "init" as well.

JayaKrishnaNamburu:
> Or maybe just move the logic from yargs inside of "build" function so it automatically reused in "init" as well.

This is exactly what i wanted to do too, gotten inside the `linear-gradient` stuff and missed it. Will update the PR 👍

JayaKrishnaNamburu:
@istarkov @TrySound updated the logic now to use from `build-utils`. Tested it, works fine and fixes the bug too.

istarkov (approved):

PR #4112, opened 2024-09-12
refactor: rewrite backgrounds section with repeated style

TrySound:

Switched to new style engine and repeated style component for backgrounds section. A lot of utilities are replaced with more generic solution.
Screenshot 2024-09-12 at 10 23 07


kof (changes requested):
These are broken

- hide/show layers
- delete
- sort

kof (approved):
didn't retest

PR #4113, opened 2024-09-12
fix: Inserting Slots within Slots from the Marketplace causes infinite loop and crash

### LINKED ISSUE #4111: Inserting Slots within Slots from the Marketplace causes infinite loop and crash
1. Add Marketplace item with Slot such as Ani Motion > Code
2. Click into the inserted Slot
3. Add the same Slot again from the Marketplace
4. Notice Builder freezes and will crash if you insert more

Seems to only be repro with Marketplace.

https://github.com/user-attachments/assets/fcbcd278-eda5-4cae-a979-a9dc6466aef4

istarkov:
## Description

closes #4111

- [x] - Fixes client side
- [x] - Fixes server side

https://github.com/user-attachments/assets/8bb6ffdc-2e78-45eb-b98b-e697daee822a

## Steps for reproduction

Now seems like impossible to save broken project. Tested locally on broken project.

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/project-build/src/db/build.ts
+++ b/packages/project-build/src/db/build.ts
@@ -32,7 +32,12 @@ const parseCompactData = <Item>(serialized: string) =>
 
 const parseCompactInstanceData = (serialized: string) => {
   const instances = JSON.parse(serialized) as Instance[];
+
+  // @todo: Remove after measurements on real data
+  console.time("breakCyclesMutable");
   breakCyclesMutable(instances, (node) => node.component === "Slot");
+  console.timeEnd("breakCyclesMutable");
+
   return instances;
 };
 
--- a/packages/project-build/src/db/build.ts
+++ b/packages/project-build/src/db/build.ts
@@ -25,49 +25,14 @@ import { parseDeployment } from "./deployment";
 import { serializePages } from "./pages";
 import { createDefaultPages } from "../shared/pages-utils";
 import type { MarketplaceProduct } from "../shared//marketplace";
-import { findCycles } from "../shared/graph-utils";
+import { breakCyclesMutable } from "../shared/graph-utils";
 
 const parseCompactData = <Item>(serialized: string) =>
   JSON.parse(serialized) as Item[];
 
 const parseCompactInstanceData = (serialized: string) => {
   const instances = JSON.parse(serialized) as Instance[];
-  const cycles = findCycles(instances);
-  if (cycles.length === 0) {
-    return instances;
-  }
-  // eslint-disable-next-line no-console
-  console.error("Cycles detected in instance data", cycles);
-
-  const cycleInstances = new Map<Instance["id"], Instance>();
-  const cycleInstanceIdSet = new Set<Instance["id"]>(cycles.flat());
-
-  // Pick all instances that are part of the cycle
-  for (const instance of instances) {
-    if (cycleInstanceIdSet.has(instance.id)) {
-      cycleInstances.set(instance.id, instance);
-    }
-  }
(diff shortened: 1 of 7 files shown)


packages/project-build/src/db/build.ts
```diff
@@ -25,17 +25,66 @@ import { parseDeployment } from "./deployment";
 import { serializePages } from "./pages";
 import { createDefaultPages } from "../shared/pages-utils";
 import type { MarketplaceProduct } from "../shared//marketplace";
+import { findCycles } from "../shared/graph-utils";
 
 const parseCompactData = <Item>(serialized: string) =>
   JSON.parse(serialized) as Item[];
 
+const parseCompactInstanceData = (serialized: string) => {
```
kof, line 33: sounds like this could use a test

istarkov: Done

packages/project-build/src/shared/graph-utils.test.ts
```diff
@@ -0,0 +1,148 @@
+// graph-utils.test.ts
```
TrySound, line 1: why have a filename in comment?

packages/project-build/src/shared/graph-utils.ts
```diff
@@ -0,0 +1,93 @@
+import type { Instance } from "@webstudio-is/sdk";
+
+type InstanceId = Instance["id"];
+
+// Depth-First Search (DFS) algorithm to find cycles in a directed graph
+export const findCycles = (
+  graph: Iterable<Pick<Instance, "id" | "children">>
+): InstanceId[][] => {
+  const adjacencyList: Record<InstanceId, InstanceId[]> = {};
```
TrySound, line 9: Maybe use map for it?

packages/project-build/src/shared/graph-utils.test.ts
istarkov: Auto generated test, 99% is written with vscode chat gpt

packages/project-build/src/shared/graph-utils.ts
istarkov: Not needed, we are just iterating over it

TrySound (approved):

packages/project-build/src/shared/graph-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    if (visited.has(nodeId)) {
+      return;
+    }
+
+    visited.add(nodeId);
+    path.push(nodeId);
+
+    for (const childId of adjacencyList[nodeId] || []) {
```
TrySound, line 36: Map is more type safe here

packages/project-build/src/shared/graph-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+      .map((child) => child.value);
+  }
+
+  const visited = new Set<string>();
+  const path: InstanceId[] = [];
+  const cycles: InstanceId[][] = [];
+
+  const dfs = (nodeId: string): void => {
+    if (path.includes(nodeId)) {
+      const cycleStart = path.indexOf(nodeId);
```
TrySound, line 24: could be single path.indexOf instead of both includes and insteadOf

istarkov: Usually in this algorithms here is not inlined code but some `normalizePath` useless for our purposes.

PR #4117, opened 2024-09-13
experimental: add css custom properties support to advanced styles

TrySound:

- added custom properties autocomplete in advanced styles
- fixed unit value detecting to not break var()
- show inherited custom properties in advanced styles
- dropped "types" from properties meta
- parse not top level var(), for example inside layers or function as unparsed for now

image
Screenshot 2024-09-12 at 21 53 38
Screenshot 2024-09-12 at 21 53 47

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
@@ -203,6 +203,16 @@ const AdvancedPropertyLabel = ({ property }: { property: StyleProperty }) => {
   );
 };
 
+const $customProperties = computed($definedStyles, (definedStyles) => {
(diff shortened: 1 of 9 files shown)


kof:
just to make sure, this is allowed: `var(--test) px` ? because it works with 'width' and I am not sure why :)

image

TrySound:
Yeah, it's valid in browsers 🤷

kof:
- [ ] when typing "var("  and going further with  "var(--test" it would be good to still autocomplete, basically ignoring everything but the variable name, because when I have started typing var I won't get any completion and if I had a value inserted and now delete something - again I would expect it to auto complete

kof (approved):
found only the autocompletion issue, otherwise works well

TrySound:
Will improve autocomplete in separate PR

PR #4118, opened 2024-09-13
refactor: migrate transform panels to new style engine

TrySound:

Switched translate, scale, rotate and skew panels to new style engine. Simplified style updates.

image
image
image
image

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/transforms/transform-scale.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/transforms/transform-scale.tsx
@@ -1,3 +1,4 @@
+import { useState } from "react";
 import {
   Box,
   EnhancedTooltip,
@@ -6,15 +7,7 @@ import {
   SmallToggleButton,
   theme,
 } from "@webstudio-is/design-system";
-import {
-  StyleValue,
-  toValue,
-  UnitValue,
-  type StyleProperty,
-  type Unit,
-} from "@webstudio-is/css-engine";
-import { CssValueInput } from "../../shared/css-value-input";
-import { type TransformPanelProps } from "./transform-utils";
+import type { StyleValue, StyleProperty } from "@webstudio-is/css-engine";
 import {
   BorderRadiusIcon,
   XAxisIcon,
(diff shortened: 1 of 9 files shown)


apps/builder/app/builder/features/style-panel/sections/transforms/transform-scale.tsx
```diff
@@ -266,34 +177,29 @@ export const ScalePanelContent = (props: TransformPanelProps) => {
             description={propertySyntaxes.scaleZ}
           />
           <CssValueInput
-            key="scaleZ"
             styleSource="local"
-            property={fakeProperty}
+            property={property}
```
JayaKrishnaNamburu, line 181: Screenshot 2024-09-19 at 11 12 51 AM


sclae-none-bug

We need to set the `fakeProperty` here, or else it allowing user to set keywords. But when used in `scaleX` and `scaleY` keywords can't be set of just one axis

JayaKrishnaNamburu (approved):
Looking good, and working as expected. Just a small change for `property` update for scale.

apps/builder/app/builder/features/style-panel/sections/transforms/transform-scale.tsx
TrySound: Hm, this is not a keyword. It's some weird unit in there.

JayaKrishnaNamburu: It comes from here 
https://developer.mozilla.org/en-US/docs/Web/CSS/scale#syntax

We are setting property as `scale` directly and so the `CSSValueInput` is picking the keywords available for the property directly. But in reality we don't allow to set just one value for the scale from UI right. So used `opacity` before which has the same syntax for individual values and no keywords like `none` as such.
https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#syntax

TrySound: I mean we do not pass keywords list of css input. And you are pointing to unit select. For some reason it shows keyword, but it shouldn't.

JayaKrishnaNamburu: It comes from here
https://github.com/webstudio-is/webstudio/blob/main/apps/builder/app/builder/features/style-panel/shared/css-value-input/unit-select.tsx#L50

`buildOptions` is taking keywords from the auto-generated properties list to show under it. So even if the passed `keywords` list is empty. It picks from the `UnitSelector`.

TrySound: Killed it

Issue #3448, opened 2024-05-29
Gradient UI for generating gradients


kof:
As a user I want to be able to generate a gradient from builder directly without going to other tools

https://www.figma.com/design/sfCE7iLS0k25qCxiifQNLE/%F0%9F%93%9A-Webstudio-Library?node-id=6563-7633&t=msAWXlTq6xeq2tNi-0

JayaKrishnaNamburu:
Hey @kof any more info on this,
- Like how this new selector should work, (i.e there will be a another selector which picks the colour and this just slides over to pick the inputs for the points).
- It should have a basic support for linear gradients or are we aiming at setting the type of gradient too.
- currently the design has just one slider, how that should work ? How can users select multiple colours and interact by changing them.
Any example on how it can work would help to kickstart this one 👍

kof:
Here is the full ui design link, the other one is only components https://www.figma.com/design/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=11466-66434&t=XvqeVFmbw4qwXuAQ-4

johnsicili:
A consideration: Users want to overlay a background with a semi transparent color, though CSS does not allow a background color overlay like this. They do however allow background-image. So what Webflow does is they allow users to select a _single_ color and behind the scenes they generate a "gradient" though it's just that one color listed twice.

I don't know what the gradient UI is going to look like or the proper solution, but wanted to share this use case.

https://github.com/user-attachments/assets/eac4d8af-071c-487b-bf68-ac5dacbdd341

JayaKrishnaNamburu:
@johnsicili from the design, i am thinking you are referring to this one right

https://www.figma.com/design/xCBegXEWxROLqA1Y31z2Xo/%F0%9F%93%96-Webstudio-Design-Docs?node-id=11466-66789&node-type=frame&t=zq93inXImLUZxySW-0

johnsicili:
Correct!

bryanbemis:
Here's a great reference for generating vivid color gradients in different color spaces, while avoiding washed-out areas and sharp banding artifacts: [https://www.learnui.design/tools/gradient-generator.html](https://www.learnui.design/tools/gradient-generator.html).
- The tool linked above calculates the interpolations automatically.
[...]

PR #4131, opened 2024-09-18
experimental: parse `linear-gradient` and extract `stops`, `angle`, `side-or-corner` and `hints` from the gradient.

JayaKrishnaNamburu:
## Description
This PR is to kick off the #3448 track. First of everything we need to build the gradient selector/indicator
image

Which should be able to parse the gradient and pick the `stops` and `colors` and `hints` from it. 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-data/src/property-parsers/linear-gradient.ts
```diff
@@ -0,0 +1,213 @@
...
+      type: "rgb",
+      r: value.r,
+      g: value.g,
+      b: value.b,
+      alpha: value.a,
+    };
+  }
+};
+
+export const reconstructLinearGradient = (parsed: ParsedGradient): string => {
```
TrySound, line 197: Why not just use toValue with original StyleValue?

packages/css-data/src/property-parsers/linear-gradient.ts
```diff
@@ -0,0 +1,213 @@
...
+  node: csstree.FunctionNode | csstree.Identifier | csstree.Hash
+): RgbValue | undefined => {
+  let color: string;
+  if (node.type === "Function") {
+    color = csstree.generate(node);
+  } else if (node.type === "Identifier") {
+    color = node.name;
+  } else {
+    color = `#${node.value}`;
+  }
```
TrySound, line 181: Can this be replaced with `const color = generate(node)`?

packages/css-data/src/property-parsers/linear-gradient.ts
```diff
@@ -0,0 +1,213 @@
...
+          (item.type === "Operator" && item.value === ",") ||
+          node.children.last === item
+        ) {
+          // If the gradientParts lenght is 1, then we need to check if it is angle or not.
+          // If it's angle, then the value is related to <angle> or else it is a <color-hint>.
+          if (gradientParts.length === 1) {
+            if (isAngle(gradientParts[0]) === true) {
+              angle = {
+                type: "unit",
+                value: parseFloat(gradientParts[0].value),
```
TrySound, line 77: Number.parseFloat pls

packages/css-data/src/property-parsers/linear-gradient.ts
```diff
@@ -0,0 +1,213 @@
...
+                  position.type === "Dimension")
+                  ? {
+                      type: "unit",
+                      value: parseFloat(position.value),
+                      unit:
+                        position.type === "Percentage"
+                          ? "%"
+                          : (position.unit as Unit),
+                    }
+                  : undefined,
```
TrySound, line 127: Seems like you could you a separate function for this repeated part

packages/css-data/src/property-parsers/linear-gradient.test.ts
```diff
@@ -0,0 +1,277 @@
...
+      parseLinearGradient(
+        "linear-gradient(to right, red 20%, orange 20% 40%, yellow 40% 60%, green 60% 80%, blue 80% )"
+      )
+    ).toEqual({
+      angle: undefined,
+      sideOrCorner: {
+        type: "keyword",
+        value: "to right",
+      },
+      stops: [
```
TrySound, line 173: I usually collapse long data like this into single line and let prettier do the magic

packages/css-data/src/property-parsers/linear-gradient.ts
JayaKrishnaNamburu: I think it's best to save gradients as `unparsed` in the data-layer. Because they already come under as a `tuple` for background.
And secondary we parse the `gradient` just for showing it in the component. So, it's best to read and save it back as `unparsed` itself. And to `reconstruct` is just to save it back as `unparsed to the dat layer. 

The actual implementation can be differ when we start building the feature.  Right now, i feel keeping it as `ParsedGradient` for the component works best to keep the `stops` and `hints` and `angle` at disposal.

packages/css-data/src/property-parsers/linear-gradient.ts
```diff
@@ -0,0 +1,200 @@
...
+            const value = gradientParts
+              .map((item) => csstree.generate(item))
+              .join(" ");
+            sideOrCorner = { type: "keyword", value };
+          }
+
+          // if there is a color-stop in the gradientParts, then we need to parse it for position and hint.
+          const colorStop = gradientParts.find(isColorStop);
+          if (colorStop !== undefined) {
+            const [_, position, hint] = gradientParts;
```
TrySound, line 101: I would still name _ variable for clarity

TrySound (approved):

PR #4132, opened 2024-09-18
feat: make pages tree accessible from keyboard

TrySound:
Added new implementation for tree component in design system which has out of the box keyboard support without tying with dnd.

Now only home page is tabbable, all others are accessible with arrows. Right arrow can both expand folder and access "settings" button. Both left and right arrows are able to travel between folders/pages.

https://github.com/user-attachments/assets/8d3264ca-1069-4168-97d9-afd7bdd4759e


kof:
- [ ] folder when empty doens't need arrow for collapsing

kof (approved):
seems to work fine, only one minor bug

johnsicili (approved):
Nice!

PR #4134, opened 2024-09-18
feat: auto-detect font weight based on selected font family

kof:
## Description

I rewrote the font weight display logiс in the following way:

1. We now detect for each font  which weights are supported
2. Unsupported weights are lighter gray in the list, but user can still choose them, we just hint that the weight may not be supported, because with font stacks, you can have different fonts supporting different weights, we are testing the one user can see right now

## Steps for reproduction

1. try switching between different stacks and see that when hovering the font width that is light gray, its not affecting canvas, meaning our detection was correct and that weight is not supported
3. same thing for any custom font

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili (changes requested):
I don't think it's working properly for `inherit`

1. Add heading
2. Set family to inherit
3. See weight only has 400
4. Go to body see weight has 400 + 600

Otherwise looks good. Tested with variable fonts too

https://github.com/user-attachments/assets/0481e224-2a64-4d6b-9d11-19dca9f136da

kof:
its probably best to not try to detect for inherit and show them all, because we really don't know what its gonna be, since it can change in any part of the tree

apps/builder/app/canvas/canvas.tsx
```diff
@@ -143,6 +145,7 @@ const DesignMode = () => {
   useEffect(subscribeInstanceSelection, []);
   useEffect(subscribeInstanceHovering, []);
   useEffect(subscribeInspectorEdits, []);
+  useMount(subscribeFondLoadingDone({ signal }));
```
TrySound, line 148: what with signal?

johnsicili (approved):
Looks good well done

apps/builder/app/canvas/canvas.tsx
kof: done

apps/builder/app/canvas/canvas.tsx
```diff
@@ -145,7 +145,10 @@ const DesignMode = () => {
   useEffect(subscribeInstanceSelection, []);
   useEffect(subscribeInstanceHovering, []);
   useEffect(subscribeInspectorEdits, []);
-  useMount(subscribeFondLoadingDone({ signal }));
+  useMount(subscribeFontLoadingDone({ signal: abortController.signal }));
+  useEffect(() => () => {
+    abortController.abort();
```
TrySound, line 150: Ok, so you pass one abort c0ntroller signal and abort another? See, you recreate it on every render. You can unsubcribe in useMount callback. Do not expose this just to use abort controller.

TrySound: We use subscribe(): unsubscribe pattern everywhere, why change?

kof: This way you can pass the same signal to all and reuse the abort controller. This is a new pattern we discussed with Ivan, since we learned about abort controller. Seems like a much better option than just returning a callback. If you want to discuss this, lets do in dm

kof: Oh you are right about where I create the controller, it should be only created once, fixing it

apps/builder/app/canvas/canvas.tsx
```diff
@@ -130,24 +130,25 @@ const useElementsTree = (
...
   useScrollNewInstanceIntoView();
   useSelectedInstance();
-  useEffect(updateCollaborativeInstanceRect, []);
-  useEffect(subscribeInstanceSelection, []);
-  useEffect(subscribeInstanceHovering, []);
-  useEffect(subscribeInspectorEdits, []);
-  useMount(subscribeFontLoadingDone({ signal: abortController.signal }));
-  useEffect(() => () => {
-    abortController.abort();
+  useMount(() => {
```
TrySound, line 142: just use explicit useEffect(() => {}, []). No dependencies from react here.

TrySound (approved):

aleclarson:
Did this break variable fonts with `wght` axis? I can't set arbitrary `font-weight`.

apps/builder/app/builder/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
```diff
@@ -1,45 +0,0 @@
...
-        wght: { name: "wght", min: 100, default: 400, max: 1000 },
-      },
-    } as const;
-    const asset = createServerAsset(meta);
-    expect(isSupportedFontWeight(asset, "1", meta.family)).toBe(false);
-    expect(isSupportedFontWeight(asset, "100", meta.family)).toBe(true);
-    expect(isSupportedFontWeight(asset, "500", meta.family)).toBe(true);
-    expect(isSupportedFontWeight(asset, "1000", meta.family)).toBe(true);
-    expect(isSupportedFontWeight(asset, "1001", meta.family)).toBe(false);
-  });
```
aleclarson, line 44: did this test get moved somewhere or just deleted?

kof:
> Did this break variable fonts with wght axis? I can't set arbitrary font-weight.

@aleclarson does it not correctly detect the available font weights? or did you want to use some font weight that's not in the list at all, like 450 or something?

apps/builder/app/builder/features/style-panel/controls/font-weight/is-supported-font-weight.test.ts
kof: isSupportedFontWeight function was removed in favor of a new approach which tests for font support instead of using font data

aleclarson:
> > Did this break variable fonts with wght axis? I can't set arbitrary font-weight.
> 
> @aleclarson does it not correctly detect the available font weights? or did you want to use some font weight that's not in the list at all, like 450 or something?

Screenshot 2025-04-02 at 14 32 48

It should be a number input, not a dropdown, IIUC.

kof:
> Yes, I want to use an arbitrary value like 450.

We basically didn't account for this. Can you educate us on a practical use case?

kof:
Got convinced that enabling editing custom font weight for variable fonts is important! Now the question if we should check if the font is variable and only then render a combobox or always render it.

Downside of the combobox is smaller click surface on the icon vs select.

PR #4153, opened 2024-09-23
fix: allow user to define a custom font stack

kof:
## Description

Previously defining a custom stack - comma separated list of fonts, was broken, we only used the first font

## Steps for reproduction

1. comma separated fonts, more than one
2. see that its rendered in dev tools and that we don't add a default fallback to it

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-engine/src/core/to-value.ts
+++ b/packages/css-engine/src/core/to-value.ts
@@ -6,10 +6,12 @@ export type TransformValue = (styleValue: StyleValue) => undefined | StyleValue;
 
 const fallbackTransform: TransformValue = (styleValue) => {
   if (styleValue.type === "fontFamily") {
-    const fonts = SYSTEM_FONTS.get(styleValue.value[0])?.stack ?? [
-      styleValue.value[0],
-      DEFAULT_FONT_FALLBACK,
-    ];
+    const fallback =
+      styleValue.value.length > 1
+        ? // Its a custom stack, we won't add a fallback
+          styleValue.value
+        : [...styleValue.value, DEFAULT_FONT_FALLBACK];
+    const fonts = SYSTEM_FONTS.get(styleValue.value[0])?.stack ?? fallback;
     const value = Array.from(new Set(fonts));
 
     return {
--- a/packages/css-engine/src/core/to-value.ts
+++ b/packages/css-engine/src/core/to-value.ts
@@ -9,34 +9,23 @@ const fallbackTransform: TransformValue = (styleValue) => {
     return;
   }
 
+  // By default we assume its a custom font stack.
+  let { value } = styleValue;
+
   // Shouldn't be possible, but just in case.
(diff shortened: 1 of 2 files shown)


packages/css-engine/src/core/to-value.ts
```diff
@@ -6,10 +6,12 @@ export type TransformValue = (styleValue: StyleValue) => undefined | StyleValue;
...
   if (styleValue.type === "fontFamily") {
-    const fonts = SYSTEM_FONTS.get(styleValue.value[0])?.stack ?? [
-      styleValue.value[0],
-      DEFAULT_FONT_FALLBACK,
-    ];
+    const fallback =
+      styleValue.value.length > 1
+        ? // Its a custom stack, we won't add a fallback
+          styleValue.value
+        : [...styleValue.value, DEFAULT_FONT_FALLBACK];
```
TrySound, line 13: Honestly not a fan of integrating font stack into normal font family values. Makes it impossible to use actual font with such name.

Would change this logic this way
1. use provided value
2. if value is singular try to expand into stack
3. add default fallback if no generic-family in the end (https://developer.mozilla.org/en-US/docs/Web/CSS/font-family)

kof: Basically its either a stack name or the stack itself.

kof: Ok so what you are saying is that we should check first the length of the input, if its 1 item, then we check if its a stack name, if its more than one item, we treat it as a stack. If its 1 item but doesn't match a stack name - its a singular font family name without a fallback, so we add a default fallback to it.

kof: this is exactly what I am doing there, but the logic is written in a way that's too unobvious, going to rewrite it so that this is clear

TrySound (approved):

packages/css-engine/src/core/to-value.ts
```diff
@@ -5,18 +5,39 @@ import type { StyleValue } from "../schema";
...
+      return {
+        type: "fontFamily",
+        value: stack,
+      };
+    }
+    // It's a specific font family.
     return {
       type: "fontFamily",
-      value,
+      value: [styleValue.value[0], DEFAULT_FONT_FALLBACK],
```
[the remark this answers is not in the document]
kof: why is thisa problem?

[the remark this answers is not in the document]
TrySound: Maybe not a problem
```
font-family: sans-serif, sans-serif;
```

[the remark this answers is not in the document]
kof: oh, yes I totally forgot to call Set and make those unique, adding a test

PR #4159, opened 2024-09-25
experimental: add gradient controller to update color stops using ui

JayaKrishnaNamburu:
The `GradientControl` is coming along. There are a lot of edge-cases when it comes to gradients and how lenient browsers can be with them. So, documenting and implementing them alone. This is still a in-progress PR. But if you want to give a try, you can check the storybook. 

Screenshot 2024-09-25 at 10 32 47 AM
Screenshot 2024-09-25 at 10 32 43 AM
Screenshot 2024-09-25 at 10 32 38 AM
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-control.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-control.tsx
@@ -1,0 +1,135 @@
+import { toValue, UnitValue } from "@webstudio-is/css-engine";
+import { Root, Range, Thumb, Track } from "@radix-ui/react-slider";
+import { useEffect, useState, useCallback } from "react";
+import {
+  reconstructLinearGradient,
+  type GradientStop,
+  type ParsedGradient,
+} from "@webstudio-is/css-data";
+import { styled, theme, Flex } from "@webstudio-is/design-system";
+
+type GradientControlProps = {
+  gradient: ParsedGradient;
+  onChange: (value: ParsedGradient) => void;
+};
+
+const defaultAngle: UnitValue = {
+  type: "unit",
+  value: 90,
+  unit: "deg",
+};
+
+export const GradientControl = (props: GradientControlProps) => {
+  const [stops, setStops] = useState<Array<GradientStop>>(props.gradient.stops);
+  const [selectedStop, setSelectedStop] = useState<number | undefined>();
+  const positions = stops.map((stop) => stop.position?.value) as number[];
+  const background = reconstructLinearGradient({
+    stops,
+    sideOrCorner: props.gradient.sideOrCorner,
+    angle: defaultAngle,
+  });
+
+  useEffect(() => {
+    const newStops: Array<GradientStop> = [];
+    for (const stop of props.gradient.stops || []) {
+      if (stop.color !== undefined && stop.position?.value !== undefined) {
+        newStops.push({
+          color: stop.color,
+          position: stop.position,
+        });
+      }
+    }
+    setStops(newStops);
+  }, [props.gradient]);
+
+  const handleValueChange = useCallback(
+    (newPositions: number[]) => {
+      const newStops: GradientStop[] = stops.map((stop, index) => ({
+        ...stop,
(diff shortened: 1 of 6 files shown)


apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-control.tsx
```diff
@@ -0,0 +1,171 @@
+import { toValue, UnitValue } from "@webstudio-is/css-engine";
```
kof, line 1: If we are building a lower level component, without any application logic, then it should go into design system pacakge

JayaKrishnaNamburu: Currently in the `design-system` package we are not using `css-engine` pacakge. Is it ok if i add it there ? I am using types from there and it's the only reason i added it into builder.

kof: I would say you can decouple the things you are using from css engine, seems very minimal, but curious what @TrySound would say

apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-control.tsx
```diff
@@ -0,0 +1,375 @@
+import { toValue, UnitValue, type RgbValue } from "@webstudio-is/css-engine";
+import { Root, Range, Thumb, Track } from "@radix-ui/react-slider";
```
TrySound, line 2: Let's use Slider instead of Root for clarity

TrySound:
I like how gradient style tries to balance new dots instead of manually centering, maybe not needed for our case
https://github.com/user-attachments/assets/a339afc6-6d18-4597-9eb0-518dcc03b52e

TrySound:
I would add outline for focus-visible only
image

TrySound:
I instinctively do alt+click to delete thumbs

TrySound:
I would do color picker as popover, tooltip feels super unpredictable

kof:
> I instinctively do alt+click to delete thumbs

I think its not a bad idea for us to support this, since our users know this combo

TrySound (approved):

kof:
Total list of issues:
- [x] click on a stop handle shouldn't move it 
- [x] click on track in any place should be adding a stop, not move the selected stop, not use an artificial minimal distance to the previous stop
- [ ] simple linear-gradient(red, red) should work
- [x] tooltip "Add color stop" missing
- [x] neither figma nor webflow has hints, why do we need them? I don't know.

JayaKrishnaNamburu:
> click on track in any place should be adding a stop, not move the selected stop, not use an artificial minimal distance to the previous stop

Looks like this is how all the sliders behave too. Its add for me too while working with gradient selector. But, let's see react-aria. Even it moves the closest thumb, when we click somewhere on the range
https://react-spectrum.adobe.com/react-aria/Slider.html#reusable-wrappers

Do you think, it's the actual standard spec and trying to prevent the movement is anti pattern ?

kof:
@JayaKrishnaNamburu idk if preventing the movement is supported from the spec, like using some preventDefault, but it doesn't really matter, because in our case its not really a slider, we are just trying to see if slider gets us the UX we need to avoid reimplementing accessibility ourselves.  So the question is really - can you hide the track or not let it listen to clicks in some other way or not

JayaKrishnaNamburu:
> So the question is really - can you hide the track or not let it listen to clicks in some other way or not
Using radix, if we are doing `preventDefault` for clicks. It's avoiding the thumb movement altogether. But i see the browser based `input[range]` doesn't have these type of weird behaviours. So, i will give a try with that this weekend. Ranger slider might not be the best choice, as all of them are moving on clicks. Or either stops to move if prevented for clicks.

kof:
yeah, but you already have your own click position logic, so why not hide the range, avoid radix for these, and based on where the click happened just add the new stop, click - your own logic, drag - slider logic

PR #4165, opened 2024-09-26
refactor: reimplement navigator tree with new component

TrySound:

What to test

- hover
- select
- hiding (with show button) item should be transparent
- slot should be purple
- hide slot fragment
- render collections
- enter should trigger edit instance text
- support editing instance label with cmd+e
- should prevent dragging rich text child
- should prevent dragging non detachable instances (accordion header)
- should support reparenting after the last child


packages/design-system/src/components/tree.tsx
```diff
@@ -215,24 +215,45 @@ export type TreeDropTarget = {
...
+    };
+  }
+  if (treeDropTarget?.afterLevel !== undefined) {
+    return {
+      type: "reorder-below",
+      currentLevel: treeDropTarget.afterLevel,
+      indentPerLevel: BARS_GAP,
+    };
+  }
+  return null;
```
kof, line 256: is this old? should be using undefined, basically no return

packages/design-system/src/components/list-position-indicator.tsx
```diff
@@ -16,6 +16,7 @@ const OVERLAP = 2;
 
 const Container = styled(Box, {
   position: "absolute",
+  zIndex: 1,
```
kof, line 19: would isolation: isolate work here?

TrySound: I don't think so, I just need to get it higher than outline

johnsicili (changes requested):
Works really well!

1. Adding new nested items should be expanded by default (add box, add heading, can't see heading because box is collapsed). @kof fixed this IIRC when he was doing shortcuts and search for components.
2. Related to the first, if you click on an item that is nested, you can't see anything selected in the navigator. Same with hovering
3. When there are enough instances in the navigator to fill the viewport, and you add a new component, the new one is not scrolled into view in the navigator

TrySound:
@johnsicili
fixed 1 and 2 though, we never expanded on hover, should we?
Scrolling works for me, could you record a video pls?

johnsicili:
Didn't realize the hover was existing behavior. Idk what's right here so we can leave it.

Here is the no scroll... seems to happen on page load and if I select the parent via the canvas

https://github.com/user-attachments/assets/1b749bf2-6417-4419-8c2d-884e5719b19e

kof (approved):

TrySound:
Tree scroll seems like fixed. Though found we have a weird problem with auto scroll on canvas. Reproduced in main as well so not a blocker for this PR.

johnsicili (approved):
Nice! Honestly I think I like it better without smooth scroll. It's faster

PR #4175, opened 2024-09-27
experimental: add :root support

TrySound:

Now we have a way to change rem and define global css variables

- show root node in navigator behind flag
- show style panel when root is selected
- inherit styles from root
- show advanced styles defined on root in descendents
- show browser styles for root (`html { display: block }`)
- show initial values instead of `inherit` in root typography
- render root styles on canvas
- property label shows root as the source

Note: generated sites are not supported yet
Screenshot 2024-09-27 at 19 24 41

Screenshot 2024-09-27 at 19 09 51

### DIFF
--- a/apps/builder/app/builder/features/inspector/inspector.tsx
+++ b/apps/builder/app/builder/features/inspector/inspector.tsx
@@ -2,6 +2,7 @@ import { useRef } from "react";
 import { computed } from "nanostores";
 import { useStore } from "@nanostores/react";
 import type { Instance } from "@webstudio-is/sdk";
+import { rootComponent } from "@webstudio-is/react-sdk";
 import {
   theme,
   PanelTabs,
@@ -103,7 +104,9 @@ export const Inspector = ({ navigatorLayout }: InspectorProps) => {
 
   const availablePanels = {
     style: documentType === "html" && (meta?.stylable ?? true),
-    settings: true,
+    // @todo hide root component settings until
+    // global data sources are implemented
+    settings: selectedInstance.component !== rootComponent,
   };
 
   return (
@@ -119,7 +122,7 @@ export const Inspector = ({ navigatorLayout }: InspectorProps) => {
             value={
               availablePanels[activeInspectorPanel]
                 ? activeInspectorPanel
-                : "settings"
+                : Object.keys(availablePanels)[0]
             }
             onValueChange={(panel) => {
               $activeInspectorPanel.set(panel as keyof typeof availablePanels);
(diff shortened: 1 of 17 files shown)


johnsicili (changes requested):
Amazing work!!!

Nothing big here. Decide what you think is worth fixing.

1. Per DMs, would be nice for Body to not have default styles like font size and others too like line height so we can control these values globally.
2. When scrubbing values on root, the canvas doesn't reflect the changes until you blur
3. When selecting root and trying to add a component, nothing happens, not even an error. It's not a big deal, but maybe we should do something in this case
4. Related to previous PR, clicking and dragging Body shows the dragging UI, though Body isn't actually moved. Root behaves correctly.

johnsicili (approved):
Changes look great! This is exciting

Last thing for a separate PR is to discuss with @kof about setting default styles on :root instead of Body so users can define font size, line-height, and font family globally on the root without Body overriding because specificity. That's only from the user perspective, I don't have any insights on the proper/technical way of doing this :)

kof:
Questions:
1. Why are we calling it :root in the navigator, :root is a css selector, in html, it means HTML tag. It would be consistent with body and other things to call it html
1. Same in the tooltip, :root vs html
1. Nesting level: on one hand adding a persistent html root that always takes space is not optimal, on the other hand, its logically the highest parent .... not sure what to do here
1. Should realistically user be restricted what css properties are allowed on html? I just added padding and its so weird to see it but idk, maybe I am just not used to it and this is all goo
1. Noticing the variable name on the root looks strange `var(--\37 qJPVZc6ns5sbZ1kOWx_b--paddingTop, 56px)` , what's the "\37 "
1. Just noticing that presets like font family are on body, so setting font family on root doesn't make them inherit it. Would it make sense to have the presets for all those inheritable properties on root instead? I am not sure

Todos:

- Title in the variable tooltip could be just "CSS Variable"
image

TrySound:
1. Wanted to separate it somehow from other instances because this one is global for all pages.
3. Can be fixed before release. Not a problem. For now root behind the flag.
4. font-size is configures rem here
6. maybe, but in another PR

TrySound:
> Noticing the variable name on the root looks strange var(--\37 qJPVZc6ns5sbZ1kOWx_b--paddingTop, 56px) , what's the "\37 "

I guess it is encoded `-` from nanoid, shouldn't be a problem

kof:
> 1. Wanted to separate it somehow from other instances because this one is global for all pages.

Yes, definitely good idea, since its not just like other tags, its global, same thing on every page. This is a new concept in webstudio.

> 3. font-size configures rem here

Def. font-size is a unique thing on the rootl. I wonder what we should do about the rest. Is there a list of css properties that's even considered legal on the root? are they all just supported officially? I have no idea for example why padding works there, it is a little bit weird. I haven't looked into the spec yet to see if there is anything that explains how hmtl tag is being treated and what we need to disable or enable.

It definitely doesn't feel like a regular tag. Maybe we need to only have the advanced panel on the root and only enable a subset of properties.

Having Advanced panel only has the benefit of having a lot of space to manage lots of global variables.

PR #4177, opened 2024-09-28
experimental: generate root styles for published sites

TrySound:

- render layout hack (display: grid; min-height: 100%) as root preset styles
- generate root user styles on published sites
- moved typography preset from body to the root
- bumped a couple of deps

Screenshot 2024-09-28 at 23 25 08


johnsicili (changes requested):
Preset root styles (at least grid and height) don't show in the Style Panel until clicking on the body/possibly any instance that has styles.

1. Refresh page
2. Check root styles see no grid
3. Click body
4. Click root see grid


https://github.com/user-attachments/assets/54f1ce64-75ba-4e89-8d02-ae577382f3c4

TrySound:
@johnsicili was a separate bug Ivan fixed, rebased. Now should work as expected.

johnsicili (approved):

istarkov (approved):

PR #4193, opened 2024-10-01
experimental: support copying root styles from marketplace

TrySound:

Now we have a way to deliver custom properties to user projects.

### DIFF
--- a/apps/builder/app/shared/instance-utils.test.tsx
+++ b/apps/builder/app/shared/instance-utils.test.tsx
@@ -3177,15 +3177,15 @@ describe("insert webstudio fragment copy", () => {
 
 describe("support copy paste of :root styles", () => {
   const css = (strings: TemplateStringsArray, ...values: string[]) => {
-    let css = "";
-    strings.forEach((string, i) => {
-      css += string + (values[i] || "");
+    let cssString = "";
+    strings.forEach((string, index) => {
+      cssString += string + (values[index] ?? "");
     });
     const styleSourceSelections: StyleSourceSelections = new Map();
     const styleSources: StyleSources = new Map();
     const styles: Styles = new Map();
     const parsed = mapGroupBy(
-      parseCss(css),
+      parseCss(cssString),
       (parsedStyleDecl) => parsedStyleDecl.selector
     );
     for (const [selector, parsedStyles] of parsed) {
--- a/apps/builder/app/shared/instance-utils.test.tsx
+++ b/apps/builder/app/shared/instance-utils.test.tsx
@@ -5,6 +5,10 @@ import {
   collectionComponent,
   type WsComponentMeta,
 } from "@webstudio-is/react-sdk";
+import type { Project } from "@webstudio-is/project";
+import { createDefaultPages } from "@webstudio-is/project-build";
+import { $, renderJsx } from "@webstudio-is/sdk/testing";
+import { parseCss } from "@webstudio-is/css-data";
 import * as defaultMetas from "@webstudio-is/sdk-components-react/metas";
 import type {
   Asset,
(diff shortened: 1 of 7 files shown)


istarkov (approved):

apps/builder/app/shared/instance-utils.test.tsx
```diff
@@ -3165,6 +3171,101 @@ describe("insert webstudio fragment copy", () => {
       availableDataSources: new Set(),
     });
     expect(data.instances.size).toEqual(4);
-    expect(newInstanceIds.size).toEqual(4);
+    expect(newInstanceIds.size).toEqual(5);
+  });
+});
+
+describe("support copy paste of :root styles", () => {
+  const css = (strings: TemplateStringsArray, ...values: string[]) => {
```
kof, line 3179: this test looks quite complex for a test, looks like we are missing some abstraction

apps/builder/app/shared/instance-utils.test.tsx
```diff
@@ -3165,6 +3171,101 @@ describe("insert webstudio fragment copy", () => {
...
+    let css = "";
+    strings.forEach((string, i) => {
```
kof, line 3181: please
```suggestion
    strings.forEach((string, index) => {
```

apps/builder/app/shared/instance-utils.test.tsx
```diff
@@ -3165,6 +3171,101 @@ describe("insert webstudio fragment copy", () => {
...
     });
     expect(data.instances.size).toEqual(4);
-    expect(newInstanceIds.size).toEqual(4);
+    expect(newInstanceIds.size).toEqual(5);
+  });
+});
+
+describe("support copy paste of :root styles", () => {
+  const css = (strings: TemplateStringsArray, ...values: string[]) => {
+    let css = "";
```
kof, line 3180: having function called css and this variable, is not very nice for the reader

apps/builder/app/shared/instance-utils.test.tsx
```diff
@@ -3165,6 +3171,101 @@ describe("insert webstudio fragment copy", () => {
...
+      }
+    }
+    return { styleSourceSelections, styleSources, styles };
+  };
+
+  test("should add :root local styles", () => {
+    const oldProject = getWebstudioDataStub({
+      ...renderJsx(<$.Body ws:id="oldProjectBodyId"></$.Body>),
+      ...css`
+        :root__oldprojectlocalid {
```
kof, line 3215: hmm, why do we need this second part starting with underscores, I wonder if there is a better way, don't want this to be user facing at some point

TrySound: doesn't matter, this is figure

apps/builder/app/shared/instance-utils.test.tsx
TrySound: this is temporary to make actual test more readable

apps/builder/app/shared/instance-utils.test.tsx
TrySound: this is an abstraction around the logic I need to test.
Maybe it will go to module level in the future, didn't want to spend time on refactoring

kof: I think we have now 2 very similar functions, another one i wrote in plugin-webflow.test.tsx called toCss and inlined it in the file with the same reasoning as you now

kof: problem with these is that next time we will create another one and another one, because we don't know whats inside all individual modules unless its used everywhere

TrySound: Anyway, test is simpler with this. Will refactor the rest of tests some time later. Maybe find the right abstraction. But not in context of this PR.

kof (approved):

PR #4200, opened 2024-10-02
experimental: allow pasting styles into style source

TrySound:

He added support for paste style declarations into style source or state. It will let us create collections of custom properties like open props with a few clicks.

The downside is browser prompt with permission confirmation because we use "unsafe" navigator.clipboard.readText() instead of paste event.

https://github.com/user-attachments/assets/8e0153b6-5b98-4f07-bb2c-aac6bc51c6d3

### DIFF
--- a/apps/builder/app/builder/features/style-panel/style-source-section.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source-section.tsx
@@ -11,6 +11,7 @@ import {
   type StyleSources,
   getStyleDeclKey,
 } from "@webstudio-is/sdk";
+import { parseCss } from "@webstudio-is/css-data";
 import {
   Flex,
   Dialog,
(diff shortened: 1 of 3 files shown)


kof:
This is a good start, lets decide what we do if user pastes an entire css rule

TrySound:
Nothing, this is for us. Behind the flag.

TrySound:
Related yes, though I'm not solving that issue but just added a way for us to test big amount of variables to improve ui.

johnsicili (approved):
Quick test for internal use and looks really useful. If it's a quick change, I'd remove the paste option from all other instances except Global Root as it only seems to work there.

Now it's more clear on what things we can do to improve the UI to accommodate design systems with vars such as

1. maybe expand option like we have for code
2. search forsure (the style sources _looks_ like a search)
3. color picker or at least color preview
4. might be nice to only see css vars rather than mixing in advanced too 🤷 
5. I don't love the wrapping when the var is long, but maybe expand window could help
6. Might make sense to add a copy option for a single var to use... I would need to build a site using this to see my workflow but I could see me forgetting my vars, referencing the global list, then wanting to copy it. Though maybe autocomplete will solve this

TrySound:
let's put this into css variables issue

johnsicili:
Done. Added to an "Ideas" section there

PR #4205, opened 2024-10-03
experimental: show color picker in advanced panel

TrySound:

Now users can tinker colors in css variables

Screenshot 2024-10-03 at 14 47 26


kof (approved):
looks like we can delete color-control now?

kof (changes requested):
spacing seems broken 
image

TrySound:
> looks like we can delete color-control now?

No, here I made advanced value an adaptive control. Color control restricts color only.

Works locally :)

image

kof:
this component has size property with 2 versions, style panel in general uses a bigger one than advanced panel

kof:
image
image

kof:
in advanced, the picker button needs to be smaller, so that it has 2px margin

PR #4206, opened 2024-10-03
experimental: show color picker in advanced panel

TrySound:

Now users can tinker colors in css variables

Screenshot 2024-10-03 at 14 47 26

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
@@ -1,3 +1,4 @@
+import { colord } from "colord";
 import { useEffect, useMemo, useRef, useState, type ReactNode } from "react";
 import { computed } from "nanostores";
 import { matchSorter } from "match-sorter";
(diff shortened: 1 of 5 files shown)


kof (approved):

johnsicili (approved):
It's amazing! All small stuff. Do whatever you think is easy or important.

1. Color keywords like "blue" are no longer allowed (nbd)
no-colors-keywords

2. Scrub is allowed on color picker
I recall talking with @kof in the page about disallowing scrub on NaN and the solution was that he made it less sensitive. I think at a minimum it would be nice to disallow scrub on the color picker. If too much work right now lets do it post launch.

https://github.com/user-attachments/assets/f6f6fd52-ebe8-4e37-bbe3-ba459c14c875

3. When clearing out a color, the input is too small IMO

https://github.com/user-attachments/assets/e72c3658-ee80-4640-b2d1-b11481c907d2

4. When changing the color in the color picker, theres a delay in when it shows in the var

https://github.com/user-attachments/assets/a0fbeadb-3854-4697-a28d-c5eca6bec459

PR #4208, opened 2024-10-03
refactor: enable isolated declarations in icons

TrySound:
Enforced isolated declarations which requires annotating all inputs and outputs for exports like flow enforced some time ago.


kof (approved):

kof:
Just curious, what is this solving?

TrySound:
In a long perspective typescript performance

kof:
so that it doesn't have to analyze dependency graph if the export function has input/output typed? Is that already in-place?

TrySound:
I guess so. Not sure how in practice but at least it will let us use oxc to generate typings instantly. Btw since all outputs will be defined we won't need to have fake declaration generation while typechecking to check schema edge cases. So typecheck will be faster.

istarkov (approved):

PR #4212, opened 2024-10-03
experimental: support css variables in dedicated sections

TrySound:

Added support for entering and autocomplete of css variables to dedicated style sections like size and typography.

- color control
- text control

Also improved autocomplete to search with both var(--name) and --name.

Screenshot 2024-10-04 at 01 51 12

### DIFF
--- a/apps/builder/app/builder/features/style-panel/controls/color/color-control.tsx
+++ b/apps/builder/app/builder/features/style-panel/controls/color/color-control.tsx
@@ -1,12 +1,14 @@
+import { useStore } from "@nanostores/react";
 import type { StyleProperty } from "@webstudio-is/css-engine";
 import { ColorPicker } from "../../shared/color-picker";
 import { styleConfigByName } from "../../shared/configs";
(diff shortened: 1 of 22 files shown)


kof (approved):

johnsicili (changes requested):
OMG you're moving fast.

Looks good. Autocomplete is amazing! I searched `var--letter` and it still pulled up my letter spacing vars!

## Needs fixing
1. Border vars autocomplete not working (color, radius, width) though manual entry works. If this was not intended to be supported here let me know and I'll approve.

## Minor
1. Would be nice to have type checking so I can't use a color var in my font size field. I think browsers handle this well though so nbd.
2. I see this is for dedicated sections with color and text control so I guess it's known there is no support for things like transition duration and any of the layered css styles

TrySound:
Borders, easings, gradients will do separately. As for type checking, would leave it for after launch. Can be complicated.

johnsicili (approved):

PR #4213, opened 2024-10-04
experimental: support nested and attribute selectors

kof:
## Description

Needed to support all wf presets for fixing this https://github.com/webstudio-is/webstudio/pull/4155

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/css-data/src/parse-css.ts
+++ b/packages/css-data/src/parse-css.ts
@@ -161,6 +161,7 @@ export const parseCss = (css: string, options: ParserOptions = {}) => {
           if (node.type === "MediaQuery" && node.children.size > 1) {
             invalidBreakpoint = true;
           }
+          ``;
         },
       });
       const generated = csstree.generate(this.atrule.prelude);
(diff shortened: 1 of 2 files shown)


packages/css-data/src/parse-css.ts
```diff
@@ -161,50 +161,71 @@ export const parseCss = (css: string, options: ParserOptions = {}) => {
...
-          name = `:${nameNode.name}`;
-        } else {
-          continue;
+
+    for (const node of this.rule.prelude.children) {
+      if (node.type !== "Selector") {
+        continue;
+      }
+      let selector: Selector | undefined = undefined;
+      node.children.forEach((node) => {
```
TrySound, line 183: for/of

packages/css-data/src/parse-css.ts
```diff
@@ -161,50 +161,71 @@ export const parseCss = (css: string, options: ParserOptions = {}) => {
...
+        let name: string = "";
+        let state: string | undefined;
+        switch (node.type) {
+          case "TypeSelector":
+            name = node.name;
+            break;
+          case "ClassSelector":
+            // .a {} vs a.b {}
+            name = selector ? `.${node.name}` : node.name;
```
TrySound, line 192: I don't get it. You just put class as tag selector here

kof: in case its a single class, we don't set the dot, to make it look nicer as a token, but when its  a combo class like .a.b we turn this into a.b ... any better ideas?

kof: we are kinda converting regular selectors into token names, so with things like combinators and attribute selectors we can do anything we want, in the end its just a name, I guess it would be more straight forward to stay as close as possible to the original thing so that user can rename if necessary

TrySound: This is misleading, caz parser can be used functionally. It is not specific to webflow paste. Better trim dot when generate tokens specifically.

TrySound (approved):

PR #4214, opened 2024-10-04
fix: Unexpected end of JSON input in Webhook Form

### LINKED ISSUE #4050: Unexpected end of JSON input in Webhook Form
When sending a webhook form to https://tapeapp.com webhook, we get "Unexpected end of JSON input" error.

I assume we either expect a json but get something else or there are wrong headers being sent from the app. Repro in dm

https://test-2h51i.wstd.io/

submit the form, see network response

Either way we should tolerate wrong headers and wrong responses in the webhook. We basically only care if its some http error code or 200 and we don't really need to parse the response

istarkov:
## Description

closes #4050

This PR modifies the response handling logic to attempt parsing all resource data as JSON. If an error occurs during parsing, an error response is only returned if the Accept header is explicitly set to application/json. 

204 responses are skipped too.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk/src/resource-loader.ts
```diff
@@ -1,4 +1,5 @@
 import type { ResourceRequest } from "./schema/resources";
+import Negotiator from "negotiator";
```
TrySound, line 2: do we really need negotiator to check content-type?

TrySound (approved):

PR #4220, opened 2024-10-04
refactor: autocomplete variables lazily

TrySound:

Here improved combobox component to accept
options factory instead of ready to use array.
This function is invoked only when combobox is open.

As a result each of 500 css variables will no longer subscribe to available variables which should reduce scripting time and lagging.

Here's local test

Screenshot 2024-10-05 at 02 03 02

Screenshot 2024-10-05 at 02 05 16

Preview test

Screenshot 2024-10-05 at 02 23 58
Screenshot 2024-10-05 at 02 22 44


kof (approved):

kof:
This is amazing. Indeed there is tons of options, computing them all upfront is expensive

johnsicili (approved):

PR #4225, opened 2024-10-05
experimental: support css variables for borders

TrySound:

Added autocomplete for css variables in border width, color and radius. Sadly no support for blob and drawn radius because they require shorthand.

Screenshot 2024-10-06 at 01 26 28


johnsicili (approved):
Let's go!

PR #4226, opened 2024-10-06
experimental: support css variables in space section

TrySound:

Here migrated space section to new style engine and added support for autocompleting css variables. Rendering css variables was already in place.

Screenshot 2024-10-06 at 12 28 48

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/position/inset-control.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/position/inset-control.tsx
@@ -48,7 +48,6 @@ const Cell = ({
         isOpen={isPopoverOpen}
         property={property}
         onClose={onPopoverClose}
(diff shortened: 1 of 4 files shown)


johnsicili (approved):
1. Took me a bit to realize that hitting enter selects the autocomplete and then enter again saves it/closes it. It's the second enter that enables me to press shift + enter to apply it to all sides. I feel like it should all happen on the first enter but I assume this is just the behavior of that component.
2. Longer var names are just not ideal in this ui. Maybe there needs to be an option to have it looks like the border radius UI.. idk though as that just odd to have two UIs for it
Screenshot 2024-10-06 at 5 41 19 AM

kof:
Another option would be to actually show computed value in there, then showing var() when user clicks into the field

PR #4228, opened 2024-10-06
experimental: show computes values in tooltip when alt is pressed

TrySound:

One of the worst thing with our internal theme is I don't know what actual values used for example in theme.spacing[5] without going into theme file.

Here added info about computed value in property tooltip. It computes initial, inherit, unset, currentColor and css variables.

Screenshot 2024-10-06 at 13 25 32

### DIFF
--- a/apps/builder/app/builder/features/style-panel/property-label.tsx
+++ b/apps/builder/app/builder/features/style-panel/property-label.tsx
@@ -1,6 +1,7 @@
 import { useStore } from "@nanostores/react";
 import { useState, type ReactNode } from "react";
 import { AlertIcon, ResetIcon } from "@webstudio-is/icons";
+import { isFeatureEnabled } from "@webstudio-is/feature-flags";
 import {
   hyphenateProperty,
   toValue,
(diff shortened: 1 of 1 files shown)


kof:
Def. very needed but the question is how to show it nicer, more compact, few options I see:
- modifier key switching between defined and computed
- icon-button to switch between computed and defined
- switch to computed on hover over the value
- a shorter persistent notation: `width: inherit; (auto)`

TrySound:
> a shorter persistent notation: width: inherit; (auto)

Or something like this
```css
width: inherit; /* auto */
```

TrySound:
but would require wider tooltip

kof:
> > a shorter persistent notation: width: inherit; (auto)
> 
> Or something like this
> 
> ```css
> width: inherit; /* auto */
> ```

thought about it too, but since we are not using CSS syntax highlighter, we can just render a value in a different color without using comments syntax

But primarily I don't know if user wants to see both values at the same time or would be ok seeing them on hover or over a switch click

johnsicili (approved):
This is a common occurrence:

Screenshot 2024-10-05 at 7 44 39 PM

Tooltip solves it simply.

If more time can be dedicated here I think two things would help:
1. A style panel with longer fields as vars are going to be the new standard
2. A quicker way to see the computed value like a keyboard shortcut... tooltip takes a bit to show and then your eyes have to find the value in the longhand format

But still I think this is fine for now

kof:
> tooltip takes a bit to show

only if you hover, you can always click on label to show it immediately

kof:
Another way to see quickly a computed value would be to focus on the value and press a modifier.

kof:
After thinking a bit, I think we should do this:
- [ ] in the tooltip show computed value on option/alt keydown and hide it on keyup, so the computed version is visible while modifier is pressed
- [ ] same thing on the inputs, this way I don't need to even open a tooltip and have a very fast access to a computed version. Here the only question is if we should do this for the focused property only or for the entire style panel.

TrySound:
https://github.com/user-attachments/assets/b6fc27ad-821f-4517-8fb5-d5106d9e8aba

johnsicili:
This is nice.

> > tooltip takes a bit to show
> 
> only if you hover, you can always click on label to show it immediately
This isn't the case for all inputs like flex gap and spacing. Beyond it taking more time, the other downside is you have to scan through properties and with longhand properties it takes time.

PR #4239, opened 2024-10-07
experimental: support gradient variables

TrySound:

Here used computed values with resolved variables to get the list of gradients and preview in backgrounds list.

Though the new gradient field sets whole background-image property when variables are used.

Screenshot 2024-10-07 at 14 20 20


kof:
Shouldn't the code preview have the variable there?

kof:
can user paste a gradient in code below that uses variables for any/every part of it?

kof (approved):

TrySound:
Sure paste is possible, but it will be "unparsed". Here the change is about supporting background-image: var(--my-gradient) where --my-gradient can have repeated gradient values.

johnsicili (approved):
Initial review looks good. I'll be playing with these in more real use cases today once merged

PR #4240, opened 2024-10-07
fix: some scrub bugs

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4196 https://github.com/webstudio-is/webstudio/issues/4237

- fixed z-index scrubbing by rounding integrer number
- fixed background size by fallback to non layers value

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
+++ b/apps/builder/app/builder/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
@@ -44,6 +44,13 @@ export const parseIntermediateOrInvalidValue = (
   }
 
   if ("unit" in styleValue && styleValue.unit === "number") {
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/style-panel/shared/css-value-input/parse-intermediate-or-invalid-value.ts
```diff
@@ -44,6 +44,13 @@ export const parseIntermediateOrInvalidValue = (
   }
 
   if ("unit" in styleValue && styleValue.unit === "number") {
+    // when unit is number some properties supports only integer
+    // for example z-index
+    styleInput = parseCssValue(property, `${Math.round(Number(value))}`);
```
kof, line 49: I think you can only round it for some properties like z-index, but not generally, as you may want to have 1,5rem or something

kof (approved):

PR #4243, opened 2024-10-07
refactor: add advanced transform options

TrySound:
Here moved advanced transform options behind popover instead of adding with + button. It is slightly less annoying.

Also migrated transforms section to new style engine with proper css variables support. Will enable autocomplete in another PR.

Screenshot 2024-10-08 at 01 23 32


johnsicili:
I very much like everything in the advanced section but I don't like the default state... too much empty space. We could almost collapse by default but there's a use case for perspective without any transform properties.

I think it looks too bad as is though.

Screenshot 2024-10-07 at 3 35 29 PM

A better UI altogether, I think, would have the various icons for rotate, translate, etc, right there to fill the space, and you can add them by clicking on them... but this is entirely unique to our UI.

Can we make it like the Object Position property where it still has a label?

image

@kof thoughts?

TrySound:
Screenshot 2024-10-08 at 02 01 35

johnsicili:
I can get behind this! Though option click label doesn't reset and the label doesn't have any tooltip

kof:
When there are no transforms, the Advanced Transforms won't look very well. Maybe placing Just the button on the left from the + in the section header is a better option

TrySound:
Is this bad?
image

kof:
The current problem you are solving is this overloaded look with a lot of UI. But the important fact here is that these properties are rarely used and most people won't ever have a problem with that.

image

kof:
But when these things are defined, its quite important to notice that they are, because you will otherwise wonder why something is working unexpectedly

kof:
WF has it in the header fwiw 
image

TrySound:
> But when these things are defined, its quite important to notice that they are, because you will otherwise wonder why something is working unexpectedly

How is it a concern with current solution?

https://github.com/webstudio-is/webstudio/pull/4243#issuecomment-2398084401

TrySound:
Screenshot 2024-10-08 at 16 43 09

kof (approved):

PR #4252, opened 2024-10-08
refactor: upgrade ui libraries

TrySound:
Just minor releases

- fonts
- floating-ui
- radix
- react-aria
- react-error-boundary


kof:
sounds like a risk to get more subtle bugs with no benefits ... I would rather upgrade when we see the need and test that one specific thing well instead of doing bulk updates without knowing what it can break

TrySound:
> I would rather upgrade when we see the need

never 😄

kof:
right now before launch it sounds very dangerous in particular

PR #4259, opened 2024-10-09
fix: render background styles after builder reload

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/4244

When started rendering all preview var() fallbacks as unparsed. Styles renderer was not able anymore to rerender those.

So here at least temporary just added computed with transform value to always rerender when assets are changed.


TrySound:
Hm, nope. Didn't work

johnsicili (approved):

PR #4270, opened 2024-10-11
experimental: support css variables in transforms

TrySound:

Screenshot 2024-10-11 at 23 34 48

One thing I noticed, our implementation does not compose scale and rotate well.
According to spec rotate should be computed before scale but in our implementation rotate goes after scale.

https://www.w3.org/TR/css-transforms-2/#ctm


johnsicili (approved):
Let's go!

PR #4275, opened 2024-10-13
experimental: improve repeated styles with css variables

TrySound:

Here added a few improvements to repeated style to handle variables properly. For now all changes affects only backgrounds section.

- single layer can use css variable with multiple gradients
- with multiple layers each can have css variable with single gradient
- when single variable inside multiple layers has multiple gradients all thumbnails become transparent, it should be and indicator of invalid gradients
- fixed toggling hidden considering variables
- fixed deleting layers considering variables
- now when single layer has variable with multiple gradients only first gradient can be deleted or toggled

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/backgrounds/background-content.tsx
@@ -89,12 +89,10 @@ const setGradientProperty = (
   }
   const unpackedItem = newItem.type === "layers" ? newItem.value[0] : newItem;
   if (items.length === 1 && unpackedItem.type === "var") {
-    console.log(unpackedItem);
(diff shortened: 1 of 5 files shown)


johnsicili (approved):
Great just one thing probably needing fixing

1. Upon adding a 2nd gradient, both gradient names change to "None". Gradient appearance remains intact 

Screenshot 2024-10-13 at 9 46 38 AM

TrySound:
Will this work?

image

johnsicili:
YES!

johnsicili:
Looks amazing. Just used it. Works well too.

Btw in the overview task I have one in ideas to only show `--names` and exclude `var()` from the UI. I think it's way cleaner, especially given the limited space. 

There still are places for `var()` like in the actual input fields.

PR #4281, opened 2024-10-14
experimental: autocomplete variables in backgrounds code

TrySound:

Using dedicated inputs to autocomplete variables makes UI more complex. Though we already have "Code" fields which can be used for the same purpose.

Here added css fragment editor to background gradient.

Screenshot 2024-10-14 at 15 56 54


johnsicili (changes requested):
1. Real time rendering appears to have gone away when arrowing through the vars.
2. (minor) It's not until blur does the variable name render in the style panel.

Otherwise looks good to me

TrySound:
Yeah, there is no preview in css editor. It is basically the same as expression editor, not like other inputs. I would leave it as is til launch and improve later. Pls approve

johnsicili (approved):
Okay but high prio to bring it back. They are really difficult to use without preview... its a guessing game

PR #4290, opened 2024-10-15
fix: show custom option in easing dropdow only when used

### LINKED ISSUE #4197: Custom easing transition not working
When selecting `custom` in easing function, it just goes to "ease".

I haven't used this before, so not sure the expectation, but maybe a field I can enter the function 🤷 

https://github.com/user-attachments/assets/e80b90e4-4632-4f5d-ac53-9035d978ed03

JayaKrishnaNamburu:
fixes #4197 

## Description

This PR improves the UX for `custom` option in the easing functions. Currently it is displaying every time, with the PR we show them only when value added by the users are `custom` and hides it when the entered value doesn't come under custom category.

## Steps for reproduction

- Add a transition on a node and then select a value from the dropdown.
- When the dropdown is open and the easing that is added is a valid one. The `Custom` option should be hidden.
- When a custom value is added under the text-area, then the select box should show `Custom` value.

https://github.com/user-attachments/assets/938a8452-ae77-48d9-a395-db6b5ca23851

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/transitions/transition-timing.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/transitions/transition-timing.tsx
@@ -88,9 +88,11 @@ export const TransitionTiming = ({
               </SelectItem>
             );
           })}
-          <SelectItem key="custom" value="custom">
-            Custom
-          </SelectItem>
+          {value === "custom" && (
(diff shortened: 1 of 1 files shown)


kof (approved):
looking great!

PR #4305, opened 2024-10-18
fix: allow inserting into textual collection items

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4294


kof (approved):
didn't test

PR #4314, opened 2024-10-19
fix: After modifying style, it takes two clicks to select the next instance

### LINKED ISSUE #4306: After modifying style, it takes two clicks to select the next instance
1. modify a style
2. click another instance on canvas
3. see it didn't register
4. click again
5. see it registered

https://github.com/user-attachments/assets/0eb3ff89-37ad-4384-be25-5ee776e07798

### LINKED ISSUE #4309: Style Source looses focus after adding Token
The cursor remains for about half a second; then it loses focus, so I can't add multiple back-to-back. Or, if I hit delete to backspace, it deletes the instance as that's what becomes focused.

1. Add token from autocomplete
2. Wait a second
3. Try adding another or hit backspace
4. See focus elsewhere

https://github.com/user-attachments/assets/ed8fe1f3-9685-4fc1-8d11-07604e87a499

istarkov:
## Description
Style Source looses focus after adding Token

closes #4309
closes #4306

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/shared/inert.ts
+++ b/apps/builder/app/canvas/shared/inert.ts
@@ -1,11 +1,15 @@
 let resetTimeoutHandle: number | undefined = undefined;
 
 const resetAutoDisposeInert = () => {
+  if (resetTimeoutHandle === undefined) {
+    return;
+  }
   document.body.removeAttribute("inert");
   clearTimeout(resetTimeoutHandle);
   resetTimeoutHandle = undefined;
 };
 
+let lastPointerEventTime = Date.now();
 // 1000 ms is a reasonable time for the preview to reset.
 // Anyway should never happen after user has finished preview changes (can happen during preview changes)
 const AUTO_DISPOSE_INERT_TIMEOUT = 1000;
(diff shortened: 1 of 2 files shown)


apps/builder/app/canvas/shared/inert.ts
```diff
@@ -31,3 +43,10 @@ const setAutoDisposeInert = (timeout: number) => {
  */
 export const setInert = () => setAutoDisposeInert(AUTO_DISPOSE_INERT_TIMEOUT);
 export const resetInert = () => setAutoDisposeInert(DISPOSE_INERT_TIMEOUT);
+
+// window.self !== window.top means we are on canvas
+if (typeof window !== "undefined" && window.self !== window.top) {
+  window.addEventListener("pointerdown", () => {
```
kof, line 49: don't we need to clean up all these even listeners at somepoint? For example hot reload will probably cause issues here and there if there is no init and cleanup

kof (approved):

apps/builder/app/canvas/shared/inert.ts
istarkov: Somehow we need, but everytime we have tried to use module.hot it became much more worse. 
Current code should not do anything bad even in case of hot reload imo

PR #4317, opened 2024-10-19
feat: support css variables in transitions

TrySound:

- stop filtering when open combobox suggestions
- prevent showing selected description when highlight css variable
- support css variables in duration
- support css variables in delay

Also replaced select with css value input for easing. Had to get rid from out easings presets though now we can use easings from css variables which I think a better trade.

Presets will make sense with dedicated UI for easings like https://www.easing.dev.

Screenshot 2024-10-19 at 21 55 02


johnsicili (approved):
YES

PR #4318, opened 2024-10-19
fix: Prevent tooltip on focus

### LINKED ISSUE #4293: After using expression editor, canvas loses focus
After I use the expression editor, I can't scroll on the canvas until I click on it again.

1. Ensure the page has some room to scroll (eg 200vh)
2. Open expression editor
3. Click out of the expression editor, including clicking canvas directly
4. Try scrolling canvas
5. See no scroll 🙈 

https://github.com/user-attachments/assets/3a1ef1b6-cb5c-479c-9765-f9dd1777aed9

Borderline P2 bug. If too much effort maybe it's P2.

istarkov:
## Description

closes #4293

One of solution also
```
        onCloseAutoFocus={(event) => {
          event.preventDefault();
        }}
```

on dialog and popover content, but it's lost element focus.

This solution doesn't lost focus. Just prevent tooltip from being shown on focus.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/tooltip.tsx
+++ b/packages/design-system/src/components/tooltip.tsx
@@ -103,7 +103,11 @@ export const Tooltip = forwardRef(
           asChild
           {...triggerProps}
           onFocus={(event) => {
-            // Prevent tooltip from opening on focus
+            // Prevent the tooltip from being shown on focus, even though this goes against accessibility guidelines.
+            // The only valid use case for showing tooltips on focus is for users who cannot use a mouse.
+            // However, since the editor cannot be used without a mouse, it is better to have a working button
+            // than to attempt to support this scenario, which introduces issues like breaking canvas scrolling or displaying
+            // the tooltip when closing a dialog/popover.
             event.preventDefault();
             triggerProps?.onFocus?.(event);
           }}
--- a/packages/design-system/src/components/tooltip.tsx
+++ b/packages/design-system/src/components/tooltip.tsx
@@ -13,7 +13,6 @@ import { Box } from "./box";
 import { Text } from "./text";
 import type { CSS } from "../stitches.config";
 import { theme } from "../stitches.config";
-import { composeEventHandlers } from "@radix-ui/primitive";
 
 export const TooltipProvider = TooltipPrimitive.TooltipProvider;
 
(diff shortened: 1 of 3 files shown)


kof (approved):
please add a comment about WHY we decided to do that

packages/design-system/src/components/tooltip.tsx
```diff
@@ -13,7 +13,7 @@ import { Box } from "./box";
 import { Text } from "./text";
 import type { CSS } from "../stitches.config";
 import { theme } from "../stitches.config";
-import { disableCanvasPointerEvents } from "../utilities";
+import { composeEventHandlers } from "@radix-ui/primitive";
```
TrySound, line 16: We already have such utility here
image

packages/design-system/src/components/tooltip.tsx
```diff
@@ -78,18 +78,21 @@ export const Tooltip = forwardRef(
...
+     *    (This doesn't work well in Chrome, as scrolling stops working on elements hovered with pointer-events: none,
+     *    even after removing pointer-events.)
+     * 2. Close the tooltip on onMouseLeave.
+     *    (This breaks some grace area behavior, such as closing the tooltip when moving the mouse from the content to the trigger.)
+     *
+     * The simpler solution with fewer side effects is to close the tooltip on mouse leave.
+     */
+    const handleMouseEnterComposed = composeEventHandlers(() => {
+      setOpen(false);
+    }, props.onMouseLeave);
```
TrySound, line 95: Doesn't this do the same?
```
(event) => {
  setOpen(false);
  props.onMouseLeave?.(event)
}
```

istarkov: In current implementation yes. But what's wrong with using it.

TrySound: Extra direct dependency to maintain. Too little value

TrySound: And frankly I can only guess how it behaves. Explicit function call is just so much easier to read.

TrySound (approved):

PR #4323, opened 2024-10-20
feat: support paste styles in advanced section

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3399 https://github.com/webstudio-is/webstudio/issues/3540

Here added two more features to advanced panel

1. When search for property, user can enter css declarations like `width: 100px; height: 200px;` and press enter instead of searching for property and editing property after creation.

2. When search for property, user can paste css declarations from some theme editor or figma, this is useful interop with other tools

Note: this does not support breakpoints, states or tokens. Declarations are inserted only within currently selected style source and state.

https://github.com/user-attachments/assets/1f37f5d1-11b9-4f05-ae4d-a7c647ae14b6

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
@@ -234,7 +234,7 @@ const AdvancedSearch = ({
           <InputField
             {...combobox.getInputProps()}
             autoFocus={true}
-            placeholder="Find or create a property"
+            placeholder="Add styles"
             suffix={<NestedInputButton {...combobox.getToggleButtonProps()} />}
             onPaste={(event) => {
               const data = event.clipboardData.getData("text");
--- a/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
@@ -236,14 +236,6 @@ const AdvancedSearch = ({
             autoFocus={true}
             placeholder="Add styles"
             suffix={<NestedInputButton {...combobox.getToggleButtonProps()} />}
-            onPaste={(event) => {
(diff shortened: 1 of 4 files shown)


kof:
- [ ] we should probably update the placeholder in that input to better reflect this new capability. Maybe "Add CSS"?

kof:
- [ ] when I pasted, I didn't expect it to insert them all right away, I expected to see them in the input then hit enter. This beggs the question if input is the right element for this, I think what you want here is either code editor component or just auto growing textarea like we have in the settings

kof (changes requested):
A few things with ux need improvement, otherwise works great! Great idea turning that input into effectively a paste editor

TrySound:
I wouldn't want to to be "css editor" caz advanced panel is already an editor itself. I see it more like a command line where user can write one-liners. Styles are applied with form on "Enter", multiline editor will kill this simplicity.

kof:
> user can write one-liners

yeah, though these are not one-liners, its more like the editor console

TrySound:
Terminal in editor? Drop history, it's one liner.

kof:
When you paste multiline css block it becomes multiline

kof:
Single line or multi line, its a detail though, yes, conceptually something like the cli, though we need to make pasted multiline block editable comfortably before sending

TrySound:
Sending where? User already copied what is needed. Why it needs to be edited? How user should apply multiline changes? Enter doesn't work anymore. Add "save" button?

TrySound:
It's actually similar to paste markdown. We do not prompt user with this markdown before applying.

kof:
After thinking. I think the mindset here should be primarily around a single-line usage. So when pasted multiline content, textarea would grow, allowing you to see and edit, but enter would submit, shift+enter would add a newline.

TrySound:
Downshift completely fails with textarea

kof (approved):

PR #4333, opened 2024-10-22
feat: Add resource control to webhook form action

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/4093 https://github.com/webstudio-is/webstudio/issues/3871

Here added new resource control for out properties. It is replaced action and method properties webhook form. Now users will be able to additionally provide headers. In another PR headers will let user to customize body format.

Also migrated "action" in existing webhook forms to resource prop type.

Screenshot 2024-10-22 at 16 07 38

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
+++ b/apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
@@ -217,11 +217,9 @@ export const ResourceControl = ({
               }
               // attempt to save form on close
               if (areAllFormErrorsVisible(form.current)) {
-                console.log("submit");
                 form.current?.requestSubmit();
                 setIsResourceOpen(false);
               } else {
-                console.log("validate");
                 form.current?.checkValidity();
                 // prevent closing when not all errors are shown to user
               }
--- a/apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
+++ b/apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
@@ -1,0 +1,267 @@
+import { nanoid } from "nanoid";
+import { computed } from "nanostores";
+import {
+  forwardRef,
+  useId,
+  useRef,
+  useState,
+  type ComponentProps,
+} from "react";
+import {
+  EnhancedTooltip,
+  Flex,
+  InputField,
+  NestedInputButton,
+  theme,
+} from "@webstudio-is/design-system";
+import { useStore } from "@nanostores/react";
+import { encodeDataSourceVariable, Prop, Resource } from "@webstudio-is/sdk";
+import { GearIcon } from "@webstudio-is/icons";
+import { $resources, $selectedPage } from "~/shared/nano-states";
+import { updateWebstudioData } from "~/shared/instance-utils";
+import { FloatingPanel } from "~/builder/shared/floating-panel";
+import {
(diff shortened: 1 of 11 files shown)


kof:
- [ ] Icon position is off: 
image

kof:
Haven't tested, but looks great, lets list what this enables from use cases perspective, so that once @johnsicili is back we can record an edu video, maybe replacing the old form video.

Are all of these done? https://github.com/webstudio-is/webstudio/issues/4093

kof:
So far I see 2 use cases we need to demonstrate:

1. As a user sometimes I need to use an API that only accepts json data and requires tokens in headers. One such example is signing up for newsletter in splunk https://docs.useplunk.com/api-reference/contacts/subscribe

2. Using response json and rendering something in return as a success or error.

Am I correct to assume both use cases are supported now?

kof:
Got it, so this unlocks a use case like useplunk.

TrySound:
Yes, though I'm planning a small launch with all those features requested so webhook forms would make more sense to users.

kof:
Btw you should try to post something to https://docs.useplunk.com/api-reference/contacts/subscribe to see if we are still missing something there

sartipaz:
Hi, we are doing an ecommerce site (with a lot of handmade javascript) and we really need this feature, as I can see is already working and passes all the CI checks, I will mean a lot if we get those new configs for the headers.

Right now we are going to pass (in some hidden fields) the values we need in the headers thru zapier and then move the hidden fields to the corresponding header just to make the site work, but it is a huge security issue.

apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
```diff
@@ -0,0 +1,265 @@
...
+        type: "resource",
+        value: resource.id,
+      },
+    });
+    data.resources.set(resource.id, resource);
+  });
+};
+
+const areAllFormErrorsVisible = (form: null | HTMLFormElement) => {
+  if (form === null) {
```
istarkov, line 155: Looks like huge error if form is not available at the moment

apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
```diff
@@ -0,0 +1,265 @@
...
+  for (const element of form.elements) {
+    if (
+      element instanceof HTMLInputElement ||
+      element instanceof HTMLTextAreaElement
+    ) {
+      // field is invalid and the error is not visible
+      if (
+        element.validity.valid === false &&
+        // rely on data-color=error convention in webstudio design system
+        element.getAttribute("data-color") !== "error"
```
istarkov, line 168: looks like dirty hack

istarkov (approved):

apps/builder/app/builder/features/settings-panel/controls/resource-control.tsx
TrySound: It is though it works and I didn't figure out another way.

johnsicili (approved):
Looks good. I didn't test headers as the webhook I tested with didn't need them.

IMO the action input field should allow editing as most webhooks require just a URL. The settings should be there to make further mods.

kof:
@TrySound what is this waiting for?

PR #4334, opened 2024-10-22
feat: Block editing with arrow keys

istarkov:
## Description

- Double click now sets cursor to the click position
- Up/Down arrows on a first/last line trying to preserve cursor x position (visible in case of block above block)
- Right/Left arrows switch blocks if cursor is at block begin/end.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/navigator/navigator-tree.tsx
+++ b/apps/builder/app/builder/features/navigator/navigator-tree.tsx
@@ -584,11 +584,6 @@ export const NavigatorTree = () => {
                       ? rawTheme.colors.foregroundReusable
                       : undefined,
                   },
-                  onKeyDown: (event) => {
-                    if ($textEditingInstanceSelector.get() !== undefined) {
-                      event.preventDefault();
-                    }
-                  },
                 }}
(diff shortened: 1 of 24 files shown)


apps/builder/app/builder/features/navigator/navigator-tree.tsx
```diff
@@ -583,6 +584,11 @@ export const NavigatorTree = () => {
                       ? rawTheme.colors.foregroundReusable
                       : undefined,
                   },
+                  onKeyDown: (event) => {
+                    if ($textEditingInstanceSelector.get() !== undefined) {
+                      event.preventDefault();
+                    }
```
TrySound, line 590: Is it still necessary after removing focus on select?

istarkov: Pretty sure that yes, as it had intersection with key down during the editing

TrySound (approved):
Brilliant

PR #4367, opened 2024-11-04
experimental: add command panel

TrySound:

Press cmd+k or ctrl+k and see yourself. Or just watch the video.

Based on https://cmdk.paco.me used by shadcn.

https://github.com/user-attachments/assets/46bb807a-c6af-487f-b57c-09fb8f6e5741


johnsicili:
This is really nice. Works well!!

Some things I noticed.. not sure if you want to address now:

- Sort order on Components seems random
- Select inserted component (should emulate the behavior of selecting components from add panel imo)
- Scroll into view after insert

TrySound:
First 2 and 3, not sure what order do you expect in 1. Would alphabetical make sense?

TrySound:
Ah, forgot we have order 🤦

MilanBgd:
Oh God Bogdan!! This is what I'm waiting for in order to accelerate the design process. I don't like to click on "+" button. The Cmd+K is just perfect! Hope it will be release soon :-) Congrats for your work!

johnsicili (approved):
1. I still don't see this instance scrolling into view after insert. Is this related to some other issue I vaguely recall we have that's not related to this PR?

Otherwise good to go!!

TrySound:
> I still don't see this instance scrolling into view after insert. Is this related to some other issue I vaguely recall we have that's not related to this PR?


Navigator scroll into view works but canvas one seems does not work with regular insert as well. So not related to this PR.

kof:
- [ ] when home is the only page

image

kof (approved):

oslogrolls:
Unless this uses another build than the latest release, the Hotkey does nothing on Windows.

kof:
@oslogrolls its behind a feature flag, to activate type: `localStorage.features = 'command'` in the console and reload

kof:
@oslogrolls that is if you are trying in production, here on dev environment from the link it should be working without manually setting the feature flag

oslogrolls:
Thank you. Works now!

PR #4375, opened 2024-11-05
feat: select the page when opening the page settings

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4372

Editing another page settings is confusing to users and complicates the logic. Win-win


johnsicili (approved):
Nice!

Issue #3994, opened 2024-08-22
Content Mode


kof:
As a builder, I want to let a customer edit text and images without giving them the full power of the builder.

Here is how it can work:

## Step 1. Add Content Edit mode to the panel (similar to Preview button at the top panel).
- [x] Remove Style panel
- [x] At left show Pages, Navigator, Assets 
- [x] Disable Binding editing.
- [x] Disable Canvas drag.
- [x] Disable Tree drag.
- [x] In the properties panel, show only Link and Img tag default properties (but not ID or Class), do not allow add new properties. 
- [x] Do not show/allow bindings editing (text editing in binded instances too).
- [x] Disable copy/paste
- [x] Add EditContent button
- [x] Enable text editing functionality.
- [x] Probably allow text editing in a single click instead of double click in Edit Mode.
- [x] Make edit styles (i.e. show "text" cursor on editable text)

 Roll out without feature flag or public announcement.

## Step 2. Add "content editor" Share link.
- [x] Do some system rest API checks to not allow patching styles, props (only img and link are allowed)
- [x] Free users can't share Content Edit Link
- [x] All shared links from Free users must not have Content Edit mode

Roll out without feature flag or public announcement.

## Step 2.5 Add combo for modes
- [x] - Switch preview, design, content using combo like switch
- [x] - Padding right at menu is not ok (#4484)
- [x] - Order Design/Build/Preview (#4484)
- [ ] - Rename modes (Visible Text initially)
- [x] - Do not reset to preview automatically (#4484)
- [x] - Button zone (#4487)
 
## Step 3. Allow content Editors to upload and modify images

- [x]  Roll out without feature flag or public announcement.

## Step 4. Introduce the `EditableBlock` component.

```
EditableBlock
   - Templates (Designers can add components within a controlled template area.)
     - Header
     - Article 
     - Anything 
  - Header Edited By user
  - Article Edited By user   
  - Article Edited By user   
```   

- [x]  The content editor can copy components from the `Templates` area directly into the `EditableBlock`, making them immediate children of the `EditableBlock` _(UI inspired by Notion or potentially with a “+” button at the bottom of each block, open for discussion)_.

- [x]  **Component Reordering**: Users can reorder components vertically within the EditableBlock. _(Note: Dragging on the main canvas is not yet supported. Lack of main canvas dragging is acceptable in the editor at this stage.)_ _(Done at #4483)_
- [x] Allow delete  block children (_done at #4483)
[...]

PR #4376, opened 2024-11-06
feat: Add builderMode, add mode "edit"

istarkov:
## Description
Part 1 Of #3994 

This PR introduces a feature flag, `contentEditableMode`, that hides the `EditorButton` in the top bar. This button is now intended for **testing purposes only** and will **never be used in production**.

image

## Most Notable Things In The Content Edit Mode

1. **Notion-like Editing**: Elements are editable with a single click.
2. **Restricted Element Editing**: Binded elements, such as expressions, are non-editable.
3. **Selective Property Editing**: Only properties for `img` and `link` elements are editable.

Additional changes:
- **Copy/Paste and Drag/Drop**: Disabled to maintain a controlled testing environment.

Refer to the issue `Part 1` for further details on upcoming changes and planned enhancements.

## Steps for reproduction

Play play 

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -30,6 +30,7 @@ import {
   $imageLoader,
   $textEditingInstanceSelector,
   $isDesignMode,
+  $isContentMode,
 } from "~/shared/nano-states";
 import { $settings, type Settings } from "./shared/client-settings";
 import { builderUrl, getCanvasUrl } from "~/shared/router-utils";
@@ -59,7 +60,10 @@ import { migrateWebstudioDataMutable } from "~/shared/webstudio-data-migrator";
 import { Loading, LoadingBackground } from "./shared/loading";
 import { mergeRefs } from "@react-aria/utils";
 import { CommandPanel } from "./features/command-panel";
-import { initCopyPaste } from "~/shared/copy-paste/init-copy-paste";
+import {
+  initCopyPaste,
+  initCopyPasteForContentEditMode,
+} from "~/shared/copy-paste/init-copy-paste";
 
 registerContainers();
 
@@ -283,6 +287,7 @@ export const Builder = ({
   const isCloneDialogOpen = useStore($isCloneDialogOpen);
   const isPreviewMode = useStore($isPreviewMode);
   const isDesignMode = useStore($isDesignMode);
+  const isContentMode = useStore($isContentMode);
 
   const { onRef: onRefReadCanvas, onTransitionEnd } = useReadCanvasRect();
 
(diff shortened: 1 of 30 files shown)


apps/builder/app/shared/nano-states/misc.ts
```diff
@@ -303,7 +303,17 @@ export const $hoveredInstanceSelector = atom<undefined | InstanceSelector>(
   undefined
 );
 
-export const $isPreviewMode = atom<boolean>(false);
+const builderModes = ["design", "preview", "edit"] as const;
```
TrySound, line 306: design -> build 😉

istarkov: decided to use "design" as in webflow build mode is our new "edit" mode

kof: gets a bit confusing, to me if its build/design is the same thing

kof: edit is very generic without any prefix like "content" ... I would actually call that "content" mode or "contentEditing"

kof: "edit" is semantically everything: design, build, content etc, all of it is edit

johnsicili (changes requested):
Really love the single click edits!!

- [x] - 1. (high prio) Duplicate works, should not be allowed (cmd d)
2. (just noting) Seems odd to have the right panel when the majority of the time it just consumes valuable space. I don't have ideas right now. Maybe in the future we'll add link and image editing in canvas and can do without right panel for clients
- [ ] - 3. (known issue) Cursor shows select on non editable text
4. (known issue) Links are hard to edit when adding them inline
5. (just noting) I don't love image dimension fields exposed because I often reduce the dimensions to better optimize. But if they replace the image then I guess it doesn't matter.
6. (just noting) They can edit text content on XML. IDK if this is a problem.
- [x] - 7. (high prio) Breakpoints can be modified
- [x] - 8. (high prio) Can still add components via "A" command and "CMD K" 
9. (low prio) Allow editors to swap out Vimeo URL
- [ ] - 10. (low prio) Maybe disable CSS preview

apps/builder/app/builder/builder.tsx
```diff
@@ -294,24 +297,31 @@ export const Builder = ({
   const [loadingState, setLoadingState] = useState(() => $loadingState.get());
 
   useEffect(() => {
+    if (!isDesignMode) {
+      return;
+    }
```
TrySound, line 302: why not $isDesignMode.subscribe inside effect?

apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
```diff
@@ -152,6 +153,28 @@ export const usePropsLogic = ({
...
+
+  /**
+   * In content edit mode we show only Image and Link props
+   * In the future I hope the only thing we will show will be Components
+   */
+  const isPropVisible = (propName: string) => {
+    const contentModeWhiteList: Partial<Record<string, string[]>> = {
+      Image: ["src", "width", "height", "alt"],
+      Link: ["href"],
+      RichTextLink: ["href"],
```
TrySound, line 166: don't like targeting component names, we need to make it configurable from sdk in the future, maybe add todo?

TrySound (approved):

apps/builder/app/builder/features/settings-panel/props-section/use-props-logic.ts
kof: def. need to change this before launching/considering it done

apps/builder/app/builder/builder.tsx
istarkov: Attached only in design mode. Before was subscribe I spent 1h trying to get why and how it;s working, and pretty sure it didn't work well

kof:
looking great!

PR #4398, opened 2024-11-10
refactor: generalize insertable lookup

TrySound:
Here replaced a bunch boilerplaty code with more general findClosestInsertable. This also reduced amount of explicit instance selector calls. Prevented inserting list item into body. Refactored tests with jsx.

### DIFF
--- a/apps/builder/app/shared/instance-utils.ts
+++ b/apps/builder/app/shared/instance-utils.ts
@@ -35,7 +35,6 @@ import {
 import {
   $props,
   $styles,
-  $selectedInstanceSelector,
   $styleSourceSelections,
   $styleSources,
   $instances,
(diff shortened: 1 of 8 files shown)


apps/builder/app/shared/instance-utils.ts
```diff
@@ -1621,3 +1589,40 @@ export const findClosestSlot = (
     }
   }
 };
+
+export type Insertable = {
+  parentSelector: InstanceSelector;
+  position: number | "end";
+};
+
+export const findClosestInsertable = (
```
kof, line 1598: what's the difference between findClosestInsertable and findClosestDroppableTarget?

TrySound: findClosestDroppableTarget is previous implementation which did not access stores. Here the utility access stores for simplicity of usage. findClosestDroppableTarget is used only in this modules, not exported anymore. Will probably refactor further later.

kof: oh, you probably need to inline that into that function, because its confusing from naming perspective

kof (approved):

PR #4407, opened 2024-11-12
experimental: add tabbable command actions

TrySound:

Here adding new idea into command search.
User searches for an object and can navigate actions over this object. In case of pages user can open page or open pangel settings panel.

https://github.com/user-attachments/assets/a5d6c039-ba6b-4dbb-b0b4-bb9f1782dd20


johnsicili (approved):
Really good idea! 

BTW I noticed when you add a component when another panel besides navigator is open, the navigator doesnt open

TrySound:
Made tab trigger another command panel 😉 
https://github.com/user-attachments/assets/5ccd38f2-6c32-4bde-8654-c69d29a727bf

PR #4432, opened 2024-11-16
fix: stop failing when retrieve mdn url for css variables

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4431


johnsicili (approved):
Nice!

PR #4434, opened 2024-11-17
experimental: add shortcuts group to command panel

TrySound:

Added most of our shortcuts to command panel.

Screenshot 2024-11-17 at 17 28 55


TrySound:
Some I removed because they are not useful or do no work with command panel.
- breakpoints already in the panel
- lexical formatting
- edit text will loose focus
- open command panel
- click canvas
- cancel drag

Maybe those should not be commands

kof (approved):
- [ ] option+down doesn't work when you are in shortcuts category

PR #4449, opened 2024-11-20
feat: make loading screen background dark

kof:
## Description

A better looking loading screen

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.css
+++ b/apps/builder/app/builder/builder.css
@@ -6,6 +6,11 @@ body {
   overflow: hidden;
   overscroll-behavior: contain;
   -webkit-font-smoothing: antialiased;
+  /* 
+    This is the top bar and loading screen color (--colors-backgroundTopbar).
+    We are setting it to avoid a white screen flash when opening project from the dashboard.
+   */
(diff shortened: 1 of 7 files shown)


johnsicili:
Is there a way to combat this white screen? Even if its just setting the default color to the same of the loader, that will look better. White screen with nothing seems like there is some issue.

https://github.com/user-attachments/assets/f9a92055-5b48-4a40-acbf-ff1a7543c226

kof:
done, check it out

johnsicili:
Looks better. I don't know if its related to this PR, but when I access a Project via a share link, the loading screen doesn't go away

kof:
reverted back the fix I was trying to make in parallel to the background change, which is about waiting for selected instance, before closing the overlay, currently its still having some movement after its loaded I wanted to avoid for a better effect, but seems too hard

@johnsicili good catch with this one

johnsicili (approved):

istarkov (approved):

PR #4451, opened 2024-11-20
refactor: add new way to synchronize immerhin stores

TrySound:
Here added SyncClient class which will be base for our new sync engine. For now it can synchronize one immerhin store.

Event emitter is shared with canvas directly to avoid indirection of iframe postMessage api.

This also partially fixes the issue with messages pollution and conflict with react devtools.


istarkov (approved):

apps/builder/app/builder/builder.tsx
```diff
@@ -294,6 +298,11 @@ export const Builder = ({
   useSetWindowTitle();
 
   const iframeRefCallback = mergeRefs((element: HTMLIFrameElement | null) => {
+    if (element?.contentWindow) {
+      // added to iframe window and stored in local variable right away to prevent
+      // overriding in emebedded scripts on canvas
+      element.contentWindow.syncEmitter = builderClient.emitter;
```
kof, line 304: you are adding this syncEmitter variable on the window/global ... that should be at least some lengthy name aka `__webstudioSomething__`

apps/builder/app/shared/sync-client.ts
```diff
@@ -0,0 +1,127 @@
...
+}>;
+
+type SyncClientOptions = {
+  role: "leader" | "follower";
+  store: Store;
+  emitter?: SyncEmitter;
+};
+
+export class SyncClient {
+  clientId = nanoid();
```
kof, line 27: do you need a nanoid generator here or could be just a simple counter or a Math.random call without dependencies?

TrySound: I need unique id to communicate with other tabs or users.

apps/builder/app/builder/builder.tsx
TrySound: Doesn't really matter, can rename

apps/builder/app/shared/sync-client.ts
TrySound: Currently, clientId is used to district own from received messages.

PR #4453, opened 2024-11-20
refactor: add sync object pool

TrySound:
Sync client is getting some shape. Here I abstracted all immerhin related api into SyncObject and added SyncObjectPool to combine multiple objects. Objects are identified by unique name.

### DIFF
--- a/apps/builder/app/shared/sync-client.ts
+++ b/apps/builder/app/shared/sync-client.ts
@@ -54,7 +54,7 @@ export class ImmerhinSyncObject implements SyncObject {
 export class SyncObjectPool implements SyncObject {
   name = "SyncObjectPool";
   objects = new Map<string, SyncObject>();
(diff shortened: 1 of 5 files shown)


apps/builder/app/shared/sync-client.ts
```diff
@@ -4,55 +4,136 @@ import type { Change, Store } from "immerhin";
...
+      callback({ id, object: this.name, changes });
+    });
+    signal.addEventListener("abort", unsubscribe);
+  }
+}
+
+export class SyncObjectPool implements SyncObject {
+  name = "SyncObjectPool";
+  objects = new Map<string, SyncObject>();
+  constructor(...objects: SyncObject[]) {
```
kof, line 57: better to just use an array instead of spreading the arguments, more explicit

kof (approved):

PR #4467, opened 2024-11-24
fix: show flex child for collection item

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4458

Collection is not stylable and was checked for display: flex but instead should be skipped.


johnsicili (approved):
Nice

PR #4469, opened 2024-11-25
refactor: remove instance selector from component hooks

TrySound:
We need to remove reliance on instance selector iteration as much as possible. In most cases we need only instance key which is a JSON.stringify or instance selector to access computed props and variables.

Here got rid of instance selector in component hooks and instead made api to accept "instance data" with instance id, key and component inside.


kof (approved):
sounds good, didn't test

PR #4475, opened 2024-11-27
refactor: add server as storage to sync engine

TrySound:
Another layer in sync engine is composable storage.

Now you can specify a chain of storages. Each can listen transactions to update some state. Though subscription for state is invoked one by one until non-undefined is provided.

See tests to understand API.

Here I also migrated our server data loading and synchronization to this storage api.

### DIFF
--- a/apps/builder/app/builder/shared/sync/sync-server.ts
+++ b/apps/builder/app/builder/shared/sync/sync-server.ts
@@ -1,13 +1,16 @@
 import { useEffect } from "react";
 import { atom } from "nanostores";
+import type { Change } from "immerhin";
 import type { Project } from "@webstudio-is/project";
 import type { Build } from "@webstudio-is/project-build";
 import type { AuthPermit } from "@webstudio-is/trpc-interface/index.server";
 import * as commandQueue from "./command-queue";
 import { restPatchPath } from "~/shared/router-utils";
 import { toast } from "@webstudio-is/design-system";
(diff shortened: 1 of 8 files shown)


kof (approved):

apps/builder/app/builder/shared/sync/sync-server.ts
```diff
@@ -305,6 +311,22 @@ const useSyncProject = ({
   }, [projectId, authPermit]);
 };
 
+export class ServerSyncStorage implements SyncStorage {
+  name = "server";
+  sendTransaction(transaction: Transaction<Change[]>) {
+    if (transaction.object === "server") {
+      pendingTransactions.push(transaction);
```
istarkov, line 318: I have feeling that we can just do
`commandQueue.enqueue({ type: "transactions", transactions, projectId });`
here

Not getting why we have setInterval

istarkov: See in poll logic we already have sleep insterval. There is no need in intermediate pending with additional interval

istarkov (commented):
I would remove setInterval logic

apps/builder/app/builder/shared/sync/sync-server.ts
TrySound: Indeed

istarkov (approved):

PR #4480, opened 2024-11-28
fix: make publish error more descriptive

TrySound:
This should fix users confusion when they try to publish from development version.

image


kof:
Why not link "Webstudio CLI" instead of separate line?

kof (approved):

TrySound:
> Why not link "Webstudio CLI" instead of separate line? Right now you expect user to copy the url instead of clicking?

toast is a plain text

kof:
> toast is a plain text

oh shit, I remember there waas something about sending the toast from the canvas to the parent with jsx, maybe this issue doesn't exist any more?

istarkov (approved):

kof:
Idea - if we already have markdown parser anyways, those notifications could support md, so we don't need to add jsx there, but a proper link etc would still be possible!

PR #4487, opened 2024-11-30
feat: Plus button hover polygon and menu icons

istarkov:
## Description

ref #3994

Menu Icons are now from meta

image

Red is hover polygon (used to not lost hover) (small quad before was not enough)

image

image

image

image

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/workspace/canvas-tools/outline/editable-block-instance-outline.tsx
+++ b/apps/builder/app/builder/features/workspace/canvas-tools/outline/editable-block-instance-outline.tsx
@@ -21,6 +21,7 @@ import {
   toast,
   Kbd,
   Text,
+  iconButtonSize,
 } from "@webstudio-is/design-system";
 import { EditableBlockChildAddButtonOutline } from "./outline";
 import { applyScale } from "./apply-scale";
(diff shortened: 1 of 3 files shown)


TrySound (approved):

apps/builder/app/builder/features/workspace/canvas-tools/outline/editable-block-instance-outline.tsx
```diff
@@ -352,8 +356,8 @@ export const EditableBlockChildHoveredInstanceOutline = () => {
       <Flex
         css={{
           width: "min-content",
-          height: "min-content",
           pointerEvents: isMenuOpen ? "none" : "all",
+          clipPath: `polygon(0% 0%, 100% 0%, 100% 100%, 0% 24px)`,
```
kof, line 360: that 24px needs to be somehow reused from the values that define trigger size etc, otherwise if we change something there, there is no way we will not forget to update this value as well, because this thing is basically invisible

kof (approved):
works nicely, but please don't hardcode 24px

PR #4498, opened 2024-12-03
feat: Add initial text to the Content Block

istarkov:
## Description

ref #3994 

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/react-sdk/src/core-components.ts
```diff
@@ -162,6 +162,19 @@ const blockMeta: WsComponentMeta = {
...
+          label: "Initial",
+          component: "Box",
+          children: [
+            {
+              type: "text",
+              value:
+                "Content Block component designates regions on the page where pre-styled instances can be inserted in Content mode",
+            },
+          ],
+        },
```
kof, line 177: Here is the final content we want to use when user inserts content block by default:

```
[Content Block](https://wstd.us/content-block) component designates regions on the page where pre-styled instances can be inserted in Content mode. 

1. In Content mode, you can edit any direct child instances that were pre-added to the Content Block, as well as add new instances predefined in Templates.

2. To predefine instances for insertion in Content mode, switch to Design mode and add them to the Templates container.

3. To insert predefined instances in Content mode, click the + button while hovering over the Content Block on the canvas and choose an instance from the list.
```

It contains a paragraph, a link, a list and list items.
Same components should be also available by default inside the Templates container.

istarkov: Can you create example on main and give link? I still don't undertand what's needed

kof (approved):

PR #4507, opened 2024-12-04
refactor: migrate dnd to new constraints

TrySound:
Replaced legacy utilities with matchers

List item no longer possible to place into list item, thanks to new parent matcher.

Dnd in navigator and on canvas can be tested


TrySound:
@kof this is previous PR with these matchers
https://github.com/webstudio-is/webstudio/pull/4499

PR #4515, opened 2024-12-05
feat: Create xml-time component to convert time to iso string in xml documents

istarkov:
## Description

image

Tested with fixtures.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/xml-time.tsx
+++ b/packages/sdk-components-react/src/xml-time.tsx
@@ -1,0 +1,62 @@
+import { forwardRef, useContext, type ElementRef } from "react";
+import { ReactSdkContext } from "@webstudio-is/react-sdk/runtime";
+
+const DEFAULT_DATE_STYLE = "short";
+const INITIAL_DATE_STRING = "dateTime attribute is not set";
+const INVALID_DATE_STRING = "";
+
+type XmlTimeProps = {
+  dateStyle: "long" | "short";
+  datetime: string;
+};
+
+const parseDate = (datetimeString: string) => {
+  if (datetimeString === "") {
+    return;
+  }
+  let date = new Date(datetimeString);
+
+  // Check if the date already in valid format, e.g. "2024"
+  if (Number.isNaN(date.getTime()) === false) {
+    return date;
+  }
+
+  // If its a number, we assume it's a timestamp and we may need to normalize it
+  if (/^\d+$/.test(datetimeString)) {
+    let timestamp = Number(datetimeString);
+    // Normalize a 10-digit timestamp to 13-digit
+    if (datetimeString.length === 10) {
+      timestamp *= 1000;
+    }
+    date = new Date(timestamp);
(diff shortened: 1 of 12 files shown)


packages/sdk-components-react/src/xml-time.tsx
```diff
@@ -0,0 +1,62 @@
...
+const DEFAULT_DATE_STYLE = "short";
+const INITIAL_DATE_STRING = "dateTime attribute is not set";
+const INVALID_DATE_STRING = "";
+
+type XmlTimeProps = {
+  dateStyle: "long" | "short";
+  datetime: string;
+};
+
+const parseDate = (datetimeString: string) => {
```
kof, line 13: couldn't we reuse the logic from the other time component? it feels like this stuff needs lots of tests

packages/sdk-components-react/src/xml-time.ws.ts
```diff
@@ -0,0 +1,22 @@
...
+  type WsComponentMeta,
+  type WsComponentPropsMeta,
+} from "@webstudio-is/react-sdk";
+
+import { props } from "./__generated__/xml-time.props";
+
+export const meta: WsComponentMeta = {
+  category: "xml",
+  type: "container",
+  description: "Converts machine-readable date and time to ISO format.",
```
kof, line 13: machine-readable? or just unix time stamp? what are all the formats it supports?

packages/sdk-components-react/src/xml-time.tsx
istarkov: Something with build:args on reuse. It calcs exports as props. So I would spend tooo many time on reuse. Here is just few lines

packages/sdk-components-react/src/xml-time.ws.ts
istarkov: Have no idea, took same description as it was in time

TrySound (approved):

packages/sdk-components-react/src/xml-time.tsx
kof: so we can't import any modules from that package? that would be crazy @TrySound do you know what's going on?

istarkov: We can't export something not related to component i.e. props component itself otherwise type builder assumes this as props.
I just don't want waste time to find how to fix/improve it. As this code is so small that does not need any reuse.

kof: I understand the urge and maybe we can merge this, but I think its a big problem in general, not only for this, I don't want us to be fixing this thing after users find a bug without a way to ensure there are no regressions and not having tests for it, working with dates is always a pain ...

kof (approved):

PR #4516, opened 2024-12-05
fix: Prevent focus when selecting form label in builder

### LINKED ISSUE #4513: Prevent focus when selecting form label in builder
When selecting a form label that as the `for` property, the field is focused preventing editing the label via the canvas.

We should prevent the default behavior here when within a building mode.

1. Add form label and input
2. Give label a `for` value that matches the input's `id`
3. Try selecting label in builder to edit

https://github.com/user-attachments/assets/c9babc59-b19a-49aa-a697-ec91bbaba71c

istarkov:
## Description

Prevent forwarding the click event on an input element when the associated label has a "for" attribute

In real browser emulates click event on input when label is clicked.

closes #4513

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/interceptor.ts
+++ b/apps/builder/app/canvas/interceptor.ts
@@ -71,12 +71,12 @@ const switchPageAndUpdateSystem = (href: string, formData?: FormData) => {
 export const subscribeInterceptedEvents = () => {
   const handleClick = (event: MouseEvent) => {
     // Prevent forwarding the click event on an input element when the associated label has a "for" attribute
-    if (event.target instanceof Element) {
-      if (event.target.closest("label[for]")) {
-        if (!$isPreviewMode.get()) {
-          event.preventDefault();
-        }
-      }
+    if (
+      event.target instanceof Element &&
+      event.target.closest("label[for]") &&
+      !$isPreviewMode.get()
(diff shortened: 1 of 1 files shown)


apps/builder/app/canvas/interceptor.ts
```diff
@@ -70,6 +70,15 @@ const switchPageAndUpdateSystem = (href: string, formData?: FormData) => {
 
 export const subscribeInterceptedEvents = () => {
   const handleClick = (event: MouseEvent) => {
+    // Prevent forwarding the click event on an input element when the associated label has a "for" attribute
+    if (event.target instanceof Element) {
+      if (event.target.closest("label[for]")) {
+        if (!$isPreviewMode.get()) {
```
kof, line 76: do this first before you call other things

kof (approved):

apps/builder/app/canvas/interceptor.ts
kof: also why is this a nested if instead of a single condition :)

TrySound (approved):

PR #4533, opened 2024-12-07
fix: Do not show empty content blocks in preview mode

istarkov:
## Description

ref #3994

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/features/build-mode/block.tsx
+++ b/apps/builder/app/canvas/features/build-mode/block.tsx
@@ -11,6 +11,7 @@ import * as React from "react";
 import {
   $instances,
   $isDesignMode,
+  $isPreviewMode,
   $selectedInstanceSelector,
 } from "~/shared/nano-states";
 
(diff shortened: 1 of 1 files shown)


apps/builder/app/canvas/features/build-mode/block.tsx
```diff
@@ -81,7 +83,9 @@ export const Block = React.forwardRef<
   return (
     <div ref={ref} style={editableBlockStyle} {...props}>
```
kof, line 84: why do you want to render the div still?

kof (approved):

apps/builder/app/canvas/features/build-mode/block.tsx
istarkov: data-ws-id, selectors, ref's are used in various algorithms we have. Every our control has some kind of root element with all that stuff

PR #4540, opened 2024-12-09
feat: Change spacing alt-click behavior to reset only one value

kof:
## Description

- When using alt modifier, on a spacing value, currently you can't reset one value only, you would reset the opposite value too.
- When alt-click happens outside of label-button, no reset should happen
- when using input in spacing, hodling alt/shift and then pressing enter allows to set opposite and all sides

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

johnsicili:
Can you make it so that when you option click outside the input, it pulls up the input? On shift it works like this.

Reason: I want to pull up the input with option so that when I save it, both sides save. Without that the workflow is reset current side, option open input, save both sides.

https://github.com/user-attachments/assets/894a4d6e-ed79-4e3c-b747-f31d3c132ceb

johnsicili (changes requested):
Good except can't reset all sides (e.g. shift + option + click)

kof:
> Good except can't reset all sides (e.g. shift + option + click)

I could make that work too, wasn't planing on it though, essentially if modifier + click, then reset and don't show input

johnsicili:
But how would you reset all sides then? 4 clicks? 

Also I think it's confusing visually when I hold option and hover over a label - it highlights both sides but the action only affects one side.

kof:
> But how would you reset all sides then? 4 clicks?

shift+click

kof:
> Also I think it's confusing visually when I hold option and hover over a label - it highlights both sides but the action only affects one side.

Yeah, wouldn't be too concerned about it, maybe hover state on the label would add a bit clarity, like on other labels

kof:
but yeah, technically the highlighted option state should disappear when hovering label

johnsicili:
> shift+click

This opens up the input.

So shift + click then backspace then enter?

kof:
oh your are right, it should be alt + shift to reset all sides

johnsicili (approved):
Yep looks good. Only imperfection is when hovering over one of the numbers and holding option highlights both sides. Everything looks good, glad to have it back

PR #4546, opened 2024-12-10
fix: navigator dnd bugs

TrySound:
- fixed dropping into non-container instances (like image)
- fixed dropping instances from another block


istarkov (approved):
🔥

PR #4549, opened 2024-12-10
refactor: upgrade to react 19

TrySound:
Let's give it a try.


kof:
after this launch later this week or monday next week pls

kof:
do we want to upgrade or are we blocked on something and need to close?

PR #4554, opened 2024-12-11
fix: relax list item constraint

TrySound:
Latest constraints improvements let us support direct child constraints. Which can be useful for very coupled components. Though in case of list and list item should be avoided because list item can be used inside of slot and collection.


kof (approved):

kof:
I think constraints need to be able to express non-direct parents, e.g. ul > div > li, where li has to have a parent ul somewhere but not necessarily direct parent

PR #4555, opened 2024-12-11
fix: show whole block tree to support radix components

TrySound:
We decided to show whole block tree so clients could open popover or tab content and edit text inside.

image


kof (approved):

istarkov:
image
In latest Content Block if Latest Child is open I can't drag anything on the last place.

johnsicili (approved):

PR #4566, opened 2024-12-11
feat: Allow maximizing CSS value for complex values

kof:
## Description

When the value is not a simple keyword or unit, we now show a maximize button in the css value input and that lets you edit comfortably a long value.

When dialog gets closed, value gets applied.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -157,7 +157,7 @@ const getChromeLayout = ({
 
   if (navigatorLayout === "undocked" && activeSidebarPanel !== "none") {
     return {
-      gridTemplateColumns: `auto ${theme.spacing[30]} 1fr ${theme.spacing[30]}`,
+      gridTemplateColumns: `auto ${theme.sizes.sidebarWidth} 1fr ${theme.sizes.sidebarWidth}`,
       gridTemplateAreas: `
             "header header header header"
             "sidebar navigator main inspector"
@@ -167,7 +167,7 @@ const getChromeLayout = ({
   }
 
   return {
-    gridTemplateColumns: `auto 1fr ${theme.spacing[30]}`,
+    gridTemplateColumns: `auto 1fr ${theme.sizes.sidebarWidth}`,
     gridTemplateAreas: `
           "header header header"
           "sidebar main inspector"
--- a/apps/builder/app/builder/builder.tsx
+++ b/apps/builder/app/builder/builder.tsx
@@ -157,7 +157,7 @@ const getChromeLayout = ({
 
   if (navigatorLayout === "undocked" && activeSidebarPanel !== "none") {
     return {
-      gridTemplateColumns: `auto ${theme.sizes.sidebarWidth} 1fr ${theme.sizes.sidebarWidth}`,
+      gridTemplateColumns: `auto ${theme.spacing[30]} 1fr ${theme.spacing[30]}`,
       gridTemplateAreas: `
             "header header header header"
             "sidebar navigator main inspector"
@@ -167,7 +167,7 @@ const getChromeLayout = ({
   }
 
   return {
-    gridTemplateColumns: `auto 1fr ${theme.sizes.sidebarWidth}`,
+    gridTemplateColumns: `auto 1fr ${theme.spacing[30]}`,
     gridTemplateAreas: `
           "header header header"
           "sidebar main inspector"
--- a/apps/builder/app/builder/shared/code-editor-base.tsx
+++ b/apps/builder/app/builder/shared/code-editor-base.tsx
@@ -416,29 +416,3 @@ export const EditorDialog = ({
     </Dialog>
   );
 };
-
-export const CodeEditorBase = ({
-  title,
-  open,
-  content,
-  onOpenChange,
-}: {
-  title?: ReactNode;
-  open?: boolean;
-  content: ReactNode;
-  onOpenChange?: (newOpen: boolean) => void;
-}) => {
-  return (
-    <EditorDialogControl>
-      {content}
-      <EditorDialog
-        open={open}
-        onOpenChange={onOpenChange}
-        title={title}
-        content={content}
-      >
-        <EditorDialogButton />
-      </EditorDialog>
-    </EditorDialogControl>
-  );
-};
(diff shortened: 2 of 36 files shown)


kof:
> Only show it when it meets a character threshold

Already had this implemented, turned out bad idea because it comes and goes and this flashing leads to jumping and is annoying

kof:
> There are css functions like clamp or calc

var() can also contain a very long variable, in fact most of them in craft don't fit in a field like width

kof:
> No autocomplete

True, probably need to add autocompletion, we can do that

kof:
> Can't see rendered values until you click out

I see 2 options:
1. add a shortcut, e.g. cmd+s and make it also work in all other editors like expression
2.  let blur of the input do this as well (not sure how practical but we do this in expressions too)

kof:
Additionally maybe it makes sense to set the initial position and size to a smaller size and somewhere above the style panel to avoid covering the canvas

Or alternatively just let user keep the position/size after they resize/move .... thoughts? This would be also good for all dialogs

johnsicili:
> True, probably need to add autocompletion, we can do that

I think we have code for this as it works in in some areas IIRC

johnsicili:
> Additionally maybe it makes sense to set the initial position and size to a smaller size and somewhere above the style panel to avoid covering the canvas Or alternatively just let user keep the position/size after they resize/move .... thoughts? This would be also good for all dialogs

Idk if dialog is the solution here though. I would try to see how many inputs overflow when not showing var() and then figure out out of those, how many don't benefit from the input scroll. It's mainly CSS functions I think. Maybe some colors too but those have more room for bigger widths

kof:
One thing I can think of is showing the maximize button only on hover and when focused, so normally you don't see it, we actually do the same thing in other areas.

kof:
Actually we even show it only on hover, not on focus, so it doesn't take extra space and if it covers important text, you just move your cursor away

johnsicili:
I don't like it. You're making me think to much as a user. Plus, the UX is not as good in the dialog.

What if we somehow expanded the field to be 100% and move the others out of the way? Might be too busy but something that just works

kof:
> What if we somehow expanded the field to be 100% and move the others out of the way? Might be too busy but something that just works

Thought about this too, will solve majority of cases but not all. In some complex cases you really want more space to write a complex css value in a calc or complx oklch thing. We also need space for AI in all code inputs so its easier to have a universal UI that handles these things and that has universally AI added to the interface.

kof:
In fact I think majority of cases will be fine with the default input without even opening it, most cases are not complex enough even if they are longer, so its fine, all vars and most calc() calls will be fine without maximizing. Maximizing is really for complex values.

johnsicili:
I see maximize button on any input that has a variable regardless of state (e.g. hover): 

Screenshot 2024-12-11 at 1 47 40 PM

Secondly, we are reducing the available space for text with this button:

Before:

Screenshot 2024-12-11 at 1 48 07 PM

After: 
Screenshot 2024-12-11 at 1 48 18 PM

>  all vars and most calc() calls will be fine without maximizing. 

I don't think I've used calc once where I didn't need a bigger space. Calc nearly alway uses a variable so at a minimum it will be this for me:

`calc(var(--size-2) * 2)`

Resulting in just this visible `calc(var(--size`

Calc is the exact scenario I would _want_ something like the proposed solution.

kof:
> I see maximize button on any input that has a variable regardless of state (e.g. hover):

Yes, I just realized we need to do this on hover only

kof:
> Secondly, we are reducing the available space for text with this button:

Yes, i won't reduce space, it will cover the content, just like in other places

kof:
> Calc is the exact scenario I would want something like the proposed solution.

exactly, though calc on itself is relatively rare, that's why I said most cases will be fine with just input

TrySound (approved):

johnsicili:
1. Say I want to use calc in an empty input. I must type in some dummy calc to then get the expand window. Not sure what to do here, but its a key place where this is useful. Maybe in the unit dropdown we have like "Fn" or something and you click that and it sets the field in function mode aka no units and expand button.
2. Seems unnecessary when the current input is small
Screenshot 2024-12-12 at 7 13 19 AM
3. The existing expand window in gradients reflects changes in realtime on the canvas. However, the expand window in this PR does not. 1. we should stay consistent and 2. its much better for experimentation to have changes reflect without clicking out

apps/builder/app/builder/shared/code-editor-base.tsx
```diff
@@ -420,14 +420,14 @@ export const EditorDialog = ({
 export const CodeEditorBase = ({
```
TrySound, line 420: Everything inside this component already reusable and composable. There is no sense in this component as abstraction like this.

kof: done

apps/builder/app/builder/builder.tsx
```diff
@@ -157,7 +157,7 @@ const getChromeLayout = ({
 
   if (navigatorLayout === "undocked" && activeSidebarPanel !== "none") {
     return {
-      gridTemplateColumns: `auto ${theme.spacing[30]} 1fr ${theme.spacing[30]}`,
+      gridTemplateColumns: `auto ${theme.sizes.sidebarWidth} 1fr ${theme.sizes.sidebarWidth}`,
```
TrySound, line 160: You can do PR with this first

johnsicili (approved):

apps/builder/app/builder/builder.tsx
```diff
@@ -157,7 +157,7 @@ const getChromeLayout = ({
 
   if (navigatorLayout === "undocked" && activeSidebarPanel !== "none") {
     return {
-      gridTemplateColumns: `auto ${theme.sizes.sidebarWidth} 1fr ${theme.sizes.sidebarWidth}`,
+      gridTemplateColumns: `auto ${theme.spacing[30]} 1fr ${theme.spacing[30]}`,
```
TrySound, line 160: rebase?

kof: done

PR #4593, opened 2024-12-16
fix: Pasting gradient into color field breaks CSS parser

### LINKED ISSUE #4578: Pasting gradient into color field breaks CSS parser
When pasting linear-gradient(green,red) into color, parser throws an error we don't catch and don't show to the user error state

istarkov:
## Description

closes #4578

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


TrySound (approved):

kof (approved):
don't you think we should do a try/catch somewhere in case of some other unpredictable scenario lilke that that will. throw from the parser?

PR #4594, opened 2024-12-16
experimental: Commands UI design tweaks

kof:
## Description

- Bigger border radius for the dialog and floating panels as well for consistency
- Bigger search container height, same as titles in dialogs

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/design-system/src/components/command.tsx
```diff
@@ -30,23 +30,23 @@ import { Button } from "./button";
...
   border: `1px solid ${theme.colors.borderMain}`,
+  borderRadius: theme.borderRadius[7],
   // clip selected item background
   overflow: "clip",
   // remove input border bottom when no command matches
-  [inputBorderBottomSize]: "0px",
+  [inputBorderBottomSize]: 0,
   "&:has([cmdk-group]:not([hidden]))": {
-    [inputBorderBottomSize]: "1px",
+    [inputBorderBottomSize]: 1,
```
TrySound, line 49: You sure `1 solid red` will work?

packages/design-system/src/components/command.tsx
```diff
@@ -30,23 +30,23 @@ import { Button } from "./button";
 import { Popover, PopoverContent, PopoverTrigger } from "./popover";
 import { Kbd } from "./kbd";
 
-const panelWidth = "400px";
-const itemHeight = "32px";
+const panelWidth = 400;
+const itemHeight = 32;
```
TrySound, line 34: please let's not rely on stitches magic where it is not colocated

packages/design-system/src/components/command.tsx
```diff
@@ -101,8 +101,8 @@ export const Command = (props: CommandProps) => {
 
 const CommandDialogContent = styled(DialogContent, {
   position: "absolute",
-  top: 140,
-  left: `calc(50% - ${panelWidth} / 2)`,
```
TrySound, line 105: there is nothing wrong with having it in calc

johnsicili (approved):
Good. Drag handles are a little to close to the edge imo but very minor

Screenshot 2024-12-16 at 7 01 17 AM

packages/design-system/src/components/command.tsx
kof: no, but we compile it to 1px, so we can use numbers, right?

packages/design-system/src/components/command.tsx
kof: hmm ok

packages/design-system/src/components/command.tsx
TrySound: who compiles it? Stitches magic? I have no idea if it compiles custom properties or not. Too magic.

PR #4603, opened 2024-12-17
fix: stop requiring dialog title and description

TrySound:
Ref https://discord.com/channels/955905230107738152/1318292995023306803/1318292995023306803

Here removed constraints for dialog title and dialog description in dialog content. Users often delete them though radix starts throwing warnings about it for a while now.


kof:
Radix will still have warnings if these are deleted, right? I think radix requires them for accessiblity reasons and simply not providing it will be a problem for acessibility. Probably some form of title still needs to be provided but in a way that's hidden or something?

kof (approved):

PR #4604, opened 2024-12-17
refactor: split jsx testing utils into template package

TrySound:
We can start using testing utilities for components meta. It will drastically simplify maintenance of components, already works pretty well for tests.

Here just replaced @webstudio-is/sdk/testing with
@webstudio-is/template.


kof (approved):

kof:
NICE, can you make one meta to see how this works pls? I can help rewrite them

TrySound:
There are still missing pieces here

1. builtin local styles
2. cli to generate serializable version

PR #4613, opened 2024-12-18
feat: Floating panel v2

kof:
Closes https://github.com/webstudio-is/webstudio/issues/4572

## Description

- All floating panels are now using a dialog and are draggable (bindings and expressions not migrated yet, separate PR)
- Logic for positioning the panels is now in our control and easier to manage, because we have lots of special cases
- Deleted FloatingPanelPopover, now everything uses either FloatingPanel or Popover, difference is in positioning logic, FloatingPanel is using Dialog
- Big center positioned dialogs like code for html embed allow having another panel open at the same time, wile all other floating panels close the other panel when clicked
- Most extended settings in style panel are now positioned above style panel, not on the left

## Steps for reproduction

1. click button
3. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/shared/floating-panel.tsx
```diff
@@ -74,18 +84,20 @@ export const FloatingPanelProvider = ({
...
+  content: ReactNode;
+  children: ReactNode;
+  maximizable?: boolean;
+  resize?: ComponentProps<typeof DialogContent>["resize"];
+  width?: number;
+  height?: number;
+  // - container - aligns the dialog above the container like left or right sidebar
+  // - left - aligns the dialog on the left side of the container
+  // - center - aligns the dialog in the center of the screen
+  align?: "left" | "center" | "container";
```
kof, line 98: @TrySound I am still not quite happy with these semantics

apps/builder/app/builder/shared/floating-panel.tsx
```diff
@@ -96,32 +108,116 @@ export const FloatingPanel = ({
...
+    x: undefined,
+    y: undefined,
+    width,
+    height,
+  });
+
+  const rectRef = useRef(rect);
+  rectRef.current = rect;
+
+  useEffect(() => {
```
kof, line 142: need to rewrite without using effect

TrySound (approved):

johnsicili:
## 1
1. Set resource
2. Drag window
3. Open bindings for URL
4. See new window not relative to first window

Screenshot 2024-12-18 at 4 00 12 PM

## 2
Not sure if they should move together

https://github.com/user-attachments/assets/4e9b3aa5-e5e3-40ca-bf1c-e40e9bbd654b

## 3
Idk if related but project setting code actions are gone like closing window

https://github.com/user-attachments/assets/51983bbc-ad9a-4c4e-9802-7539b38179e0

## 4
Discoverability is hard

Breakpoint window not draggable, project settings window not draggable, but to me they look the same.

Though idk how much I'll be using this/how important it is from a user perspective.

## 5

Open bindings in page settings, window is off.

Screenshot 2024-12-18 at 4 07 58 PM

kof:
1. Second window opens relative to the trigger that was clicked, not sure if it was the case when you tested, but seems as expected. Also these 2 are open at the same time only because the first one had errors and couldn't close

kof:
4. yeah, in case of project settings enabling draggable now, others, when there is no title/close button I think its intuitive to not expect them to be draggable. Maybe removing close button from publish and share dialogs will make this more clear in this case, they are basically popovers, not dialogs, but yeah

I felt like using the move cursor to the title bar is a bit too distracting, figma is not doing it either where its draggable, but idk

kof:
5. fixed

TrySound (approved):
maybe ok

PR #4621, opened 2024-12-19
feat: RichText edit

istarkov:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/features/navigator/navigator-tree.tsx
```diff
@@ -655,6 +655,7 @@ export const NavigatorTree = () => {
                   onClick: () => selectInstance(item.selector),
                   onFocus: () => selectInstance(item.selector),
                   onKeyDown: (event) => {
+                    console.info("onKeyDown", event.key);
```
TrySound, line 658: is it necessary?

johnsicili:
Content Feedback @istarkov 

1. I can see a use case for insert within like if you are in a list, add a link or something, it will get inserted after. Low priority.
2. Select the default text when adding. E.g., add paragraph, select text so I can start typing without CMD A backspace. Might be difficult though if we can’t detect default text. Just thinking about writing a page of content having to cmd a delete each time after add
3. I think enter should drop the cursor to the next “row”. Seems enter is pretty useless inside paragraphs, headings, etc. If anything, only shift enter should be used. Plus not allowing enter will prevent misuse of line breaks in paragraphs. So maybe we should do something in between auto adding a new paragraph and adding line breaks upon enter?
4. Seems odd to have to click into an existing instance to add a new one. Not sure what to do here.. maybe what Notion does.
5. Come to think of it, Slash command is useful when it auto adds a paragraph. It’s like a way to convert the existing paragraph to something else. What if we just make “enter” pull up the add dialog? I think it might be more intuitive than adding a / after text like “end of my sentence./“
6. Content mode feedback: Maybe we shouldn’t allot text edit in boxes I don’t think. I just realized when using slash, it put my cursor in an empty box.

Screenshot 2024-12-19 at 10 01 53 AM

8. Content mode feedback: CMD K still works in content mode
9. When using slash in content mode and not in a content block, it’s just a slash. Maybe we need some indication what’s going on here.

istarkov:
> 2 Select the default text when adding. E.g., add parag...

If you have an empty paragraph there is no need to `select all`. 
image

> 3 Enter ....
not implemented here is `/` only

> 4 Seems odd to have to click into an existing ...
What this mean?

> 6 Content mode feedback: Maybe we shouldn’t allot text edit in boxes I don’t think.
Until boxes are editable and they are, there is no way to distinguish it vs any other editable content. Some flag probably.

> 8 CMD K still works in content mode
cmd k implementation bug.

> 9 When using slash in content mode and not in a co
Same in webstudio

johnsicili:
4 is the biggest feedback so let me explain.

I think we should use "enter" instead of "slash".

Slash is a really great command, but in every implementation of it that I've seen, you use it on a blank line, which is essentially a paragraph.

For our implementation, we are using it to add a new block.

The difference is in the common implementation of it, slash is used to convert blank lines, which are paragraphs, to another element.

In our implementation, slash is used to _create_ a new line and component at the same time.

So I think unless we have a paragraph by default on enter, then "Enter" should be used to add something new. Type, enter, add component. 

Also, typing slash within existing content feels odd.

Just a little video show how slash is used on new lines:
https://github.com/user-attachments/assets/9c7d146c-8092-46c5-8798-44f597b3ca22

apps/builder/app/builder/features/workspace/canvas-tools/block-editor-context-menu.tsx
```diff
@@ -0,0 +1,228 @@
...
+  border: "none",
+  overflow: "hidden",
+  padding: 0,
+});
+
+const InertController = ({
+  // value,
+  onChange,
+}: {
+  // value: boolean;
```
TrySound, line 38: no need?

apps/builder/app/builder/features/workspace/canvas-tools/block-editor-context-menu.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const timeout = setTimeout(() => {
+      handleChange(false);
+    }, 0);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [handleChange]);
+
+  return null;
```
TrySound, line 53: just `return`?

apps/builder/app/builder/features/workspace/canvas-tools/outline/block-instance-outline.tsx
```diff
@@ -465,40 +363,46 @@ export const BlockChildHoveredInstanceOutline = () => {
...
-            </Tooltip>
+                setButtonOutline(undefined);
+                $blockChildOutline.set(undefined);
+                $hoveredInstanceSelector.set(undefined);
+                $hoveredInstanceOutline.set(undefined);
+              }}
+              css={{
+                borderStyle: "solid",
+                borderColor: isAddMode
+                  ? `oklch(from ${theme.colors.backgroundPrimary} l c h / 0.7)`
```
TrySound, line 400: 😉

apps/builder/app/builder/features/workspace/canvas-tools/block-editor-context-menu.tsx
```diff
@@ -0,0 +1,228 @@
...
+                return { repeat: prev.repeat + 1, templates: [] };
+              }
+
+              return { repeat: 0, templates: filteredTemplates };
+            });
+
+            setValue(filteredTemplates[0]?.[1] ?? undefined);
+          }
+
+          break;
```
TrySound, line 124: put "break" inside of block to reduce nesting

apps/builder/app/builder/features/workspace/canvas-tools/block-editor-context-menu.tsx
```diff
@@ -0,0 +1,228 @@
...
+  templates,
+}: {
+  cursorRect: DOMRect;
+  anchor: InstanceSelector;
+  templates: [instance: Instance, instanceSelector: InstanceSelector][];
+}) => {
+  const [inert, setInert] = useState(true);
+  const modifierKeys = useStore($modifierKeys);
+  const scale = useStore($scale);
+  // const textEditorContextMenuCommand = useStore($textEditorContextMenuCommand);
```
TrySound, line 74: no need?

apps/builder/app/canvas/features/text-editor/text-editor.tsx
```diff
@@ -1004,6 +1297,11 @@ export const TextEditor = ({
 
   const handleChange = useEffectEvent((editorState: EditorState) => {
     editorState.read(() => {
+      // Otherwise editorState.read captures focus
+      if ($textEditorContextMenu.get() !== undefined) {
+        //return;
```
TrySound, line 1302: no need?

TrySound (approved):

apps/builder/app/builder/features/workspace/canvas-tools/block-editor-context-menu.tsx
istarkov: alt text

istarkov: image

PR #4624, opened 2024-12-20
fix: support command panel only in design mode

TrySound:
We do not want users to mess with command panel in content or design mode.


kof (approved):

apps/builder/app/builder/shared/commands.ts
```diff
@@ -496,6 +496,9 @@ export const { emitCommand, subscribeCommands } = createCommandsEmitter({
       hidden: true,
       defaultHotkeys: ["meta+k", "ctrl+k"],
       handler: () => {
+        if ($isDesignMode.get()) {
```
kof, line 499: this reads wrong

kof (changes requested):
yep you made the opposite of what's needed)

kof (approved):

PR #4625, opened 2024-12-20
feat: allow finding tokens

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3901 https://github.com/webstudio-is/webstudio/issues/4608

Here added new command panel group: tokens. For now only support selecting which opens a list of instances the token is defined on.

https://github.com/user-attachments/assets/f6c5849a-b2f5-449b-9749-3faa65b7f33f

### DIFF
--- a/apps/builder/app/builder/features/command-panel/token-group.tsx
+++ b/apps/builder/app/builder/features/command-panel/token-group.tsx
@@ -1,0 +1,224 @@
+import { useState } from "react";
+import { matchSorter } from "match-sorter";
+import { computed } from "nanostores";
+import { useStore } from "@nanostores/react";
+import {
+  CommandGroup,
+  CommandGroupHeading,
+  CommandInput,
+  CommandItem,
+  CommandList,
+  Flex,
+  ScrollArea,
+  Text,
+  toast,
+  useSelectedAction,
+} from "@webstudio-is/design-system";
+import type { Instance, Instances, StyleSource } from "@webstudio-is/sdk";
+import {
+  $instances,
+  $pages,
+  $registeredComponentMetas,
+  $selectedStyleSources,
+  $styleSources,
+  $styleSourceSelections,
+} from "~/shared/nano-states";
+import { getInstanceLabel } from "~/shared/instance-utils";
+import type { InstanceSelector } from "~/shared/tree-utils";
+import { $awareness } from "~/shared/awareness";
+import { $commandContent, closeCommandPanel } from "./command-state";
+
+export type TokenOption = {
+  terms: string[];
+  type: "token";
+  token: Extract<StyleSource, { type: "token" }>;
+  usages: number;
+};
+
+const $styleSourceUsages = computed(
+  $styleSourceSelections,
+  (styleSourceSelections) => {
+    const styleSourceUsages = new Map<StyleSource["id"], Set<Instance["id"]>>();
+    for (const { instanceId, values } of styleSourceSelections.values()) {
+      for (const styleSourceId of values) {
+        let usages = styleSourceUsages.get(styleSourceId);
+        if (usages === undefined) {
+          usages = new Set();
+          styleSourceUsages.set(styleSourceId, usages);
+        }
+        usages.add(instanceId);
+      }
+    }
+    return styleSourceUsages;
+  }
+);
+
+export const $tokenOptions = computed(
+  [$styleSources, $styleSourceUsages],
+  (styleSources, styleSourceUsages) => {
+    const tokenOptions: TokenOption[] = [];
+    for (const styleSource of styleSources.values()) {
+      if (styleSource.type !== "token") {
+        continue;
+      }
+      tokenOptions.push({
+        terms: ["tokens", styleSource.name],
+        type: "token",
+        token: styleSource,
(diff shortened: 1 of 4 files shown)


apps/builder/app/builder/features/command-panel/token-group.tsx
```diff
@@ -0,0 +1,224 @@
...
+              </CommandItem>
+            ))}
+          </CommandList>
+        </ScrollArea>
+      </Flex>
+    </>
+  );
+};
+
+export const TokenGroup = ({ options }: { options: TokenOption[] }) => {
```
kof, line 189: Why do we need all these token specific components? Why can't tokens, instances, or variables search all be using the same components?

apps/builder/app/builder/features/command-panel/token-group.tsx
```diff
@@ -0,0 +1,224 @@
...
+import { $commandContent, closeCommandPanel } from "./command-state";
+
+export type TokenOption = {
+  terms: string[];
+  type: "token";
+  token: Extract<StyleSource, { type: "token" }>;
+  usages: number;
+};
+
+const $styleSourceUsages = computed(
```
kof, line 38: I understand that we use computed everywhere else, but search is a different case, you don't need to compute this stuff on every change, because its only needed when search gets open, so this computation is wasted as long as the search is not open

kof: IMO search data should be gathered only when search interface opens + all data should be normalized to one format, so that the search works the same way for all types of data

kof: Like a sqlite database that boots when you need it

apps/builder/app/builder/features/command-panel/token-group.tsx
TrySound: Caz all of them have different logic to implement, different groups. "Different functions or functions with conditions inside."

apps/builder/app/builder/features/command-panel/token-group.tsx
TrySound: Computed are invoked lazily when something is subscribed to them

TrySound: This is why all stored are subscribed inside of dialog

apps/builder/app/builder/features/command-panel/token-group.tsx
kof: you mean for actions? that could be a config per group that has an action name and action handler

apps/builder/app/builder/features/command-panel/token-group.tsx
kof: weird, I thought computed subscribes all the time

TrySound: It would completely kill performance

kof: when I use $something.get() it doesn't know if there is a subscription, will get invoke computed as well then?

apps/builder/app/builder/features/command-panel/token-group.tsx
TrySound: Each group look differently. Has different requirements. We don't need to make it look the same.

apps/builder/app/builder/features/command-panel/token-group.tsx
TrySound: One time subscription to recompute all stuff when .get is invoked.

apps/builder/app/builder/features/command-panel/token-group.tsx
kof: Hmm It sounds like lots of code while we don't know if we need this level of customization, I would expect different categories to have different actions, but nothing else

TrySound: image
image
image

kof (approved):

johnsicili (approved):
Nice. I think it would be helpful to show the page in commands where we usually show the shortcuts.

Eg

Heading 1                      About

Might help understand usage better at a glance.

kof:
> Nice. I think it would be helpful to show the page in commands where we usually show the shortcuts.
> 
> Eg
> 
> Heading 1 About
> 
> Might help understand usage better at a glance.

I was also thinking of making a special shortcuts overview page, but I think this is a better idea:

1. we need a help center anyways, so we are probably going to link the docs in the iframe for this
2.shortcuts is already in the docs

If we do this, showing shortcuts help is basically opening help center with shortcuts page open, thoughts?

johnsicili:
Actually I meant have info in the panel about where those tokens are used:

Screenshot 2024-12-22 at 10 46 23 AM

So on the right side it would say the page that token is on

kof:
> Actually I meant have info in the panel about where those tokens are used:

Those are instances. Bogdan is adding a category title "Instances", so that there is no confusion and also a button "back" to get back to the previous view

kof:
Showing which page the instance is on can be difficult, especially if its on many pages, but one can separately search the Instance name and see this.

PR #4643, opened 2024-12-22
chore: Improved components list organization

kof:
## Description

1. Rename Text category to Typography
2. Move Content Block to typography
3. move Time to a standalone new "Localization" category
4. move Media section next after Typography

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/time.ws.ts
+++ b/packages/sdk-components-react/src/time.ws.ts
@@ -16,14 +16,13 @@ const presetStyle = {
 } satisfies PresetStyle<"time">;
 
 export const meta: WsComponentMeta = {
-  category: "data",
+  category: "localization",
   type: "container",
   description:
     "Converts machine-readable date and time to a human-readable format.",
   icon: CalendarIcon,
   states: defaultStates,
   presetStyle,
-  order: 5,
 };
 
 export const propsMeta: WsComponentPropsMeta = {
--- a/packages/react-sdk/src/components/component-meta.ts
+++ b/packages/react-sdk/src/components/component-meta.ts
@@ -24,6 +24,7 @@ export const componentCategories = [
   "media",
   "data",
   "forms",
+  "localization",
   "radix",
   "xml",
   "hidden",
--- a/packages/react-sdk/src/components/component-meta.ts
+++ b/packages/react-sdk/src/components/component-meta.ts
@@ -20,9 +20,9 @@ export type WsComponentPropsMeta = z.infer<typeof WsComponentPropsMeta>;
 
 export const componentCategories = [
   "general",
-  "text",
-  "data",
+  "typography",
   "media",
+  "data",
   "forms",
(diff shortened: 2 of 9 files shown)


johnsicili (approved):

packages/sdk-components-react/src/time.ws.ts
```diff
@@ -16,14 +16,13 @@ const presetStyle = {
 } satisfies PresetStyle<"time">;
 
 export const meta: WsComponentMeta = {
-  category: "data",
+  category: "localization",
```
TrySound, line 19: Single component category doesn't look good. And it is still fits with "data". Why add now?

TrySound:
List and list item are also typography

TrySound:
The content block could go into general

kof:
> List and list item are also typography

I agree, I guess I was influenced by Webflow's choice there ... but thinking about it lists is basically part of content authoring ...

kof:
> Then content block could go into general

Hmm difficult choice, I was trying to put all the content authoring things into one place but yeah Content Block is more than Typography.

What if we called it "Content" instead of Typography and kept things like we have them there now + lists?

kof:
The problem with "Content" is that everything is content, including media category

kof:
or alternatively keeping it at Typography and Content Block there as well. Thinking of: content block is more than text, but in the end this component is enabling user to edit text in more complex layouts, but in the end the goal is text editing, which is closer to typography than to anything else ... I think it should stay in Typography

packages/sdk-components-react/src/time.ws.ts
kof: The reason anybody would use Time component is localizing date/time format. They don't need it if they want to just type some date and that has also very little to do with the fact data is remote. Remote nature of data in this case is coinscidence.

IMO user will think: "How do I localize the date" when trying to find that component in the panel

kof: I agree single component doesn't look good, lets add currency localization, this is very needed soon for shopify

kof:
Regarding list and list item, the question is if having it in typography isn't confusing for someone who wants to use a list to build a menu of some kind

TrySound:
> Regarding list and list item, the question is if having it in typography isn't confusing for someone who wants to use a list to build a menu of some kind

There was never a clear reason why people use lists for menus. They actually harm a11y. So it better be confusing.

kof:
> > Regarding list and list item, the question is if having it in typography isn't confusing for someone who wants to use a list to build a menu of some kind
> 
> There was never a clear reason why people use lists for menus. They actually harm a11y. So it better be confusing.

I guess historically this was the way to go, but we can move past that now

Issue #4595, opened 2024-12-16
Rich Text component


kof:
Content block is a better version of Rich Text, here is what's missing

- [x] slash command - lists the available instances inside templates, same as + button.  #4621
- [x] enter - creates a paragraph if paragrpah is available in templates
then `/` on empty paragraph if it's empty after adding a new block then we delete current block #4627

- [x] if there is no paragraph - newline
- [x] updated list of template instances according to wf richtext
- [ ] decide on the image block template to have a bunch of different  options for the layout
- [x] backspace on empty paragraph - deletes the paragraph and moves caret to the previous one #4627

- [x] UL/LI support #4633
- [x] Would be nice to auto exit after entering within an empty list item. #4641
- [x] Move `editablePlaceholderComponents` into meta https://github.com/webstudio-is/webstudio/blob/cd27371cc21445b0b841c9942047629ac652b99d/apps/builder/app/canvas/shared/styles.ts#L70
- [x] Placeholders #4644

Add:

- [x] - Select all text on new  #4627

PR #4644, opened 2024-12-23
experimental: Use placeholders for editable elements

istarkov:
## Description

ref #4595

### TODO:
- [x] - Found bug that deleting last LI does not delete UL
- [x] - Consider removing default texts for paragraph link blockquote and headings i.e. `Heading text you can edit`

https://p-cfc5b051-3b7a-4a59-a994-44d8d2f8b836-dot-styles.development.webstudio.is/

The following components are considered non-collapsible in **Edit** and **Content** modes:  
*Paragraph*, *Heading*, *ListItem*, *Blockquote*, *Link*.

We are adding placeholders for these components when they are empty, using the following logic (partially inspired by Notion):  

1. **If a component is empty and not being edited**, we display the same label as shown in the components tree.  
   image

2. **If a component is empty and being edited**, the behavior depends on its location:  
   - **Inside a content block:**  
     - For *Paragraph*, we display: `Write something or press '/' for commands...`.  
     - For all other components, we display the label.  
   - **Outside a content block:**  
     - Always display the label.  

image

### Update: Default Content Block Component Changed

The default content block component has been updated. Add the new component

### PS
On Publish and Preview previous behaviour.

## Steps for reproduction

1. click button
4. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/canvas/shared/styles.ts
+++ b/apps/builder/app/canvas/shared/styles.ts
@@ -72,6 +72,7 @@ export const editablePlaceholderComponents = [
   "Heading",
   "ListItem",
   "Blockquote",
+  "Link",
 ];
 
 const editablePlaceholderSelector = editablePlaceholderComponents
--- a/apps/builder/app/canvas/shared/styles.ts
+++ b/apps/builder/app/canvas/shared/styles.ts
@@ -61,6 +61,25 @@ export const mountStyles = () => {
   helpersSheet.render();
 };
 
+/*
+div[contenteditable]:has(p > br:only-child) {
+  background-color: red;
+}
+*/
+
+const globalHelperStyles = [
+  `:is(p, h1, h2, h3, h4, h5, h6)[${idAttribute}]:empty::before {
+    content: '\\200B';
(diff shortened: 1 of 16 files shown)


johnsicili:
I like this a lot!

1. Make instead of use / for commands it should say 'templates'?
2. Make default UL and OL only have 1 LI (multiple aren't needed since adding new ones are possible via enter)

johnsicili:
Bug fix looks good

apps/builder/app/canvas/shared/styles.ts
```diff
@@ -61,7 +64,44 @@ export const mountStyles = () => {
   helpersSheet.render();
 };
 
+/**
+ * Opinionated list of non collapsible components in the builder
+ */
+export const editablePlaceholderComponents = [
```
kof, line 70: Hmm, maybe instead of that it should be collapsible setting in meta?

kof (approved):

PR #4662, opened 2024-12-27
refactor: rewrite sdk react templates with jsx

TrySound:
Here added templates for all sdk react components (except forms, no variables support yet). Dropped content-embed.ws.ts as there is no ContentEmbed component.


kof (approved):
didn't even know content embed doens't have any custom logic to it, why did we decide to have it as a separate component back then?

TrySound:
Descendants are not available to users yet

PR #4672, opened 2024-12-29
fix: relax detaching constraint

TrySound:
Every time somebody is facing invalid constraint issue whole tree is blocked from deleting instances except the problematic one with actual issue.

Here rewrote detachability check to validate only ancestors and avoid traversing potentially problematic siblings.


kof (approved):

johnsicili (changes requested):
1. Add nav menu
2. Del menu content
3. See error 

https://github.com/user-attachments/assets/630a51e3-f825-40d2-948a-64833e4ce368

2. Also body in prod is "detachable" as it shows the dragging ability

istarkov (approved):

TrySound:
Fixed 1 and 2. And yes we replaced "detachable" flag with constraints so instead of draggability we check if instance can be dropped considering all constraints.

PR #4676, opened 2024-12-30
feat: support folding text fragments in code editor

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/3636

Enabled folding for dialog editors only to avoid bloating space in settings panel.

Screenshot 2024-12-30 at 19 42 03
Screenshot 2024-12-30 at 19 42 55


johnsicili (approved):
good one

PR #4685, opened 2024-12-31
feat: Mark empty style sources

kof:
[Discussion](https://discord.com/channels/955905230107738152/1323031888281075793)

## Description

As a designer, I want to know at a glance if the style source has no defined values. Cases where this is useful:
1. Is anything defined on Local? E..g you sometimes forget to switch the token and define things on local by accident
2. Is a token not having any value after it was created because value went on some other token? again, possibly because token wasn't selected

Empty token should be a red flag, should basically never happen, but empty local source would be most of the time

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili (changes requested):
I run into this a lot. On local more than anything. However I don't like the approach.

## The dot doesn't communicate well
1. We use dot to indicate a state _has_ styles. Now we are also using it to say that the style source does _not_ have styles.
2. The dot by itself doesn't make much sense. The dot on states is intuitive; change a style, it shows up.

Screenshot 2024-12-30 at 6 50 16 PM


## Solution

Maybe use the dropdown to add a dot when it _does_ have a style. Such as next to "Local" or the Token name.

Downside of this is that it takes a click to see, though possible this is okay because it's infrequently a thing and the benefits of it aligning with the other dots could outweigh any cons.

apps/builder/app/builder/features/style-panel/style-source/style-source-input.tsx
```diff
@@ -191,47 +225,50 @@ const TextFieldBase: ForwardRefRenderFunction<
...
-            <Flex align="center" justify="center">
-              <LocalStyleIcon />
-            </Flex>
-          ) : (
-            item.label
-          )}
-        </StyleSourceControl>
-      ))}
+      {value.map((item) => {
+        return (
```
TrySound, line 229: Seems like no need in extra nesting

kof:
@johnsicili the white dot was meant to be as a "level indicator being low", where blue dot is the thing is full, but yeah, the difference may not be clear

apps/builder/app/builder/features/style-panel/style-source/style-source-input.tsx
oslogrolls: John makes a great point here. 
Having collapsed panels and the states-pulldown, where you use dots to indicate “has styles” and adding a similar dot to indicate the opposite, “has no styles” makes no sense.

Simply inverting the dot-Logic in Style Sources (Local and Tags get a dot when they have styles) would, however, create a lot of noise – as the sole reason for Tokens to exist is holding data. Dots would therefore show up pretty much for any selected item. For that reason, I suggested **rather marking the exception** or error.

Placing a dot in the States dropdown, as indicated by John (see rough sketch) would clearly be an improvement – yet this required a click each time. 

Frame 56

I had [originally proposed](https://discord.com/channels/955905230107738152/1323031888281075793) several other ways to indicate the Empty state – they would not clash with the symbols / logic you already use.

I also added a new, subtle option at the bottom (a little cutout).
2024-12-31_10h55_34

My personal favourite is fully rounded. In a well-built project this style would **never** show for Tokens. Round would show you an error. For Local round would be the default. As it only appears on the left, it also indicates the end of the row.
2024-12-30_08h52_25

oslogrolls:
> I run into this a lot. On local more than anything. However I don't like the approach. The dot doesn't communicate well
John makes a great point here. 
Having collapsed panels and the states-pulldown, where you use dots to indicate “has styles” and adding a similar dot to indicate the opposite, “has no styles” makes no sense.

Simply inverting the dot-Logic in Style Sources (Local and Tags get a dot when they have styles) would, however, create a lot of noise – as the sole reason for Tokens to exist is holding data. Dots would therefore show up pretty much for any selected item. For that reason, I suggested **rather marking the exception** or error.

Placing a dot in the States dropdown, as indicated by John (see rough sketch) would clearly be an improvement – yet this required a click each time. 

Frame 56

I had [originally proposed](https://discord.com/channels/955905230107738152/1323031888281075793) several other ways to indicate the Empty state – they would not clash with the symbols / logic you already use.

I also added a new, subtle option at the bottom (a little cutout).
2024-12-31_10h55_34

My personal favourite is fully rounded. In a well-built project this style would **never** show for Tokens. Round would show you an error. For Local round would be the default. As it only appears on the left, it also indicates the end of the row.
2024-12-30_08h52_25

kof:
@oslogrolls I am afraid all these shapes are just not going to communicate clearly the matter ... user won't know what this means

kof:
Here is another idea:
1. add the blue dot right from Token/Local in the dropdown menu
2. additionally as a secondary indication do this so one can see if local has styles on local at a glance, because local is kinda the only case where you want to know if it has styles.
3. additionally we could add a red border as an indication of "error state", because for a token, not having any styles, should be considered an error state

image
image

image

image

kof:
Another idea is to add to a token something super obviously visible, when token has styles, that icon disappears

image

oslogrolls:
> Another idea is to add to a token something super obviously visible when token has no styles, when token has styles, that icon disappears

This would mean, the circle had to disappear on Local with data, though.

kof:
> This would mean, the circle had to disappear on Local with data, though.

For a full consistency yes, but local is special in any case, so we could do an exception here and show circle with a dot.

kof:
image

image

image

kof:
Also moved the dot on states to the right, so that its consistent and also better for alignment.
image

johnsicili:
I think this approach is better. I like local.

Bug:

1. Add paragraph
2. Add token that has styles
3. Change local styles
4. See no indicator on local


https://github.com/user-attachments/assets/ccc734bb-a75a-4f14-a66e-84e80b6f147b

kof:
@johnsicili bug should be fixed now

kof:
Also stopped capitalizing token names, since these are user-defined and should be left as-is
image

apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
```diff
@@ -241,6 +241,19 @@ const StyleSourceState = styled(Text, {
   },
 });
 
+const LocalStyle = styled(LocalStyleIcon, {
+  variants: {
+    hasStyles: {
+      true: {},
+      false: {
+        "& [fill]": {
```
TrySound, line 249: Too much reliance on icon internals. I would inline it and use css variable to hide circle.

TrySound (approved):

johnsicili (approved):

apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
kof: yeah, what you you do? add some explicit attribute?

PR #4692, opened 2025-01-01
fix: restore missing components in command panel

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4689

While refactoring with new component templates
forgot to register them in command panel.

Looks like the prod is affected a little too


kof (approved):
needs hotfixing?

PR #4708, opened 2025-01-03
fix: Floating panel and dialog bugs

kof:
## Description

- [ ] fix maximize/minimize size/positioning
   - open, move, maximize, minimize
   - should be in original position
   

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/shared/css-value-input/value-editor-dialog.tsx
+++ b/apps/builder/app/builder/features/style-panel/shared/css-value-input/value-editor-dialog.tsx
@@ -92,7 +92,7 @@ export const ValueEditorDialog = ({
       title="CSS Value"
       placement="bottom"
       height={200}
-      width={Number(rawTheme.sizes.sidebarWidth)}
+      width={parseFloat(rawTheme.sizes.sidebarWidth)}
       content={
         <CssFragmentEditorContent
           autoFocus
--- a/packages/design-system/src/components/dialog.tsx
+++ b/packages/design-system/src/components/dialog.tsx
@@ -8,6 +8,8 @@ import {
   createContext,
   useContext,
   useState,
+  useEffect,
+  useCallback,
 } from "react";
 import * as Primitive from "@radix-ui/react-dialog";
 import { css, theme, type CSS } from "../stitches.config";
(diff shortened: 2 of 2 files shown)


apps/builder/app/builder/features/style-panel/shared/css-value-input/value-editor-dialog.tsx
```diff
@@ -92,7 +92,7 @@ export const ValueEditorDialog = ({
       title="CSS Value"
       placement="bottom"
       height={200}
-      width={Number(rawTheme.sizes.sidebarWidth)}
+      width={parseFloat(rawTheme.sizes.sidebarWidth)}
```
TrySound, line 95: Let's take both Number.parseFloat

packages/design-system/src/utilities.ts
```diff
@@ -94,9 +94,9 @@ export const useResize = ({
   onResizeEnd,
   timeout = 300,
 }: {
-  onResizeStart?: () => void;
-  onResize?: () => void;
-  onResizeEnd?: () => void;
+  onResizeStart?: (entries: Array<ResizeObserverEntry>) => void;
+  onResize?: (entries: Array<ResizeObserverEntry>) => void;
+  onResizeEnd?: (entries: Array<ResizeObserverEntry>) => void;
```
TrySound, line 99: where entries are used?

kof: I was in the middle of refactoring and used them, then removed the usage, thought its still useful potentially since the utility is generic.

johnsicili (changes requested):
## 1. Bindings open in middle

1. Select an instance on the canvas
2. Next click, open a binding
3. See it in the middle of the page 

_Bindings open in unpredictable spots often. Will need to test this a lot_

https://github.com/user-attachments/assets/67d3dce4-3649-4796-a8a6-a38a2d62b059

## 2. Style panel scroll with open dialog

1. Open complex value
2. Scroll style panel
3. Click or drag dialog
5. See jump

https://github.com/user-attachments/assets/925350be-4fb2-4590-affb-40cafecfd9bf

Let me review again once these are fixed. These are super error prone and unpredictable IMO and we need to be careful with them. From a user perspective, they've caused my harm than help as of now.

kof:
@johnsicili seems like deployment didn't go through, you tested some buggy version

kof:
second issue is valid

johnsicili (approved):

PR #4709, opened 2025-01-04
fix: Fix code selection  in property tooltip

kof:
## Description

In property tooltip I want to be able to select and copy the code.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili (changes requested):
It doesn't work on layered props such as background, filter, and box shadow. Nothing gets selected

TrySound:
maybe switch to popover?

kof:
Turns out we have not one tooltip like this but many in different sections and I fixed only one usage.

kof:
And yes, we would have been better off with a Popover in this case, tooltips are not designed to have clickable/selectable stuff in the content, but it has already too much other logic that would potentially need a lot of testing if I chnage it now to popover.

johnsicili (approved):

istarkov (approved):

PR #4711, opened 2025-01-04
fix: remove "show" toggles on descendant component

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4670


kof (approved):
Another option would have been to add this ability to meta to have show toggle available, just like we do that with props

TrySound:
It's ok to have custom behavior for core components. I doubt this is the case for react components.

istarkov (approved):

PR #4721, opened 2025-01-06
refactor: rewrite webhook form template with jsx

TrySound:
Big one

- added `new Variable(name, initialValue)` to defined unique variable
- added `expression\`${variable}\`` to access variables by reference
- added `new Action(args, expression\`\`)` support
- added `ws:show` prop which is converted to data-ws-show under the hood

All these features are necessary to represent all webhook form features. They also may be handy in tests.


kof (approved):
didn't test

PR #4722, opened 2025-01-06
fix: Various Scrub bugs and issues

### LINKED ISSUE #4511: Clicking into input doesn't always select all text
When clicking into an input, I want all text to be selected, though sometimes only some of it is selected — in which case I have to CMD A, then do what I need.

The culprit _may_ be clicking while moving the mouse. It's something that happens so subtly though I think it should be accounted for.

Repro:
1. Click into and out of fields quickly
2. Doesn't seem to happen when they are slow clicks/no mouse movement occurs

See in the video how sometimes only partial inputs are selected upon click:

https://github.com/user-attachments/assets/2bc38d1a-ec8c-4cfd-a983-0e395e08c653

istarkov:
## Description

Need to be heavily tested

closes #4511 

Fixes:
- [x] - Resize cursor wasn't removed after scrub on some cases image
- [x] - Esc now works as expected.
- [x] - Selection https://github.com/webstudio-is/webstudio/issues/4511 
- [ ]  - Works in Safari and FF
- [ ]  - Should work in case if pointer lock permission is not provided (hard to test as can't revoke chrome permission)

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/design-system/src/components/primitives/numeric-gesture-control.ts
+++ b/packages/design-system/src/components/primitives/numeric-gesture-control.ts
@@ -15,6 +15,20 @@
  */
 
 import { clamp } from "@react-aria/utils";
+import { css } from "../../stitches.config";
+
+const scrubUI = css({
+  "*": {
+    userSelect: "none!important",
+    touchAction: "none!important",
+  },
+});
+
+const cursorUI = css({
+  "*": {
+    cursor: "ew-resize!important",
+  },
+});
 
 export type NumericScrubDirection = "horizontal" | "vertical";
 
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/primitives/numeric-gesture-control.ts
```diff
@@ -229,124 +345,147 @@ export const numericScrubControl = (
...
+    disposeOnCleanup(() => {
+      const timerId = window.setTimeout(() => {
+        requestPointerLockSafe(targetNode)
+          .then(() => {
+            // do nothing
+
+            const cursorNode = (targetNode.ownerDocument.querySelector(
+              "#numeric-guesture-control-cursor"
+            ) ||
+              new Range().createContextualFragment(`
```
TrySound, line 437: new DOMParser().parseFromString(value, "application/xml").documentElement

TrySound (approved):
❤️

PR #4725, opened 2025-01-06
fix: insert properly from marketplace

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4723

We removed body from template data but didn't fix roots. Later insertion logic started to rely on these roots.


kof:
Can we make a test for this scenario?

TrySound:
Extract from all children at once instead of removing body as root but leaving all its data like props and styles there

johnsicili (approved):
Minimally tested

PR #4738, opened 2025-01-10
refactor: replace date-fns with use-relative-time

TrySound:
See https://github.com/nkzw-tech/use-relative-time

We use only single function from date-fns while the package is like 22MB. Here's small alternative for react based on intl.

Added jsx support to ours toasts inside of builder and RelativeTime component to use it inside of callbacks.


packages/design-system/src/components/toast.tsx
```diff
@@ -383,9 +383,12 @@ export const Toaster = () => {
 type Options = Pick<ToastOptions, "duration" | "id" | "icon">;
 
 export const toast = {
-  info: (value: string, options?: Options) => hotToast(value, options),
-  error: (value: string, options?: Options) => hotToast.error(value, options),
-  warn: (value: string, options?: Options) => hotToast.custom(value, options),
-  success: (value: string, options?: Options) =>
+  info: (value: JSX.Element | string, options?: Options) =>
```
kof, line 386: why JSX.Element and not ReactNode?

kof (approved):

packages/design-system/src/components/toast.tsx
TrySound: Because react-hot-toast does not accept ReactNode

PR #4745, opened 2025-01-13
refactor: add animation package

TrySound:
Just an empty one


istarkov (approved):

packages/sdk-components-animation/LICENSE-3RD-PARTY
```diff
@@ -0,0 +1,13 @@
+MIT License
```
kof, line 1: that shouldn't be here

PR #4750, opened 2025-01-16
fix: support more curl flags

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4661

- support -G to enforce get method and treat data as search params
- post data has application/x-www-form-urlencoded content type by default
- -u flag allows basic http authentication
- support more data flags like --data-ascii, --data-raw, --data-urlencode


kof (approved):

johnsicili (approved):
Works well thanks

PR #4751, opened 2025-01-16
feat: Add head component

istarkov:
## Description

Allows Head component

Hidden under `headSlotComponent` flag

Vercel fixture now have a HTMLRewriter example. Improves development.

## Todo In Next PR
- Head Title support
- Icons
- Allow reset `as` field

- [x] - BG is now white always
- [x] - It's shown as fixed element to not break layout 

##  Implementation details

This component has more priority above PageSettings meta tags
e.g. If some tag exists in PageSettings and in HeadSlot the HeadSlot wins

## Steps for reproduction

Create HeadSlot

image

- [x] - Visit https://webstudio-fixture-project-a-0su3o.wstd.work/head-tag
check server response.

image

See `og:title` is not duplicated.

- [x] - Check Elements Panel

image

See no duplicated. All tags exists.

- [x] - Click Home. 
See no meta tags are in Elements panel from `/head-tag` page

- [x] - Click back button

image

See all meta is rendered and is not duplicated. 

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/sdk-components-react/src/head-link.ws.ts
```diff
@@ -0,0 +1,22 @@
+import { Link2Icon } from "@webstudio-is/icons/svg";
```
TrySound, line 1: maybe combine into single module like radix components?

istarkov: yep will do

istarkov: Decided to not. Easier as now

fixtures/webstudio-remix-vercel/custom-template/vite.config.ts
```diff
@@ -0,0 +1,19 @@
+import { defineConfig } from "vite";
```
TrySound, line 1: I use .template folder in react-router-docker to distinct from other folders

johnsicili:
Good stuff!! 🔥 

IMO we should set a background to this and the XML page because when the theme is dark it's hard to read.

Also set to full width

Screenshot 2025-01-18 at 7 23 28 AM

istarkov:
> Good stuff!! 🔥
> 
> IMO we should set a background to this and the XML page because when the theme is dark it's hard to read.
> 
> Also set to full width
> 
> Screenshot 2025-01-18 at 7 23 28 AM


div is already a full width.

For background I can set something like white with 50% opacity would be visible in most cases

istarkov:
Probably we need to show it as absolute element above content. So it will not break the existing layout

fixtures/webstudio-custom-template/app/routes/[script-test]._index.tsx
```diff
@@ -16,7 +16,11 @@ import {
   formIdFieldName,
   formBotFieldName,
 } from "@webstudio-is/sdk/runtime";
-import { ReactSdkContext } from "@webstudio-is/react-sdk/runtime";
+import type { PageMeta } from "@webstudio-is/sdk";
```
Comment (TrySound) [inline fixtures/webstudio-custom-template/app/routes/[script-test]._index.tsx:19]: is it used now?

istarkov:
Background fixed now white

packages/sdk-components-react/src/head-slot.tsx
```diff
@@ -0,0 +1,78 @@
...
+  useContext,
+} from "react";
+import { XmlNode } from "./xml-node";
+
+export const defaultTag = "head";
+
+export const HeadSlot = forwardRef<
+  ElementRef<"div">,
+  { "data-ws-expand"?: boolean } & Omit<
+    ComponentProps<typeof defaultTag>,
```
TrySound, line 19: I don't think we actually support any props in head slot and should not generate them

istarkov: Does not used anywhere on prod, just for hooks

istarkov: Done

TrySound (approved):

PR #4754, opened 2025-01-16
fix: merge generated classes and class prop properly

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4749

Classes were merged without considering ternaries like this

```js
"cls1" + " " + condition ? "cls2" : ""
```

which made condition part always truthy

Fixed by generating template literal instead of concatenation operator.


istarkov (approved):

johnsicili (approved):
Can't test on a published site. If you want to push a staging branch I can double check but all good if you're confident in it

TrySound:
https://github.com/webstudio-is/webstudio/pull/4755

PR #4756, opened 2025-01-17
refactor: rewrite core components templates with jsx

TrySound:
Here refactored all core components with jsx templates and moved them to sdk package. Fixed all imports as well.


kof (approved):
nice!

istarkov (approved):

PR #4763, opened 2025-01-20
fix: avoid imports of not installed packages when setup SSG

TrySound:
SSG templates have @webstudio-is/image imports which is not installed when prebuild.ts imports constants.mjs.

Here removed URL.canParse checks in favor of netlify and vercel domains configuration.


kof (approved):
why not inline the canParse utility? also there is URL.canParse

TrySound:
It's just not needed at all. Users should configure optimization of remote images.

PR #4767, opened 2025-01-21
refactor: replace image-size with image-meta

TrySound:
Image-size is barely supported, works only in node.js.

Image-meta is a fork from unjs community, without dependencies, can work in browsers, we already use it inside of ipx.

### DIFF
--- a/packages/asset-uploader/package.json
+++ b/packages/asset-uploader/package.json
@@ -11,12 +11,12 @@
   },
   "dependencies": {
     "@aws-crypto/sha256-js": "^5.2.0",
-    "@smithy/signature-v4": "^4.2.3",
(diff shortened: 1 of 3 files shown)


kof:
They copied code from the original package without even mentioning it in the readme ... red flag!!!

kof:
essentially they just copied one function and made it a separate package that works with streams, image-size has the same function to work with a buffer + has a fs api, that's why you thought its node only

kof:
image-size also had comits just last month

kof:
I think this behavior is shameless from image-meta authors

TrySound:
Mentioned in license, all good
https://github.com/unjs/image-meta/blob/main/LICENSE#L5-L6

kof (approved):

PR #4770, opened 2025-01-21
experimental: Head Slot Icons

kof:
## Description

image

image

-  additionally replaced blockquote icon
- changed the order of the components in general

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


johnsicili:
## Prevent multiple Head Slots on one page? 

Seems if they both define the same info, they'll both print out.

Screenshot 2025-01-21 at 12 50 26 PM

## Add component desc.. suggestion:

The Head Slot component lets you customize page-specific head elements (like canonical URLs) which merge with your site's global head settings, with Head Slot definitions taking priority over Page Settings. For site-wide head changes, use Project Settings instead.

## Page Settings Meta
If we don't migrate this right away, should we at least conditionally show it only if existing meta is defined? Or nah?

kof:
I don't think its a problem to have multiple head slots, in the end its the same concept as slots cc @istarkov

johnsicili (approved):
I think it's fine but will provide some feedback if you choose to do anything with it.

These three icons are confusing as they all show some part of a box but don't work well together:
Screenshot 2025-01-21 at 6 06 24 PM

Also can you add the component desc. I suggested one in the thread

istarkov (approved):

kof:
> Seems if they both define the same info, they'll both print out.

yes, though this is technically same as if you add 2 divs with the same text ... up to the user

kof:
added description, merging ...

PR #4771, opened 2025-01-22
feat: forbid data variables with the same name on instance

TrySound:

We don't want to allow data variables on one instance with the same name. Though we will allow child variables to mask parent ones in another PR.

To test:
- create variable var1 on box
- try to create another variable with var1
- create variable var2
- try to rename it to var1
- create variable on body with var1


kof (approved):
What happens with the existing variables which have the same name?

TrySound:
Nothing, users will get the error when try to edit them. Though I doubt anybody name variables the same.

PR #4779, opened 2025-01-23
refactor: drop stylable flag in component meta

TrySound:
There is no need in additional flag if we add preset style to every stylable component already.


kof (approved):

kof:
A little implicite, but overall I agree

PR #4780, opened 2025-01-23
refactor: add renderData utility to template package

TrySound:
renderData is similar to renderTemplate but returns maps of data which is used by most of builder utilities and component generator.

Here migrated to it css and component generator and used it to generate data sources where possible.


kof:
I was confused about the name at first because data is output, render takes stilll jsx, maybe jsxToData would be a better name?

kof (approved):

TrySound:
render is the key here IMO

PR #4781, opened 2025-01-23
feat: Dashboards projects search 

kof:
## Description

- [ ] Search the projects
- [ ] Use arrow keys directly from the search or by focusing the project
- [ ] Hit enter to open

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/dashboard/projects/project-card.tsx
+++ b/apps/builder/app/dashboard/projects/project-card.tsx
@@ -163,12 +163,12 @@ export const ProjectCard = ({
   },
   hasProPlan,
   publisherHost,
+  ...props
 }: ProjectCardProps) => {
   const [isRenameDialogOpen, setIsRenameDialogOpen] = useState(false);
   const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
   const [isShareDialogOpen, setIsShareDialogOpen] = useState(false);
   const [isHidden, setIsHidden] = useState(false);
-  const { thumbnailRef, handleKeyDown } = useProjectCard();
   const handleCloneProject = useCloneProject(id);
   const [isTransitioning, setIsTransitioning] = useState(false);
 
@@ -195,7 +195,7 @@ export const ProjectCard = ({
   const linkPath = builderUrl({ origin: window.origin, projectId: id });
 
   return (
-    <Card hidden={isHidden} tabIndex={0} onKeyDown={handleKeyDown}>
+    <Card hidden={isHidden} {...props}>
       <CardContent
         css={{
           background: theme.colors.brandBackgroundProjectCardBack,
@@ -219,17 +219,9 @@ export const ProjectCard = ({
         />
 
         {previewImageAsset ? (
-          <ThumbnailLinkWithImage
-            to={linkPath}
-            name={previewImageAsset.name}
-            ref={thumbnailRef}
-          />
+          <ThumbnailLinkWithImage to={linkPath} name={previewImageAsset.name} />
         ) : (
-          <ThumbnailLinkWithAbbr
-            title={title}
-            to={linkPath}
-            ref={thumbnailRef}
-          />
+          <ThumbnailLinkWithAbbr title={title} to={linkPath} />
         )}
         {isTransitioning && <Spinner delay={0} />}
       </CardContent>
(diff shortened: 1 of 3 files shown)


istarkov (approved):

johnsicili (changes requested):
1. On first login I see this bad layout, refresh works fine:

Screenshot 2025-01-25 at 10 19 31 PM

2. Why on the side? Header seems appropriate because what's below it is what you're filtering and tabbing after search would work. Right now tab after search goes to left nav. And it's just odd.. you look at projects, then want to filter so you go to a different panel.

3. I think related to the new login with secret, type 0000, hit enter, and enter selects Google instead of submitting the pass

kof:
1. this background bug - I can't reproduce it any more, I saw it before and fixed it, but I am not sure how you are still getting it, seems like you are on the latest version and yet ....

kof:
3. can't repro, hitting enter logs me in with pass ....

kof:
2. The idea is to have a single search box for all things we are going to add over time, same as in figma

kof:
Reagrding accessibility with tab - yeah not sure, next tabbable in line is navigation, but yeah after searching something you might want to tab into found items

kof:
Actually just realized why The search position looks weird, its because I didn't show search results from all things, but just stayed in the same view, e.g. projects or templates.

The current position of search actually requires to show a new view, "search", which would show all things in one view. Not sure if I should move the search for now to the header or add the search view. Search view is better long-term.

kof:
I think I am gonna add the search view now, its not hard.

kof:
Another alternative could be using cmd+k interface here like we do in the builder, clicking on search would open cmd+k panel ....

kof:
Hmm, problem is in this case search needs to either have all data loaded upfront or be done server-side.

kof:
Hmm, ok I think I am going to do the following:

1. keep the search position
2. add the search view
3. keep the client-side search for snappiness (don't think its going to be too much data any time soon)
4. load both: starter templates list and projects on any route, so that its always available for the search
5. add the route for /search?q=abc

johnsicili:
Issue summary

1. On login screen, type pass, hit enter, see it selects Google
2. On login, I get the bad layout
3. On logout I get an error
4. Still don't love several tabs to get to the things you are searching. I think the same thing about search components. Search then tab should go to the first item in results
5. On search, usually typing quickly, I get a frozen dashboard for like 3 seconds

https://github.com/user-attachments/assets/32ba8c3c-a212-43d5-926d-484f89866b10

johnsicili (changes requested):
Left comments in previous comment

johnsicili:
1. I can't login with a new user anymore. E.g login with 0000:blah. Infinite redirect

https://github.com/user-attachments/assets/38f50bea-5b13-4bf1-8ca9-7e89eadb26ba


2. Still when searching _sometimes_ it takes 3-5 seconds and freezes. See previous comment for a video

The other bugs appear to be resolved.

kof:
@johnsicili  great catch

apps/builder/app/auth/login.tsx
```diff
@@ -35,7 +34,14 @@ export const Login = ({
 }: LoginProps) => {
   globalStyles();
   return (
-    <Flex align="center" justify="center" css={{ height: "100vh" }}>
+    <Flex
+      align="center"
+      justify="center"
+      css={{
+        height: "100vh",
```
istarkov, line 41: Not your change but usually 100vh means very badly written layout

apps/builder/app/dashboard/search/search-field.tsx
```diff
@@ -0,0 +1,47 @@
...
+          handleCancelSearch();
+          return;
+        }
+        if (isSearchRoute === false) {
+          navigate(
+            {
+              pathname: dashboardPath("search"),
+              search: `?q=${value}`,
+            },
+            // Remember the last path to return to on abort
```
istarkov, line 33: on cancel, broke my mind trying to find abort

johnsicili (approved):

istarkov (approved):

apps/builder/app/dashboard/search/search-field.tsx
kof: you mean you were searching for onAbort callback but its called onCancel?

apps/builder/app/auth/login.tsx
kof: what would you do?

PR #4790, opened 2025-01-26
feat: match variables by name when copy paste

TrySound:

Here improved copy paste experience between expressions. All expressions while editing have are no longer encoded with ids. For example `system.search.name` is the same.
Though invalid js characters are encoded with code point like this `Collection Item.title` becomes `Collection$32$Item.title` when copy into textual editor.

And this less obscure name can be copied between different lists with the same `Collection Item` name.


johnsicili (approved):

johnsicili:
Nice I like it

kof (approved):

PR #4798, opened 2025-01-28
feat: save variable/resource even with validation errors

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4753

Variable validation especially resources shouldn't be too destructive. User may configure complex resource and get it all wiped out because url is invalid or header has empty value.

Here I tweaked the logic to require only not duplicated "name" and everything else is saved even when invalid.

This will not break anything in generation process and will still show user an error.

Screenshot 2025-01-28 at 14 18 11

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/variable-popover.tsx
+++ b/apps/builder/app/builder/features/settings-panel/variable-popover.tsx
@@ -813,7 +813,14 @@ export const VariablePopoverTrigger = ({
                 if (requiresUpgrade) {
                   return;
                 }
-                if (event.currentTarget.checkValidity()) {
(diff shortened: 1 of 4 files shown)


johnsicili (approved):
I like this. 

It takes two clicks to get out of a dialog that has validation error fyi

kof:
> It takes two clicks to get out of a dialog that has validation error fyi 

I believe the idea was before saving with an error state, we want user to notice there is an error

kof (approved):

TrySound:
Yep, this is intended. Dialog is closed only when all errors are shown to user.

PR #4800, opened 2025-01-28
refactor: support resources in jsx templates

TrySound:
Added full resource configuration in templates.
Will be useful for testing.


kof (approved):
nice!

PR #4801, opened 2025-01-28
refactor: drop react and warn-once from CLI

TrySound:
Here removed image package from CLI which depends on react and warn-once. CLI can be a little smaller without it.


kof (approved):
no react in cli? that must be not just a little

PR #4808, opened 2025-01-30
feat: unset variables when create or rename data variables

TrySound:

Now user can write expressions with not existing variables and create these variables later. Builder will automatically traverse all expressions and bind to newly created variable by name.

Additionally user will see suggestions with all unset variables in subtree.

Screenshot 2025-01-30 at 12 35 21


kof (approved):
AMAZING.

johnsicili (approved):

TrySound:
image

johnsicili:
"pick a not defined" doesn't sound good but used yours as inspo:

Enter a variable name, or select from the dropdown if any variables have been used in expressions but not yet created.

johnsicili:
Didn't see that it's conditional now. Maybe: Enter a new variable or select a variable that has been used in expressions but not yet created.

PR #4821, opened 2025-02-02
fix: render nothing on website when collection is invalid

TrySound:
User got 500 on published site. Turned out collection got invalid data (not array). We need to keep site working in such cases, missing piece will be more intuitive than 500 on whole website.


johnsicili (approved):
Didn't test (can't publish) but logically seems great

PR #4824, opened 2025-02-03
feat: restore variables by name when copy-pasting

TrySound:

Unset variables when copy and restore when paste.

Now you can copy instances with references to `system` variable and it will be connected on another page as well when you paste.


kof (approved):

johnsicili (approved):
This is great!!

PR #4833, opened 2025-02-05
refactor: drop legacy netlify templates from CLI

TrySound:
Remix will no longer be used for netlify. And less choice for user. Now we have only netlify and ssg netlify templates.

image


kof (approved):
didn't test

PR #4849, opened 2025-02-09
feat: rebind masked variables

TrySound:
Here improved creating variables to override deeper bindings when new variable overrides ancestor variable similar to css variables behavior.

Here's a video with example

https://github.com/user-attachments/assets/41f93f5b-fd2d-4a3b-97d1-076f0ded5f01

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
+++ b/apps/builder/app/builder/features/settings-panel/resource-panel.tsx
@@ -59,7 +59,7 @@ import {
   $selectedPage,
 } from "~/shared/awareness";
 import { updateWebstudioData } from "~/shared/instance-utils";
-import { restoreTreeVariablesMutable } from "~/shared/data-variables";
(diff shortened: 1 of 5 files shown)


kof (approved):

johnsicili:
Hey I'm seeing variable IDs output. Am I doing something wrong? Thought we got rid of those.

Seems like something is wrong... I think I would expect only one "hello" to show up as the second overrides the first but in my video I see both and can select them independently.

https://github.com/user-attachments/assets/797b4719-6aab-483a-9ce1-76c18cf7df1f

johnsicili (approved):
Nice that works now.

This _might_ cause confusion having both vars there, selecting the "first", but the second valuing showing.

Screenshot 2025-02-09 at 11 22 09 AM

Though if people are defining the same variable twice then they should understand.

TrySound:
Will fix it in another PR

PR #4855, opened 2025-02-11
fix: remove masked duplicates in binding popover

TrySound:
Here improved computing variables logic to exclude masked variables from bindings popover list.

Also rewrote related tests with jsx template.

Screenshot 2025-02-11 at 17 49 00


johnsicili (approved):
nice

kof (approved):

PR #4860, opened 2025-02-12
feat: allow deleting variables used in expressions

TrySound:

Before we disabled deleting variables which are used in expressions. Now we will ask confirmation and unset this variable. This way user will not be locked if variable is lost and we will show diagnostics (in the future) where all unset variables are.

Screenshot 2025-02-12 at 15 38 14


kof:
Also is there a way to show any hint which instance or variable is using it right in that dialog?

kof:
At least we could show instance names, at beest we would make instance names clickable to jump to that instance

kof (approved):

TrySound:
image

PR #4867, opened 2025-02-13
feat: drop legacy CLI templates

TrySound:
Here dropped vanilla and legacy vercel templates.
Replaced vanilla with docker in publish dialog.

Screenshot 2025-02-13 at 18 34 53

### DIFF
--- a/apps/builder/app/builder/features/topbar/publish.tsx
+++ b/apps/builder/app/builder/features/topbar/publish.tsx
@@ -56,7 +56,6 @@ import {
 import { AddDomain } from "./add-domain";
 import { humanizeString } from "~/shared/string-utils";
 import { trpcClient, nativeClient } from "~/shared/trpc/trpc-client";
-import { isFeatureEnabled } from "@webstudio-is/feature-flags";
 import type { Templates } from "@webstudio-is/sdk";
 import DomainCheckbox, { domainToPublishName } from "./domain-checkbox";
 import { CopyToClipboard } from "~/builder/shared/copy-to-clipboard";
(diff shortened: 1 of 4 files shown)


kof:
Why do we call it Vanilla instead of Docker?

TrySound:
Because we have ssg completely unrelated to docker on the same option.

kof:
Can you show the options we have now? I haven't checked
I think I misunderstand ...

TrySound:
The choice works for both ssg and ssr. Above ssg download, below it ssr instructions.

image

kof:
Oh I see so we say vanilla but if you want a remix app you use docker, but if you want static hmtl/css you click download. So we kinda lost the option of building the js app without docker and without spcific configs for vercel/netlify

TrySound:
Maybe split vanilla into static and docker and show only one part of dialog for each. For vercel and netlify show both parts.

kof:
Or keep Vanilla but show 2 code options in step 3: with docker and without docker

TrySound:
Vanilla server doesn't make much sense and I don't want to support it as additional template

kof:
> Vanilla server doesn't make much sense and I don't want to support it as additional template

Ok, but how can someone create a project with remix and without a docker?

TrySound:
and do what with it?

TrySound:
image
image

kof:
> and do what with it?

Its not up to us, they may want to start the node environment anywhere, without using docker. Who knows, this is the default in any case

kof:
> image image

this looks good, still don't understand how are we communicating regular app checkout

TrySound:
> Its not up to us, they may want to start the node environment anywhere, without using docker. Who knows, this is the default in any case

they can just use docker template, it is the same but with dockerfile inside

kof:
> they can just use docker template, it is the same but with dockerfile inside

Where is this communicated?

What are the CLI commands to use without docker? lets say I don't have and don't want to install docker.

kof:
Why can't we provide the cli commands to create the app without docker?

TrySound:
> What are the CLI commands to use without docker? lets say I don't have and don't want to install docker.
```
$ webstudio
```

kof:
I think this needs to be expoed in that export UI, so that there are all logical options presented

TrySound:
Vanilla is a bad option. Users don't understand what it is. And it does not give any real value. We don't want to support bad react-router configuration by users.

kof (approved):
discussed adding some way of knowing there is a way for a local dev without docker

johnsicili:
Didn't review. Don't have much to add here besides I agree that it should be known that "Docker" is really just vanilla with a Dockerfile so someone can host on some other serverless platform besides the two we provide options for.

PR #4870, opened 2025-02-14
feat: support global data variables

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/4166 https://github.com/webstudio-is/webstudio/issues/3686

Added "Settings" section to global root. Now users can create data variables which are available for all pages and inside slots.

https://github.com/user-attachments/assets/be2cf862-08ec-427e-aea3-076fa03a4437

### DIFF
--- a/apps/builder/app/builder/features/inspector/inspector.tsx
+++ b/apps/builder/app/builder/features/inspector/inspector.tsx
@@ -2,7 +2,6 @@ import { useRef } from "react";
 import { computed } from "nanostores";
 import { useStore } from "@nanostores/react";
 import type { Instance } from "@webstudio-is/sdk";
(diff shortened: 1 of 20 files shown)


kof (approved):
Yes! Finally!

What about system?

TrySound:
Another PR

johnsicili:
Good work!!!

I noticed when duplicating or copy pasting an instance that references a global var, the data doesn't display even though the binding still seems to be correct.

https://github.com/user-attachments/assets/df72a5b0-ea09-4feb-aa20-426439c6b37f

1. Bind Global data var value to a heading
2. Duplicate heading
3. See no text on duplicate

_Still testing but passing this along for now_

johnsicili (approved):

Issue #4626, opened 2024-12-20
Update message on the mode selection dropdown when projects are loaded from url's


JayaKrishnaNamburu:
When projects loaded from shared links, there is only `Design` mode or `Preview` mode. But the message here says `Select Design or Content mode`. It should be only `Design mode`. 

<img width="298" alt="Screenshot 2024-12-20 at 5 39 05 PM" src="https://github.com/user-attachments/assets/d66eaf70-67be-4171-b2fa-3dcf86a4df1c" />

<img width="1797" alt="Screenshot 2024-12-20 at 5 37 31 PM" src="https://github.com/user-attachments/assets/deb327db-5d32-441d-a6f3-7ef4682f0428" />

PR #4881, opened 2025-02-16
update the message on mode selection dropdown.

victorchrollo14:
## Description

fixes issue  #4626 

added a function that returns the default message after checking what all modes are enabled, if both the content mode and design mode is enabled then we display "Select Design or Content mode", if content mode is disabled then we display "Select Design mode" ie in the share page where links generated by free users has the content mode disabled.

but the function doesn't handle a case where both the content and design mode are disabled, since there didn't seem to be any scenario where both are disabled. in such a scenario we should probably just hide the drop down menu that let users select content or design mode.

## Steps for reproduction

1. set USER_PLAN to empty string in env.
2. then generate a share link on the builder page.
3. open the share link and you can the see message is "Select Design mode" in mode selection dropdown, since only design mode is enabled and content mode is disabled.

## Code Review

  - conceptual review ( feature-correctness)
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/topbar/builder-mode.tsx
+++ b/apps/builder/app/builder/features/topbar/builder-mode.tsx
@@ -69,6 +69,18 @@ export const BuilderModeDropDown = () => {
     setActiveMode(undefined);
   };
 
+  const getDefaultMessage = () => {
+    if (
+      isContentModeAllowed &&
+      isFeatureEnabled("contentEditableMode") &&
+      isDesignModeAllowed
+    ) {
+      return "Select Design or Content mode";
+    }
+
+    return !isDesignModeAllowed ? "Select Content mode" : "Select Design mode";
+  };
+
   return (
     <Flex align="center">
       <Tooltip
@@ -145,7 +157,7 @@ export const BuilderModeDropDown = () => {
             <Box css={{ width: theme.spacing[25] }}>
               {activeMode
                 ? menuItems[activeMode].description
-                : "Select Design or Content mode"}
+                : getDefaultMessage()}
             </Box>
           </div>
         </DropdownMenuContent>


apps/builder/app/builder/features/topbar/builder-mode.tsx
```diff
@@ -69,6 +69,18 @@ export const BuilderModeDropDown = () => {
...
+  const getDefaultMessage = () => {
+    if (
+      isContentModeAllowed &&
+      isFeatureEnabled("contentEditableMode") &&
+      isDesignModeAllowed
+    ) {
+      return "Select Design or Content mode";
+    }
+
+    return !isDesignModeAllowed ? "Select Content mode" : "Select Design mode";
```
kof, line 81: try to avoid negations when possible

```suggestion
    return isDesignModeAllowed ? "Select Design mode" : "Select Content mode";
```

kof: overall this function could be expressed as a map, something like: 

```js
const messages = {
  both: "Select Design or Content mode",
  content: "Select Content mode",
  design: "Select Design mode" 
}

const mode = ContentModeAllowed &&
      isFeatureEnabled("contentEditableMode") &&
      isDesignModeAllowed ? "both" : isDesignModeAllowed ? 'design' : 'content'
```

Something like this, we tend to prefer state machines over "ifs"

kof: I am fine merging it though as-is, just a tip for the future

victorchrollo14: > overall this function could be expressed as a map, something like:
> 
> ```js
> const messages = {
>   both: "Select Design or Content mode",
>   content: "Select Content mode",
>   design: "Select Design mode" 
> }
> 
> const mode = ContentModeAllowed &&
>       isFeatureEnabled("contentEditableMode") &&
>       isDesignModeAllowed ? "both" : isDesignModeAllowed ? 'design' : 'content'
> ```
> 
> Something like this, we tend to prefer state machines over "ifs", even though here you still have a conditional

thanks for the advice. I've come across state machines in automata theory, but haven't come across using that in code, is there any article you'd recommend reading.

istarkov:
Seems like no reason to show select at all in such situation?
cc @kof 
image

kof:
> Seems like no reason to show select at all in such situation?
agree

PR #4882, opened 2025-02-16
chore: Unreleased bugs with advanced section

kof:
## Description

- [x] Can't click a property on main in advanced
    https://github.com/user-attachments/assets/a2446685-e7dc-4c01-bf03-4ee64b8f5e16
- [x] looks like property name validation is broken. I add "dispaaayyy" and it accepeted

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
+++ b/apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
@@ -1,0 +1,56 @@
+import {
+  properties,
+  parseCss,
+  type ParsedStyleDecl,
+} from "@webstudio-is/css-data";
+import { type StyleProperty } from "@webstudio-is/css-engine";
+import { camelCase } from "change-case";
+import { lexer } from "css-tree";
+
+/**
+ * Does several attempts to parse:
+ * - Custom property "--foo"
+ * - Known regular property "color"
+ * - Custom property without -- (user forgot to add)
+ * - Custom property and value: --foo: red
+ * - Property and value: color: red
+ * - Multiple properties: color: red; background: blue
+ */
+export const parseStyleInput = (css: string): Array<ParsedStyleDecl> => {
+  css = css.trim();
+  // Is it a custom property "--foo"?
+  if (css.startsWith("--") && lexer.match("<custom-ident>", css).matched) {
+    return [
+      {
+        selector: "selector",
+        property: css as StyleProperty,
+        value: { type: "unset", value: "" },
+      },
+    ];
+  }
+
+  // Is it a known regular property?
+  const camelCasedProperty = camelCase(css);
+  if (camelCasedProperty in properties) {
+    return [
+      {
+        selector: "selector",
+        property: css as StyleProperty,
+        value: { type: "unset", value: "" },
+      },
+    ];
+  }
+
+  // Is it a custom property "--foo"?
+  if (lexer.match("<custom-ident>", `--${css}`).matched) {
+    return [
+      {
+        selector: "selector",
+        property: `--${css}`,
+        value: { type: "unset", value: "" },
+      },
+    ];
+  }
+
+  return parseCss(`selector{${css}}`);
+};
--- a/apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
+++ b/apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
@@ -6,6 +6,7 @@ import {
 import { type StyleProperty } from "@webstudio-is/css-engine";
 import { camelCase } from "change-case";
 import { lexer } from "css-tree";
+import { styleConfigByName } from "../../shared/configs";
 
 /**
  * Does several attempts to parse:
@@ -52,5 +53,17 @@ export const parseStyleInput = (css: string): Array<ParsedStyleDecl> => {
     ];
   }
 
-  return parseCss(`selector{${css}}`);
+  const styles = parseCss(`selector{${css}}`);
+
+  for (const style of styles) {
+    // somethingunknown: red; -> --somethingunknown: red;
+    if (
+      style.value.type === "unparsed" &&
+      style.property.startsWith("--") === false
+    ) {
+      style.property = `--${style.property}`;
+    }
+  }
+
+  return styles;
 };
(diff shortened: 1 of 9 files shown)


johnsicili (changes requested):
1. Paste in/type `gap: 2px`, see "unset" as the value.
Screenshot 2025-02-16 at 2 29 32 PM

2. Paste in/type `gap: 2px`, see you can only select shorthands (normally it converts to longhand) 

Screenshot 2025-02-16 at 2 30 04 PM

3. It still accepts invalid names

Screenshot 2025-02-16 at 2 30 58 PM

apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
```diff
@@ -0,0 +1,56 @@
...
+  const camelCasedProperty = camelCase(css);
+  if (camelCasedProperty in properties) {
+    return [
+      {
+        selector: "selector",
+        property: css as StyleProperty,
+        value: { type: "unset", value: "" },
+      },
+    ];
+  }
```
TrySound, line 42: isn't this what parseCss does?

packages/css-engine/src/core/rules.ts
```diff
@@ -52,15 +52,16 @@ const mergeDeclarations = (declarations: Iterable<Declaration>) => {
...
+export const generateStyleMap = (
+  style: StyleMap,
+  {
+    indent = 0,
+    transformValue,
+  }: {
+    indent?: number;
+    transformValue?: TransformValue;
+  } = {}
+) => {
```
TrySound, line 64: Is it more readable now?

kof: in the usage yes, because most usages  don't provide the options

kof: debatable ofcourse, but felt like doing the right thing

apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
kof: Yes, but the parser will also let non-existing properties pass through and it will also not convert "bla" to "--bla" which is something I discovered people may want as they don't know  of forget that a variable needs "--" when adding property in advanced.

I could add all that logic directly in to parseCss, but this is so biased towards how UX in advanced panel needs to be that I am not sure this is a good idea to make parseCss behave in such a special way

kof: Idk, maybe its ok to make our parseCss behave this way? wdyt?
basically all it needs is to not let unknown properties get through and instead convert them to custom properties by adding dashes

packages/css-engine/src/core/rules.ts
kof: btw I thik the name of that function is super confusing, it should be generateCssFromStyleMap or styleMapToCss

kof: generateStyleMap semantically reads as it will return a style map

TrySound: I would prefer to completely get rid of it

apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
kof: no, blablubb is going to be a valid proeprty, I am literally fixing this right now with this PR as reported by John

packages/css-engine/src/core/rules.ts
kof: get rid of what? what's the replacement for that function?

apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
TrySound: Are you sure it does not output `{ type: 'invalid', value: input }`? Maybe need to filter by type before checking length.

kof: yes, that's why I had to write this function

kof: we could make the parseCss though check the list of properties with our list, so it would return type invalid, then I could wrap that functionality into that specific function I have now with additional UX for auto variables conversion

kof:
@johnsicili how exactly did you manage to get it to insert invalid property name?

kof:
> 1. Paste in/type gap: 2px, see "unset" as the value.

Can't reproduce, how did you get that? is it possible you somehow tested the wrong branch? here is what I get


image

apps/builder/app/builder/features/style-panel/sections/advanced/parse-style-input.ts
TrySound: Maybe just filter out invalid properties inside of parseCss and go with previous approach?

kof: invalid as in missing in our list of known properties?

kof: I wanted to still know which once are invalid and provide an option to the user create a custom property out of them

kof: when its just one property provided by user and empty array returned from parser - this is easy to assume there was nothing valid etc, but then maybe its still better to have a way to know when there are multiple properties that are invalid and make them all automatically into custom properties, this way we don't have to completely ignore them

johnsicili:
> how exactly did you manage to get it to insert invalid property name?

I know forsure it will add via typing/pasting full declaration `displayyyyy: 2px`

johnsicili:
>  Should I add to the dropdown additionally the same shorthand notation user has so they understand they don't have to choose the shorthand provided?

Yes. I did not think of hitting escape but even with knowing that, it requires an unnecessary step so adding to the autocomplete is a good idea

johnsicili:
> Can't reproduce, how did you get that? is it possible you somehow tested the wrong branch? here is what I get


https://github.com/user-attachments/assets/45d117ff-eabc-485d-bb58-508209adda95

kof:
fixed, when user enters a proeprty that's unknown/wrong, it will be added but the value marked as error. Simply to avoid situation where stuff I enter disappears.

displayyyyy: 2px

image

TrySound (approved):

kof:
On the second thought, to make it consistent with "entering proeprty without dashes and automatically turning it into a custom property" it seems better to have invalid properties automatically turned into custom properties always, instead of adding them with unset value

Now this: `displayyyyy: 2px` will turn into

image

PR #4888, opened 2025-02-18
refactor: decouple system store from variables

TrySound:
We are moving away from binding system to variables and instead gonna have one global variable.

Here moved system values into separate store coupled to pages instead of variables.

Also fixed suggestions in address bar, probably broke while design system refactorings.

Testing
- submit form with one named input
- check system value is rendered
- check system value can be inspected


kof (approved):
done conceptual review

johnsicili:
@TrySound  It doesn't appear form submits are putting the values into system. I added a from, one named text input, submit button and upon submitting nothing showing in system var (within the builder).

Might be good to push a staging branch to test a published site too.

TrySound:
This is builder only change. Does not affect published sites.

Issue #4729, opened 2025-01-07
Can't hit space in the instance name setting at the end


kof:
1. select any instance
2. settings
3. try hitting space at the end, middle works

PR #4891, opened 2025-02-18
fix: remove trim from onchange event and trim the value onBlur (#4729)

victorchrollo14:
## Description

fixes issue #4729 
the issues was happening since the value during onchange event had a trim in the end that removed the space from the end and beginning.

we removed the trim from onchange event and moved it to onBlur, so while typing we can enter space at the end and beginning, but when the element looses focus we trim the value.

## Steps for reproduction

1. select an instance
2. click on settings in right side panel
3. edit the name and you can add a space at the end and beginning.

## Code Review

  - conceptual review (feature-correctness)
  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/settings-section.tsx
+++ b/apps/builder/app/builder/features/settings-panel/settings-section.tsx
@@ -50,8 +50,11 @@ export const SettingsSection = () => {
           key={selectedInstance.id}
           placeholder={placeholder}
           value={localValue.value}
-          onChange={(event) => localValue.set(event.target.value.trim())}
-          onBlur={localValue.save}
+          onChange={(event) => localValue.set(event.target.value)}
+          onBlur={(event) => {
+            localValue.set((event.target as HTMLInputElement).value.trim());
+            localValue.save;
+          }}
         />
       </HorizontalLayout>
     </Row>


apps/builder/app/builder/features/settings-panel/settings-section.tsx
```diff
@@ -50,8 +50,11 @@ export const SettingsSection = () => {
           key={selectedInstance.id}
           placeholder={placeholder}
           value={localValue.value}
-          onChange={(event) => localValue.set(event.target.value.trim())}
-          onBlur={localValue.save}
+          onChange={(event) => localValue.set(event.target.value)}
+          onBlur={(event) => {
+            localValue.set((event.target as HTMLInputElement).value.trim());
+            localValue.save;
```
TrySound, line 56: save is not invoked

victorchrollo14: > I would do trimming inside of useLocalValue callback instead

I tried this initially but the text was getting trimmed after a few milliseconds when we are typing.

https://github.com/user-attachments/assets/0debeb0f-de96-4861-a499-03e88894f4a6

we could use something like the following 
export a new function inside the useLocalvalue hook.
``` 
  const finalSave = (value: Type) => {
    localValueRef.current = value;
    setRefresh((refresh) => refresh + 1);
    save();
  };
```

then run it on the onBlur event as follows, which would trim out the white space and also call the save function to sync the value on the server. what do you think @TrySound 

```
<InputField
          id={id}
          /* Key is required, otherwise when label is undefined, previous value stayed */
          key={selectedInstance.id}
          placeholder={placeholder}
          value={localValue.value}
          onChange={(event) => localValue.set(event.target.value)}
          onBlur={(event) =>
            localValue.finalSave(
              (event.target as HTMLInputElement).value.trim()
            )
          }
        />
```

apps/builder/app/builder/features/settings-panel/shared.tsx
```diff
@@ -175,6 +175,12 @@ export const useLocalValue = <Type,>(
     }
   };
 
+  const finalSave = (value: Type) => {
```
kof, line 178: this doesn't sound like great addition to the api, semantics with this are messy.

apps/builder/app/builder/features/settings-panel/settings-section.tsx
```diff
@@ -50,8 +50,12 @@ export const SettingsSection = () => {
           key={selectedInstance.id}
           placeholder={placeholder}
           value={localValue.value}
-          onChange={(event) => localValue.set(event.target.value.trim())}
-          onBlur={localValue.save}
+          onChange={(event) => localValue.set(event.target.value)}
+          onBlur={(event) =>
+            localValue.finalSave(
```
kof, line 55: @TrySound maybe take over and see what's the best solution, because we are not going anywhere

kof: to me it seemed like the initial version, where he was setting the trimmed value on blur and then calling save was a good call

PR #4894, opened 2025-02-19
refactor: generate presets with css cascade layers

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4796

CSS Cascade Layers are already widely support for almost 3 years. See https://developer.mozilla.org/en-US/docs/Web/CSS/@layer

We can switch to it instead of wrapping selectors with `:where` all preset selectors.

For older browsers impact will be insignificant as in most cases we use presets only to normalize browser styles.

Generated code now looks like this

```css
@layer presets {
  div.w-box {
    display: block;
  }
}
```

Here's how presets look in chrome devtools. As you can see browser styles are defined as layer too.

Screenshot 2025-02-19 at 17 14 23

### DIFF
--- a/packages/sdk/src/css.ts
+++ b/packages/sdk/src/css.ts
@@ -90,11 +90,12 @@ export const generateCss = ({
   assetBaseUrl,
   atomic,
 }: CssConfig) => {
-  const globalSheet = createRegularStyleSheet({ name: "ssr" });
-  const sheet = createRegularStyleSheet({ name: "ssr" });
+  const fontSheet = createRegularStyleSheet({ name: "ssr" });
+  const presetSheet = createRegularStyleSheet({ name: "ssr" });
+  const userSheet = createRegularStyleSheet({ name: "ssr" });
 
-  addFontRules({ sheet: globalSheet, assets, assetBaseUrl });
-  globalSheet.addMediaRule("presets");
+  addFontRules({ sheet: fontSheet, assets, assetBaseUrl });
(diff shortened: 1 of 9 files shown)


packages/sdk/src/css.ts
```diff
@@ -201,21 +203,33 @@ export const generateCss = ({
...
   }
 
+  const fontCss = fontSheet.cssText;
+  // render presets inside of cascade layer to let user completely override all properties
+  // user agent (browser) styles work in the same way
+  // for example a { color: black } overrides a:visited as well
+  const presetCss = presetSheet.cssText.replaceAll(
+    "@media all ",
+    "@layer presets "
+  );
```
kof, line 218: that doesn't look right, should be part of the engine. properly and specified the right way ... e.g. instead of     presetSheet.addMediaRule("presets") presetSheet.addLayer('presets')

TrySound: It's not user facing feature and only done inside of generateCss function. Shouldn't bring any side effects.

kof: why doing it at all then? anyways if you insist - fat fat todo comment please with all that stuff you just said

TrySound: Sure. With more usage of this feature we will design the api properly. For now requirements are not clear except for this little use case.

> why doing it at all then?

The feature fits this case well and looks better in devtools.

johnsicili (approved):

kof (approved):

PR #4896, opened 2025-02-21
feat: add global system variable in builder

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4166

Now system is global variable and can be accessed in slots.

Legacy one can be deleted

image

### DIFF
--- a/apps/builder/app/builder/features/pages/page-settings.tsx
+++ b/apps/builder/app/builder/features/pages/page-settings.tsx
@@ -1365,36 +1365,26 @@ const NewPageSettingsView = ({
 
 const createPage = (pageId: Page["id"], values: Values) => {
   serverSyncStore.createTransaction(
-    [$pages, $instances, $dataSources],
-    (pages, instances, dataSources) => {
+    [$pages, $instances],
+    (pages, instances) => {
       if (pages === undefined) {
         return;
       }
       const rootInstanceId = nanoid();
-      const systemDataSourceId = nanoid();
       pages.pages.push({
         id: pageId,
         name: values.name,
         path: values.path,
         title: values.title,
         rootInstanceId,
-        systemDataSourceId,
         meta: {},
       });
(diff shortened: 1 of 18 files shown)


johnsicili (changes requested):
## Duplicate 
Click Global Root, click on system var, then click out and it duplicates.

Once it duplicates the system var doesn't have any data and it seems stuck there. Can't delete it.

https://github.com/user-attachments/assets/f8a4233e-faf6-45f4-bd78-3368d3530d2f

## Info about delete legacy

Can we add one of those hints that informs the user about this option? I think the answer is always "it's safe to delete" if I understand the logic correctly and the user does not need to do anything else.

The first bug is preventing me from doing testing because both of my legacy projects with lots of good testing stuff have the corrupted global system var. Will need to test more once fixed. If the fix isn't retro active, that's okay, I can create a new project.

TrySound:
Fixed saving system variable. It can no longer can be edited.

> Info about delete legacy

This option is more for us to cleanup templates and our own projects. Legacy system variable should not affect users.

kof:
If user is trying to use the legacy system variable, we are still using the right one, right? So its not gonna get them confused about "why is this not working inside slots?"?

TrySound:
Slots will have access only to global one. Body and deeper will have access only to page one defined on body. And deleting the one on body will replace references to global one just like with regular variables. So we just treat page system variable like regular one in this case.

kof (approved):
conceptual review

johnsicili (approved):
Good work!

I still think there should be info about Delete Legacy Variable.

Users will see it and wonder if they should do it and if there are any implications for it.

Maybe something like "The legacy variable is safe to delete as references will fallback to the system variable on the Global Root."

TrySound:
Or I'll just leave it as "Delete Variable" to not confuse anybody

PR #4901, opened 2025-02-22
fix: prevent inserting into instance with placeholder

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4897

Added "placeholder" field to component meta to signal the instance accepts should not be insertion target. This reverts us to previous behavior even with empty instances.

Also slightly refactored placeholder rendering to use attr() instead of extra css variable.


istarkov (approved):

johnsicili (approved):
This is good. Creating a new task for marketplace insert

PR #4904, opened 2025-02-23
feat: unset global variables when delete

TrySound:
Global variables will be replaced with "name" or unset when deleted from data variables section.

There are few cases to test

1. variable is unset on a few pages when delete
2. variable is unset on inside of slots when delete


kof (approved):
conceptual review done

johnsicili (approved):

Issue #4203, opened 2024-10-03
Overflow clip is the new overflow hidden


kof:
This article explains why https://www.bram.us/2024/02/14/scroll-driven-animations-you-want-overflow-clip-not-overflow-hidden/

## Todo
- [ ] replace the button for hidden with the button for clip
- [ ] ensure good tooltip text
- [ ] add an icon https://iconbuddy.com/search?q=crop
- [ ] ensure existing hidden value used by users is in advanced panel

victorchrollo14:
hey @kof . will this be ok for the tooltip 
"Content that exceeds the container's boundaries is clipped  at the overflow clip edge and hidden without scrollbars"


This is the description given on mdn docs
![Image](https://github.com/user-attachments/assets/437ea38c-0c5d-477f-b725-1146f2c530d9)

kof:
We already have this: "Clips the content with no scrollbars." ... not good?

victorchrollo14:
what do you suggest then?. the clip property is similar to hidden with a few difference like we have the overflow-clip-margin property that let's you define how much content we can see beyond the clipped container, it also doesn't create scroll container as mentioned in the above blog. 

what about "Content that exceeds container's size is clipped and you can set the length of content visible beyond the clip using overflow-clip-margin" 

or shorten this a bit like 
"Content exceeding container size is clipped, with overflow-clip-margin controlling how much extra content shows."

kof:
I am not sure what we are discussing, this issue is about adding this property to the visual panel. It already works via advanced panel. The description I mentioned is already used there

victorchrollo14:
sorry about that, I didn't read the description for clip in advaned panel, I thought we were gonna add a new description for overflow: clip. 

I've replaced overflow hidden button with clip button, will raise a pr then.

PR #4914, opened 2025-02-26
feat: add overflow clip in styles panel

victorchrollo14:
## Description

closes ( #4203 )

- added a new crop svg to icons package and generated the cropIcon component.
- replaced the eye icon with the crop icon and also changed hidden to clip in the overflow toggle group.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


istarkov (approved):

istarkov:
Super, last 2 weeks I had 2 serious issues with hidden vs clip!!!

istarkov:
The only issue is the builder and clipped content.
In case of hidden scroll on instance select is working, in case of clip there is no scroll.
So clipped content will always be invisible.

kof:
@istarkov ohh, we actually have a use case with the slider where we use overflow hidden and scroll to the right image from the navigator thanks to overflow hidden.

istarkov:
Same usecase ruined some animations if hidden is set ;)
Its a builder logic issue.

istarkov:
What if only in the builder if element is selected we temporary replace clip on hidden for all ancestors. In case of no animations are in progress.

kof:
> What if only in the builder if element is selected we temporary replace clip on hidden for all ancestors. In case of no animations are in progress.

will probably sooner or later lead to a situation where something is broken in that particular state, e.g. if something relies on clip but hidden is used instead

packages/icons/icons/crop.svg
```diff
@@ -0,0 +1,11 @@
+<svg
```
kof, line 1: need to make sure the icon gets added  to figma first, then via the same process added here

kof: adding it now to figma

kof: adding now here

TrySound:
We agreed to get rid of "scroll" and keep both "clip" and "hidden" Caz both have use cases with animations

TrySound:
Thanks!

PR #4918, opened 2025-02-26
fix: insert marketplace sections after selected instance

TrySound:
Closes https://github.com/webstudio-is/webstudio/issues/4902

Rapid prototyping requires slightly different logic for inserting from marketplace so sections could be added one after another.

### DIFF
--- a/apps/builder/app/builder/features/marketplace/templates.tsx
+++ b/apps/builder/app/builder/features/marketplace/templates.tsx
@@ -57,7 +57,11 @@ const insertSection = ({
   }
   const insertable = findClosestInsertable(fragment);
   if (insertable) {
-    insertable.position = "after";
+    // numeric position means the instance already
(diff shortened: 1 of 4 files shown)


johnsicili (approved):
Thanks for doing all this. Didn't realize the insertion logic wasn't just in the marketplace but needed a receiving end. Either way, needed to get done... cool demo incoming

apps/builder/app/builder/features/marketplace/templates.tsx
```diff
@@ -57,6 +57,11 @@ const insertSection = ({
   }
   const insertable = findClosestInsertable(fragment);
   if (insertable) {
+    // numeric position means the instance already
+    // insertd after or even into ancestor
```
kof, line 61: should have mentioned this is specifically for marketplace

PR #4929, opened 2025-02-28
fix: navigate hash urls natively

TrySound:
Fixes https://github.com/webstudio-is/webstudio/issues/4830

Remix incorrectly navigates hashes and clear search params. Here fixed this by using native `<a>` and tweaked the logic in builder.

Test navigating hashes (when search params are present) in builder and on published site.


johnsicili:
Hey this actually fixes https://github.com/webstudio-is/webstudio/issues/3974 :)

The only problem I'm seeing is the query params are stripped still. Am I doing something wrong?

https://github.com/user-attachments/assets/a1c674d9-6e49-480b-bfd8-debda9e4d8bd

TrySound:
Ah, right. Well page+id selector still doesn't work because it relies on specific page. And when you put for example into slot we cannot use "current page matches specified one" heuristic. But "#my-id" string url should work.

johnsicili (approved):
Ah makes sense! Good stuff thanks

PR #4930, opened 2025-02-28
refactor: switch to hyphenated preset styles

TrySound:
Replaced camel case with kebab in all presets


TrySound:
All covered with unit tests

PR #4936, opened 2025-02-28
fix: optimize saving global variables

TrySound:
Reduced amount iterations over all instances
to optimize rebinding variables on save.


kof (approved):
haven't tested

johnsicili (approved):

PR #4942, opened 2025-03-04
refactor: simplify xml generation

TrySound:
- got rid of svg hacks, suffix seems work well
- filtered data instead of generating empty components
- fixed sitemap in features fixture


istarkov (approved):
We need dedupeMeta for dev purposes

PR #4952, opened 2025-03-06
experimental:  Use css editor in keyframes

kof:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
+++ b/apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
@@ -1,7 +1,10 @@
-import { parseCss } from "@webstudio-is/css-data";
-import { StyleValue, toValue } from "@webstudio-is/css-engine";
 import {
-  Text,
+  StyleValue,
+  toValue,
+  type CssProperty,
+  type CssStyleMap,
+} from "@webstudio-is/css-engine";
+import {
   Grid,
   IconButton,
   Label,
@@ -9,6 +12,9 @@ import {
   Tooltip,
   ScrollArea,
   theme,
+  SectionTitle,
+  SectionTitleButton,
+  SectionTitleLabel,
   Box,
 } from "@webstudio-is/design-system";
 import { MinusIcon, PlusIcon } from "@webstudio-is/icons";
@@ -18,10 +24,9 @@ import {
   CssValueInput,
   type IntermediateStyleValue,
 } from "~/builder/features/style-panel/shared/css-value-input";
-import { toKebabCase } from "~/builder/features/style-panel/shared/keyword-utils";
-import { CodeEditor } from "~/builder/shared/code-editor";
 import { useIds } from "~/shared/form-utils";
 import { calcOffsets, findInsertionIndex, moveItem } from "./keyframe-helpers";
+import { CssEditor, type CssEditorApi } from "~/builder/shared/css-editor";
 
 const unitOptions = [
   {
@@ -125,21 +130,28 @@ const Keyframe = ({
   onChange: (value: AnimationKeyframe | undefined) => void;
 }) => {
   const ids = useIds(["offset"]);
+  const apiRef = useRef<CssEditorApi>();
 
-  const cssProperties = useMemo(() => {
-    let result = ``;
-    for (const [property, style] of Object.entries(value.styles)) {
-      result = `${result}${toKebabCase(property)}: ${toValue(style)};\n`;
-    }
-    return result;
-  }, [value.styles]);
+  const styleMap: CssStyleMap = useMemo(
+    () =>
+      new Map(
+        Object.keys(value.styles).map((key) => [
+          key as CssProperty,
+          value.styles[key],
+        ])
+      ),
+    [value.styles]
+  );
 
   return (
     <>
       <Grid
         gap={1}
         align={"center"}
-        css={{ gridTemplateColumns: "1fr 1fr auto" }}
+        css={{
+          gridTemplateColumns: "1fr 1fr auto",
+          paddingInline: theme.panel.paddingInline,
+        }}
       >
         <Label htmlFor={ids.offset}>Offset</Label>
         <OffsetInput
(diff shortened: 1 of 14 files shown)


apps/builder/app/builder/shared/css-editor/css-editor.tsx
```diff
@@ -150,13 +150,13 @@ const AdvancedPropertyValue = ({
             onChange={(styleValue) => {
               const options = { isEphemeral: true, listed: true };
               if (styleValue) {
-                onSetProperty(camelCasedProperty)(styleValue, options);
+                onSetProperty(property)(styleValue, options);
```
TrySound, line 153: can be removed everywhere I think

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
```diff
@@ -125,21 +131,28 @@ const Keyframe = ({
...
-    for (const [property, style] of Object.entries(value.styles)) {
-      result = `${result}${toKebabCase(property)}: ${toValue(style)};\n`;
-    }
-    return result;
-  }, [value.styles]);
+  const styleMap: CssStyleMap = useMemo(
+    () =>
+      new Map(
+        Object.keys(value.styles).map((key) => [
+          key as CssProperty,
```
TrySound, line 140: untyped by design?

kof: I don't like the entire value type def, but its made by @istarkov and I am not sure if its some temp thing, needs to be discussed, I am not 100% in favor of Map/Set and avoid using objects.

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
```diff
@@ -157,25 +170,49 @@ const Keyframe = ({
...
-            const parsedStyles = parseCss(`selector{${cssText}}`);
-            onChange({
-              ...value,
-              styles: parsedStyles.reduce(
-                (r, { property, value }) => ({ ...r, [property]: value }),
-                {}
-              ),
-            });
+          onSetProperty={(property) => {
+            return (newValue) => {
```
TrySound, line 195: I would remove this carrying from CssEditor api

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
```diff
@@ -157,25 +170,49 @@ const Keyframe = ({
...
+          }}
+          onDeleteAllDeclarations={() => {
+            onChange({ ...value, styles: {} });
           }}
         />
       </Grid>
+      <Box css={{ paddingInline: theme.panel.paddingInline }}>
+        <Button
+          onClick={() => {
+            apiRef.current?.showAddStyleInput();
```
TrySound, line 208: 😢

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
kof: I just wanted to keep the exact same interface as we have with setProperty and void having different variations of it.

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
kof: this imperative style in this case is way better than all the logic mess with managing external and internal state, because that's a true mess when you need to trigger showing the element internally and externally

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
kof: e.g. we already use CssStyleMap in some cases, maybe that can be used instead here as well?

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
TrySound: I would remove that carrying everywhere 😄

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
TrySound: I mean, invokers would here well 😸

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-keyframes.tsx
istarkov, line 140: Should be transformed to camelCase as `const parsedStyles = parseCss();` now converts to kebab.
If you expect camelCase.
And parsers working with kebab too.
So in existing json is a mix of camel and kebab. (In the future will be kebab)
Also variables should not be camel cased.
@TrySound knows more about.

istarkov: CSSPropery should gone from type system. Not a lot of value, and it's very very long union type what makes ts slow.

istarkov: If we want to make it typed, we should use opaque types for unions like this.

TrySound (approved):

apps/builder/app/shared/style-object-model.ts
```diff
@@ -286,6 +286,9 @@ export type ComputedStyleDecl = {
   cascadedValue: StyleValue;
   computedValue: StyleValue;
   usedValue: StyleValue;
+  // We will delete it once we have added additional filters to advanced panel and
+  // don't need to differentiate this any more.
+  listed?: boolean;
```
kof, line 291: not sure if I should do this or create a separate type

istarkov:
When I hover over Add Style menu items 

Selected Outline disappears when I hover over properties without value and appears again when I hover over props without value

Screen Recording 2025-03-13 at 09 10 56

apps/builder/app/builder/features/style-panel/shared/model.tsx
```diff
@@ -221,7 +222,7 @@ export const getDefinedStyles = ({
 
 export const $definedStyles = computed(
   [
-    $selectedInstancePathWithRoot,
+    $selectedInstancePath,
```
TrySound, line 225: are you sure nothing will break?

istarkov (approved):

apps/builder/app/shared/style-object-model.ts
TrySound: Is it used anywhere?

apps/builder/app/builder/features/style-panel/shared/model.tsx
```diff
@@ -297,14 +294,16 @@ export const $definedComputedStyles = computed(
         state: styleSourceSelector?.state,
         property,
       });
+      computedStyleDecl.listed = listed;
```
TrySound, line 297: a bit unexpected place for it

apps/builder/app/builder/features/style-panel/shared/model.tsx
kof: no, but I tried o find out, didn't seem like broken, pls check as well

kof:
> When I hover over Add Style menu items
> 
> Selected Outline disappears when I hover over properties without value and appears again when I hover over props without value
> 

Makes sense, property without value can't be previewed on canvas, when we preview, we remove outlines

apps/builder/app/builder/features/style-panel/sections/advanced/stores.ts
```diff
@@ -19,64 +14,39 @@ const initialProperties = new Set<CssProperty>([
   "user-select",
 ]);
 
-export const $advancedStylesLonghands = computed(
-  [
-    // prevent showing properties inherited from root
-    // to not bloat advanced panel
```
TrySound, line 25: it was not generalized for a reason

apps/builder/app/builder/features/style-panel/shared/model.tsx
kof: ok I tried, but couldn't find a better solution for now

apps/builder/app/shared/style-object-model.ts
kof: couldln't find a better solution for now, everything else results in additional types and overhead

PR #4955, opened 2025-03-06
fix: Fix token state selector

kof:
## Description

Select the state, see it stays selected

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
@@ -332,7 +332,7 @@ export const StyleSourceControl = ({
             isEditing={isEditing}
             tabIndex={-1}
             onClick={onSelect}
-            bleed
+            bleed={source === "local"}
           >
             <Flex align="center" justify="center" gap="1">
               {source === "local" ? (
--- a/apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
@@ -215,6 +215,12 @@ const StyleSourceButton = styled("button", {
       },
       false: {},
     },
+    bleed: {
+      true: {
+        position: "absolute",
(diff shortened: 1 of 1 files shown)


apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
```diff
@@ -318,14 +324,15 @@ export const StyleSourceControl = ({
         disabled={disabled}
         aria-current={selected && state === undefined}
         role="button"
```
TrySound, line 326: i don't think role=button should be here

kof: me neither, this entire component is badly made, no idea what we were thinking

apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
```diff
@@ -318,14 +324,15 @@ export const StyleSourceControl = ({
...
-        onClick={onSelect}
         hasError={error !== undefined}
       >
         <Flex grow css={{ padding: theme.spacing[2] }}>
```
TrySound, line 329: this padding can be put on flex inside of button

TrySound: no bleeding necessary

kof: can't, check token name editing, it will loose the "border", so you won't see the token

TrySound (approved):

apps/builder/app/builder/features/style-panel/style-source/style-source-control.tsx
kof: could add the border back during editing on the container but not sure this is any better

PR #4966, opened 2025-03-07
fix: render empty page when xml is empty

TrySound:


packages/react-sdk/src/component-generator.ts
```diff
@@ -408,33 +408,37 @@ export const generateWebstudioComponent = ({
 }) => {
   const instance = instances.get(rootInstanceId);
   if (instance === undefined) {
-    return "";
+    // return "";
```
kof, line 411: forgotten?

kof (approved):

PR #4976, opened 2025-03-10
experimental: Animation publishing

istarkov:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


packages/css-engine/src/runtime.ts
```diff
@@ -0,0 +1,2 @@
+export { toValue } from "./core/to-value";
+export { hyphenateProperty } from "./core/to-property";
```
TrySound, line 2: hyphenateProperty shouldn't be necessary as we output all properties already hyphenated

vite.sdk-components.config.ts.timestamp-1741617650893-3f20c73be5f32.mjs
```diff
@@ -0,0 +1,35 @@
+// ../../vite.sdk-components.config.ts
```
TrySound, line 1: I think this file is unnecessary

packages/css-engine/src/runtime.ts
istarkov: Renamed all seems like can be deleted

istarkov: removed

TrySound (approved):

PR #4985, opened 2025-03-12
fix: format publishing tooltip properly

TrySound:
Screenshot 2025-03-12 at 11 29 27

### DIFF
--- a/apps/builder/app/builder/features/topbar/publish.tsx
+++ b/apps/builder/app/builder/features/topbar/publish.tsx
@@ -127,7 +127,9 @@ const ChangeProjectDomain = ({
       }
       suffix={
         <Grid flow="column" align="center">
-          <Tooltip content={error !== undefined ? error : statusText}>
+          <Tooltip
+            content={error !== undefined ? error : <Text>{statusText}</Text>}
+          >
             <Flex
               align="center"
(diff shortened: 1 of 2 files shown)


istarkov (approved):

apps/builder/app/builder/features/topbar/publish.tsx
```diff
@@ -127,7 +127,9 @@ const ChangeProjectDomain = ({
       }
       suffix={
         <Grid flow="column" align="center">
-          <Tooltip content={error !== undefined ? error : statusText}>
+          <Tooltip
```
kof, line 130: if content text is string, its already being rendered as Text:
`              {typeof content === "string" ? <Text>{content}</Text> : content}` in Tooltip

apps/builder/app/builder/features/topbar/publish.tsx
```diff
@@ -127,7 +127,9 @@ const ChangeProjectDomain = ({
       }
       suffix={
         <Grid flow="column" align="center">
-          <Tooltip content={error !== undefined ? error : statusText}>
```
kof, line 130: could be just `<Tooltip content={error ?? statusText}>`

PR #4990, opened 2025-03-12
fix: show unresolved inherited values in style panel

TrySound:
Per user request we will not show computed value when inherited from ancestor.
Chrome devtools seem to show unresolved values in "inherited" section.

image


kof (approved):
we should also show the computed value on alt in the input itself, not just in the tooltip

PR #4997, opened 2025-03-13
refactor: simplify css value input sizes

TrySound:
- removed non-unit values from browser styles
- merged browser styles and unit sizes into single store; both objects used in cize conversion
- use typed cssom to parse units instead of custom parser (sorry firefox)


apps/builder/app/canvas/instance-selected.ts
```diff
@@ -78,6 +79,30 @@ const calculateUnitSizes = (element: HTMLElement): UnitSizes => {
   };
 };
 
+const calculatePropertySizes = (element: HTMLElement) => {
+  const computedStyle = getComputedStyle(element);
+  const propertySizes: PropertySizes = {};
+  for (const property of Object.keys(propertiesData)) {
```
kof, line 85: why not
```suggestion
  for (const property in propertiesData) {
```

kof (approved):

PR #5000, opened 2025-03-14
feat: use ressetable blue label for properties

TrySound:
- properties now have blue labels when value is specified
- clicking on label shows tooltip
- alt+click does reset/delete


apps/builder/app/builder/features/settings-panel/shared.tsx
```diff
@@ -40,6 +39,17 @@ import type { BindingVariant } from "~/builder/shared/binding-popover";
 import { humanizeString } from "~/shared/string-utils";
 import { $selectedInstanceKeyWithRoot } from "~/shared/awareness";
 
+export const showAttributeMeta: PropMeta = {
```
kof, line 42:
```suggestion
export const showPropMeta: PropMeta = {
```

apps/builder/app/builder/features/settings-panel/shared.tsx
```diff
@@ -234,75 +239,67 @@ export const useLocalValue = <Type,>(
...
+      wrap="wrap"
+      css={{
+        columnGap: theme.spacing[5],
+        rowGap: theme.spacing[3],
+        paddingBlock: theme.spacing[2],
+      }}
+    >
+      <Box
+        css={{
+          // wrap label and input when label is more than ~9 characters
```
kof, line 289: you could use a 9ch

kof: if you wanted it to be basde on characters amount

TrySound: I don't. It's hard enough to make it wrap

kof:
- [ ] doesn't have the reset hint in the tooltip for some reason
image

TrySound:
Required fields cannot be reset or deleted. I think we had the same logic before

johnsicili (approved):
This is awesome! Can you make the attribute copy-able in the tooltip?

This has been a request in the past so we can easily copy at least the property (ideally property value but no way to paste right now the whole "declaration").

The tooltips in CSS allow copy, Oleg did it IIRC.

TrySound:
Do we have an issue about this? I would leave it for another PR.

PR #5019, opened 2025-03-19
fix: sort mixed min and max breakpoints properly

TrySound:
Added a couple of tests with style engine to figure out proper sorting.


kof (approved):

kof:
Was it not completed?

PR #5021, opened 2025-03-19
fix: reparent within another slot instance

TrySound:
Ivan found a bug. Reparent slot child into another instance of the same slot got this child removed.

Now refactored whole thing (almost). The main difference is same parent move will not do delete/paste but will just update child position. This fixes bug above and makes other logic slightly simpler.

Rewritten reparentInstance with reparentInstanceMutable to simplify testing and make it composable (if ever need). Added more tests.


kof (approved):
sounds like its all covered by tests and doestn't need manual testing

PR #5030, opened 2025-03-22
refactor: restrict publish count per account

TrySound:


kof (approved):

packages/project/src/trpc/project-router.ts
```diff
@@ -63,6 +63,38 @@ export const projectRouter = router({
...
     return projectIds.map((project) => project.id);
   }),
+
+  userPublishCount: procedure.query(async ({ ctx }) => {
+    try {
+      if (
+        ctx.authorization.type !== "user" &&
+        ctx.authorization.type !== "token"
+      ) {
+        throw new Error("Not authorized");
```
kof, line 73: I would do early returns everywhere here to reduce the nesting level, the giant try/catch is reducing readability in my subjective opinion

TrySound: try/catch will be there anyway to catch postgrest error

PR #5036, opened 2025-03-23
feat: navigate to pro features on publish dialog

TrySound:
Users don't know where to find pro features accidentally created while being on pro tier or added with marketplace.

Here wrapped items in used pro featured with links (buttons actually) which navigate to instances.

https://github.com/user-attachments/assets/7dd879fa-4e1b-4f66-9dd3-18308cd62f58

### DIFF
--- a/apps/builder/app/builder/features/topbar/publish.tsx
+++ b/apps/builder/app/builder/features/topbar/publish.tsx
@@ -1,3 +1,4 @@
+import stripIndent from "strip-indent";
 import { computed } from "nanostores";
 import {
   useEffect,
@@ -35,10 +36,12 @@ import {
   PopoverTitle,
   PopoverClose,
   PopoverTitleActions,
+  css,
+  textVariants,
 } from "@webstudio-is/design-system";
-import stripIndent from "strip-indent";
-import { $publishDialog } from "../../shared/nano-states";
 import { validateProjectDomain, type Project } from "@webstudio-is/project";
+import { $awareness, type Awareness } from "~/shared/awareness";
+import type { InstanceSelector } from "~/shared/tree-utils";
 import {
   $authTokenPermissions,
   $dataSources,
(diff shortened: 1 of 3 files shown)


apps/builder/app/builder/features/topbar/publish.tsx
```diff
@@ -212,28 +219,41 @@ const ChangeProjectDomain = ({
...
   (pages, dataSources, instances) => {
-    const features = new Set<string>();
+    const features = new Map<string, undefined | Awareness>();
+    if (pages === undefined) {
+      return features;
+    }
     // specified emails for default webhook form
     if ((pages?.meta?.contactEmail ?? "").trim()) {
-      features.add("Custom contant email");
+      features.set("Custom contant email", undefined);
```
kof, line 228:
```suggestion
      features.set("Custom contact email", undefined);
```

apps/builder/app/builder/features/topbar/publish.tsx
```diff
@@ -646,6 +666,12 @@ const refreshProject = async () => {
   toast.error(result.error);
 };
 
+const buttonLinkClass = css({
```
kof, line 669: need to create a button link component to avoid this mess, but for now lets stick with how its done everywhere else:

```
          <Link
            className={buttonStyle({color: "neutral"})}
            underline="none"
            href={}
            target="_blank"
          >
            Open project
          </Link>
```

kof (approved):

apps/builder/app/builder/features/topbar/publish.tsx
kof: that's weird, you could have just used a link and made prevent default :)

TrySound: There is no url to navigate to

PR #5045, opened 2025-03-25
feat: Implement cmd+enter shortcut to focus style source input

kof:
## Description

After adding an instance user wants to focus the style source input via shortcut to add tokens.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


apps/builder/app/builder/shared/style-source/style-source-input.tsx
```diff
@@ -48,6 +48,7 @@ import {
   type ReactNode,
```
TrySound, line 48: Why move whole section out of style panel? You could move to shared only element store

PR #5049, opened 2025-03-25
feat: substitute variables in calc and box-shadow

TrySound:
Here improved style engine logic to substitute variables in unparsed values which is essential in cases like box-shadow or calc with variables.

Screenshot 2025-03-26 at 02 03 03
Screenshot 2025-03-26 at 01 59 40


kof:
image


If I delete everything from the textarea and unfocus, should it reenable all inputs?

kof:
Found a bug with disabled color input, it seems to not be actually disabled and when I focus/unfocus it even though it looks disabled - changes the mode to computed

https://share.descript.com/view/BoIEmNG4xUq

TrySound:
> If I delete everything from the textarea and unfocus, should it reenable all inputs?

Nope, invalid value

> Found a bug with disabled color input, it seems to not be actually disabled and when I focus/unfocus it even though it looks disabled - changes the mode to computed

disabling would be less user friendly though yes we oversave on every blur even when value is not changed

kof (approved):
Do you want to fix the bug with color input or merge? I guess blur wouldn't be triggered if the focus was on the input itself, but we probably have the focus on a container around input, so blur happens even if the input is disabled

TrySound:
Autofocusing code when var or unparsed seems working

PR #5052, opened 2025-03-25
feat: make spacing input autogrow

kof:
## Description

When user opens a long value in spacing we now show an input that can big much bigger, max size 240px for the input. On the other hand it will be small if value is small AND it will autogrow

image

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/shared/input-popover.tsx
+++ b/apps/builder/app/builder/features/style-panel/sections/shared/input-popover.tsx
@@ -45,11 +45,13 @@ const Input = ({
 
   return (
     <CssValueInput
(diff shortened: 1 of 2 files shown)


TrySound (approved):

johnsicili (approved):
Nice one. Field sizing has rough support but I suppose the tradeoff is worth it

kof:
> Nice one. Field sizing has rough support but I suppose the tradeoff is worth it

Chromium has it, so that's the main thing.

PR #5057, opened 2025-03-27
fix: skip template prompt if `--template` is used

aleclarson:
The `webstudio init` command was prompting for a template even if `--template` was used. This fix avoids that behavior.

<sup>Simple fix so I skipped the PR template 🙈 </sup>


TrySound:
Tested. All works. Thanks!

TrySound (approved):

PR #5058, opened 2025-03-27
fix: use non-zero exit code for cancelled init flow

aleclarson:
A non-zero exit code indicates that the process did not complete. Alternatively, we could use a 130 exit code, which commonly refers to Control+C termination.


TrySound (approved):

TrySound:
Thanks!

PR #5063, opened 2025-03-27
feat: add "Accessible Label" setting

aleclarson:
## Description

Allow the user to set the [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) attribute.

This is maybe a controversial feature, but it would be useful for post-processing that I'm doing on the generated output. 😅 

Screenshot 2025-03-27 at 14 54 23
Screenshot 2025-03-27 at 14 56 51

## Before merging

- [x] tested locally ~~and on preview environment (preview dev login: 0000)~~
- [ ] added tests
- [ ] ~~if any new env variables are added, added them to `.env` file~~

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/property-label.tsx
+++ b/apps/builder/app/builder/features/settings-panel/property-label.tsx
@@ -17,7 +17,11 @@ import { showAttribute } from "@webstudio-is/react-sdk";
 import { updateWebstudioData } from "~/shared/instance-utils";
 import { $selectedInstance } from "~/shared/awareness";
 import { $props, $registeredComponentPropsMetas } from "~/shared/nano-states";
-import { humanizeAttribute, showAttributeMeta } from "./shared";
+import {
+  ariaLabelAttributeMeta,
+  humanizeAttribute,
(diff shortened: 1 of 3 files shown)


kof:
User can add aria-label to any instance, not sure why aria-label needs to become a system property.


image

kof:
It would be nice to expose aria-label on all components where likelyhood of using it is high as initialProperties, so that user doesn't need to add it, its already visible

TrySound:
Is it necessary for all components or only buttons and links which often have icons as content?

TrySound:
So yes the best option is specifying aria-label where needed
https://github.com/webstudio-is/webstudio/blob/main/packages/sdk-components-react/src/button.ws.ts#L33

System props and core components are used to manage their behaviour with builder and cli instead of generic code generation.

kof:
I meant adding it to initial only on components where users would be using aria label potentially, buttons, links and similar, I don't remember if we have it in initialProps in all of them, maybe its a non-issue.

PR #5065, opened 2025-03-28
fix: backspace when StyleSourceInput is empty

aleclarson:
## Description

At some point, the local token was moved to the start of the style source array, but the backspace logic was never updated. This led to the 2nd-to-last source being detached, which is unexpected.

## Steps for reproduction

1. Select a layer with 2+ style sources
2. Press the Command+Enter keys to focus the style sources input.
3. Press the Backspace key and notice the 2nd-to-last source gets detached, but it *should* be the last source getting detached

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/style-source/style-source-input.tsx
+++ b/apps/builder/app/builder/features/style-panel/style-source/style-source-input.tsx
@@ -535,7 +535,7 @@ export const StyleSourceInput = (
         label === "" &&
         props.editingItemId === undefined
       ) {
-        const item = value[value.length - 2];
+        const item = value[value.length - 1];
         if (item) {
           props.onDetachItem?.(item.id);
         }


TrySound:
I think current behavior makes sense

1. local style source cannot be deleted
2. caret is placed between tokens and local style source so it is expected backspace should remove things on the left

With your change caret removes a thing on the right and then tokens cannot be removed at all


https://github.com/user-attachments/assets/15be088a-c587-489f-8e2e-34c4b7e10e1a


> At some point, the local token was moved to the start of the style source array,

This should not be possible. We enforce local styles to be placed in the end. If placed at the start it must be a bug.

kof:
I agree with @TrySound, should we close?

aleclarson:
I'm gonna take another look in a few days. There's definitely a bug here, even if my solution is incorrect.

aleclarson:
We don't use this feature anymore, so I'm gonna close this.

PR #5068, opened 2025-03-28
experimental: add field label for animation section

TrySound:
Added generic field label to give animation labels similar experience to property labels.
Screenshot 2025-03-28 at 15 30 34


TrySound:
Markdown

Screenshot 2025-03-28 at 17 10 22

kof:
You probably want to adjust the padding slightly, seems like the list is rendered with default browser styles and padding is too much

PR #5070, opened 2025-03-28
fix: prevent dragging folder inside of itself

TrySound:
User reported the case which probably bugged us a few times but we could not figure out why.

Here added guard to prevent reparanting page folder inside of itself or child folders.


kof (approved):
didn't test

PR #5074, opened 2025-04-01
feat: enforce html content model

TrySound:

This is a big step into native html elements support. I scraped content model from html spec and now enforcing all html tags structure so we no longer need to write custom restrictions.

Now paragraph cannot be nested into heading. Box cannot be nested into span Text though still can in div Text. Any interactive elements cannot be nested automatically. List item should have a list as direct parent (collection and slot are exception because they don't have elements inside).

Most rules are defined in this table
https://html.spec.whatwg.org/multipage/indices.html#elements-3


apps/builder/app/shared/content-model.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+  // for example ul has "li" as children category
+  if (allowedCategories.includes(tag)) {
+    return true;
+  }
+  // instance does not match parent constraints
+  if (isIntersected(allowedCategories, categoriesByTag[tag]) === false) {
+    return false;
+  }
+  // prevent accepting interactive when ancestor enforced non-interactive elements
```
istarkov, line 57: what this means? Why this can occure, explanation and examples needed

apps/builder/app/shared/content-model.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+    const tagByInstanceId = getTagByInstanceId(props);
+    const meta = metas.get(instance.component);
+    const tag = tagByInstanceId.get(instance.id) ?? getMetaTag(meta);
+    allowedCategories = getTagChildren(tag, allowedCategories);
+  }
+  return allowedCategories;
+};
+
+export const isTreeSatisfyingContentModel = ({
```
istarkov, line 129: plz give explanation of whats happening here why we need getTagChildren etc

apps/builder/app/shared/content-model.ts
```diff
@@ -0,0 +1,175 @@
...
+    return false;
+  }
+  // prevent accepting interactive when ancestor enforced non-interactive elements
+  if (
+    allowedCategories.includes("non-interactive") &&
+    categoriesByTag[tag].includes("interactive")
+  ) {
+    return false;
+  }
+  // prevent accepting form when ancestor enforced non-form elements
```
istarkov, line 64: also explanation why its not covered by main algorithm

apps/builder/app/shared/content-model.test.tsx
```diff
@@ -0,0 +1,356 @@
+import { expect, test } from "vitest";
+import { coreMetas } from "@webstudio-is/sdk";
+import * as baseComponentMetas from "@webstudio-is/sdk-components-react/metas";
+import { $, renderData, ws } from "@webstudio-is/template";
+import { isTreeSatisfyingContentModel } from "./content-model";
+
+// @todo maybe change api to do better errors
```
kof, line 7: would be great to also add an example of what's wrong right now, later if someone else wants to work on it, they won't know what this means

TrySound:
> What if Box is dropped in first and then user selects tag span?

We need tag control which will allow switching to only satisfying tags. Also it will replace tag prop with instance.tag.

packages/html-data/bin/crawler.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+  return result;
+};
+
+export const getAttr = (
+  node: NodeWithChildren,
+  name: string
+): undefined | Attribute => {
+  return node.attrs.find((attr) => attr.name === name);
```
kof, line 35: why the first attr and not the last one for example? is it obvious?

kof (approved):

packages/html-data/bin/crawler.ts
TrySound: Do you expect multiple attributes to have the same name?

kof: generally possible, but not sure how relevant here

TrySound: 🤷

PR #5075, opened 2025-04-01
experimental: Animation descriptions

kof:
## Description

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/props-section/animation/animation-section.tsx
+++ b/apps/builder/app/builder/features/settings-panel/props-section/animation/animation-section.tsx
@@ -41,6 +41,9 @@ const animationTypeDescription: Record<AnimationAction["type"], string> = {
   view: "View-based animations occur when an element enters or exits the viewport. They rely on the element’s visibility rather than the scroll position.",
 };
 
+const insetDescription =
+  "Adjusts the animation’s start/end position relative to the scrollport. Positive values move it inward (delaying start or hastening end), while negative values move it outward (starting animation before visibility or continuing after disappearance).";
+
 const animationTypes: AnimationAction["type"][] = Object.keys(
   animationTypeDescription
 ) as AnimationAction["type"][];
@@ -247,7 +250,7 @@ export const AnimationSection = ({
           padding: theme.panel.paddingInline,
         }}
       >
-        <FieldLabel description="Allows to preview the animation on the canvas in design mode.">
+        <FieldLabel description="Even if its off, you can preview the animation by selecting the item in the instance.">
           Run on canvas
         </FieldLabel>
         <Tooltip content={value.isPinned ? "Off" : "On"}>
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/features/settings-panel/props-section/animation/animation-section.tsx
```diff
@@ -251,7 +247,7 @@ export const AnimationSection = ({
           padding: theme.panel.paddingInline,
         }}
       >
-        <Text variant={"titles"}>Animation</Text>
```
johnsicili, line 254: Can this be a field label instead? Enables reset and also a static tooltip that makes more sense IMO. Like this:

"Enable this to always run the animation on the canvas, even when the animation instance is not selected."

kof: indeed

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-section.tsx
```diff
@@ -383,12 +369,12 @@ paragraph
...
             css={{ gridTemplateColumns: "1fr 1fr" }}
           >
-            <FieldLabel>
+            <FieldLabel description="Inset defines the point where an element’s left/top edge enters the viewport, setting the start position for the animation’s progress.">
               {value.axis === "inline" || value.axis === "x"
                 ? "Left Inset"
                 : "Top Inset"}
             </FieldLabel>
-            <FieldLabel>
+            <FieldLabel description="Inset defines the point where an element’s right/bottom edge enters the viewport, setting the start position for the animation’s progress.">
```
johnsicili, line 377: I see two issues with this. 1) I think it can be negative so not just entering the viewport but also outside of it and 2) I think the Top and Bottom might need different descriptions.

> The view-timeline-inset [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) property is used to specify one or two values representing an adjustment to the position of the scrollport (see [Scroll container](https://developer.mozilla.org/en-US/docs/Glossary/Scroll_container) for more details) in which the subject element of a named view progress timeline animation is deemed to be visible. Put another way, this allows you to specify start and/or end inset (or outset) values that offset the position of the timeline.

https://developer.mozilla.org/en-US/docs/Web/CSS/view-timeline-inset

johnsicili (approved):
Some minor stuff

apps/builder/app/builder/features/settings-panel/props-section/animation/animation-section.tsx
johnsicili: > If the value is positive, the position of the animation's start/end will be moved inside the scrollport by the specified length or percentage.
If the value is negative, the position of the animation's start/end will be moved outside the scrollport by the specified length or percentage, i.e. it will start animating before it appears in the scrollport, or finish animating after it leaves the scrollport.

kof: Inset defines the point where an element’s edge enters the viewport, setting the start position for the animation’s progress.

works for you universally for all insets?

johnsicili: Eh for a short one this

> Adjusts the animation’s start/end position relative to the scrollport.

If we want more context:

> Adjusts the animation’s start/end position relative to the scrollport. Positive values move it inward (delaying start or hastening end), while negative values move it outward (starting animation before visibility or continuing after disappearance).

PR #5080, opened 2025-04-02
refactor: add tag property control

TrySound:
https://github.com/webstudio-is/webstudio/issues/3632

Here added dedicated tag property control which will automatically migrate Box, Text and Heading components to instance.tag internally and will be used for ws:element component.

Later will extend tag control with tag descriptions and auto inferring matching tags.

Screenshot 2025-04-02 at 15 56 46


kof (approved):
This is awesome!

PR #5085, opened 2025-04-03
fix: allow placing labelable elements inside of labels

TrySound:
Fixes the issue reported in discord (https://discord.com/channels/955905230107738152/1356788803624501370/1357299346558816306)

Found the spec actually allows to place button and input inside a label. So now we can generalize this logic.

Also content model validation no longer prevents inserting invalid structure and only warns user is something went wrong. This will let us to never block the user with false positives and maybe even figure out the issue on their own.

Screenshot 2025-04-04 at 01 28 53


kof (approved):
didn't test

PR #5092, opened 2025-04-04
refactor: paste new tag with webflow and markdown

TrySound:

Here migrated webflow and markdown paste plugins to instance.tag to avoid adding more legacy "tag" prop.

Also fixed content model for instance.tag and rewrote its tests with ws:element.


kof (approved):
didn't test

PR #5101, opened 2025-04-07
refactor: add component content model

TrySound:
Extended content model to support categories defined in component meta. Most components has transparent children categories to pass through constraints from parent, this way we allow descendants with non category.

Here migrated only core and base components.
Will extend more for animations in another PR.


kof (approved):
didn't test

PR #5109, opened 2025-04-09
refactor: use content model to detect rich text

TrySound:

Here I finally got rid from "rich-text-child" necessity. With html elements in the tree it will no longer work so we need a new way to work with rich text.

Added a few new utilities to content model which detect rich text by text inside or any textual tags. Span and links can still be detected as containers rather than children when not surrounded by text and other textual elements.


kof (approved):
didn't test

PR #5118, opened 2025-04-11
fix: autocomplete properties with invalid values

TrySound:
The logic assumed invalid values should become custom properties. Here replaced that heuristic with property name check.

Screenshot 2025-04-11 at 19 11 54


johnsicili (approved):

kof (approved):
didn't test

PR #5121, opened 2025-04-12
feat: avoid generating unused preset styles

TrySound:
Now we have first-class support for html tags and able to figure out which tags are rendered so unused preset styles can be removed from websites.

For example

```
<Box ws:tag="div">
  <Box ws:tag="span">
  </Box>
</Box>
```

will render only div.w-box and span.w-box though there are more presets for for other tags like main, section etc.

Before all tags were generated


istarkov (approved):

kof (approved):
didn't test but sounds great

PR #5123, opened 2025-04-14
fix: show token name when convert local styles to token

TrySound:
Ref https://discord.com/channels/955905230107738152/1360809139810599063/1360809139810599063

There were two problems here

1. empty token name collapsed and was not editable; fixed it with nbsp
2. the label set with dom didn't have proper timing; fixed it by rendering label by react


kof:
Found another little problem https://share.descript.com/view/Y1WqH3E2BN4

kof (approved):

PR #5126, opened 2025-04-15
experimental: split keyframe styles into popover

TrySound:
We want to add a few predefined fields in every keyframe. Here we split each keyframe into separate popover to avoid overloading the ui.

Screenshot 2025-04-16 at 01 53 41


johnsicili (approved):
Good. Downside to this is it's hard to see which keyframe has which properties. Most use cases each keyframe uses the same property so not a big deal

kof (approved):

PR #5129, opened 2025-04-16
refactor: bump radix

TrySound:
https://www.radix-ui.com/primitives/docs/overview/releases#april-8-2025


kof:
Looks like tooltips need to be tested in particular well with style panel logic we have and other places

kof (approved):

PR #5132, opened 2025-04-17
refactor: detect empty tags and components

TrySound:
Here replaced "embed" and "control" components with content model. Many components now detect no content based on html content model like with img or input tags. In other cases component meta can explicitly specify empty array to not accept anything.

```
meta = {
  contentModel: {
    category: 'instance',
    children: []
  }
}
```


packages/html-data/bin/html-crawler.ts
```diff
@@ -50,7 +50,7 @@ const childrenCategoriesByTag: Record<string, string[]> = {};
     const children = parseList(getTextContent(row.childNodes[4]));
     for (const tag of elements) {
       categoriesByTag[tag] = categories;
-      childrenCategoriesByTag[tag] = children;
+      childrenCategoriesByTag[tag] = children.includes("empty") ? [] : children;
```
kof, line 53: missed one

kof (approved):

packages/html-data/bin/html-crawler.ts
TrySound: It is html table parser

PR #5140, opened 2025-04-18
refactor: describe descendants in component content model

TrySound:
Here replaced "transparent" keyword with more explicit "descendants" list in component content model.

This will enable use case with child constraint while allowing other instances to be children like here

```
  contentModel: {
    category: "instance",
    children: [
      "instance",
      "AnimateText",
      "StaggerAnimation",
      "VideoAnimation",
    ],
  },
```

and descendants are described like this

```
  contentModel: {
    category: "instance",
    children: ["instance"],
    descendants: ["VimeoSpinner", "VimeoPlayButton", "VimeoPreviewImage"],
  },
```

### DIFF
--- a/apps/builder/app/shared/content-model.ts
+++ b/apps/builder/app/shared/content-model.ts
@@ -246,10 +246,6 @@ const getAllowedParentCategories = ({
   metas: Metas;
   instanceSelector: InstanceSelector;
 }) => {
-  // @todo
-  if (instanceSelector.length === 0) {
-    // return;
-  }
   const instanceId = instanceSelector[1];
   const instance = instances.get(instanceId);
   if (instance === undefined) {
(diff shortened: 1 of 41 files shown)


apps/builder/app/shared/content-model.ts
```diff
@@ -234,24 +218,47 @@ const computeAllowedComponentCategories = ({
...
+  metas,
+  instanceSelector,
+}: {
+  instances: Instances;
+  metas: Metas;
+  instanceSelector: InstanceSelector;
+}) => {
+  // @todo
+  if (instanceSelector.length === 0) {
+    // return;
```
kof, line 251: might be good to comment what this handles

kof (approved):

kof:
Btw is there a way to describe direct descendants vs descendants in some subtree? E.g. if you want to enforce that direct children of the component are X vs it just needs a child X anywhere inside of that component.

TrySound:
@kof this is what PR about, see examples above. It allows specific children or descendants which has `category: none` and cannot be placed anywhere else.

PR #5160, opened 2025-04-24
perf: reuse aria attributes between all components

TrySound:

We have a universally generated list of aria attributes and can reuse it instead of generating for every component and bloating our client bundle.

Now users will have to download ~700kB less javascript when load builder.

```diff
- 5748    build/client/assets
+ 5036    build/client/assets
```

Screenshot 2025-04-24 at 12 01 46


packages/html-data/bin/aria.ts
```diff
@@ -24,6 +24,13 @@ type Attribute = {
   options?: string[];
 };
 
+const overrides: Record<string, Partial<Attribute>> = {
+  "aria-label": {
+    description:
```
kof, line 29: why is it mauallly written?

kof (approved):
I don't understand why separating specifically aria attributes and not all common/universal html attributes

packages/html-data/bin/aria.ts
TrySound: Because step didn't have a good one. AI didn't generate a good one either (we had the same manual override).

TrySound:
Because aria attributes has own rules based on roles. Regular attributes are based on tags. We are not doing it right at the moment though will improve in the future.

PR #5162, opened 2025-04-24
feat: add context menu to animations

TrySound:
Now user can copy and paste specific animations between animation groups.

Screenshot 2025-04-24 at 23 41 34

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/props-section/animation/animations-select.tsx
+++ b/apps/builder/app/builder/features/settings-panel/props-section/animation/animations-select.tsx
@@ -1,4 +1,4 @@
-import { useState, useMemo, type ReactNode } from "react";
+import { useState, useMemo, type ReactNode, useRef } from "react";
 import {
   theme,
   DropdownMenu,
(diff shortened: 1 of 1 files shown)


kof (approved):
Works great, I would suggest 2 improvements:

1. ability to copy all animations, similar to how we have it in advanced section, so this way if I have many animations, I don't have to copy them 1 by one. Call it "Paste animations" - this works for one or many

2. Add meta to the json similar to what we have on the instance: "@webstudio/instance/v0.1", so its easier to identify what this is and we can also version it if we need a schema change at some point

kof:
this schema would be great: `{'@webstudio/animation/v0.1': Array<Animation> }`

TrySound:
I don't think versioning is important for small short term copy paste here

kof:
> I don't think versioning is important for small short term copy paste here

Yeah, we don't know, still nicer to have a consistent structure with an identifier

kof:
Interesting thing I noticed is that description is part of the object:
`{"@webstudio/animation/v0.1":[{"name":"Video Ranges","description":"Fade in the element as it scrolls into view.","keyframes":[{"offset":0,"styles":{"opacity":{"type":"unit","unit":"number","value":0}}}],"timing":{"easing":"linear","fill":"backwards","rangeStart":["cover",{"type":"unit","value":0,"unit":"%"}],"rangeEnd":["cover",{"type":"unit","value":100,"unit":"%"}]}}]}`

PR #5175, opened 2025-05-01
refactor: cleanup legacy descriptions

TrySound:
Here dropped legacy descriptions generated with AI. All attributes are using descriptions scraped from html spec. Though many component properties redefine these attributes without descriptions. So to workaround just always fall back to attributes descriptions. Covers most cases.


packages/sdk-components-react-radix/src/__generated__/popover.props.ts
```diff
@@ -17,7 +17,6 @@ export const propsPopoverContent: Record<string, PropMeta> = {
     type: "string",
     defaultValue: "center",
     options: ["center", "start", "end"],
-    description: "Specifies the horizontal alignment of the element.",
```
kof, line 20: Is this one gonna be correct for example?

kof (approved):

packages/sdk-components-react-radix/src/__generated__/popover.props.ts
TrySound: Previously take from legacy html attribute though not necessary means the same in radix.

PR #5181, opened 2025-05-03
feat: support paste from other browsers

TrySound:
Users started to build own marketplaces. And we found paste something copied from other browser like firefox -> chromium based app doesn't work. This is because OSs guarantee only text/plain to be stored in system clipboard.

Here I changed copy paste logic to copy as text/plain and paste as both application/json and text/plain if possible.

### DIFF
--- a/apps/builder/app/shared/copy-paste/init-copy-paste.ts
+++ b/apps/builder/app/shared/copy-paste/init-copy-paste.ts
@@ -3,7 +3,7 @@ import {
   $authTokenPermissions,
   $textEditingInstanceSelector,
 } from "../nano-states";
-import * as instance from "./plugin-instance";
+import { instanceText, instanceJson } from "./plugin-instance";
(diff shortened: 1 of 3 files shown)


kof (approved):

kof:
We should probably check if we have other places that still use application/json

apps/builder/app/shared/copy-paste/init-copy-paste.ts
```diff
@@ -64,7 +64,7 @@ const validateClipboardEvent = (event: ClipboardEvent) => {
 
 const defaultMimeType = "application/json";
```
kof, line 65: should that still be the defaultMimeType?

TrySound:
Webflow 😄

kof:
> Webflow 😄

I mean some other places, e.g. resource copy

TrySound:
We don't have resource copy yet

PR #5183, opened 2025-05-04
fix: set xml node background to white

### LINKED ISSUE #5015: Use white bg and padding on XML pages
We launched Headslot with a white background and padding here:

https://github.com/webstudio-is/webstudio/blob/8d05217cf300368ddbdf4578369dcac580c7fa27/packages/sdk-components-react/src/head-slot.tsx#L26

That made it viewable on sites with a dark theme.

We still have the problem on sitemaps though:

Image

JayaKrishnaNamburu:
## Description

fixes #5015 
Added `backgroundColor` to `white` for `XMLNode` so the content is visible even when the `GlobalRoot` is set to something else.

## Steps for reproduction

- Create a demo project.
- Create a XML with sitemap.xml
- Now, switch back to the home page and set the `backgroudnColor` of the global root to some random colour.
- Then, switch back to the sitemap page. Now, the contents are not visible as the background is inherited from the GlobalRoot. This fix will help in setting to white whenever a xml node is used.

## Code Review
  - test it on preview

## Before requesting a review
- [x] made a self-review

## Before merging

### DIFF
--- a/packages/sdk-components-react/src/head-slot.tsx
+++ b/packages/sdk-components-react/src/head-slot.tsx
@@ -26,7 +26,6 @@ export const HeadSlot = forwardRef<
     <div
       ref={ref}
       style={{
-        backgroundColor: "rgba(255,255,255,1)",
         padding: "8px",
         position: "fixed",
         top: 0,
--- a/packages/sdk-components-react/src/xml-node.tsx
+++ b/packages/sdk-components-react/src/xml-node.tsx
@@ -71,7 +71,7 @@ export const XmlNode = forwardRef<ElementRef<"div">, Props>(
     };
 
     return (
-      <div {...props}>
+      <div {...props} style={{ backgroundColor: "rgba(255,255,255,1)" }}>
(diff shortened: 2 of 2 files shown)


packages/sdk-components-react/src/head-slot.tsx
```diff
@@ -26,7 +26,6 @@ export const HeadSlot = forwardRef<
     <div
       ref={ref}
       style={{
-        backgroundColor: "rgba(255,255,255,1)",
```
TrySound, line 29: Why remove here?

JayaKrishnaNamburu: Just felt redundant as XmlNode is going to apply it anyways

TrySound: Ah, got it. Xml node is used by head slot.

TrySound (approved):

TrySound:
Thanks!

PR #5189, opened 2025-05-07
experimental: paste inline styles

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/4404

Inline styles are easy to support and tailwind templates occasionally use them for unsupported stuff. Also fixed space collapsing.

https://github.com/user-attachments/assets/35e7a8d7-ef5b-496c-9a6d-0b05e113e503


kof (approved):
only question I have is what if user actually wants inline styles :)

TrySound:
@kof We don't support them neither in style panel nor in settings panel 😉

PR #5194, opened 2025-05-09
fix: allow deleting from input in preview mode

TrySound:
Fixes https://discord.com/channels/955905230107738152/1369211497590554675

Delete on canvas is separateed from the one in builder so we can safely remove preventDefault to allow entering or deleting from inputs in preview mode.


kof (approved):
Are you sure? 

This was the original change https://github.com/webstudio-is/webstudio/commit/e9fa0cd53010967c1b3b3dffaa0434a96d64e6d9

TrySound:
Yep, fixes canvas doesn't affect builder

PR #5204, opened 2025-05-12
experimental: add matching tag when insert element component

TrySound:

Now element does not provide default tag and instead builder will try to find tag matching its parent.

This way we can insert element, give it "ul" tag
and then insert another element into it which will match only "li"

https://github.com/user-attachments/assets/3ea87624-9524-4f2e-91b2-d658d16d7a1b

### DIFF
--- a/apps/builder/app/shared/instance-utils.ts
+++ b/apps/builder/app/shared/instance-utils.ts
@@ -311,13 +311,13 @@ export const insertWebstudioElementAt = (insertable?: Insertable) => {
   let matchingTag: undefined | string;
   for (const tag of tags) {
     element.tag = tag;
-    const isMatching = isTreeSatisfyingContentModel({
+    const isSatisfying = isTreeSatisfyingContentModel({
       instances: newInstances,
       props,
       metas,
       instanceSelector: [element.id, ...insertable.parentSelector],
     });
-    if (isMatching) {
+    if (isSatisfying) {
       matchingTag = tag;
       break;
     }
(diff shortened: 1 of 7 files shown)


apps/builder/app/shared/instance-utils.ts
```diff
@@ -258,6 +261,80 @@ export const insertInstanceChildrenMutable = (
...
+    id: nanoid(),
+    component: elementComponent,
+    children: [],
+  };
+  const newInstances = new Map(instances);
+  newInstances.set(element.id, element);
+  let matchingTag: undefined | string;
+  for (const tag of tags) {
+    element.tag = tag;
+    const isMatching = isTreeSatisfyingContentModel({
```
kof, line 314: I was confused by the "matching", to me this is isSatisfying or something, matching sounds to me like you are searching for a tag that will match a specific other tag

kof (approved):
didn't test

packages/sdk/src/__generated__/tags.ts
```diff
@@ -1,4 +1,6 @@
 export const tags: string[] = [
```
kof, line 1: this was sorted alphabetically before, the need to have div/span first is unobvious and you can't rely on alphabetic order any more in case something was expecting alphabetic

TrySound: Nothing relied on it. We just give div and span more priority when auto insert

PR #5208, opened 2025-05-14
experimental: adapt tag control to html content model

TrySound:

Here improved tag control to allow updating with tag satisfying content model.

For example when ul has li it can only be changed to ol and menu. li inside ul cannot be changed with anything.

Also switch to combobox when there is more than 10 satisfying tags.

https://github.com/user-attachments/assets/ad42bba7-9be1-4f50-9b1a-6f6cda52f3e5


kof (approved):
very cool

PR #5212, opened 2025-05-16
feat: add default time for time component

### LINKED ISSUE #3510: Use current date by default in time component
As a user I want to display the current date in the footer for the copyright. There will be other use cases where you want a current date, but I don't know any use case for not having a date.

To avoid hydration errors:
- [ ] render serverside current date but always with 0 seconds, this way we reserver approximately the needed space for the client render, this will also work for most dates that don't display seconds
- [ ] render in effect on the client the current date

JayaKrishnaNamburu:
fixes #3510 

## Description

Adding a new default `new Date.now()` support to the time component. So, users can just drop and use the component whenever they want to show the current time. Rendering the default time using seconds as `0` on the server side. And then replacing the value on the client side. Things that need to be taken care of

- Time component should support any static values that is passed to it.
- If no value is passed, it should use the `new Date()` and respect all the configs on the component.
- If no value is passed, it should render the date, month and year.

## Steps for reproduction

- Add a time element to the canvas and pass a static value.
- Add a time element to the canvas without any defaults.

Currently there is a small flicker for seconds when the project is hydrated since we render with `0` seconds on the server side and replace with proper seconds on the client side. The flicker is expected IMO, as we are resetting and using it for render for space concerns.
default-time-with-hydration-seconds

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk-components-react/src/time.tsx
+++ b/packages/sdk-components-react/src/time.tsx
@@ -1,4 +1,10 @@
-import { forwardRef, type ComponentProps, type ElementRef } from "react";
+import {
+  forwardRef,
+  useEffect,
+  useState,
+  type ComponentProps,
+  type ElementRef,
+} from "react";
 
 const languages = [
   "af",
@@ -281,7 +287,15 @@ type Country = (typeof countries)[number];
 type DateStyle = Intl.DateTimeFormatOptions["dateStyle"] | "none";
 type TimeStyle = Intl.DateTimeFormatOptions["timeStyle"] | "none";
 
-const INITIAL_DATE_STRING = "dateTime attribute is not set";
+const createInitialDate = () => {
+  const date = new Date();
+  date.setSeconds(0);
+  date.setMilliseconds(0);
+  return date;
+};
+
+const INITIAL_DATE_STRING =
+  typeof window === "undefined" ? createInitialDate() : null;
 const INVALID_DATE_STRING = "";
 
 const DEFAULT_LANGUAGE = "en";
@@ -365,6 +379,9 @@ export const Time = forwardRef<ElementRef<"time">, TimeProps>(
     },
     ref
   ) => {
+    const [clientSideDate, setClientSideDate] = useState<string | Date | null>(
+      null
+    );
     const locale = `${languageOrDefault(language)}-${countryOrDefault(
       country
     )}`;
@@ -374,8 +391,21 @@ export const Time = forwardRef<ElementRef<"time">, TimeProps>(
       timeStyle: timeStyleOrUndefined(timeStyle),
     };
 
+    useEffect(() => {
+      if (dateTime === INITIAL_DATE_STRING) {
+        setClientSideDate(new Date());
+      }
+    }, [dateTime]);
+
+    const dateToUse =
+      typeof window !== "undefined" &&
+      clientSideDate !== null &&
+      dateTime === INITIAL_DATE_STRING
+        ? clientSideDate
+        : dateTime;
+
     const datetimeString =
-      dateTime === null ? INVALID_DATE_STRING : dateTime.toString();
+      dateToUse === null ? INVALID_DATE_STRING : dateToUse.toString();
 
     const date = parseDate(datetimeString);
     let formattedDate = datetimeString;
@@ -389,7 +419,12 @@ export const Time = forwardRef<ElementRef<"time">, TimeProps>(
     }
 
     return (
-      <time ref={ref} dateTime={datetimeString} {...props}>
+      <time
+        ref={ref}
+        dateTime={datetimeString}
+        {...props}
+        suppressHydrationWarning
+      >
         {formattedDate}
       </time>
     );


JayaKrishnaNamburu:
Any ideas on how to handle or if we want to handle when server and client are in different time-zones. Here is the current behaviour. If when they both are in different time-zones. Should we take into consideration of the clients time-zone in server during SSR process ?

Because, on canvas we use clients timezone. On server it is server based. Again on client side from deployed site. It shows the client side timezone with a flicker.
server-client-timezone

kof:
another option as opposed to suppressing and potentially rendering a different date could be reading the date from the server and using it to construct the new Date object

JayaKrishnaNamburu:
> another option as opposed to suppressing and potentially rendering a different date could be reading the date from the dom and using it to construct the new Date object

Yes doing the same once the component is fully hydrated. But still there will be a flicker from the server side as we render on server with 0 seconds or any seconds for that matter as they change.

I think the big bug would be with the flicker with timezones between server and client.

kof:
Can't you set the seconds/ms/tz to a single value when its the default value so that its always matching and then read it?

JayaKrishnaNamburu:
If there is no default by user. Component is using Date.now() inside server. So, basically it picks the server timezone. 

https://github.com/webstudio-is/webstudio/pull/5212/files

Looks like there are ways to avoid such behaviours

https://github.com/remix-run/remix/discussions/9039#discussioncomment-8767904

Will try few things out and update the PR tonight 👍🏽

JayaKrishnaNamburu:
Just dropping it for reference, even in the case of static value, the date is rendered using the server time. Even if we input the timezone.

Screenshot 2025-05-16 at 9 22 04 PM


Screenshot 2025-05-16 at 9 22 12 PM

kof:
I am not sure what you mean, but can you avoid creating a new date on the client entirely and use the server-rendered date fully? e.g. supply the original date in an attribute because text is formatted so it may not have parts of the date

JayaKrishnaNamburu:
> I am not sure what you mean, but can you avoid creating a new date on the client entirely and use the server-rendered date fully

We can, but the catch is while server rendering it. It needs to be aware of where the client is located in. As the SSR uses the server time. So, if both client and server are in entirely different timezones. You will only see server time and miss the client timezone if not rendered again on client side. One easy way to do this is marking the component as client component. But we will loose ssr benefit for it.

kof:
> We can, but the catch is while server rendering it. It needs to be aware of where the client is located in. As the SSR uses the server time. 

Let it user server time, use the same time on the client

kof:
> But we will loose ssr benefit for it.

Making it pure client-side rendered date could be an option too, I don't know if SSR is actually important here, but if its used in the hero section, it will jump around

JayaKrishnaNamburu:
Updated the component now with the default time behaviour. This will no reflect the current time
Here is a project
https://p-5212e8bc-eb36-4bee-9218-77885cf13e01-dot-default-time.staging.webstudio.is/
Deployed version
https://jk-time-element-ck8hm.wstd.work/

This the current behaviour from the `main` branch
https://date-time-g5ijg.wstd.io

The only side effect is, this will currently throws an error with `Mismatch of hydration` error. Which is happening on `main` too. So, left is as is. I tried to fix it, by adding `supressHydrationWarning` flag. But, this is preventing the first re-render for static values. Let me know, if the behaviour is looking ok.

The only other options is
- Completely mark the component as client component (Can create layout shits after hydration). 

cc @kof @TrySound @istarkov

kof:
Why can't we use server-rendered date including everything, including timezone?

JayaKrishnaNamburu:
> Why can't we use server-rendered date including everything, including timezone?

We are not taking timezone from users right now. If they can provide it from the settings tab on the right. We can use that to render the users timezone. So, that flicker reduces more, and maybe sometimes at max to seconds. If at all it happens.
Screenshot 2025-05-20 at 9 27 10 PM

kof:
If there is a tz provided by user - it should be used, otherwise a timezone on the server should be used for both server and client-side, I still don't see the problem

JayaKrishnaNamburu:
@kof i thought of that too. But, it still doesn't cover all the cases. For example, If users don't provide any time related data in the string like `+5:30` or `Indian Standard Time` etc. We can't detect that in server as well. Right now, the canvas takes the browser time. And displays it. And when deployed, it takes the server time, and hydrates with client time again (throwing hydration error). 

Screenshot 2025-05-22 at 8 20 25 PM

So, if we are ok with hydration errors in console. It's pretty much straight forward. It's only a issue when we want to avoiding those hydration error messages 😅 

Screenshot 2025-05-22 at 8 24 40 PM

That's why i am not leaned towards taking `timeZone` from the prop string and using it. It doesn't solve all cases. My suggestion is. We can add a additational dropdown. Which shows list of all timezones. By default it selects the users browser timezone. And we can take that prop and render in server and client. And even if they want to explicitly change it, they can still do it.

kof:
If something isn't provided, use the default server-side, then read it from html on the client

kof:
Parsing a time string including timezone can't be a big deal. Please don't write a regex for that. Find a tested solution.

kof:
Weird!

In that case either just client-only rendering and accepting the potential visible rerender on hydration or reducing this thing to just a year and month so that timezone doesn't matter

kof:
I quickly asked gpt, it suggested https://github.com/taylorhakes/fecha/blob/master/src/fecha.ts

kof:
Another one it suggested is https://github.com/iamkun/dayjs

JayaKrishnaNamburu:
@kof i gave multiple shots at this. Even with different approaches. Almost most of them get effected with the browser or the server env. The fast easy way could be to have a fixed height to reduce the jump if we mark this as client component.

kof:
@JayaKrishnaNamburu that's not going to work, if this issue is too hard, pls don't waste your time.

JayaKrishnaNamburu:
Ok, feel free to close this one. I will start looking into other issues that are on me 👍🏽

kof:
Successor https://github.com/webstudio-is/webstudio/pull/5434

PR #5227, opened 2025-05-21
fix: make all radix components uncontrolled

TrySound:
Here improved handling radix states by making all components uncontrolled yet still synchronizing external state with local one.


kof (approved):

apps/builder/app/shared/content-model.ts
```diff
@@ -424,7 +424,6 @@ export const richTextPlaceholders: Map<undefined | string, string> = new Map([
   ["blockquote", "Blockquote"],
   ["li", "List item"],
   ["a", "Link"],
-  ["span", "Span"],
```
kof, line 427: accidental?

TrySound: Nope, span is often used empty as non-textual container in radix components

PR #5228, opened 2025-05-22
feat: add wrap in element command

TrySound:

Here "Wrap in Element" is replacing "Wrap in Box" as more general option.


kof (approved):
To release this we need to release the element first or otherwise keep both in place?

TrySound:
Not quite blocking. Just another entry point for element. Though I will remove flag before release.

kof:
But then don't delete wrap in a Box if Box is still there and Element isn't

TrySound:
Only in components list

PR #5230, opened 2025-05-23
feat: migrate forms components to html elements

TrySound:

This is big one. So I switched to html elements whole forms section in components. Though here I introduced some deviation from specification

**Textarea**

Initial value is represented as text content
```html
<textarea>content</textarea>
````

Though it is trickier to manage text content as state and all frameworks rely on value property of textarea dom interface.

**Select**

Similar story. There is no value attribute and state is represented by selected boolean attribute on options

```html
<select>
  <option value="value1" selected></option>
  <option value="value2"></option>
</select>
```

We replace selected attribute on option with value attribute on selected when paste.

So basically all form elements are using "value" and "checked" attributes which can be easily mapped to html and all frameworks.


kof (approved):
Do you want me to test?

PR #5231, opened 2025-05-24
feat: add element component

TrySound:

Here unflagged element component in components panel and html paste.

Additionally

- migrate link template to html element
- added :local-link state to links which is replaced with [aria-current=page] when css is generated, sadly not yet supported in any browser
- migrated many templates from legacy components to html elements
- put tailwind paste behind flag


kof (approved):
🎉

kof:
image
why do we have these attributes on a div?

PR #5240, opened 2025-05-30
perf: improve instance selection performance

TrySound:
We have html element mutations in a few places and each can add ~250ms of repaint time on large projects. Here I got rid of it in two places

1. Collapsed logic enforced document height to preserve scroll position. Instead I created an absolutely positioned div which enforced scroll where collapsed elements are reset.
2. Scroll into view logic defined position absolute on scroll container. Though it's not necessary on html element so just ignored it.

### DIFF
--- a/apps/builder/app/canvas/collapsed.ts
+++ b/apps/builder/app/canvas/collapsed.ts
@@ -10,6 +10,7 @@ import {
 import { $selectedBreakpoint } from "~/shared/nano-states";
 import { $selectedPage } from "~/shared/awareness";
 import { serverSyncStore } from "~/shared/sync";
+import { doNotTrackMutation } from "~/shared/dom-utils";
 
 const isHtmlTag = (tag: string): tag is HtmlTags =>
   htmlTags.includes(tag as HtmlTags);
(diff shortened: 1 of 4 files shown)


apps/builder/app/canvas/collapsed.ts
```diff
@@ -246,14 +231,25 @@ const recalculate = () => {
...
-
-  // Mark that we are in the process of recalculating collapsed elements
-  document.documentElement.style.setProperty(
-    markCollapsedMutationProperty,
-    `true`
-  );
-  document.documentElement.style.height = `${document.documentElement.scrollHeight}px`;
+  // use nested element to avoid full page repaint and freeze on big projects
+  let collapsedElement = document.querySelector(
+    "#ws-collapsed"
```
kof, line 236: could probably just memoize the reference somewhere instead of dom query

kof (approved):

apps/builder/app/canvas/collapsed.ts
TrySound: HMR will loose it most of the time

PR #5242, opened 2025-06-03
feat: add replace with element command

TrySound:

This can be a nice addition and a way to migrate legacy components.

https://github.com/user-attachments/assets/8354d4bd-e60b-4cf3-b662-5c8e974d9537


kof (approved):
Niceeee! What happens if one changes from a link to a div with href attribute?

kof:
Oh I see it keeps the attribute. User needs to remove bad attributes manually.

TrySound:
In html all attributes are good 😇

kof:
> In html all attributes are good 😇

I mean invalid, can't put href on a div for example

PR #5260, opened 2025-06-09
fix: prevent typing only inside of textarea and input on canvas

TrySound:
Accidentally broke when fixed links download attribute


kof (approved):

TrySound:
Already hotfixed

PR #5270, opened 2025-06-11
fix: Rename Image File in fixtures/webstudio-features/public/assets

shootyourbricks:
## Description

Renaming file in fixtures/webstudio-features/public/assets as the previous name was too long and caused github cloning to fail. https://github.com/webstudio-is/webstudio/issues/5267

## Steps for reproduction

1. Use github desktop to clone westudio-is/webstudio repository.
2. Cloning will fail due to too long of a file name.

## Code Review

hi @kof , I need you to let me know if there are any file dependencies for this file. I didn't notice any, so I only changed the name of the file.


luke-monroe:
@TrySound Any chance you can look at this? It's pretty simple.

PR #5289, opened 2025-06-22
feat: improve automatic domain configuration

TrySound:
- added official entrijs with types and mimimal runtime (no dependencies)
- automatically add `www.` prefix to non-cloudflare root domains
- enabled automatic www redirect to setup with entri (though does not work seamlesly, namecheap requires ssl certificate, godaddy requires manual setup by user)
- toast error when click on "Configure automatically" on free plan


kof:
> enabled automatic www redirect to setup with entri (though does not work seamlesly, namecheap requires ssl certificate, godaddy requires manual setup by user)

What do we do with this? how will user know that their root domain is not redirected to www?

TrySound:
They will check and start complaining. We cannot really do much about it. Entri does as much as possible. The rest is up to user.

kof:
> They will check and start complaining. We cannot really do much about it. Entri does as much as possible. The rest is up to user.

If we know it didn't work, can we tell the user?

TrySound:
Do we know?

kof (approved):

PR #5292, opened 2025-06-26
fix: Create .gitattributes to fix error due to line endings on Windows

shootyourbricks:
Added Git Attributes file to ensure that line endings are the same regardless of the OS (This was causing problems for Windows + Docker)

## Steps for reproduction (Windows)

1. Clone webstudion builder and open with vs code, docker, etc.
2. every single file will have a git status of "m" (modified) due to changed line endings

## Code Review

 hi @kof, I need you to take a look and make sure this is an ideal fix

It seems to work great on my end.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,43 @@
+# Enforce LF endings across all text files
+* text=auto eol=lf
+
+# --- TypeScript & JavaScript ---
+*.ts     text eol=lf
+*.tsx    text eol=lf
+*.js     text eol=lf
+*.jsx    text eol=lf
+
+# --- CSS ---
+*.css    text eol=lf
+
+# --- Shell Scripts ---
+*.sh     text eol=lf
+
+# --- SQL (PL/pgSQL scripts) ---
+*.sql    text eol=lf
+
+# --- Docker ---
+Dockerfile* text eol=lf
+*.dockerignore text eol=lf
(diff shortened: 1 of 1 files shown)


TrySound:
Seems like other popular repos have lf configured in .editorconfig. Can you please try if it works?

https://github.com/sveltejs/svelte/blob/main/.editorconfig#L1-L4

istarkov:
https://github.com/search?q=repo%3Afacebook%2Freact%20gitattributes&type=code

Most libs does not have this at all. Also we can't somehow test this as no one of devs have windows.

istarkov:
IMO should be short like React if needed

luke-monroe:
@TrySound @istarkov Looks like it does work with the editor.config instead of git attributes

editor.config
```diff
@@ -0,0 +1,15 @@
...
+indent_style = tab
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+
+[test/**/expected.css]
+insert_final_newline = false
+
+[package.json]
+indent_style = space
```
TrySound, line 15: everything except for end_of_line can be removed

luke-monroe:
@TrySound ok I changed what you said. looks like it still works.

TrySound:
Thank you!

PR #5322, opened 2025-07-12
feat: automatically paste tailwind when html has inception mark

TrySound:
We want to reduce friction in pasting from inception and will automatically process tailwind classes and generate styles when paste into builder without using command panel.

### DIFF
--- a/apps/builder/app/shared/copy-paste/plugin-html.ts
+++ b/apps/builder/app/shared/copy-paste/plugin-html.ts
@@ -1,7 +1,10 @@
 import { generateFragmentFromHtml } from "../html";
 import { insertWebstudioFragmentAt } from "../instance-utils";
+import { generateFragmentFromTailwind } from "../tailwind/tailwind";
 import { denormalizeSrcProps } from "./asset-upload";
 import type { Plugin } from "./init-copy-paste";
+
+const inceptionMark = `<!-- Generated with Inception by Webstudio -->`;
 
 export const html: Plugin = {
   name: "html",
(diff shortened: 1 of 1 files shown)


apps/builder/app/shared/copy-paste/plugin-html.ts
```diff
@@ -1,14 +1,20 @@
 import { generateFragmentFromHtml } from "../html";
 import { insertWebstudioFragmentAt } from "../instance-utils";
+import { generateFragmentFromTailwind } from "../tailwind/tailwind";
 import { denormalizeSrcProps } from "./asset-upload";
 import type { Plugin } from "./init-copy-paste";
 
+const inceptionMark = `<!-- Generated with Inception by Webstudio -->`;
```
kof, line 7: I would use something like our version identifier for paste, might be helpful later on to have some version in it: aka `@webstudio/instance/v0.1` but `@webstudio/tailwind/v0.1`

kof: This way the intent of the comment is also clear and it has some structured data in it instead of a natural language string

kof (approved):

apps/builder/app/shared/copy-paste/plugin-html.ts
```diff
@@ -1,14 +1,20 @@
 import { generateFragmentFromHtml } from "../html";
 import { insertWebstudioFragmentAt } from "../instance-utils";
+import { generateFragmentFromTailwind } from "../tailwind/tailwind";
 import { denormalizeSrcProps } from "./asset-upload";
 import type { Plugin } from "./init-copy-paste";
 
+const inceptionMark = `<!-- @webstudio/incation/1 -->`;
```
kof, line 7: typo!!!!

PR #5358, opened 2025-08-04
fix: access available features from product plans

TrySound:
This will let us to configure available features for each plan.


kof (approved):
Looking great, didn't test

TrySound:
We'll test in prod 😄

PR #5365, opened 2025-08-09
feat: show when asset is used

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/572 https://github.com/webstudio-is/webstudio/issues/3261

- made settings gear visible when asset is used anywhere
- when asset is used delete button shows dialog with all usages

image

image

Screenshot 2025-08-09 at 11 30 37


bryanbemis:
This feature looks great. _I suggest a change to how the user triggers showing all uses before deleting._

As a user, I want to review a list of asset uses before deciding whether to delete—2 decisions (review, then cancel or confirm deletion). 

Right now you must click a "Delete" button (Asset Details dialog) to see uses, then click another "Delete" button (Delete Asset dialog) to confirm. This could be confusing because when the user wants to see the list of uses, there's no indication that they can do that by clicking the first "Delete" button.

Suggested change: **Change the wording on the Asset Details dialog from "Delete" to "Review uses."** This matches the user's **intent** at that point. 

uses

The Delete Asset dialog can remain the same because the user can cancel (by clicking the dialog close button) if they don't want to delete, or they can confirm deletion by clicking the existing Delete button.

kof:
There is partial truth to what you suggested, though you introduced another issue: Review uses doesn't indicate that you will be able to delete the asset. In fact now deleting asset is completely gone and won't be found.

kof:
I see 2 potential solutions to this:

1. Call the button "Review & delete" though here we are still mixing 2 intents
2. Make a separate button to review where its been used, e.g. the 1 uses could be a link or button that opens dialog to show where its used. Then keep the delete button as-is and show the review again, this is great in any case to see before deletion.

kof:
I think a simpler solution here is "Review & delete" button in default color, without trash icon.

kof:
If the image has 0 usages, we can call this straight away "Delete" as it is right now and not show review, just straight deleting it with one click.

PR #5369, opened 2025-08-12
feat: allow user to add asset description

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3263

Now users can control "alt" for all image uses with asset description.

https://github.com/user-attachments/assets/c0e5ae4b-d2b0-4ec1-bf93-c385816369d1

### DIFF
--- a/apps/builder/app/shared/nano-states/props.ts
+++ b/apps/builder/app/shared/nano-states/props.ts
@@ -259,20 +259,6 @@ export const $propValuesByInstanceSelector = computed(
 
       if (props) {
         for (const prop of props) {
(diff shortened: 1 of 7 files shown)


kof:
- [x] use variant=wrapped image

apps/builder/app/shared/nano-states/props.ts
```diff
@@ -259,20 +259,6 @@ export const $propValuesByInstanceSelector = computed(
 
       if (props) {
         for (const prop of props) {
-          if (prop.type === "asset" && prop.name === "width") {
```
kof, line 262: huh, why did we have it before?

PR #5381, opened 2025-08-24
feat: cache resources

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/4077

How it works

- Cache-Control opts in caching (Cache-Control: max-age=3600)
- cached response is matched by url, method and body
- cache is invalidated with Cache-Control change
- responses stored with the cache are not replicated across regions
- responses are not shared between projects

Link for testing
https://test-cache-hcevl.wstd.work/?message=hello

### DIFF
--- a/packages/sdk/src/resource-loader.ts
+++ b/packages/sdk/src/resource-loader.ts
@@ -108,9 +108,9 @@ const getCacheKey = async (request: Request) => {
 };
 
 export const cachedFetch = async (
+  namespace: string,
   input: RequestInfo | URL,
-  init: RequestInit,
-  namespace: string
+  init?: RequestInit
 ) => {
   if (globalThis.caches) {
     const request = new Request(input, init);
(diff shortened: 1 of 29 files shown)


packages/sdk/src/resource-loader.ts
```diff
@@ -89,3 +90,56 @@ export const loadResources = async (
...
+  const method = request.method;
+  const body = await request.clone().text();
+  // invalidate cache when cache-control is changed
+  const cacheControl = request.headers.get("Cache-Control");
+  const resourceHash = hash(`${method}:${body}:${cacheControl}`);
+  url.searchParams.set("ws-resource-hash", resourceHash);
+  return url;
+};
+
+export const cachedFetch = async (
```
kof, line 110: does this need a few unit tests? because if we break this we won't easily notice

kof (approved):
think about a polyfill for node and co

PR #5383, opened 2025-08-27
feat: new resources dialog

TrySound:

- moved response preview into resource dialog
- increased dialog size and centered it
- added maximize button to resource dialog
- removed inspect option from dropdown
- refactored centering dialogs without transform to fix nested popovers positioning

https://github.com/user-attachments/assets/901e086b-afcd-42e9-8866-d62d6cb662d2


kof:
Hmm, I would have expected the maximize button to be on the dialoog itself, so that one can keep looking/changing left side and still have a better view of the data

PR #5384, opened 2025-08-31
feat: add search params list to resource UI

TrySound:

Editing search params separately from url can improve experience working with external CMS or databases like baserow.

Also made headers list more compact and aligned with search params.
Validation is simplified, empty pairs will be just ignored without blocking user with validation errors.

https://github.com/user-attachments/assets/3102d2c8-a902-4e86-8011-d71346fb5491

### DIFF
--- a/apps/builder/app/builder/features/settings-panel/curl.test.ts
+++ b/apps/builder/app/builder/features/settings-panel/curl.test.ts
@@ -4,6 +4,7 @@ import { generateCurl, parseCurl, type CurlRequest } from "./curl";
 test("support url", () => {
   const result = {
     url: "https://my-url/hello-world",
+    searchParams: [],
     method: "get",
     headers: [],
   };
@@ -19,6 +20,7 @@ test("support multiline command with backslashes", () => {
   `)
   ).toEqual({
     url: "https://my-url/hello-world",
+    searchParams: [],
     method: "get",
     headers: [],
   });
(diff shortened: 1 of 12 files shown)


kof:
- [ ]  Something happened to new variable panel position

image

kof:
- [x] empty state feels like something is missing, probably needs to say "No search params" and "No headers" to make visually clear that these are sections

image

kof:
Ideally we would show the search params in the URL as you add them via inputs and when URL changes inputs get added/removed automatically

image

kof:
I can imagine having autocompletion for standard header names would be very helpful

kof:
- [x] when adding a search param via + button, I would autofocus the first input

image

TrySound:
>  Something happened to new variable panel position

Same as resource variable

> Ideally we would show the search params in the URL as you add them via inputs and when URL changes inputs get added/removed automatically

Can be deceiving for user. URL is also an actual input. Dunno why put there.

kof:
> Same as resource variable

Not sure what you mean, its now in the middle but it should be next to the settings panel

TrySound:
Resource panel is variable panel and it is centered now for better preview. Works with system and big json data too.

image
image

kof:
What's a little off is that the panel is so small and sits in the middle of the screen, we could potentially solve this by showing the data preview always, so its always not small

kof:
Another thing I noticed is after switching from a Resource with data in it to a string, you still see data in the preview and only once I refresh I see the value preview

image
image

PR #5387, opened 2025-09-01
feat: support json in resource search params and headers

TrySound:

This should improve experience when working with baserow filters or similar backends.

Screenshot 2025-09-01 at 12 19 35


TrySound:
tested with webhook.site
image

PR #5400, opened 2025-09-18
fix: reduce patch size to let users duplicate huge pages

TrySound:
User is trying to duplicate huge page and gets "413 Content Too Large". Here I'm trying to reduce patch request size by removing revise patches which are not used on server.


TrySound:
This reduced payload size

```diff
- 5477222
+ 4089733
```

PR #5403, opened 2025-09-23
feat: enabling safe functions in expression engine

zehjotkah:
## Description

allowing toLowerCase(), replace(), and split() to be used in the expression editor. 

Real life use-case explained in the discord message. 
https://discord.com/channels/955905230107738152/1418928616648998992/1419623805541814393

## Steps for reproduction

1. use any of the mentioned functions in the expression editor
2. they're currently not allowed even though they'd be safe to use 
3. those functions (and presumably others as well) can be used in hosted webstudio by deploying my changes on a locally hosted webstudio editor, using the functions, and copying the pages containing the functions to the hosted version. 
-> validation check bypass 

## Code Review

  - conceptual review (does webstudio WANT to allow these functions?)
  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/packages/sdk/src/expression.test.ts
+++ b/packages/sdk/src/expression.test.ts
@@ -418,19 +418,19 @@ describe("transpile expression", () => {
         expression: "title.toLowerCase()",
         executable: true,
       })
-    ).toEqual("title?.toLowerCase()");
+    ).toEqual("title?.toLowerCase?.()");
     expect(
       transpileExpression({
         expression: "user.name.replace(' ', '-')",
         executable: true,
       })
-    ).toEqual("user?.name?.replace(' ', '-')");
+    ).toEqual("user?.name?.replace?.(' ', '-')");
     expect(
       transpileExpression({
         expression: "data.title.split('-')",
         executable: true,
       })
-    ).toEqual("data?.title?.split('-')");
+    ).toEqual("data?.title?.split?.('-')");
   });
 
   test("transpile chained string methods with optional chaining", () => {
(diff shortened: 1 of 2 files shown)


kof:
My only wish is that we would support the entire js there and not a few functions, because:
1. we can't tell this is just javascript without it
2. AI won't be able to help unless it knows the whitelisted functions
3. people will always ask why X works and Y doesn't

zehjotkah:
I'd love webstudio to support the entire js in the expression editor! 
With this PR I wanted to make the minimum changes necessary for my usecase.

zehjotkah:
> My only wish is that we would support the entire js there and not a few functions, because:
> 
>     1. we can't tell this is just javascript without it
> 
>     2. AI won't be able to help unless it knows the whitelisted functions
> 
>     3. people will always ask why X works and Y doesn't

That's a fair point. A random list of functions is confusing, and you're right, people will always wonder why one thing works and another doesn't.

The problem with allowing everything is the security risk, right? Stuff like `fetch` or accessing `document` could open up XSS holes, and things like `while(true){}` could crash the builder, if I understand it correctly.

So what if we set a clear rule for what's allowed? We could support all the standard methods on primitives (Strings, Arrays, etc.) as long as they don't mutate the original data.

That way, it feels like you're just writing JavaScript for data manipulation. Methods like `.map()` or `.slice()` would be fine, but something like `.pop()` or `.splice()` wouldn't, because they change the original array and could cause weird side effects.

My PR for `split()`, `replace()`, and `toLowerCase()` fits this rule. It solves a real problem for building dynamic pages. Right now, those functions do work if you paste them in from a local webstudio fork where the validation check is turned off, so the linter is already out of sync with the runtime. 

What do you think? I'm happy to expand the PR to include more of these "safe" methods if you agree on this as the general rule.

packages/sdk/src/expression.test.ts
```diff
@@ -222,24 +222,38 @@ describe("lint expression", () => {
...
+      "split",
+      "at",
+      "endsWith",
+      "includes",
+      "startsWith",
+      "toUpperCase",
+      "toLocaleLowerCase",
+      "toLocaleUpperCase",
+    ];
+    for (const method of stringMethods) {
```
TrySound, line 237: try test.each instead so each method would have separate case in test report.
https://vitest.dev/api/#test-each

packages/sdk/src/expression.test.ts
```diff
@@ -220,6 +220,71 @@ describe("lint expression", () => {
...
+    ).toEqual([]);
+  });
+
+  test("forbid unsafe method calls", () => {
+    expect(
+      lintExpression({
+        expression: `arr.pop()`,
+        availableVariables: new Set(["arr"]),
+      })
+    ).toEqual([error(0, 9, "Functions are not supported")]);
```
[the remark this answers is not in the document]
zehjotkah: Thanks! Made the error messages more clear.

packages/sdk/src/expression.test.ts
```diff
@@ -346,6 +411,42 @@ describe("transpile expression", () => {
...
     expect(errorString).toEqual(`Unexpected token (1:0) in ""`);
   });
+
+  test("transpile string methods with optional chaining", () => {
+    expect(
+      transpileExpression({
+        expression: "title.toLowerCase()",
+        executable: true,
+      })
+    ).toEqual("title?.toLowerCase()");
```
TrySound, line 421: It should add optional chaining to function calls as well so calling method on some object would not give runtime errors. `title?.toLowerCase?.()`. We do the same in collection `array?.map?.(item`

zehjotkah:
do you want me to change anything else? 
Thanks!

TrySound:
All good. We were just a occupied with inception recently. Though you can try to add autocomplete for these methods if you want.

zehjotkah:
> you can try to add autocomplete for these methods if you want.

like this? 
Bildschirmfoto 2025-10-09 um 16 45 38

TrySound:
Yes, maybe even with parentheses for clarity like this `toLowerCase()`

TrySound:
Btw I will merge this. You can start new PR

TrySound:
Thanks!

kof:
Good job @zehjotkah !

PR #5419, opened 2025-10-05
fix: upload unoptimized inception images

TrySound:
This fixes the issue with uploading avif which is not supported as image source.


kof (approved):
so with this we are not downloading the image from optimizer, but instead using a raw one, right?

TrySound:
Yep

PR #5421, opened 2025-10-09
feat: expression string method autocomplete

zehjotkah:
## Description

Added autocomplete support for safe string and array methods in the expression editor, as suggested by @TrySound  in PR #5403.

## Changes

- Exported `allowedStringMethods` and `allowedArrayMethods` from `packages/sdk/src/expression.ts`
- Enhanced the expression editor's autocomplete to suggest safe methods when typing after a variable (e.g., system)
- Methods are displayed with parentheses for clarity (e.g., `toLowerCase()`, `replace()`, `split()`)
- Autocomplete shows helpful labels: "string method" or "array method"

## How to Test

1. Open any project in the builder
2. Click the expression editor icon next to a text property
3. Type a variable name followed by a dot (e.g., `system.`)
4. You should see autocomplete suggestions for safe string methods like `toLowerCase()`, `replace()`, `split()`, etc.
5. Type a few letters to filter (e.g., `title.to` shows only methods starting with "to")
6. Select a method and it will be inserted with parentheses

## Related

- Builds on PR #5403 which enabled these safe methods in the expression engine

Bildschirmfoto 2025-10-09 um 22 33 03

### DIFF
--- a/apps/builder/app/builder/shared/expression-editor.tsx
+++ b/apps/builder/app/builder/shared/expression-editor.tsx
@@ -258,11 +258,8 @@ const scopeCompletionSource: CompletionSource = (context) => {
     const isString = typeof target === "string";
     const isArray = Array.isArray(target);
 
-    // Add string methods if target is a string or we can't determine the type
-    if (
-      isString ||
-      (!isArray && typeof target !== "number" && typeof target !== "boolean")
-    ) {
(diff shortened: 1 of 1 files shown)


TrySound:
Great!
image
image

TrySound (approved):

Issue #5378, opened 2025-08-22
CSS variable second parameter gets cut out


kof:
When entering second parameter in advanced section of a css variable, it gets visually removed in the value input, but is actually correctly saved: display: var(--anchor-display, none) 

<img width="486" height="116" alt="Image" src="https://github.com/user-attachments/assets/09136208-596e-43dc-9ac6-53c1ec72e7c7" />

<img width="462" height="538" alt="Image" src="https://github.com/user-attachments/assets/c3be953e-9f57-4bdd-9771-738a8eedeeba" />

michaelokolo:
Please can I work on this.

kof:
@michaelokolo  Sure

michaelokolo:
@kof and @TrySound, please find the PR for your review. I completed it three weeks ago but forgot to leave a comment here earlier.

niatalwalkar:
Hello! I noticed the earlier PR for this issue hasn’t had activity for a bit, so I attempted a fix and submitted a new PR. Please let me know if you'd like anything changed or combined!

PR #5422, opened 2025-10-10
fix: CSS variable second parameter gets cut out (#5378)

michaelokolo:
## Description

This pull request fixes a bug where the second parameter (fallback value) in the Advanced section of CSS variables is visually removed from the value input field after saving, even though it is correctly stored internally.

Related issue: [#5378](https://github.com/webstudio-is/webstudio/issues/5378)

## Changes

1. Updated the ` itemToString` function in `css-value-input.tsx`  to correctly handle cases where a fallback value exists.
2. Implemented a ternary check that:
    - Returns the full variable string when a fallback is present.
    - Returns only the main variable when no fallback exists.

This ensures that the fallback value remains visible in the UI after saving.

## How to Test
1. Open any project.
2. Navigate to the Advanced section under Style.
3. Click the plus (+) icon to add a new style.
4. In the Add styles input, type or select display.
5. Set the value to: `var(--anchor-display, none)`.
6. The value remains displayed as `var(--anchor-display, none)` after saving (previously, the fallback , `none` would disappear).

open source pix


kof:
Looks like we fixed it already and forgot to close this pr, please reopen if its not true!

PR #5465, opened 2025-11-28
feat: support modern color spaces in color inputs

TrySound:
Ref https://github.com/webstudio-is/webstudio/issues/3574

Here improved css value parser to parse and store any color space. oklch(), hsl(), color() and others can be used in any color input. Color picker converts to rgb on the fly and do not support any other color space at the moment.

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-utils.test.ts
+++ b/apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-utils.test.ts
@@ -542,17 +542,15 @@ describe("ensureGradientHasStops", () => {
     const result = ensureGradientHasStops(gradient);
     expect(result.stops).toHaveLength(2);
     expect(result.stops[0]?.color).toEqual({
-      type: "rgb",
-      r: 0,
-      g: 0,
-      b: 0,
+      type: "color",
+      colorSpace: "srgb",
+      components: [0, 0, 0],
       alpha: 1,
     });
     expect(result.stops[1]?.color).toEqual({
-      type: "rgb",
-      r: 0,
-      g: 0,
-      b: 0,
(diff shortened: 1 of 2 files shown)


apps/builder/app/builder/features/style-panel/shared/css-value-input/css-value-input.tsx
```diff
@@ -932,6 +932,7 @@ export const CssValueInput = ({
                             {toValue(item.fallback)}
                           </Text>
                         )}
+                        {/* @todo */}
```
kof: better write `@todo colors` or something to be able to find it

TrySound: Left it so you notice it and remind me to fix it 😅

packages/css-data/src/parse-css-value.ts
```diff
@@ -71,10 +72,6 @@ export const isValidDeclaration = (
     return keywordValues[property].includes(value);
   }
 
-  if (property === "transition-behavior") {
```
kof: what are we replacing it with?

packages/css-data/src/parse-css-value.ts
```diff
@@ -299,7 +314,13 @@ const parseLiteral = (
...
       node.name === "hsla" ||
       node.name === "rgb" ||
-      node.name === "rgba"
+      node.name === "rgba" ||
+      node.name === "oklch" ||
+      node.name === "oklab" ||
+      node.name === "lch" ||
+      node.name === "lab" ||
+      node.name === "hwb" ||
+      node.name === "color"
```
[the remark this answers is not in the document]
TrySound: This is the single place

PR #5513, opened 2025-12-14
fix: Pass in params into action

G-OD:
## Description

This fixes an issue where you can't pass system params into a webhook form action.

## Steps for reproduction

1. Create a webhook form
2. Define an action with system.params in it
3. Get undefined

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests


kof:
Looking into this, can you please describe a practical use case? for me to better understand what this unlocks

G-OD:
> Looking into this, can you please describe a practical use case? for me to better understand what this unlocks

One use case maybe having to post to another endpoint depending on what URL path you're on. Lets say you have a comment section for a blog post and you have the endpoint /posts/123, and to post a comment you need to post to /posts/123/comment. This would require you to dynamically create the action URL using the data variables (in this case the system params).

G-OD:
Hello there. Is it possible to get this reviewed? If not it's probably worth removing the variables from the action URLs as the UI suggests this is possible.

PR #5568, opened 2026-01-16
feat: When asset is deleted or uploaded - refresh asset resource

kof:
## Description

When user is using asset resource and binding it on canvas, for example video file bound to video tag to the src attribute and then deletes the asset, it is expected that canvas reflects the deleted asset as well as the variable.

## Steps for reproduction

1. click button
2. expect xyz

## Code Review

  - test it on preview

## Before requesting a review

- [ ] made a self-review

## Before merging

- [ ] added tests

### DIFF
--- a/apps/builder/app/shared/resources.ts
+++ b/apps/builder/app/shared/resources.ts
@@ -1,30 +1,13 @@
 import { atom, computed } from "nanostores";
-import hash from "@emotion/hash";
 import type { DataSource, Resource, ResourceRequest } from "@webstudio-is/sdk";
 import { restResourcesLoader } from "./router-utils";
 import { computeExpression } from "./data-variables";
 import { fetch } from "./fetch.client";
+import { getResourceKey } from "./resource-utils";
 
 const MAX_PENDING_RESOURCES = 5;
 
-export const getResourceKey = (resource: ResourceRequest) => {
-  try {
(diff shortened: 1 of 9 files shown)


apps/builder/app/shared/resources.ts
```diff
@@ -1,30 +1,13 @@
...
-import hash from "@emotion/hash";
 import type { DataSource, Resource, ResourceRequest } from "@webstudio-is/sdk";
 import { restResourcesLoader } from "./router-utils";
 import { computeExpression } from "./data-variables";
 import { fetch } from "./fetch.client";
+import { getResourceKey } from "./resource-utils";
 
 const MAX_PENDING_RESOURCES = 5;
 
-export const getResourceKey = (resource: ResourceRequest) => {
```
TrySound: Doesn't look like used anywhere else. Not necessary to split into separate module.

TrySound (approved):

apps/builder/app/shared/resources.ts
kof: indeed

kof: turns out it is used in multiple places

TrySound: Yeah, in file you removed it from and in test of that file

PR #5569, opened 2026-01-16
feat: Allow Collection component to accept both - arrays and objects

kof:
## Description

Now user can pass to collection an array like this (this worked before): `[{a: 1}]` and also object like this `{key: {a: 1}}`.

What this means is that now you can just iterate over objects returned from some api, you don't need arrays-only any more.

As an item, the value of each object key will be provied.


kof:
duplicate of https://github.com/webstudio-is/webstudio/pull/5570

PR #5598, opened 2026-02-02
fix: Breakpoint editor showing empty values after reopening

zehjotkah:
fixes https://github.com/webstudio-is/webstudio/issues/5597

When adding a new breakpoint, editing its name and settings, then closing and reopening the menu, the breakpoint appeared with empty values instead of the saved data.

Root cause: The allBreakpoints array prioritized addedBreakpoints local state over store data. When a breakpoint was added, it remained in addedBreakpoints with empty values even after being saved to the store with updated values.

Fix: Changed priority to show store breakpoints first, only showing addedBreakpoints entries that haven't been saved yet. Also added cleanup to clear addedBreakpoints when the menu closes.

### DIFF
--- a/apps/builder/app/builder/features/breakpoints/breakpoints-editor.tsx
+++ b/apps/builder/app/builder/features/breakpoints/breakpoints-editor.tsx
@@ -177,6 +177,96 @@ const BreakpointEditorItem = ({
   );
 };
 
+type BreakpointsEditorContentProps = {
+  onDelete: (breakpoint: Breakpoint) => void;
+};
+
+/**
+ * Content component that holds local state for newly added breakpoints.
+ * This component unmounts when the popover closes, automatically cleaning up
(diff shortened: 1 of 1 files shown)


kof:
Do we really need to change the order or just clean the local state when dialog is closed? Essentially the root cause is that the local state should be in a component that unmounts upon closing, so local state is cleaned up automatically, but in our case its one component that renders both, the button and the actual dialog content, so it never unmounts. I would just split this into separate components.

kof:
Added breakpoints appearing at the beginning is a UX choice, not a bug, not sure why this needed to be changed.

kof:
Every change of the  condition in the editor makes the list to sort it differently, in other words the breakpoint inside breakpoints editor is jumping around, do you see it too?

kof:
TBC in https://github.com/webstudio-is/webstudio/pull/5610

PR #5602, opened 2026-02-03
feat: Add day and month name tokens to Time element format

zehjotkah:
Fixes https://discord.com/channels/955905230107738152/955905231227609158/1467592164727849104

- Add DDDD/DDD tokens for full/short day names (e.g., Monday/Mon)
- Add MMMM/MMM tokens for full/short month names (e.g., January/Jan)
- Day and month names are localized based on selected language
- Simplify format field tooltip for better UX
- Add comprehensive tests for new tokens in multiple languages
- Maintain backward compatibility with existing format tokens


kof:
Is there anything breaking for existing users if they used previous format? If  not I think this can be merged as-is

zehjotkah:
> Is there anything breaking for existing users if they used previous format? If not I think this can be merged as-is

we just add formats which were not available before. All previous available settings continue to work exactly as before.

PR #5663, opened 2026-03-26
fix: preserve shadow labels when hiding repeated items

shreyanshkotak:
## Description

1. Fix style panel label rendering for repeated `box-shadow` and `text-shadow` items after hiding one layer.

When a shadow layer is hidden, it is omitted from serialized computed CSS. The repeated-style helper was reparsing that computed value and bailing out when the computed item count no longer matched the cascaded item count, which caused all shadow row labels to lose their details even though the CSS and edit dialogs remained correct.

This PR:
- preserves repeated-item index mapping when hidden layers compact the computed value
- returns hidden parsed items for hidden rows so labels still have data
- makes shadow labels prefer the row's parsed shadow value before falling back to computed data
- adds regression tests for hidden box-shadow and text-shadow layers

https://github.com/user-attachments/assets/bd25e302-fac1-4cea-8a4b-317cc5b448f7

## Steps for reproduction

1. Select an instance in the builder.
2. Add at least 3 `box-shadow` layers in the style panel.
3. Hide the first layer.
4. Observe that, before this fix, the labels collapse to partial values like `Outer:`.
6. Repeat the same flow with at least 3 `text-shadow` layers.
7. With this PR, hidden and visible shadow rows keep their full labels.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-utils.test.ts
+++ b/apps/builder/app/builder/features/style-panel/sections/backgrounds/gradient-utils.test.ts
@@ -638,7 +638,7 @@ describe("styleValueToColor", () => {
     };
     expect(styleValueToColor(style)).toEqual({
       type: "color",
-      colorSpace: "srgb",
+      colorSpace: "hex",
       components: [0, 0, 1],
       alpha: 1,
     });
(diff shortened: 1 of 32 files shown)


kof:
Merged, thank you!

PR #5674, opened 2026-04-02
Windows 10 LTSC Compatibility Mod

GhervasaCristian:
Commands to install and start the project. Tested under Windows 10 LTSC IoT.
-----
pnpm install
-----
docker compose -f apps/builder/docker-compose.yaml down -v

docker compose -f apps/builder/docker-compose.yaml up -d -----
pnpm migrations
-----
pnpm add -D shx -w 
-----
pnpm build
pnpm dev

### DIFF
--- a/apps/builder/docker-compose.yaml
+++ b/apps/builder/docker-compose.yaml
@@ -5,15 +5,15 @@ volumes:
 
 services:
   db:
-    image: postgres
+    image: postgres:17
     # Uncomment to log all queries
     command: ["postgres", "-c", "log_statement=all"]
     restart: always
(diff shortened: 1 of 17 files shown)


kof:
Hi, unfortunately none of our team members are using windows to develop webstudio, so we don't officially support windows. To make sure it consistently works someone would need to always test it on windows and we don't have this capacity. Even if we properly fix the things you changed in this PR - its not enough. I am going to close it for now. You can maintain your own config/fork for windows if you like.

Further: WSL exists so that POSIX commands can just work. I don't know if you tried.

Issue #3268, opened 2024-04-30
Replace an image so that it gets replaced automatically everywhere its being used


kof:
As a user when I use the same image in many places, I want to be able to replace it once and have it reflected in all places where its being used.

shreyanshkotak:
Hello @kof opened a PR related to this feature #5676

PR #5676, opened 2026-04-03
feat: replace asset option for image assets in asset details

shreyanshkotak:
## Description

Adds the ability to replace an existing image asset with a new file directly from the asset details panel, without having to delete and re-upload manually. Related to [place an image so that it gets replaced automatically everywhere its being used #3268](https://github.com/webstudio-is/webstudio/issues/3268)

When clicking the new replace button on an image asset:
1. A file picker opens filtered to image MIME types
2. The new file is uploaded via the existing upload pipeline
3. All references to the old asset (props, styles, page meta: favicon, social image, marketplace thumbnail) are atomically re-pointed to the new asset
4. The old asset's filename and description are copied to the new asset
5. The old asset is deleted

A new `ReverseIcon` from the Figma library was added to the icons package to represent the replace action.

## Steps for reproduction

1. Open the asset manager and select any image asset
2. In the asset details panel, click the new replace icon button next to the download button
3. Select a new image file from the file picker
4. Expect the asset to be replaced in place - all usages (image props, background styles, favicon, social image) should now reference the new file with the original asset's filename preserved

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  const newAsset = await waitForAsset(newAssetId);
+
+  if (!newAsset) {
+    toast.error("Replacement asset upload timed out");
+    return;
+  }
+
+  serverSyncStore.createTransaction(
```
kof, line 39: A different approach could be actually replacing the stored asset and keeping everything else untouched. Questions around this is how to refresh the rendering, in the builder as well as in prod, invalidate the caches etc.

kof: it would still have a different size potentially, file name might also be different, even description might be different

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+    [$pages, $props, $styles, $assets],
+    (pages, props, styles, assets) => {
+      const updatedNewAsset = assets.get(newAssetId);
+      if (updatedNewAsset) {
+        updatedNewAsset.filename = oldAsset.filename;
+        updatedNewAsset.description = oldAsset.description;
```
kof: this is only correct if the new image is mostly the same but changed, if its a completely different image, both of these things would be different

apps/builder/app/builder/shared/assets/replace-asset.ts
kof: so no, I think you are right, we need to replace the usages as you did

kof: In fact since we never made full url as perm link available, we can do that and in cases where the image is used via scripts, user will get the new one via asset ressource

apps/builder/app/builder/shared/assets/replace-asset.ts
kof: I think this is eventually up to the user to decide, if they want to keep the description or not, file name - would take the new one, its already user's decisionn to name a file certain way, but description - if we just throw it away, it wouldn't be expected in cases where user wants to keep it, so probably its better to keep the old description and let user handle it, in the future description can be also automated.

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+      }
+
+      for (const prop of props.values()) {
+        if (
+          prop.type === "asset" &&
+          prop.value === oldAssetId &&
+          prop.name !== "width" &&
+          prop.name !== "height"
```
kof: this reads strange, if its type asset, can prop.name be  width/height? just picking in case its bs AI generated check

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+        ) {
+          prop.value = newAssetId;
+        }
+      }
+
+      for (const [, styleDecl] of styles) {
```
kof: could be something like this, didn't check the syntax

```suggestion
      for (const styleDecl of styles.values()) {
```

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+  );
+
+  onNextTransactionComplete(() => {
+    invalidateAssets();
+  });
+
+  toast.success("Asset replaced successfully");
+};
+
+import type { StyleValue } from "@webstudio-is/css-engine";
```
kof: imports should be always on top of the file

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+      replaceAssetInStyleValue(item, oldAssetId, newAssetId);
+    }
+  }
+  if (styleValue.type === "layers") {
+    for (const item of styleValue.value) {
+      replaceAssetInStyleValue(item, oldAssetId, newAssetId);
+    }
+  }
+};
+
```
kof, line 105: very important function, must be double checked and covered by tests

kof: ah it already is

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+const waitForAsset = (
+  assetId: string,
+  timeoutMs = 60_000,
```
kof: what if network is slow and it takes longer? seems arbitrary so far we din't need this logic, why start now?

apps/builder/app/builder/shared/assets/replace-asset.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  const fileToAssetId = await uploadAssets("image", [file]);
+  const newAssetId = fileToAssetId.get(file);
+
+  if (!newAssetId) {
+    toast.error("Failed to upload replacement asset");
+    return;
+  }
+
+  const newAsset = await waitForAsset(newAssetId);
```
kof: the lines above are already checking that upload was successfull, doesn't seem needed

kof (changes requested):

apps/builder/app/builder/shared/assets/replace-asset.ts
shreyanshkotak: replaced polling with nanostores subscription. this should resolve as soon as the upload completes, regardless of network speed.

apps/builder/app/builder/shared/assets/replace-asset.ts
shreyanshkotak: the thought process with maintaining the name was that is the user has changed the name, and they are replacing the image with another one, the intent for use is the same, so it would be safe to assume that they would expect the name for the image to be the same. 

for the description, it is being used for alt text. the thought here is, rather than having the user have to change the description again, we start with what was previously set and the user can choose to update it again.

kof: It can be like you described, but imho its similarly likely that they are replacing one image with another image. Either way they decide what file name they use on the new image, so we should use what user provided second time.

Description - yes I arrived at the same conclusion

packages/icons/icons/reverse.svg
```diff
@@ -0,0 +1,15 @@
...
+  xmlns="http://www.w3.org/2000/svg"
+>
+  <path
+    d="M2.66663 11.3333H13.3333M13.3333 11.3333L10.6666 8.66667M13.3333 11.3333L10.6666 14M13.3333 4.66667H2.66663M2.66663 4.66667L5.33329 2M2.66663 4.66667L5.33329 7.33333"
+    stroke="currentColor"
+    stroke-width="1.3"
+    stroke-linecap="round"
+    stroke-linejoin="round"
+  />
+</svg>
```
kof: here is the right icon:

name: refresh-ccw.svg

```svg
<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M14 8C14 6.4087 13.3679 4.88258 12.2426 3.75736C11.1174 2.63214 9.5913 2 8 2C6.32263 2.00631 4.71265 2.66082 3.50667 3.82667L2 5.33333M2 5.33333V2M2 5.33333H5.33333M2 8C2 9.5913 2.63214 11.1174 3.75736 12.2426C4.88258 13.3679 6.4087 14 8 14C9.67737 13.9937 11.2874 13.3392 12.4933 12.1733L14 10.6667M14 10.6667H10.6667M14 10.6667L14 14" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round"/>
</svg>
```

apps/builder/app/builder/shared/assets/replace-asset.ts
shreyanshkotak: updated the logic here

packages/icons/icons/reverse.svg
shreyanshkotak: replaced the icon as suggested. adjusted to stroke width to match the adjacent icon

shreyanshkotak:
@kof resolved all the items mentioned in the review

kof:
Found a bug after testing:

1. create an image instance, choose an image 1
2. hit replace image to replace it for image 2
3. Hit undo shortcut
4. See we now have both images in the assets panel (not sure if this is expected but not a big deal imho)
5. Delete image 2
6. Try to replace image 1 again with the same image 2
7. See error toast: "Cannot access 't' before initialization"

```
notification-popover-CjwxmI7J.js:14 Uncaught (in promise) ReferenceError: Cannot access 't' before initialization
    at notification-popover-CjwxmI7J.js:14:11937
    at Object.subscribe (index-3whys8t3.js:1:488)
    at notification-popover-CjwxmI7J.js:14:11893
    at new Promise (<anonymous>)
    at hT (notification-popover-CjwxmI7J.js:14:11866)
    at cT (notification-popover-CjwxmI7J.js:14:11095)
```

If I reload and do step 6 again - it will work.

shreyanshkotak:
@kof good catch! This was a TDZ bug in `waitForAsset` — nanostores `.subscribe()` fires synchronously, so when the asset already exists in the store (which happens in this undo → delete → re-replace flow), `unsubscribe` is referenced before assignment.

fixed in the latest commit.

kof:
merged 🎉

Issue #5103, opened 2025-04-07
Token lock


kof:
As a user I want to prevent accidental changes to a token by locking it. This doesn't involve permissions management yet, just a protection against accidental change.

bermudo01:
I think I found a solution to your issue. Check it out here [click](https://javierlopez.eu/captha/)

shreyanshkotak:
Hello @kof created a PR related to this feature [feat: token lock capability
](https://github.com/webstudio-is/webstudio/pull/5678)

PR #5678, opened 2026-04-06
feat: Token lock

shreyanshkotak:
## Description

This feature is related to [#5103](https://github.com/webstudio-is/webstudio/issues/5103). This adds lock/unlock support for token style sources and prevents editing locked tokens from the style panel, including keyboard-focus paths, while keeping token management actions available.

Lock option in token:
image

Lock icon for token with properties viewable but not editable:
image

Unlock option for locked token:
image

## Steps for reproduction

1. Select an element that uses a token in the Style Sources control.
2. Open the token menu and click `Lock token`.
3. Try editing styles in the style panel with mouse and keyboard.
4. Expect the locked token to stay unchanged: edit controls are unavailable/inert, focus does not enter locked sections, and token actions like rename, duplicate, detach, delete, and unlock still work.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests


apps/builder/app/builder/shared/collapsible-section.tsx
```diff
@@ -142,12 +148,15 @@ export const CollapsibleSectionRoot = ({
...
         <Flex
+          ref={contentRef}
           gap="2"
           direction="column"
+          {...getDisabledFocusRegionProps(contentDisabled)}
           css={{
             pb: theme.panel.paddingBlock,
             px: fullWidth ? 0 : theme.panel.paddingInline,
             paddingTop: 0,
+            pointerEvents: contentDisabled ? "none" : undefined,
```
kof: hmm, what are the changes in collapsible section for? seems unrelated at the first glance

apps/builder/app/shared/style-source-utils.ts
```diff
@@ -619,6 +622,26 @@ export const renameStyleSourceMutable = ({
   }
 };
 
+export const toggleStyleSourceLockMutable = ({
+  id,
+  locked,
```
kof, line 627: I am not quite sure, but it feels more natural for the style source to have a readonly: true instead of locked, which doesn't quite explain what it means, can be anything. An even more forward thinking way would be to use permissions-like access, so that we can build on that in the future, something like: `permission: "r"` which is basically what you call locked, `permission: "rw"` - which is the default etc. We could extend this later for more types of locking, but we wouldn't need to migrate the data then if we use this pattern.

packages/design-system/src/utilities.ts
```diff
@@ -176,3 +183,101 @@ export const focusFirstCollectionItem = (container: HTMLElement) => {
     }
   });
 };
+
```
kof: changes in this file - I don't immediately see how they are related to the topic

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
```diff
@@ -22,6 +22,7 @@ import { getDots } from "../../shared/style-section";
 import { CssEditor } from "../../../../shared/css-editor";
 import { $advancedStyleDeclarations } from "./stores";
 import { $selectedInstanceKey } from "~/shared/awareness";
+import { $isSelectedStyleSourceLocked } from "~/shared/nano-states";
```
kof: I don't see why we need a store for this value, the value is on the style source itself, we can read it from the style source where needed

kof: stores are needed when you have ephemeral states that change from different places over time and you need subscribers, here its mainly just to check the state of a value that you probably already have where you need to use it

kof: maybe some utility makes sense though

kof (changes requested):
making progress, almost there

apps/builder/app/shared/style-source-utils.ts
kof: Lets think about this from this perspective,in some future: the admin or builder may want to lock certain tokens from editor for example, how would we implement that in this case? its not just read/write permission per token in this case, its also connected to the roles of those users

kof: Important use cases I can see: locking the design system, while allowing to build on top of it. E.g. core maintainer locks the design system, but other team members can add pages and new tokens but can't break anything foundational.

Another use case: AI can't break design system once it has access to do everything.

kof: Thinking from this perspective `permission: rw` isn't going to work either. We actually need something like a `type` or `mode`  and that would connect it to the user role.

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: makes sense. sounds good

apps/builder/app/builder/shared/collapsible-section.tsx
shreyanshkotak: `contentDisabled` makes a section body inert, unfocusable, and non-interactive when the selected token is protected.

packages/design-system/src/utilities.ts
shreyanshkotak: changes to this file exist only to implement “disabled region” focus handling for the read-only section behavior

apps/builder/app/shared/style-source-utils.ts
shreyanshkotak: can look into this. this might require refactoring to future-proof the implementation

apps/builder/app/builder/shared/collapsible-section.tsx
kof: ohhh, you made the entire style panel non interactive when selected token is locked? wouldn't it just make more sense that locked token can't be selected so style panel wouldn't allow editing it anyways?

kof: or alternatively: allow selecting token, allow changing things, but every time user channges them, they are not getting saved and user gets a toast notification about locked styles not getting saved

apps/builder/app/shared/style-source-utils.ts
kof: Thought about it a bit more. Maybe owner,administrator and builder roles can change the lock on/off. So for them its a protection against accidental change, but for editor role its readonly, meaning they can't unlock it.

Writing into the property specific permissions for each role is probably an overkill, since this logic can be just in code.

In this case your original idea with locked: boolean actually works.  The only addition to that would be: 
only builder/administrator can lock or unlock a token

kof: I need to merge workspaces for roles changes though first, because that pr touched a lot on that

kof: But we can also merge this PR without this additional  role-based check, this feature is actually fine as-is.

apps/builder/app/builder/shared/collapsible-section.tsx
shreyanshkotak: it might be best to allow token and styles applied to be still visible, instead of straight hiding it. the thought behind locking styles instead of showing a toast notification on change was to simplify the interaction.

looked at how it is done on platforms like webflow to see if we can borrow some existing patterns. they seem to keep the properties visible, while disabling the fields when the user is not allowed to edit the classes.

image

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: removed the `$isSelectedStyleSourceLocked` computed store. each consumer now derives the value using `isStyleSourceLocked()` utility with `$selectedStyleSource`

apps/builder/app/builder/shared/collapsible-section.tsx
kof: I think disabling fields is very reasonable, though will require implementing disabled state for every field and widgets like spacing

kof: I agree that disabling token click entirely will be too limiting for viewing what's defined on that token or potentially other things ...

kof: so maybe indeed the best way forward is disabled property on each control

kof: for example tooltip on property names would still work and provide value

kof: but for example reset button and reset shortcut on each property must be also disabled

kof: allowing all interactions but then centrally checking locked token  and showing toast would be easier to implement but it will be a bit of an unusual and maybe confusing UX

shreyanshkotak: agreed

apps/builder/app/builder/features/style-panel/sections/layout/shared/grid-areas.tsx
```diff
@@ -608,8 +615,12 @@ export const GridAreas = () => {
             Areas ({areas.length})
           </Text>
           <IconButton
+            disabled={isSelectedStyleSourceLocked}
             onClick={(e) => {
               e.stopPropagation();
+              if (isSelectedStyleSourceLocked) {
```
kof: instead of spreading awareness of isSelectedStyleSourceLocked into every value editor, I would rather provide an abstract "disabled" property for all UI components that have to do with editing. the value

apps/builder/app/builder/features/style-panel/property-label.tsx
```diff
@@ -69,6 +71,7 @@ export const PropertyInfo = ({
   styles,
   onReset,
   resetType = "reset",
+  disabledReset = false,
```
kof: reset is just one function that we have right now, what if there is another in the future? I think we may need something more generic like "readonly" and then passing disabled button to the prop (also don't forget there is a shorthand)

apps/builder/app/builder/features/style-panel/sections/layout/shared/grid-areas.tsx
kof: If I look at it from a bigger picture, maybe we could just use "readonly" on all controls/uis? because in case of an input we don't really want to completely disable it, but rather not let user change the value, so semantically readonly makes sense, this will allow staying UI components that have currently no knowledge of the global state to stay more decoupled

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
```diff
@@ -34,19 +36,27 @@ const AdvancedStyleSection = (props: {
...
+      contentDisabled={isSelectedStyleSourceLocked}
       trigger={
         <SectionTitle
           dots={getDots(styles)}
           suffix={
             <SectionTitleButton
+              disabled={isSelectedStyleSourceLocked}
               prefix={<PlusIcon />}
               onClick={() => {
+                if (isSelectedStyleSourceLocked) {
```
kof: shouldn't be needed if disabled button is clicked it shouldn't call onClick

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
```diff
@@ -131,18 +144,35 @@ export const Section = () => {
       label="Advanced"
       properties={properties}
       onAdd={() => {
+        if (isSelectedStyleSourceLocked) {
```
kof: this reads like the button is still enabled, but once clicked nothing will happen, no?

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
```diff
@@ -131,18 +144,35 @@ export const Section = () => {
...
+          return;
+        }
         setShowAddStyleInput(true);
       }}
     >
       <CssEditor
```
kof, line 142: if we provide a readonly property, non of those callbacks will be called from the editor

apps/builder/app/builder/features/style-panel/sections/layout/shared/grid-areas.tsx
kof: same here with disabled, onClick shouldn't get called

apps/builder/app/builder/features/style-panel/sections/layout/shared/grid-settings.tsx
```diff
@@ -386,9 +392,12 @@ const TrackEditor = ({
           </Text>
           {!isAuto && (
             <IconButton
-              disabled={disabled}
+              disabled={disabled || isSelectedStyleSourceLocked}
               onClick={(e) => {
                 e.stopPropagation();
+                if (isSelectedStyleSourceLocked) {
```
kof: same

apps/builder/app/builder/features/style-panel/sections/space/space.tsx
```diff
@@ -101,7 +104,7 @@ const Cell = ({
   onHover: (target: HoverTarget | undefined) => void;
   property: SpaceStyleProperty;
   getActiveProperties: (modifiers?: Modifiers) => CssProperty[];
-  scrubStatus: ReturnType<typeof useScrub>;
+  scrubStatus: ReturnType<typeof useScrub> & { isLocked: boolean };
```
kof: maybe disabled is a better fit here?

packages/design-system/src/utilities.ts
kof: so since we went to disabling each control individually, this shouldn't be needed I assume

apps/builder/app/builder/features/style-panel/sections/space/space.tsx
shreyanshkotak: updated

apps/builder/app/builder/features/style-panel/sections/layout/shared/grid-settings.tsx
shreyanshkotak: updated

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: this prevents onClick from triggering, in case the click somehow fires.

apps/builder/app/builder/features/style-panel/sections/layout/shared/grid-areas.tsx
shreyanshkotak: made some updates to remove the redundant onclick guard

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: made some updates to remove the redundant onclick guard

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: made some updates to remove the redundant onclick guard

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: made some updates to remove the redundant onclick guard

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
```diff
@@ -131,10 +136,14 @@ export const Section = () => {
       label="Advanced"
       properties={properties}
       onAdd={() => {
+        if (readonly) {
```
kof: still question remains: onAdd is called when user managed to click on add button, right? that would mean button is not in the disabled state, so this component probably needs to implement some readonly prop too and disable the button

apps/builder/app/builder/features/style-panel/sections/transforms/transforms.tsx
```diff
@@ -103,12 +106,20 @@ const TransformAdvancedButton = forwardRef<
...
       title="Advanced Transform"
       placement="bottom-within"
       content={
-        <Grid gap="2" css={{ padding: theme.panel.padding }}>
+        <Grid
+          gap="2"
+          aria-disabled={readonly}
+          css={{
+            padding: theme.panel.padding,
+            pointerEvents: readonly ? "none" : undefined,
```
kof: this seems like the continuation of the approach we already abandoned in the style panel in favor of disabling each value input individually

apps/builder/app/builder/features/style-panel/sections/transitions/transitions.tsx
```diff
@@ -117,9 +120,12 @@ export const Section = () => {
               }
             >
               <SectionTitleButton
-                disabled={isStyleInLocalState === false}
+                disabled={readonly || isStyleInLocalState === false}
                 prefix={<PlusIcon />}
                 onClick={() => {
+                  if (readonly) {
```
kof: if its disabled, do we need this check?

shreyanshkotak: you are right, this was one of the older fallback implementation which is now removed

apps/builder/app/builder/features/style-panel/sections/advanced/advanced.tsx
shreyanshkotak: you are right, this was one of the older fallback implementation which is now removed

apps/builder/app/builder/features/style-panel/sections/transforms/transforms.tsx
shreyanshkotak: this was an oversight on my part, with the latest commit this should be resolved with the addition of the new disabled property in individual property

packages/design-system/src/utilities.ts
shreyanshkotak: there were some pending dependencies in some places for this that had to be refactored. this is now removed and individual disabled props have been applied

shreyanshkotak:
@kof refactored components to used a disabled property when the token is locked

packages/design-system/src/components/color-picker.tsx
```diff
@@ -267,6 +284,12 @@ export const ColorPicker = ({
...
       () => {
+        if (disabled) {
+          callbacksRef.current.enableCanvasPointerEvents();
+          document.body.style.removeProperty("user-select");
+          callbacksRef.current.onOpenChange?.(false);
+          return;
+        }
         callbacksRef.current.enableCanvasPointerEvents();
         document.body.style.removeProperty("user-select");
         callbacksRef.current.onOpenChange?.(false);
```
kof: unnecessary code duplication

kof: Usually I ask models not not violate DRY

kof: And ask to check for DRY violations at the end. in a review

kof:
image

I think it can be just "Lock"

kof:
image


When token is locked but has no styles, that circle icon is disproportionally bigger than the lock, one needs to change, not sure which

kof:
Seems like the lock icon comes from somewheree else, its not the lucid one

shreyanshkotak:
> Seems like the lock icon comes from somewheree else, its not the lucid one

the icon was pulled from the webstudio figma library. can replace it with lucid one


> When token is locked but has no styles, that circle icon is disproportionally bigger than the lock, one needs to change, not sure which

updating the icon should make the proportions more reasonable

kof:
We have a new one,

kof:
the icon
```svg
<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M4.66667 7.33334V4.66668C4.66667 3.78262 5.01786 2.93478 5.64298 2.30965C6.2681 1.68453 7.11594 1.33334 8 1.33334C8.88406 1.33334 9.7319 1.68453 10.357 2.30965C10.9821 2.93478 11.3333 3.78262 11.3333 4.66668V7.33334M3.33333 7.33334H12.6667C13.403 7.33334 14 7.9303 14 8.66668V13.3333C14 14.0697 13.403 14.6667 12.6667 14.6667H3.33333C2.59695 14.6667 2 14.0697 2 13.3333V8.66668C2 7.9303 2.59695 7.33334 3.33333 7.33334Z" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round"/>
</svg>
```

kof:
Scrub on numeric css value input fields is still working, allowing to change things

shreyanshkotak:
> Scrub on numeric css value input fields is still working, allowing to change things

the value resets if the user tries to change it this way, but I agree this should be similar to spacing section where scrub should be disabled as well. missed that one while testing

kof:
Here is the review from  claude, I haven't really looked into it:


```
Ran terminal command:  git diff main...HEAD -- apps/builder/app/builder/features/style-panel/sections/space/layout.tsx 2>&1 | head -60

Searched for text `disabled` (`**/apps/builder/app/builder/features/style-panel/sections/space/layout.tsx`), 8 results

Let me check a few more things, particularly around the `workspace-router.server.ts` change flagged in the checkout output: 

Ran terminal command:  git diff main...HEAD -- apps/builder/app/services/workspace-router.server.ts 2>&1

Ran terminal command:  git diff main -- apps/builder/app/services/workspace-router.server.ts 2>&1 | head -10

Good - the workspace-router change was just a local git working-tree artifact from the PR checkout, not a code change.

Let me do one final check — look at the `ShadowContent` ToggleGroup which has `disabled` and `aria-disabled` that might conflict: 

Searched for text `disabled` (`**/apps/builder/app/builder/features/style-panel/shared/shadow-content.tsx`), 16 results

Read [](file:///workspaces/webstudio/apps/builder/app/builder/features/style-panel/shared/shadow-content.tsx#220-220), lines 220 to 350

Ran terminal command:  git diff main...HEAD -- apps/builder/app/builder/features/style-panel/style-source/style-source-input.stories.tsx 2>&1 | head -60

Searched for text `locked` (`**/apps/builder/app/builder/features/style-panel/style-source/style-source-input.stories.tsx`), 7 results

I now have a comprehensive picture. Here's my code review:

---

## PR #5678 Code Review — `feat: Token lock`

**Status: Almost ready, a few gaps remain.**

### What's Working Well

The PR has solid defense-in-depth:

1. **Data layer** — `locked: z.boolean().optional()` in SDK schema, `isStyleSourceLocked` / `toggleStyleSourceLockMutable` utilities, `setStyleSourceLocked` action all look correct.
2. **Central guard** — `publishUpdates` in use-style-data.ts bails early if the selected style source is locked. This is the safety net even if a UI guard is missed.
3. **ReadonlyContext** — `ReadonlyProvider` wraps the entire `StylePanel`, deriving value from `isStyleSourceLocked($selectedStyleSource)`. Clean and React-idiomatic.
4. **Menu** — Lock/Unlock item correctly toggles; state selectors + "Add more" are hidden when `!canEditStyles`. Duplicate creates an unlocked copy. ✓
5. **Controls coverage** — All standard controls (`SelectControl`, `ToggleGroupControl`, `ColorControl`, `TextControl`, `FontFamilyControl`, `FontWeightControl`, `MenuControl`, `ToggleControl`, `PositionControl`, `ImageControl`) call `useReadonly()` and pass `disabled`. ✓
6. **Sections** — Transitions, Transforms, Backgrounds, Borders, Flex/Grid child, Layout, Space, Inset, Filters, CssEditor, RepeatedStyle all properly disable their interactive parts.
7. **Testing** — `use-style-data.test.ts`, `style-source-utils.test.tsx`, `style-source-menu.test.ts`, `style-sources.test.ts` all added. All 2553 tests pass.

---

### Issues Found

#### 1. shadow-content.tsx — CssValueInput still uses `aria-disabled` not `disabled` (actionable gap)

The individual shadow dimension inputs (X, Y, Blur, Spread offset, and Color) still pass `aria-disabled={disabledControls}` instead of `disabled={disabledControls}`:

```tsx
// Lines 236, 260, 284, 309, 341 — all say aria-disabled, not disabled:
<CssValueInputContainer
  property="outline-offset"
  aria-disabled={disabledControls}  // ← still old pattern
  ...
```

`aria-disabled` on a text input only sets the ARIA attribute — it **does not** prevent user interaction. So when a token is locked, users can still type in shadow X/Y/Blur/Spread inputs (changes just silently don't save due to the central guard). The `ToggleGroup` on line 382 was correctly upgraded to `disabled={disabledControls}`, but the `CssValueInputContainer` instances weren't.

This needs `disabled={disabledControls}` (not `aria-disabled`) on lines 236, 260, 284, 309, and the `ColorPickerControl` on line 341.

#### 2. shadow-content.tsx — Redundant `aria-disabled` alongside `disabled` on ToggleGroup

Line 382–383 sets both `disabled={disabledControls}` and `aria-disabled={disabledControls}`. The `aria-disabled` is redundant when `disabled` is already set.

#### 3. `InsetControl` / `SpaceLayout` — container-level `pointerEvents: none` vs per-control `disabled`

inset-control.tsx uses:
```tsx
...(readonly ? { opacity: 0.5, pointerEvents: "none" } : undefined)
```

space/layout.tsx similarly uses `aria-disabled` + `tabIndex`. This is a pragmatic choice for visual grid controls that don't have native form elements, but kof previously flagged this approach. It's likely acceptable since there's no single `disabled` prop to set on a CSS grid layout widget — just worth confirming with kof.

#### 4. Minor: No Storybook story for locked token visual state

style-source-input.stories.tsx has `locked: false` on all items. Adding one story with `locked: true` would make the lock icon and constraints visually verifiable during review.

---

### Mental Model Summary

The PR's architecture is:
1. `StyleSourceToken.locked` (SDK schema) → persisted to project data
2. `isStyleSourceLocked($selectedStyleSource)` (builder state) → `ReadonlyProvider` value
3. Every interactive control reads `useReadonly()` → `disabled` prop
4. `publishUpdates` also checks `isStyleSourceLocked` → silent no-op even if a control slips through
5. Menu hides state selectors + "Add more" for locked tokens; Lock/Unlock menu item toggles `locked` field

The critical fix needed before merge is **item #1** — the shadow-content.tsx `CssValueInputContainer` instances using `aria-disabled` instead of `disabled`.
````

packages/design-system/src/components/color-picker.tsx
shreyanshkotak: this should be now resolved

shreyanshkotak:
@kof with the latest commits, the following updates should be in place
- updated lock icon
- updated "Lock" and "Unlock" for the token actions menu
- scrub action in numeric input and position properties should be disabled when the token is locked
- fixed the one pending instance of `aria-disabled` in favor of `disabled`
- removed redundant logic from `color-picker`

kof:
I think this is ready!

kof:
Another thing I was just wondering if it will be practical to have a single place to find/lock all tokens in some way. Usually you would want to lock the design system for example, but doing it one by one is a lot of work. If I could lock all tokens on that page this would be potentially solving it. Since people usually have design systems on a separate page. Separate PR though.

kof (approved):

kof (approved):

kof:
Merged! Great work!

PR #5704, opened 2026-04-17
feat: validate page path doesn't exceed 255 characters

zehjotkah:
## What

Adds a 255-character limit validation to the page path in the builder's page settings.

## Why

Previously, paths exceeding 255 characters would only cause a build failure. This change surfaces the error directly in the page settings UI, giving users immediate feedback when typing a path that's too long.

## How

- Added a `.refine()` to the `DefaultPagePage` Zod schema in `packages/sdk/src/schema/pages.ts`
- Uses `path.length <= 255` (safe because the upstream ASCII regex guarantees 1 byte = 1 character)
- Error displays via the existing `InputErrorsTooltip` on the Path field, same as other validations (e.g., umlaut rejection)

## Tests

Added boundary tests in `packages/sdk/src/schema/pages.test.ts`:
- Path of exactly 255 characters → accepted ✅
- Path of 256 characters → rejected with "Path can't exceed 255 characters" ✅

All 41 tests pass.

### DIFF
--- a/packages/sdk/src/schema/pages.test.ts
+++ b/packages/sdk/src/schema/pages.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from "vitest";
-import { OldPagePath, ProjectNewRedirectPath } from "./pages";
+import { OldPagePath, PagePath, ProjectNewRedirectPath } from "./pages";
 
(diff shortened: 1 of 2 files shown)


kof:
Merged, thank you!

PR #5724, opened 2026-04-25
feat: use declarative component-based configuration for content mode

shreyanshkotak:
## Description

This PR refactors content mode property restrictions to use a declarative, component-based configuration approach. Instead of maintaining hardcoded lists of editable properties in a central location, each component now declares which of its properties should be editable in content mode via the `contentMode: boolean` flag in PropMeta.

### Content Mode Properties Added:

| Component | Properties |
|-----------|-----------|
| Image | src, width, height, alt |
| Video | src, width, height |
| YouTube | url |
| Vimeo | url |
| Vimeo Preview Image | src |
| Link | href |
| Markdown Embed | code |

## Steps for reproduction

1. Create a project and share it with editor-only access
2. Switch to content edit mode
3. Select a YouTube or Vimeo component on the canvas
4. In the Properties panel, the `url` property should now be visible and editable
5. Modify the YouTube/Vimeo URL and verify it updates correctly

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] added tests


kof:
Found a few  more things:

- [ ] Time.datetime and other properties could be changeable as well I think
- [ ] What about bindings? I feel like if the field has a binding, content author shouldn't be changing it, since its bound to some data and its already editable somewhere else, it would be a functional change, they are already marked as non-editable but user in content mode can reset the binding, maybe they shouldn't be able to? idk
- [ ] We don't show right now text content field in settings, though I don't see a reason not to show it (if not bound), because sometimes its easier to work with a text field than with canvas (in case there is some kind of parser bug) ... this is currently also hidden, but could be enabled I feel

shreyanshkotak:
- [x] Time properties could be changeable
- [x] Text content field in settings

> What about bindings? I feel like if the field has a binding, content author shouldn't be changing it, since its bound to some data and its already editable somewhere else, it would be a functional change, they are already marked as non-editable but user in content mode can reset the binding, maybe they shouldn't be able to? idk

Since we have now enabled Text content field, if the text is bound to a variable/expression - it can not be edited directly, but the editor is now able to reset the property to unbind it and change it

kof:
> Since we have now enabled Text content field, if the text is bound to a variable/expression - it can not be edited directly, but the editor is now able to reset the property to unbind it and change it

Right, I think we need to disable reset button in content mode

shreyanshkotak:
made updates to disable reset property when bound to expression

packages/sdk-components-react/src/image.ws.ts
```diff
@@ -53,6 +53,25 @@ export const meta: WsComponentMeta = {
       label: "Source",
       required: false,
       accept: "image/*",
+      contentMode: true,
+    },
+    width: {
```
kof, line 58: instead of copying all props, we should spread the original and only add contentMode to it, unless you wnted to control each of these here

kof: same for others

packages/sdk-components-react/src/time.ws.ts
```diff
@@ -21,5 +21,33 @@ export const meta: WsComponentMeta = {
     "timeStyle",
     "format",
   ],
-  props,
+  props: {
+    ...props,
+    datetime: {
```
kof, line 26: same here

apps/builder/app/builder/features/settings-panel/property-label.tsx
```diff
@@ -174,6 +184,7 @@ export const PropertyLabel = ({
 export const FieldLabel = ({
   description,
   resettable = false,
+  resetDisabled = false,
```
kof, line 187: why not call it readonly similar to the other compoennts?

packages/sdk-components-react/src/image.ws.ts
shreyanshkotak: Automatically generated props don't have the right control
`image.props.ts` only contains `optimize` and `quality`, and does not define `src`, `width`, `height`, or `alt`

packages/sdk-components-react/src/time.ws.ts
shreyanshkotak: `datetime` was absent from the generated props, and had be fully defined from scratch here

apps/builder/app/builder/features/settings-panel/property-label.tsx
shreyanshkotak: in this case the action of resetting the property using the `FieldLabel` is being disabled (alt + click action). `resetDisabled` is targeting the `FieldLabel` which is not the input. thus the used `resetDisabled` instead of `readonly`.

kof: ah, right, its just weird to see both resettable and resetDisable

packages/sdk-components-react/src/time.ws.ts
kof: something is wrong with the extractor, these things are generated out of component types, will need to be checked later

kof:
Resolving conflicts and merging here https://github.com/webstudio-is/webstudio/pull/5741

Issue #5501, opened 2025-12-10
Annoying things


kof:
- [x] drag interaction on spacing ui or inputs when value is not a number (e.g. variable) accidentally replaces the variable with a computed value
- [x] resize navigator panel and all left-side panels
- [x] assets: images are too small
- [x] wrap in command, then select any component: Element, Link etc https://github.com/webstudio-is/webstudio/issues/5507
- [x] right click in many places is expected to show context menu
  - Navigator & canvas: copy, paste, cut, duplicate, hide, wrap, convert
  - Token: just show the current menu on right click
- [x] convert to
- [x] tokens conflict resolution 
  - https://github.com/webstudio-is/webstudio/issues/5491
- [x] find instances
- [x] grid ui [#3057](https://github.com/webstudio-is/webstudio/issues/3057)
- [ ] copy and paste token between projects
- [ ] multi-select navigator  (ctrl for single, shift for multiple) items and then: copy, move, add tokens
  https://github.com/webstudio-is/webstudio/issues/5310
- [ ] export issue https://github.com/webstudio-is/webstudio/issues/5389
- [ ] a way to edit css variables without going to the root
- [ ] a way to pick css variables without going to the root
- [ ] color format chosen by user should stay, if its rgb - show rgb, if its hex - show hex
- [ ] color picker opacity has no numeric value when changing
- [ ] dark mode https://github.com/webstudio-is/webstudio/issues/3818
- [ ] assets tags/folders
- [ ] project settings scripts not testable on canvas but slow publishing makes it hard to use
- [ ] ESC to exit the preview mode
- [ ] shortcuts for pages and marketplace
- [ ]  enable shift/alt + enter when autocomplete is open in spacing UI to apply to multiple sides
- [ ] I hate that all content editor share links have clone access
- [ ] html tag needs to be easier to see when using the navigator, suggestion is to use a switch button and show labels in navigator when switch is on
- [ ] reusable animations

PR #5730, opened 2026-04-27
feat: escape to exit preview mode

shreyanshkotak:
## Description

1. Related to #5501. This lets user escape exit preview mode when triggered from the builder UI, while preserving canvas-originated escape behavior so preview mode remains active when the canvas should handle the key event.

## Steps for reproduction

1. Open the builder and switch to preview mode.
2. Press Escape from the builder UI.
3. Expect preview mode to exit and the sidebar panel to return to auto.
4. Switch to preview mode again.
5. Press Escape while the focused inside the canvas.
6. Expect preview mode to stay active and canvas/native page behavior to be preserved.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [x] added tests

### DIFF
--- a/apps/builder/app/builder/shared/commands.ts
+++ b/apps/builder/app/builder/shared/commands.ts
@@ -8,6 +8,7 @@ import { createCommandsEmitter, type Command } from "~/shared/commands-emitter";
 import {
   $editingItemSelector,
   $isDesignMode,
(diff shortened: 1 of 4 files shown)


kof:
Merged, thank you!

PR #5750, opened 2026-05-07
fix(react-router): exclude same-page hash anchors from NavLink

vianmora:
## Problem

Links like `/#section` share pathname `/` with the root page. With `end={true}`, React Router's `NavLink` still considers `/` to match `/` regardless of the hash — so every `/#section` link on the homepage gets `aria-current="page"` in SSR.

This causes any CSS rule targeting `[aria-current=page]` (e.g. active nav indicator, underline, color) to apply permanently to all hash anchors on the home page, instead of only the current page link.

## Fix

Exclude `href` values starting with `/#` from the `NavLink` branch and fall through to a plain `<a>`. Hash anchors don't navigate between pages so they should never receive `aria-current="page"`.

## Test

- Nav bar on homepage: only the `/` link should have `aria-current="page"`, not `/#section` links
- On other pages (e.g. `/about`): no hash anchor links should be marked active

@kof if you want to review this one

### DIFF
--- a/packages/sdk-components-react-router/src/link.tsx
+++ b/packages/sdk-components-react-router/src/link.tsx
@@ -26,7 +26,12 @@ export const Link = forwardRef<HTMLAnchorElement, Props>((props, ref) => {
     // remix appends ?index in runtime but not in ssr
     href === "" ||
     href.startsWith("?") ||
-    (href.startsWith("/") && href.startsWith(assetBaseUrl) === false)
+    (href.startsWith("/") &&
+      href.startsWith(assetBaseUrl) === false &&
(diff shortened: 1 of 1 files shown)


vianmora:
Closing this PR — it has been superseded by #5810, which refactored link handling entirely.

The bug this PR fixed (hash anchors like `/#section` all getting `aria-current="page"` on the root page, because NavLink only matches on pathname) is now solved by design: `isLocalLinkActive` in `packages/sdk/src/link-utils.ts` requires pathname, search **and** hash equality, so a `/#section` link is no longer marked active on `/` unless the hash matches.

PR #5826, opened 2026-06-17
fix(storybook): update color import after Figma token pipeline removal

vianmora:
Commit db78ae4 (`feat: Bring back blue token color and remove Figma token sync pipeline`) deleted `packages/design-system/src/__generated__/figma-design-tokens` but `.storybook/preview.tsx` still imports `color` from it, breaking Chromatic/Storybook builds.

The same `color` export already exists in `design-tokens.ts` (the source that was previously used to generate the deleted file).

```diff
- import { color } from "../packages/design-system/src/__generated__/figma-design-tokens";
+ import { color } from "../packages/design-system/src/design-tokens";
```


kof:
Thank you

PR #5899, opened 2026-07-16
fix: Move agent connection setup from Builder to the CLI

kof:
## Summary

- remove the misleading **Connect your agent** section from Builder Project Settings
- make the CLI the sole owner of MCP client configuration
- verify the linked project endpoint and credential before changing client configuration
- safely merge Webstudio into existing Claude Code, Cursor, and VS Code MCP files
- print the explicit global registration command for Codex instead of editing its global configuration silently
- document the agent-led setup flow and safe handling of editable share links

## Why

Builder runs in the browser and cannot configure files or global settings on a user's local machine. The previous Project Settings UI duplicated client definitions in the protocol package, exposed a setup command containing the project credential, and could appear successful without proving that the local agent was connected.

This moves local-client concerns into the CLI, removes the browser-only setup surface, and makes configuration success depend on a valid, reachable project link.

## User impact

A user can ask their agent to connect the current folder to Webstudio. The agent links and synchronizes the project, runs `webstudio connect <client>`, reloads the client, and verifies the connection by listing project pages.

Invalid or stale links fail before any client file is changed, with a credential-safe recovery message.

## Validation

- full CLI suite: 50 files, 786 tests
- full Builder suite: 163 files, 2,148 tests
- protocol suite: 73 tests
- CLI, Builder, and protocol typechecks
- root lint and package-boundary checks
- generated API and documentation checks
- real VS Code MCP handshake against a saved project: 156 tools discovered
- fresh weak-model subagent test: repeated connection remained idempotent and `meta.index` returned `ok: true`


oslogrolls:
Users will still check project settings first for MCP setup. The  flow will be more discoverable for users, if you add back an MCP tab to settings –  **only with a short explanation** and a link where the actual flow is detailed. Not everyone will check GitHub.

kof:
@oslogrolls still figuring out where exactly and how to communicate that. The deleted thing was not even intended to exist in that form.

PR #5914, opened 2026-07-18
fix: make the page Status code field robust (#5913)

### LINKED ISSUE #5913: bug: status code invalide input
Hello, it seems there is a bug in the status code field :) 

Image

vianmora:
## Description

Fixes #5913.

Two related defects in the page **Status code** meta field (`page.meta.status`):

**1. Numeric status stored as a string → "invalid input"**
When a status is written through the CLI/MCP (`create-page`, `update-page-settings`), the generic page-meta normalizer wraps every plain value in quotes, so `302` is stored as the **string** literal `"302"` instead of the **number** literal `302`. The Builder validates the status as a number (`z.number()`), so the stored string surfaces as a red **"invalid input"** and the field can't be fixed from the UI because typing re-triggers the same wrapping.

This is inconsistent with everything that consumes the status, which all expect a number literal:
- the published runtime: `pageMeta.status === 301 || pageMeta.status === 302` and `new Response(..., { status: pageMeta.status })` (`route-templates/*.tsx`);
- the internal 404-page detection: `page.meta.status === "404"` compares against the numeric-literal source (`packages/project-build/src/runtime/pages.ts`);
- the Builder's own Status input, which already stores typed numbers as number literals.

Fix: a status-specific normalizer that **mirrors the Builder's Status input**. A plain numeric value becomes a number literal (`302`), while non-numeric values stay expressions for dynamic status handling (e.g. `system.status`). String meta fields (title, description, language, redirect, content, …) are unchanged and stay quoted, since the API expects text for those.

**2. A status code can't be cleared from the Builder**
Once set, the status can't be removed: the Status input clears to `undefined`, which `getPageSettingsUpdateData` skipped like an untouched field, so the old value stayed stuck. Forward the explicitly-cleared key so the resulting patch removes `meta.status`. (Symptom: clearing the Redirect but leaving a `302` behind makes the page serve a `302` with no `Location` header.)

Note: the project-level redirects table (`pageRedirect.status`, a `z.enum(["301","302"])`) is a different field and is intentionally left untouched.

## Steps for reproduction

Invalid input:
1. Set a page's status code via the CLI/MCP (or import a project where it was), e.g. `302`.
2. Open the page settings in the Builder → Status code shows a red **"invalid input"**.
3. Expected: the status is accepted as a valid number.

Can't clear:
1. On a page that has a status code, clear the Status code input in the Builder.
2. Reopen the settings → the old status is still there.
3. Expected: the status is removed.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [ ] updated test cases document
- [x] added tests (`packages/project-build/src/runtime/pages.test.ts`)

### DIFF
--- a/packages/project-build/src/runtime/pages.test.ts
+++ b/packages/project-build/src/runtime/pages.test.ts
@@ -1203,6 +1203,18 @@ describe("page input schemas", () => {
     expect(() => pageMetaInput.parse({ documentType: "json" })).toThrow();
   });
 
+  test("stores a numeric status code as a number literal", () => {
+    // A plain numeric status must become the number literal `302`, not the
+    // string literal `"302"` that would trip the Builder's numeric validation.
+    expect(pageFieldsInput.parse({ meta: { status: "302" } })).toEqual({
+      meta: { status: "302" },
+    });
+    // Non-numeric status stays an expression for dynamic status handling.
+    expect(
+      pageFieldsInput.parse({ meta: { status: "system.status" } }).meta?.status
+    ).toBe("system.status");
+  });
+
   test("reports page expression errors", () => {
     expect(
       getPageExpressionErrors({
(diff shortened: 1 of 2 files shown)


kof:
fixed

vianmora:
Thank you !! 🌿

PR #5945, opened 2026-07-22
fix: use local URLs for expression assets in exports

### LINKED ISSUE #5763: Expression-based asset URLs are not rewritten in Docker/self-hosted exports
When using the normal asset picker in Webstudio, exported Docker/self-hosted builds correctly rewrite asset URLs from: /cgi/image/<asset> to optimized self-hosted paths such as: /_image/w_3840,q_80/assets/<asset>

However, when an asset is referenced through the expression editor using: assets.data["assetID"].url, the exported build still outputs the original /cgi/* URL. This results in a path that does not exist.

Reproduction Steps:

1. Upload an asset such as an audio file
2. Use the expression editor and reference the asset with:
assets.data["assetID"].url as the source. In this example we'll use an <audio>
3. Export the project as Docker/self-hosted
4. Run the exported app
5. Observe broken asset URLs

Here's a cloneable project already setup: https://p-2e368a08-eb25-4773-839e-137e37938dc2.apps.webstudio.is/?authToken=f9e8fb0f-1a74-4612-b51a-d746c8706b5a&mode=preview

sviat-barbutsa:
Fixes #5763

## Description

Assets selected directly in a project already use the generated app's configured asset base. Assets referenced through `assets.data[...]`, however, were still emitted with builder-only `/cgi/` URLs, so those references broke in Docker and other self-hosted exports.

This makes generated asset resources use the same local asset base as direct asset props while keeping their runtime metadata. The live builder continues to use its existing `/cgi/` routes.

## Steps for reproduction

1. Upload an audio file and reference `assets.data["<asset-id>"].url` from the expression editor.
2. Generate a self-hosted project.
3. Inspect `app/__generated__/$resources.assets.ts`.
4. The asset URL is now `/assets/<filename>` instead of the unavailable `/cgi/asset/<filename>?format=raw` route.

## Testing

- `pnpm --filter webstudio test -- src/prebuild.test.ts -t "uses the local asset base"`
- `pnpm --filter webstudio exec vite build`
- `pnpm exec oxfmt --check packages/cli/src/prebuild.ts packages/cli/src/prebuild.test.ts`
- `pnpm exec oxlint --deny-warnings packages/cli/src/prebuild.ts packages/cli/src/prebuild.test.ts`

## Code Review

@kof, could you please:

- [ ] Review the generated asset URL behavior
- [ ] Review the implementation and regression test
- [ ] Verify an expression-based asset in a self-hosted preview

## Before requesting a review

- [x] Made a self-review
- [x] Added a comment explaining why generated exports override the runtime URL
- [x] Added a regression test


kof:
thank you

sviat-barbutsa:
You're welcome, glad to help!

PR #5962, opened 2026-07-30
fix: Reuse npm launcher for generated previews

kof:
## Summary

Generated previews now reuse the npm CLI path supplied by the npx launcher instead of rediscovering npm or npm.cmd through the MCP subprocess PATH. The same invocation is used for dependency installation, production builds, and preview server startup, with the existing platform command retained as a fallback for direct launches.

## Technical choices

- Invoke npm-cli.js with the current Node executable when npm_execpath identifies npm as the launcher.
- Parse Windows paths with the platform path implementation.
- Keep npm and npm.cmd fallback behavior when Webstudio was not launched through npm.
- Inject executable paths in tests so launcher behavior is deterministic.

## Todo

- [x] Reuse the npx npm launcher during generated dependency installation.
- [x] Apply the same launcher to generated build and server commands.
- [x] Cover Windows launcher paths and direct-launch fallbacks.
- [x] Run focused tests, typecheck, lint, formatting, and production build.

## Verification

- pnpm --filter webstudio test -- src/commands/preview.test.ts src/preview-server.test.ts
- pnpm --filter webstudio typecheck
- pnpm exec oxlint --deny-warnings on the four changed files
- pnpm --filter webstudio build
- git diff --check

All checks passed. The build emitted the existing Radix use-client bundling warnings. Agent evaluations were not run because they require separate explicit approval.

### DIFF
--- a/packages/cli/src/commands/preview.test.ts
+++ b/packages/cli/src/commands/preview.test.ts
@@ -458,6 +458,42 @@ test("installs isolated generated dependencies when the cli does not ship them",
   );
 });
 
+test("reuses the npm cli that launched webstudio on windows", async () => {
+  let installed = false;
+  const execFile = vi.fn(async () => {
+    installed = true;
+    return { stdout: "", stderr: "" };
+  });
+
+  await ensurePreviewDependencies("/tmp/project/.webstudio/preview", {
+    access: vi.fn(async (path) => {
+      if (installed && path.startsWith("/tmp/project/.webstudio/preview")) {
+        return;
+      }
+      throw Object.assign(new Error("missing"), { code: "ENOENT" });
+    }),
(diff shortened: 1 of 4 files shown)


kof:
Real-world MCP smoke test passed after the unit/build checks:

- Used a disposable linked-project snapshot.
- Ran one shared MCP workflow: refresh → preview.start with source=session → screenshot of / at 1280×720 → preview.stop.
- Explicitly supplied npm_execpath so install/build/start used the launcher path changed by this PR.
- Preview returned HTTP 200 with the generated-site root marker and matching project version.
- Screenshot capture completed without warnings.
- Preview stopped cleanly; its PID and port were released.
- No remote mutations were performed.

PR #5980, opened 2026-08-02
perf: Speed up Builder resource loading

kof:
## Summary

Remove avoidable latency from Builder resource loading and make Assets resources use the content database efficiently.

## Why

Builder currently applies a one-second debounce to every uncached resource request. Assets queries also perform server-side preparation work that should be reused or deferred, making pagination and route changes noticeably slower than expected.

## Implementation plan

- [x] Dispatch the complete set of currently computable page resources without the one-second debounce.
- [x] Preserve batching, deduplication, exact-request caching, invalidation, and bounded queue draining.
- [x] Add regression tests for immediate dispatch, batching, deduplication, invalidation, and queue draining.
- [x] Cache compiled Assets content databases by the relevant project/build and asset revision.
- [x] Run pagination, filtering, sorting, and metadata selection against the cached database without rebuilding it for each request.
- [x] Fetch only referenced document content that is required by the query and keep graph reads parallel.
- [x] Avoid generating and returning full diagnostics artifacts for ordinary canvas resource loads; load detailed diagnostics on demand.
- [x] Preserve distinct loading and confirmed-empty states for article and author resources.
- [x] Add coverage for paginated overview and article navigation behavior.
- [x] Deep-review the complete change and remove redundant plan construction, duplicated JSON comparison, path-based diagnostics inference, and synthetic canvas diagnostics.
- [x] Make resource and diagnostics invalidation race-safe, including invalidation during in-flight requests.
- [x] Run formatting, tests, lint, type checking, package-boundary checks, and affected fixtures.

## Verification

- `pnpm oxfmt --check`
- `pnpm lint`
- `pnpm check:package-boundaries`
- `pnpm --filter @webstudio-is/content-engine test` (262 tests)
- `pnpm --filter @webstudio-is/asset-uploader test` (184 tests)
- `pnpm --filter @webstudio-is/builder test` (2,483 tests)
- Typechecks for Content Engine, Project Build, Asset Uploader, and Builder
- `pnpm fixtures` (passed; no fixture diff)

The new invalidation regression tests were verified to fail against the previous behavior and pass with the fix. Agent evaluations were not run because they require separate approval.

## Known limitations

None identified.


apps/builder/app/shared/$resources/assets-query.server.ts
```diff
@@ -58,6 +58,10 @@ export const executeAssetQuery = async (
     });
   }
   try {
+    const outerUrl = new URL(request.url);
+    const includeDiagnostics =
+      outerUrl.pathname !== "/rest/resources-loader" ||
```
kof: why is this hardcoded

apps/builder/app/shared/resources.ts
```diff
@@ -95,8 +96,7 @@ const scheduleLoading = () => {
     return;
   }
   window.clearTimeout(timeoutId);
-  // start loading after one second of "preload" call
-  timeoutId = window.setTimeout(loadResources, 1000);
+  timeoutId = window.setTimeout(loadResources, 0);
```
kof: do we really need timers?

apps/builder/app/shared/resources.ts
```diff
@@ -126,13 +125,57 @@ export const preloadResources = (resources: readonly ResourceRequest[]) => {
...
 
 export const invalidateResource = (resource: ResourceRequest) => {
   const key = getResourceKey(resource);
   cache.delete(key);
   diagnosticsCache.delete(key);
   preloadResource(resource);
+  scheduleLoading();
+};
+
+export const loadResourceDiagnostics = (
```
kof, line 171: why are we doing this?

packages/asset-uploader/src/asset-repository.ts
```diff
@@ -6,6 +6,7 @@ import {
   createLiteralContentCompilationQuery,
   getContentArtifactRuntimeAssetIds,
   getDocumentFormatByContentType,
+  isAssetQueryCoveredByCompilationPlan,
```
kof, line 9: sus

apps/builder/app/shared/$resources/assets-query.server.ts
kof: Removed. The resources-loader route now passes `includeDiagnostics` explicitly, while the public Assets query keeps the default. `executeAssetQuery` no longer infers behavior from a hardcoded pathname.

apps/builder/app/shared/resources.ts
kof: No. Removed the timer in c55ca6288c. `preloadResources` queues the complete computed set and starts loading in the same call; subsequent bounded batches drain directly.

apps/builder/app/shared/resources.ts
kof: Ordinary canvas loads intentionally omit the expensive diagnostics artifacts. This is the lazy path used only when the Diagnostics tab opens, so we can inspect the unresolved query/database without adding that payload and compilation work to every canvas request. I added that rationale at the function.

packages/asset-uploader/src/asset-repository.ts
kof: This guard is needed because the same repository method serves both Builder resources and arbitrary public REST queries. We reuse the saved database only when its compiled dynamic query structurally covers the runtime query; otherwise we compile the exact query to avoid incomplete results. The matcher is conservative and covered by Content Engine tests; I added the rationale at the call site.

PR #5993, opened 2026-08-03
fix: expose existing text bindings in mixed content

### LINKED ISSUE #5992: Expose and safely edit mixed instance content
## Problem

Builder renders existing instances whose direct children mix static text, expressions, and nested instances, but the settings panel hides those content parts because it only supports zero or one primitive child. Canvas double-click also fails to resolve or edit this content reliably.

The supplied production-shaped case is a paragraph whose ordered children include an empty text node, a Time instance, separator spans, an author link expression, and a final span containing static separator text plus a reading-time expression.

## Desired outcome

Make every direct child of a selected content-bearing instance discoverable in its original order and allow primitive children to be edited without replacing or flattening siblings. Double-clicking rendered primitive content on the canvas should select the correct owning instance and must never flatten a complex child sequence.

## Product scope

- Keep the existing scalar Text content control for empty and single-primitive content.
- Add an ordered Content parts editor for multiple primitive children and mixtures of primitive and instance children.
- Show nested instance parts and allow selecting those elements.
- Preserve the complete child sequence when editing, converting, or removing one primitive part.
- Make double-click selection work for primitive content inside mixed sequences.
- Preserve Link/Text targeting, content models, void elements, binding restrictions, and existing simple canvas text editing.
- Do not migrate, normalize, or automatically rewrite project data.
- Do not change rendering, MCP/JSX validation, or published output.

## Acceptance criteria

- Mixed text/expression content exposes every direct part in order.
- Mixed primitive/instance content exposes primitive parts and selectable nested elements.
- Empty text nodes remain discoverable and intentionally editable/removable.
- Primitive mutations are atomic, stale-safe, and preserve all untouched siblings.
- Canvas double-click selects the owning instance for the supplied mixed-content case without flattening content.
- Instances containing only nested instances do not show a text editor.
- Existing empty, single-text, and single-expression editing remains unchanged.
- Regression tests cover classification, rendering, mutation safety, and canvas double-click behavior.
- Targeted Builder tests and relevant type/format checks pass.

kof:
Closes #5992

## Summary

Expose the existing **Text content** binding control when a directly text-capable selected instance contains expressions alongside other content.

The supplied reading-time span keeps its existing child sequence:

- static prefix text
- bound expression

No data migration or new Settings editor is introduced. Normalization happens only when the user explicitly resets a binding. Expression-bearing nodes are excluded from canvas text editing because the canvas editor cannot represent bindings safely. Literal descendants in mixed trees remain canvas-editable, while bindings remain editable through Settings.

## Implementation

- Reuse the existing scalar Text content and binding UI.
- Target expression children with indexed `instances.updateText` mutations, preserving instance boundaries and unrelated siblings.
- Fail closed when the selected instance is missing or no longer has a safe text target.
- Evaluate the targeted expression through the binding control's existing scope.
- Reset every expression on the selected instance atomically, convert evaluated values to text, and coalesce adjacent text nodes without crossing nested-instance boundaries.
- Centralize expression-free canvas text-editor targeting in the project-build lookup module.
- Route double-click, Enter, post-insert editing, cursor generation, and keyboard navigation through the shared lookup rules.
- Exclude single-expression and mixed/nested expression targets from the canvas editor in every mode.
- Preserve canvas editing for literal descendants inside otherwise mixed content.
- Share canvas fixtures and simplify binding-removal capability to an optional handler.

## Verification

- [x] Settings exposes and edits mixed text/expression targets without replacing siblings.
- [x] Reset binding converts every expression on the instance and normalizes adjacent text.
- [x] Missing and unsupported instances fail closed.
- [x] Double-click and Enter do not open bound nodes in the canvas editor.
- [x] Literal descendants within mixed trees remain canvas-editable.
- [x] Complete Builder suite passes (2,497 tests).
- [x] Complete project-build suite passes (2,070 tests).
- [x] Builder and project-build typechecks pass.
- [x] Repository lint passes.
- [x] Package-boundary validation passes.
- [x] Generated API files are clean.
- [x] Fixture inputs and outputs are unaffected.
- [x] Documentation impact reviewed: no documentation update is needed.
- [ ] Interactive browser smoke test was unavailable because the in-app browser failed to initialize.
- [ ] Agent evaluations were not run because they require separate approval.

## CI note

The previous head's Main workflow failed while pulling the E2E database image with Docker Hub `toomanyrequests: Rate exceeded`. Lost Pixel failed separately with a certificate error. Both are external infrastructure failures; branch formatting, lint, package boundaries, generated contracts/docs, tests, and typechecks were successful in that run.


kof:
Follow-up `237d66b791` fixes the disappearing-canvas/caret state. The double-click target can be promoted to an outer rich-text root; checking only the clicked instance missed expressions nested elsewhere in that proposed editing tree. The guard now traverses the full root subtree and prevents Lexical from starting whenever it contains an expression, selecting the clicked instance instead. Added a fail-before/pass-after regression for that promoted-root case.

PR #6041, opened 2026-08-07
fix: Remove hardcoded template allowlist and require authorization tokens

Kathircpe:
## Description

Removes the hardcoded template-project allowlist from `checkProjectPermit`. Marketplace projects now use their existing viewer authorization tokens, matching how the marketplace passes `authToken` through builder URLs and the builder loader.

Token authorization is also scoped to the requested project, so a valid token for one project cannot grant access to another project.

## Changes

- Removed the 15-project template allowlist and its unconditional view grant.
- Scoped `AuthorizationToken` permit queries by `projectId`.
- Added regression coverage for former template IDs, valid viewer tokens, missing tokens, and cross-project token use.
- Removed the proposed token backfill and unrelated Prisma schema drift change; existing marketplace tokens remain the source of access.

## Verification

- [x] Demonstrated the cross-project regression test fails before the project constraint.
- [x] `pnpm --filter @webstudio-is/trpc-interface test -- src/authorize/project.server.test.ts`
- [x] `pnpm --filter @webstudio-is/trpc-interface typecheck`
- [x] Reviewed documentation impact; no documentation changes are required.
- [x] Verified fixture inputs and outputs are not affected.

## Before merging

- [x] Made a self-review.
- [x] Added regression tests.
- [x] Confirmed no environment variables are added.
- [ ] Confirm GitHub checks pass.


packages/prisma-client/prisma/migrations/20260807000000_template-view-tokens/migration.sql
```diff
@@ -0,0 +1,50 @@
+-- Backfill a 'viewers' AuthorizationToken for each legacy template project
```
kof: Please remove this migration. These marketplace projects already have viewer tokens, and `ApprovedMarketplaceProduct` already selects and exposes those tokens. This copies the deleted hardcoded allowlist into a permanent data migration without evidence that any records are missing. The PR should simply remove the special authorization path and rely on the existing marketplace tokens.

kof (changes requested):
The marketplace projects already have viewer authorization tokens, and `ApprovedMarketplaceProduct` exposes those tokens to marketplace links. This PR should remove the exceptional authorization path and rely on those existing tokens instead of moving the same project allowlist into a permanent data migration.

There is also a project-scoping issue in the token authorization query: `checkProjectPermit` filters by token and relation but not by `projectId`. A marketplace token can therefore authorize the same permit for an unrelated project. Please add `.eq("projectId", input.id)` and a regression test proving that a valid token for project A is denied for project B.

The `canPublish` Prisma schema drift correction is unrelated to removing the template allowlist and should be handled separately unless this PR demonstrates that it is required for this change.

kof (approved):
Updated the PR to remove the unnecessary token backfill and unrelated Prisma schema change. Token authorization is now scoped to the requested project, with a regression test that was verified failing before the fix and passing afterward.

kof:
thank you

PR #6063, opened 2026-08-08
feat: stream s3 upload

Kathircpe:
## Problem:
 uploadToS3 read the entire file into memory before PUTting it to S3 — a spike for large uploads (notably videos), flagged in an existing todo.

## Why not naive streaming:
 S3 rejects plain Transfer-Encoding: chunked and its streaming-chunked mode requires the total size up front; our incoming streams have unknown length. Multipart and @aws-sdk/lib-storage were rejected — multipart buffers sub-5MB uploads (the whole default case) and adds 3+ signed calls, while lib-storage buffers parts in memory and breaks the package's minimal hand-rolled signer design.

## Approach:
 Spool the stream to a temp file (bounded memory, size limit aborts mid-stream, key-independent random filename), then PUT via fs.createReadStream with a known Content-Length and duplex: "half". Temp dir cleaned up in finally.

## Also fixed:
 dropped the dead body arg from signS3Request (the UNSIGNED-PAYLOAD header short-circuits hashing; passing a stream there leaked an fd per upload).

## Tests:
 unit tests (size/Content-Length, oversize abort, signer gets no body, cleanup) plus a real node:http integration test exercising the actual undici fetch + stream path.

### DIFF
--- a/packages/asset-uploader/src/clients/s3/request-headers.ts
+++ b/packages/asset-uploader/src/clients/s3/request-headers.ts
@@ -21,14 +21,12 @@ export const signS3Request = async ({
   url,
   method,
   headers,
-  body,
   query,
 }: {
   signer: SignatureV4;
   url: URL;
   method: string;
   headers: Record<string, string>;
-  body?: ArrayBuffer | Uint8Array;
   query?: Record<string, string>;
 }) =>
   await signer.sign({
@@ -38,5 +36,4 @@ export const signS3Request = async ({
     path: url.pathname,
     ...(query === undefined ? {} : { query }),
     headers: createS3SigningHeaders(url, headers),
-    ...(body === undefined ? {} : { body }),
   });
--- a/packages/asset-uploader/src/clients/s3/upload.test.ts
+++ b/packages/asset-uploader/src/clients/s3/upload.test.ts
@@ -1,0 +1,176 @@
+import { afterEach, describe, expect, test, vi } from "vitest";
+import { readdir } from "node:fs/promises";
+import { createServer, type Server } from "node:http";
+import { tmpdir } from "node:os";
+import { Readable } from "node:stream";
+import type { SignatureV4 } from "@smithy/signature-v4";
+import { AssetUploadSizeLimitError } from "../../utils/size-limiter";
+import { uploadToS3 } from "./upload";
+
+afterEach(() => vi.unstubAllGlobals());
+
+const leftoverTempFiles = async () =>
+  (await readdir(tmpdir())).filter((entry) =>
+    entry.startsWith("webstudio-asset-")
+  );
+
+describe("uploadToS3", () => {
+  test("spools the stream and uploads it with a known Content-Length", async () => {
+    const sign = vi.fn(async (request: unknown) => request);
+    const fetch = vi.fn(async () => new Response("ok", { status: 200 }));
+    vi.stubGlobal("fetch", fetch);
+
+    const data = {
+      async *[Symbol.asyncIterator]() {
+        yield new TextEncoder().encode("Post body");
+      },
+    };
+
+    const result = await uploadToS3({
+      signer: { sign } as unknown as SignatureV4,
+      name: "folder/post.md",
+      type: "text/markdown",
+      data,
+      maxSize: 1024,
+      endpoint: "https://storage.example",
+      bucket: "assets",
+      assetInfoFallback: undefined,
+    });
+
+    expect(result).toEqual({ size: 9, format: "md", meta: {} });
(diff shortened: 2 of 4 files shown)


Kathircpe:
Hey @kof , take a look at this.

kof:
## Main conceptual issues

- It shifts pressure from memory to `/tmp`; it does not eliminate buffering. Every upload is fully written before S3 receives a byte, doubling local I/O and increasing latency.
- Concurrent large uploads now compete for temporary disk capacity. Process crashes can also leave full temporary files behind.
- The default upload limit is only 4.5 MB because of Vercel’s request limit. At that size, the original memory usage is already bounded; disk spooling adds substantial machinery for a relatively small saving.
- Large uploads are precisely where serverless temporary storage, request duration, and concurrency are least dependable.
- Images and fonts are read fully back into memory for metadata, so the guarantee is type-dependent.
- More importantly, the API/CLI route already calls `arrayBuffer()` before invoking the uploader to calculate the content hash. For that path, this PR keeps the original memory spike and then adds a disk copy. See the [upload route](/Users/oleg/work/webstudio/webstudio-builder/apps/builder/app/routes/rest.assets.uploads_.$name.tsx:64).
- The abstraction discards useful information. `AssetObjectWriter` receives only an `AsyncIterable`, even though browser `File` uploads generally have a known size. The uploader is forced to rediscover the size by consuming the entire stream.
- The added temp-file implementation creates new operational obligations: disk quotas, cleanup after crashes, monitoring, permissions, and platform compatibility.

The PR is correct that AWS’s SigV4 chunked encoding does not solve unknown-size uploads: AWS still requires `x-amz-decoded-content-length`. [[AWS SigV4 streaming documentation](https://docs.aws.amazon.com/AmazonS3/latest/developerguide/sigv4-streaming.html)](https://docs.aws.amazon.com/AmazonS3/latest/developerguide/sigv4-streaming.html)

## Better incremental solution

Change the upload contract to preserve a trusted optional size:

```ts
type UploadSource = {
  data: AsyncIterable<Uint8Array>;
  contentLength?: number;
};
```

Then:

1. If `contentLength` is known:

   - Reject it immediately when it exceeds `maxSize`.
   - Stream directly to a single S3 `PutObject`.
   - Keep a counting transform to verify the actual byte count and enforce the limit.
   - Set `Content-Length` from the supplied size.

2. If the size is unknown:

   - Use disk spooling only as a fallback initially.
   - Keep it isolated behind a deliberate `spoolUnknownLengthUpload` path.
   - Add disk-capacity limits, observable cleanup failures, and stale-file cleanup.

This would avoid spooling for the normal browser-file case while retaining compatibility with unknown-length URL streams.

kof:
@Kathircpe please next time start with explaining "why" you starter working on it, what was the issue.

kof:
## Best architecture for browser uploads

For large browser files, avoid proxying the bytes through Builder:

1. Builder authorizes the upload and creates an upload ticket.
2. Browser uploads directly to S3 using a presigned PUT or multipart URLs.
3. Builder verifies/completes the upload and records the asset metadata.
4. Failed multipart uploads are aborted or expired by lifecycle policy.

A browser `File` already exposes its exact size, so this removes Builder memory, temporary disk, bandwidth, and timeout pressure altogether.

Server-originated imports with genuinely unknown sizes can use server-side multipart upload.

----

Generally I think we need to research a solution that's battle tested and not add another workaround. Meaning a library that implements the above approach well.

Kathircpe:
> @Kathircpe please next time start with explaining "why" you starter working on it, what was the issue.

@kof actually I did not face any issue. I am just picking todo comments as tasks to work on it.
If you want me to work on something, I am open to discuss about it.

kof:
> @kof actually I did not face any issue. I am just picking todo comments as tasks to work on it.
If you want me to work on something, I am open to discuss about it.

Please don't do that. Any change is a risk of breaking something else and creates more instability, if its not broken - don't fix it.

kof:
We have lots of issues, any issue must be verified by a human and carefully tested. If oyu just want to talk to agent and create prs - that's just creating more work for us.

kof:
Real contribution is making sure stuff gets better and works better, not generating code.

Kathircpe:
I get it. I thought the todo comments are need to be done.

Kathircpe:
I’d love to continue contributing—any on your backlog I can pick up today?

kof:
@Kathircpe only if oyu actually understand the changes you are making and testing them carefully. 

I think fixing bugs is a good place, because they must be reproducible first and don't require much brainstorming usually, you can sort by the label bug

https://github.com/webstudio-is/webstudio/issues?q=is%3Aissue%20state%3Aopen%20label%3Atype%3Abug

PR #6074, opened 2026-08-08
fix: Allow removal of invalid audited selectors

### LINKED ISSUE #6062: fix: Allow removal of an invalid selector reported by the style audit
## User story

As a Webstudio CLI user, I want to remediate an invalid style selector identified by the built-in audit using the semantic deletion tool.

## Summary

The style audit reports an invalid vendor pseudo-element selector, but delete-design-token-styles rejects the same selector during input validation, making the audited declaration impossible to remove semantically.

## What the agent tried

1. Run the styles audit and identify the single invalid-style-state-selector warning.
2. Inspect the unused design token and confirm the exact breakpoint, state selector, property, and value.
3. Call delete-design-token-styles in dry-run mode with the exact audited declaration coordinates.

## Expected behavior

The deletion tool should accept the existing declaration coordinates and return a destructive plan that can be reviewed and confirmed.

## Actual result

The tool returned INVALID_INPUT because its state validator rejects the audited pseudo-element as invalid.

## Recovery attempts

1. Verified that the design token and declaration exist in the current project session.
2. Used the exact property, breakpoint, and state selector returned by the audit.

## User impact

The user cannot clear a built-in style audit warning with the supported semantic tool and would need an unsafe raw patch or unrelated destructive token deletion.

## Technical context

Webstudio CLI 0.288.0. The mismatch is between audit rule invalid-style-state-selector and delete-design-token-styles validation of the deletions.state field.

## Agent environment

- Client: Codex
- Provider: OpenAI
- Model: GPT-5
- Reasoning effort: unknown
- Trigger: automatic-friction
- Category: schema-or-docs-mismatch

## Technical runtime

- CLI: 0.288.0
- Node.js: 24.12.0
- Operating system: win32 10 (x64)
- Execution mode: mcp
- API contract: public-api:iqd6ok

## Acceptance criteria

1. delete-design-token-styles can identify and remove an existing invalid selector without accepting arbitrary new invalid selectors.
2. A regression test audits an invalid selector, deletes the reported declaration, and confirms that the warning is cleared.
3. The exposed schema or error explains the supported remediation path when an existing invalid state cannot be represented.

kof:
## Summary

Allow all targeted style-deletion paths to remove an invalid state selector that already exists, including Builder UI deletion of local and token declarations and CLI/MCP design-token remediation. Creation and update validation remain strict.

## Root cause

The style audit reads legacy declarations directly and can report selectors that the current selector validator rejects. Deletion reused the creation/update validator, so the system could identify an invalid stored declaration but could not represent its coordinates for targeted removal. The first fix addressed only design-token deletion, leaving Builder UI and the generic style APIs inconsistent.

## Implementation

- [x] Give style deletions a shared deletion-only state-selector contract.
- [x] Accept an invalid selector only when the exact target, breakpoint, property, and state already exist.
- [x] Reject arbitrary invalid selectors that do not identify an existing declaration.
- [x] Reuse one matcher for local, selected/token, and design-token deletions.
- [x] Keep creation and update selector validation strict.
- [x] Add Builder UI regression coverage for token and local declaration deletion.
- [x] Add an audit regression that deletes the reported design-token selector, applies the patch, and confirms the warning is cleared.
- [x] Regenerate the public runtime operation contract.
- [x] Complete the DRY review: deletion matching, validation, key deduplication, and stored-style scanning are centralized.
- [x] Review documentation impact: no authored documentation changes are required; generated operation contracts describe the behavior.
- [x] Review fixture impact: no fixture inputs or generated fixture bundles are affected.
- [x] Run relevant tests, typechecks, lint, formatting, package-boundary, generated-API, and generated-documentation checks.
- [ ] Agent evaluations — not run because repository policy requires separate explicit approval.

## Verification

- The two Builder UI regression tests failed with `Operation input is invalid` before the shared fix and passed afterward.
- Project-build suite: 82 files, 2,097 tests passed.
- CLI suite: 60 files, 968 tests passed.
- Builder style-data target: 7 tests passed.
- Audit and styles targets: 195 tests passed.
- Project-build, Builder, and CLI typechecks passed.
- Full repository lint passed.
- Package-boundary, generated-API, and generated-documentation checks passed.
- Formatting and `git diff --check` passed.

Closes #6062

### DIFF
--- a/packages/project-build/src/runtime/styles.ts
+++ b/packages/project-build/src/runtime/styles.ts
@@ -2776,6 +2776,20 @@ export const createDesignTokenStyleUpdatePayload = ({
   };
 };
 
+const getDesignTokenStyleDeletionKey = ({
+  designTokenId,
+  deletion,
+}: {
+  designTokenId: StyleSource["id"];
+  deletion: z.infer<typeof designTokenStyleDeletionInput>;
+}) =>
+  getStyleDeclKeyFromInput({
+    styleSourceId: designTokenId,
+    breakpoint: deletion.breakpoint,
+    state: deletion.state,
+    property: deletion.property,
+  });
+
 export const createDesignTokenStyleDeletePayload = ({
   designTokenId,
   deletions,
@@ -2789,12 +2803,7 @@ export const createDesignTokenStyleDeletePayload = ({
     Array.from(styles, (styleDecl) => getStyleDeclKey(styleDecl))
   );
   const styleKeys = deletions.flatMap((deletion) => {
-    const key = getStyleDeclKeyFromInput({
-      styleSourceId: designTokenId,
-      breakpoint: deletion.breakpoint,
-      state: deletion.state,
-      property: deletion.property,
-    });
+    const key = getDesignTokenStyleDeletionKey({ designTokenId, deletion });
     return existingStyleKeys.has(key) ? [key] : [];
   });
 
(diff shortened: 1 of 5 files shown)


packages/project-build/src/runtime/styles.ts
```diff
@@ -2765,25 +2776,34 @@ export const createDesignTokenStyleUpdatePayload = ({
   };
 };
 
+const getDesignTokenStyleDeletionKey = ({
```
kof: this function seems useless

PR #6079, opened 2026-08-09
feat: move authorization token role permissions to SQL

Kathircpe:
## Description

Move authorization token role→permission mapping into SQL

The role→permission mapping for AuthorizationToken previously lived in TS (applyTokenPermissions), recomputed on every read and at risk of diverging from the intended SQL behavior. This PR codifies it in the database as the single source of truth.

## Code Review

  - test it on preview

## Before requesting a review

- [x] made a self-review

## Before merging

- [X] added tests
- [X] if any new env variables are added, added them to `.env` file

### DIFF
--- a/packages/authorization-token/src/db/authorization-token.test.ts
+++ b/packages/authorization-token/src/db/authorization-token.test.ts
@@ -14,79 +14,12 @@ import {
   create,
   update,
   remove,
-  __testing__,
 } from "./authorization-token";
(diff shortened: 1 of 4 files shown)


kof:
What's the actual problem you faced that needed this fix?

Kathircpe:
> What's the actual problem you faced that needed this fix?

The todo comment asked to implement it on the sql level.

Kathircpe:
It may be closed, since there is not a significant problem.

PR #6101, opened 2026-08-11
fix: Preserve SSG template file URL

### LINKED ISSUE #6100: SSG export fails while configuring the asset resource fetch template
## Problem

SSG export fails during project scaffolding when the published project contains an Assets query.

The CLI throws `TypeError [ERR_INVALID_URL_SCHEME]: The URL must be of scheme file` from `configureSsgAssetResourceFetch` while reading the SSG asset resource fetch template.

## Desired outcome

SSG projects with Assets queries scaffold and build successfully from the bundled CLI.

## Scope

- Preserve the SSG asset resource fetch template as a filesystem URL in the bundled CLI.
- Add regression coverage for the bundled output.
- Keep behavior for projects without an Assets query unchanged.

## Acceptance criteria

- The bundled CLI resolves the template from the installed package rather than an inlined `data:` URL.
- SSG prebuild can copy the asset resource fetch template when Assets queries are enabled.
- Relevant CLI tests, typecheck, and build pass.

## Constraints

No product behavior, schema, or migration changes are expected.

kof:
## What changed

- Read the SSG asset resource fetch template through a package-relative file URL that Vite preserves in the CLI bundle.
- Add a focused bundling regression test that packages the real template reader and verifies it can read the installed template.

## Why

Vite recognized the static `new URL("../templates/ssg/app/asset-resource-fetch.ts", import.meta.url)` expression as an asset import. Because the template is small, Vite inlined it as a `data:video/mp2t` URL. During an SSG export with an Assets query, Node passed that URL to `readFile` and threw `ERR_INVALID_URL_SCHEME`.

The template URL now uses Vite's preserve annotation, matching the CLI's other package-relative runtime files, so the bundled CLI resolves the template from the installed package.

## Impact

SSG exports that use Assets queries can scaffold the generated asset resource fetch module. Projects without Assets queries keep the existing fallback behavior.

## Todo

- [x] Reproduce the invalid URL scheme through a bundled template reader.
- [x] Preserve the package-relative template URL in the production CLI bundle.
- [x] Add and verify regression coverage.
- [x] Run CLI tests, typecheck, lint, and build.
- [x] Review documentation impact; no user-facing documentation changes are needed.
- [x] Review fixture impact; no fixture inputs or generated fixture outputs are affected.

## Verification

- `pnpm --filter webstudio test` — 61 files, 977 tests passed
- `pnpm --filter webstudio typecheck`
- `pnpm lint`
- `pnpm --filter webstudio build`
- Regression test confirmed failing with `ERR_INVALID_URL_SCHEME` before the fix and passing after it.
- Built `lib/cli.js` inspected to confirm the template remains a package-relative URL instead of a `data:` URL.

Closes #6100

### DIFF
--- a/packages/cli/src/prebuild.ts
+++ b/packages/cli/src/prebuild.ts
@@ -101,6 +101,7 @@ import { createFramework as createRemixFramework } from "./framework-remix";
 import { createFramework as createReactRouterFramework } from "./framework-react-router";
 import { createFramework as createVikeSsgFramework } from "./framework-vike-ssg";
 import { routeTemplatesDirectory } from "./framework";
+import { readSsgAssetResourceFetchTemplate } from "./ssg-asset-resource-fetch-template";
 
 export const generatedFilesManifest = join(
   ".webstudio",
@@ -111,10 +112,6 @@ const contentRuntimeBundleUrl = new URL(
   import.meta.url
 );
 const contentRuntimeFile = "$resources.asset-query-vendor.js";
-const ssgAssetResourceFetchTemplateUrl = new URL(
-  "../templates/ssg/app/asset-resource-fetch.ts",
-  import.meta.url
-);
 const appRoot = "app";
 const generatedDir = join(appRoot, "__generated__");
 const routesDir = join(appRoot, "routes");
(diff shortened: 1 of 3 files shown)


kof:
Moved to the short staging branch `ssg.staging`; continued in #6102.

PR #6247, opened 2026-08-26
chore(https): stop tracking the local-dev TLS private key

Portll:
Hi — thanks for keeping the local HTTPS helper (`https/README.md`, `https/haproxy.sh`) so easy to reproduce. While looking through the repo I noticed the private key it generates is checked in.

**What's here.** `https/haproxy.pem` and `https/privkey.pem` both contain the same live TLS private key, and it matches the committed `wstd.dev` / `*.wstd.dev` certificate. The certificate is a genuine, publicly-trusted Let's Encrypt leaf, so secret scanners (e.g. TruffleHog's `PrivateKey` detector) flag these files as a verified private key rather than as example data.

**SEV: Low** This looks intentional and the blast radius is small: `wstd.dev` resolves to `127.0.0.1`, so the certificate only ever fronts a service on the developer's own loopback. Holding the key lets someone present a browser-trusted cert *for a name that already points at localhost* — there is no external host to intercept. The current leaf is also near expiry, so it ages out on its own shortly.

**Why it's still worth fixing.**
- A committed private key trips secret scanners on every clone and in CI, which is noise for contributors and for anyone running the repo through security tooling.
- The regeneration steps are already fully documented in `https/README.md` and `https/haproxy.sh`, so the key does not need to live in git for reproduction.
- `.gitignore` already ignores `*.pem` globally; these two files were only tracked because of the explicit `!https/*.pem` re-include. Narrowing that exception keeps the *public* cert convenient while dropping the private key.

**What this PR does.**
1. `git rm --cached https/haproxy.pem https/privkey.pem` (untrack; local copies left in place so nobody's dev setup breaks).
2. Narrows `.gitignore`'s pem re-include to the public chain only (`!https/fullchain.pem`), so `haproxy.pem` / `privkey.pem` stay ignored going forward.
3. Adds a one-line note to `https/README.md` that those two files are generated locally and intentionally untracked.

**Optional, at your discretion.** Because the key did exist in public git history, the tidy follow-up is to let the current Let's Encrypt leaf expire (or revoke + reissue) rather than reuse it — given the loopback-only exposure this is housekeeping.

Happy to adjust the approach or split this up if you'd prefer. Thanks for the project!


kof:
Thanks for raising this. The underlying concern is valid: a private key for a publicly trusted certificate should not be published. However, this fix cannot be merged as written because it makes a clean checkout unusable.

The Builder reads `https/privkey.pem` synchronously from `apps/builder/vite.config.ts`, and the E2E server reads the same file from `apps/builder/e2e/serve-built-remix.ts`. Once the file is untracked:

- `pnpm dev` fails in a fresh clone.
- `pnpm e2e:builder:build` / Builder E2E fail in CI; the current builder-e2e build already exits with code 1.
- Existing local copies only protect existing worktrees, not new contributors or CI checkouts.
- The regeneration instructions are marked “Admin only” and require Webstudio Infisical and Cloudflare access, so they are not a replacement for the public contributor workflow.

Removing the files from the latest tree also does not remove the exposed key from Git history. History-scanning tools may continue to report it, and the current certificate remains compromised until it expires or is revoked.

A complete solution needs to replace the certificate provisioning path before untracking the key. Reasonable options are:

1. Generate a per-developer certificate from a local CA (for example, via `mkcert`) in an idempotent setup command, with documented one-time trust installation. Generate an ephemeral certificate for CI and configure the test browser/runtime to trust that CA or accept that test certificate.
2. Provision the certificate and key securely from secrets in both development and CI. This would also need a separate, usable path for external contributors who do not have Webstudio secret access.
3. Move local development to a hostname/certificate strategy that does not require sharing a publicly trusted private key. This is a larger change but removes the recurring certificate rotation and disclosure problem.

Whichever route is chosen, revoke or allow the current certificate to expire, issue the replacement with a new key, and do not commit that new key. Historical scanner findings should be handled separately—by an explicit allowlist for expired historical development keys or, only if justified, a coordinated history rewrite.

Portll:
Absolutely fair, and the exposure risk is minimal with a key that only really validates localhost and expires tomorrow. I decided 1-3 were out of scope for me and that the most important thing to do was raise the key issue. My preferred option as a dev is usually mkcert, but that requires an additional installation into the devenv, #2 was out of my achieveable scope (access), and #3 is my preference for webstudio longer term, but is a bigger dev decision, hence I didn't want to provide #1 and call it 'solved'.

Thanks for your prompt response.

Issue #3818, opened 2024-07-29
Redesign all colors in design system


kof:
The objective is:

1. reduce amount of colors by using relative colors with oklch
2. enable dark/light mode switch

Karan-Munjani:
Heyy @kof 
I'm a Frontend Developer turned **UI/UX Designer with 4 years of experience** (1 year in development, 3 years in design).

Throughout my design journey, I've always believed that the future would bring a super tool to enhance the frontend developer experience—but I never imagined it would be open-source.

**It would be amazing to get access to a Figma file or any existing design assets as a starting point. Also we can connect to dive deeper into the problem statement together.**

PR #6257, opened 2026-08-28
feat: add Builder dark theme and rebuild colors

### LINKED ISSUE #6267: Unify design-system and Builder browser testing
## Problem

Webstudio currently uses three overlapping test setups: Vitest with jsdom, Vitest Browser Mode, and a custom Builder E2E runner built directly on Playwright. The split duplicates browser infrastructure, makes test placement unclear, and requires maintaining execution, filtering, sharding, timeout, retry, and lifecycle behavior outside the standard runners.

## Desired outcome

Use two standard, clearly bounded test systems:

- Vitest Node for pure logic and Vitest Browser Mode for design-system components and browser behavior.
- Playwright Test for complete Builder E2E workflows.

Remove design-system jsdom and the custom Builder E2E runner rather than retaining compatibility layers.

## Scope

- Classify and migrate all design-system tests to Node or Browser Mode.
- Replace the custom Builder E2E registration and execution APIs with Playwright Test configuration and fixtures.
- Preserve database setup/reset, Builder startup, local host mapping, permissions, focused local runs, timeouts, retries, sharding, and performance reporting.
- Update package scripts, CI workflows, and testing documentation.
- Remove obsolete dependencies and infrastructure after migration.

## Constraints

- Preserve existing test intent and coverage.
- Keep pure tests fast and independent of a browser.
- Use normal Playwright Test concepts instead of recreating a second runner inside fixtures.
- Leave no custom-runner compatibility path behind.

## Acceptance criteria

- The design-system test configuration and dependencies no longer use jsdom.
- Pure design-system tests run in Vitest Node; component and browser-dependent tests run in Vitest Browser Mode.
- All Builder E2E tests run through Playwright Test.
- The previous custom suite registry and runner are removed.
- Existing database, server, domain mapping, browser permissions, filtering, sharding, retry, and local-development workflows have standard equivalents.
- CI runs the migrated suites and retains useful failure artifacts.
- Testing documentation describes the final commands and test-placement rules.
- Relevant tests, typechecks, formatting, and CI configuration validation pass.

kof:
## Summary

Rebuilds Webstudio’s color system, migrates the entire Builder to it, and introduces a complete dark theme.

A small set of theme colors and contrast controls now drives the interface through relative CSS colors:

```text
Theme colors and contrast
    ↓
Light/dark derivation bounds
    ↓
Derived color families
    ↓
Reusable semantic colors
    ↓
Component composition
    ↓
Builder interface
```

CSS is the runtime source of truth. Light and dark modes share the same theme inputs, so changing a theme color automatically updates its surfaces, content, borders, states, and component presentation. The refactor turns dark mode into a first-class Builder theme rather than a separately maintained palette.

This is an atomic migration. Legacy color names and compatibility aliases will not remain after all consumers have moved.

## Scope

- The color source, derivation model, semantic vocabulary, validation, typing, and Storybook overview.
- Every design-system component.
- Builder, dashboard, authentication, shared UI, Query Builder, stories, and tests that consume the legacy color scale.
- Removal of the legacy Stitches color scale after its final consumer is migrated.
- A complete dark theme across application UI while keeping authored canvas content independent.
- Complete light and dark visual verification.
- Unified browser-test infrastructure: Vitest Browser Mode for Design System browser contracts and standard Playwright Test for all Builder E2E workflows.

Literal colors remain only where color is the actual content or belongs to a deliberately separate system, such as user-authored values, external previews, brand artwork, data visualization, functional canvas overlays, and source assets. Every remaining exception must have an explicit owner.

## Architecture decisions

- Theme authors control six color families and three bounded contrast controls.
- Relative CSS colors derive the rest of the system.
- Dark mode changes derivation bounds instead of maintaining a second independent palette.
- Components consume a small reusable semantic vocabulary.
- Component states and exceptional presentation are composed inside components.
- Builder-specific categorical meaning uses focused composite recipes, not intent semantics or a larger global vocabulary.
- TypeScript validates public semantic CSS-variable names but does not generate runtime colors.
- No token-manager dependency, compatibility aliases, or component-specific public color tokens.
- A new semantic or theme variable is added only when existing semantics and local composition cannot represent a reusable cross-feature decision.

### Color-system foundation

- [x] Establish direct CSS as the runtime source of truth.
- [x] Define bounded theme colors and contrast controls.
- [x] Derive light and dark modes from the same theme inputs.
- [x] Reduce the semantic inventory to reusable decisions and explicit accessibility pairs.
- [x] Generate only TypeScript names for public semantic CSS variables.
- [x] Validate layers, references, cycles, consumption, naming, generated types, and browser-computed contrast.
- [x] Add the Storybook color-system overview and theme controls.
- [x] Document the architecture and its relationship to the public Craft specification.

### Design-system migration

- [x] Replace every active legacy color consumer in design-system components and Storybook chrome.
- [x] Compose component states locally without component-specific public tokens.
- [x] Preserve intentional preview and sample-content colors.
- [x] Add complete positive, negative, warning, informative, and accent accessibility pairings.
- [x] Validate light and dark scheme-bound references independently.
- [x] Verify the design-system package with typecheck, lint, formatting, Storybook, and 222 tests.

### Inventory and contract

- [x] Inventory all 319 legacy references across 107 Builder, Query Builder, story, test, and design-system files.
- [x] Classify every reference as a direct semantic replacement, local composite recipe, documented literal exception, or removable code.
- [x] Record the intentional authored-content, external-preview, functional-preview, data-visualization, brand, and source-asset exceptions.
- [x] Treat the public semantic vocabulary as closed by default and justify any proposed addition by cross-feature reuse and accessibility need.
- [x] Keep Builder code limited to typed public variables accepted by `cssVar()`.

### Theme infrastructure

- [x] Add `system` (default), `light`, and `dark` application preferences.
- [x] Persist the preference in existing client settings and resolve `system` through `matchMedia`.
- [x] Apply the resolved scheme before paint and respond to operating-system changes.
- [x] Expose the persisted theme control under **View → Color scheme**.
- [x] Keep authored canvas documents independent from the application theme.
- [x] Ensure portals, overlays, embedded surfaces, and third-party controls receive the correct interface scheme.

### Generic interface migration

- [x] Migrate shared dialogs, project settings, notifications, controls, and Query Builder.
- [x] Migrate dashboard, authentication, publishing, and shared application UI.
- [x] Migrate the Builder loading state, topbar, sidebars, navigator, workspace, footer, settings, pages, alerts, and menus.
- [x] Replace fixed interface literals alongside their owning surfaces.
- [x] Remove the old scrollbar variable and hardcoded loading color.

### Style panel and style-source recipes

- [x] Migrate ordinary style-panel inputs, labels, layouts, spacing, alignment, menus, property states, and value editors.
- [x] Model style-source provenance, kind, interaction, and error as separate axes.
- [x] Compose categorical style-source colors from neutral and accent-relative colors.
- [x] Use negative semantics only for actual errors, not overwritten provenance.
- [x] Reuse one feature-owned recipe across style-source controls, badges, menus, and values without adding global tokens.

### Canvas and editors

- [x] Migrate canvas selection, reusable outlines, grid guides, resize handles, labels, badges, and overlays.
- [x] Keep collaborator colors as data visualization while guaranteeing readable labels.
- [x] Prevent Builder theme values from leaking into authored content.
- [x] Migrate editor surfaces, selections, autocomplete, lint states, and tooltips.
- [x] Derive syntax visualization from neutral and accent-relative colors; reserve intent colors for diagnostics.

### Explicit color domains

- [x] Isolate external search and social previews from the application scheme.
- [x] Preserve user-authored CSS values, functional color previews, and checkerboards.
- [x] Review brand surfaces independently in both schemes.
- [x] Document every retained literal in the exception ledger.

### Stories, tests, and legacy removal

- [x] Migrate all affected Builder stories and tests.
- [x] Replace the design-system test that intentionally references the legacy scale.
- [x] Add theme-preference and scheme-resolution behavior tests.
- [x] Remove the legacy `design-tokens.color` registry.
- [x] Remove the Stitches `colors` scale and `rawTheme.colors`.
- [x] Remove color-name conversion code that exists only for the legacy scale.
- [x] Add a focused repository check that rejects legacy references.
- [x] Confirm no compatibility mappings or deprecated aliases remain.

### Browser-test infrastructure

- [x] Run pure Design System tests in Node and browser-dependent tests in Vitest Browser Mode with Playwright Chromium.
- [x] Migrate all 79 Builder E2E workflows from the custom harness to standard Playwright Test.
- [x] Replace filename-based shard discovery with standard Playwright sharding.
- [x] Preserve clean-backend shard retries and upload Playwright failure artifacts in CI.
- [x] Remove the custom runner, registry, and timeout infrastructure.

### Verification and documentation

- [x] Compare light-mode visual regressions directly with `main`; review every difference rather than accepting snapshots wholesale.
- [x] Add representative dark coverage for dashboard, Builder shell, style panel, dialogs, editors, canvas tools, and status states.
- [x] Test default and extreme custom-theme presets.
- [x] Verify semantic and categorical-recipe contrast.
- [x] Test first paint, persistence, system changes, portals, and canvas isolation.
- [x] Run affected tests, typechecks, lint, formatting, and Storybook checks.
- [x] Review fixture impact; no fixture inputs or generated outputs are affected.
- [x] Review authoritative `docs/` content and update it where needed.
- [x] Review the final diff and make this checklist reflect delivered scope.

## Current verification

- Design-system typecheck and generated-type freshness: passed.
- Design-system tests: 222 passed, including the full theme-parameter contrast sweep and browser component coverage.
- Builder typecheck and production build: passed.
- Theme preference and scheme-resolution tests: 10 passed.
- Scoped Oxlint, formatting, and diff checks: passed.
- Design-system Storybook build and representative light/dark visual review: passed.
- Builder E2E: 80/80 Playwright entries passed (setup plus 79 workflows).

Ref #3818
Closes #6267


packages/design-system/src/components/tooltip.tsx
```diff
@@ -13,6 +13,7 @@ import { Box } from "./box";
 import { Text } from "./text";
 import type { CSS } from "../stitches.config";
 import { theme } from "../stitches.config";
+import { cssVar } from "../colors/css-var";
```
kof: why is this in colors, its not only for colors

packages/design-system/src/design-tokens.ts
```diff
@@ -1,5 +1,164 @@
 // Design tokens used by the Webstudio design system.
 
+export const color = {
```
kof: not sure if the changes in this file were actually needed

packages/design-system/package.json
```diff
@@ -6,14 +6,19 @@
   "homepage": "https://webstudio.is",
   "type": "module",
   "scripts": {
+    "generate:color-types": "tsx src/colors/generate-color-types.ts",
+    "check:colors": "tsx src/colors/generate-color-types.ts --check",
```
kof: why do we need a separate check? shouldn't regular typescript compiler already do that check?

packages/design-system/package.json
```diff
@@ -6,14 +6,19 @@
...
     "typecheck": "tsc --noEmit -p tsconfig.typecheck.json",
     "test": "vitest run"
   },
   "devDependencies": {
+    "@babel/parser": "^7.29.7",
```
kof, line 14: why do we need babel here?

packages/design-system/package.json
```diff
@@ -6,14 +6,19 @@
...
     "@types/react": "^18.2.70",
     "@types/react-dom": "^18.2.25",
     "@webstudio-is/image": "workspace:*",
     "@webstudio-is/tsconfig": "workspace:*",
+    "colorjs.io": "^0.6.1",
+    "css-tree": "^3.2.1",
```
kof: we might already have needed utils in the css-engine package

packages/design-system/src/design-tokens.ts
kof: Agreed. Reverted the file completely; `design-tokens.ts` now has no diff in the PR.

packages/design-system/package.json
kof: It was only needed by the removed component source scanner. The scanner and direct Babel dependency are both gone.

packages/design-system/package.json
kof: Yes. Removed the separate component source scan. Normal TypeScript typechecking now validates names passed to `cssVar()`, while `check:colors` only validates the CSS source and generated declaration freshness.

packages/design-system/src/components/tooltip.tsx
kof: Agreed. Moved the generic helper to `src/css-var.ts`; component imports no longer go through the colors directory.

packages/design-system/package.json
kof: Checked `css-engine` and `css-data`. `css-engine` exposes color/value helpers but not a stylesheet AST utility, and it already has a development dependency on the design system. Importing it here would add coupling without replacing the parser, so source validation keeps the direct `css-tree` dependency. The separate `colorjs.io` dependency was removed.

packages/design-system/package.json
```diff
@@ -6,14 +6,19 @@
...
     "test": "vitest run"
   },
   "devDependencies": {
     "@types/react": "^18.2.70",
     "@types/react-dom": "^18.2.25",
+    "@vitest/browser": "^3.1.2",
     "@webstudio-is/image": "workspace:*",
     "@webstudio-is/tsconfig": "workspace:*",
+    "css-tree": "^3.2.1",
+    "playwright": "^1.50.1",
```
kof: why do we use it?

packages/design-system/package.json
kof: who is gonna use it and when?

packages/design-system/package.json
kof: `css-tree` is used by `color-source-utils.ts` to parse the complete stylesheet and nested `var()` values for type generation, structural tests, and Storybook metadata. Browser CSSOM is unavailable to the Node generator, and `css-data` does not export a full stylesheet AST utility. Keeping this as a direct dev dependency avoids regex/string parsing and does not add it to the design-system runtime bundle.

packages/design-system/package.json
kof: Good catch: the standalone command was not wired into the normal workflow. I removed `check:colors`. Generated declaration freshness now runs as part of the design-system `typecheck`, which is invoked by the repository's normal `pnpm -r typecheck` CI step; structural validation remains in the normal package tests.

packages/icons/generate.ts
```diff
@@ -19,7 +19,45 @@ const getHashedPrefix = (path?: string) => {
   return hash;
```
kof, line 19: all changes here are highly sus

packages/design-system/vitest.config.ts
```diff
@@ -10,7 +10,33 @@ export default defineConfig({
     },
   },
   test: {
-    environment: "jsdom",
```
kof, line 13: did you actualy have a very good reason for switching from jsdom to playwright? it will def. make tests slower

packages/design-system/src/css-var.ts
```diff
@@ -0,0 +1,6 @@
+import type { CssVariableName } from "./colors/__generated__/css-variable-names";
+
+export type { ThemeVariableName } from "./colors/__generated__/css-variable-names";
```
kof: these are aliases, we shouldn't do aliases, always keep the same name everywhere, pick a good one

packages/design-system/src/color-utils.ts
```diff
@@ -0,0 +1,28 @@
+import { cssVar } from "./css-var";
+
+// Selection is a component state, composed from generic theme colors instead
+// of being exposed as another theme parameter.
+export const selectionBackground = `light-dark(
```
kof: that doesn't sound like a utility but rather a component-specific style

packages/design-system/src/components/toolbar.tsx
```diff
@@ -9,12 +9,18 @@ import { separatorStyle } from "./separator";
 import { textVariants } from "./text";
 import { forwardRef, type Ref } from "react";
 import { focusRingStyle } from "./focus-ring";
+import { cssVar } from "../css-var";
+
+const toolbarBackground = cssVar("--background-primary");
+const toolbarForeground = cssVar("--foreground-primary");
+const toolbarForegroundSubtle = cssVar("--foreground-secondary");
+const toolbarBorder = cssVar("--border-default");
+const toolbarHover = cssVar("--overlay-interaction-hover");
```
kof: these js variables are probably not needed, cssVar call is not expensive

packages/design-system/src/components/toast.tsx
```diff
@@ -71,46 +72,44 @@ const AnimatedToast = styled(ToastPrimitive.Root, {
...
 });
 
-const borderAccentBackgroundColor = "--ws-toast-border-accent-background-color";
-const backgroundColor = "--ws-toast-background-color";
-const borderColor = "--ws-toast-border-color";
-const iconColor = "--ws-toast-icon-color";
+const borderAccentBackgroundColor = "--toast-border-accent-background-color";
+const backgroundColor = "--toast-background-color";
+const borderColor = "--toast-border-color";
+const iconColor = "--toast-icon-color";
```
kof: are these js variables actually needed? and also why is it not using cssVar() to get typechecked?

packages/design-system/src/components/select.tsx
```diff
@@ -138,8 +139,8 @@ export const SelectItemDescription = ({
 
       <SelectSeparator
         style={{
-          display: `var(--ws-select-description-display-top, none)`,
-          order: "var(--ws-select-description-order)",
+          display: `var(--select-description-display-top, none)`,
+          order: "var(--select-description-order)",
```
kof: similar concern, I would love to see if we can find a way to introduce typed css variables via cssVar across the entire system, not only global theme

packages/design-system/vitest.config.ts
kof: The package was not switched wholesale. Normal tests remain in the `unit` workspace with `environment: "jsdom"`; only files named `*.browser.{test,spec}.ts` run in Chromium. Those tests depend on the real CSS engine for `light-dark()`, relative colors, `color-mix()`, computed contrast, and hover state, which jsdom cannot validate. Keeping the split gives us fast jsdom tests for logic and browser coverage only where platform rendering is the behavior under test.

Issue #6270, opened 2026-08-30
Clean up unified browser testing infrastructure


kof:
## Problem

The recent testing-infrastructure migration standardized Design System browser tests and Builder E2E tests, but its final review found incomplete unification, avoidable Playwright lifecycle boilerplate, slow failure diagnostics, lost retry evidence, and stale local-runner documentation.

## Desired outcome

Testing should use the smallest practical set of runtimes, follow native Playwright fixture patterns, fail with actionable timing, preserve evidence from flaky first attempts, and validate filters before expensive environment setup.

## Scope

- Builder Vitest runtime ownership and remaining jsdom usage.
- Builder Playwright fixtures, timeouts, filtering, and CI retry artifacts.
- Testing documentation and small dead/duplicated helpers introduced by the migration.
- Preserve existing test behavior and clean-backend isolation.

## Acceptance criteria

- Browser-dependent tests use a real browser; pure tests use Node.
- Ordinary E2E tests use Playwright-managed fixtures instead of manual contexts.
- Missing UI actions fail before the full workflow timeout.
- Both initial and retry failure evidence remains inspectable.
- Invalid filters fail before Docker, database, browser installation, or Builder build work.
- The complete affected test, typecheck, lint, and formatting checks pass.

PR #6271, opened 2026-08-30
test: simplify browser testing infrastructure

kof:
## Summary

Unifies frontend testing around three intentional boundaries:

- Vitest Node projects for pure and server tests
- Vitest Browser Mode with Playwright/Chromium for DOM and component tests
- Playwright Test for Builder end-to-end workflows

The Builder no longer uses jsdom. Ordinary E2E workflows use an automatically managed isolated `page` fixture; explicit contexts remain only for simultaneous identities and suite-scoped setup.

## Technical choices

- Split Builder tests into one Node project and independent browser projects so each Chromium/Vite process has a bounded workload.
- Added shared Browser Mode configuration for the affected SDK React packages.
- Preserve the Builder's production top-window bridge ownership; Browser Mode uses explicit test-only installation seams.
- Centralized Playwright context options and kept action/navigation timeouts bounded.
- Validate Playwright filters before Docker, migrations, browser installation, or Builder builds start.
- Preserve the first failed E2E attempt's report, traces, screenshots, and videos when CI retries a shard.
- Split static checks, typechecks, CLI tests, core package tests, browser package tests, and Builder projects into parallel CI jobs with five-minute hard ceilings.
- Distribute Builder E2E tests across ten isolated shards with two workers; parallel suites create unique per-worker setup data.
- Reuse color contrast DOM/canvas samples so the exhaustive theme sweep takes about four seconds instead of timing out at 30 seconds.
- The removed cost is jsdom behavior drift, oversized serial check jobs, and unnecessary E2E startup for empty selections. Real browser startup and E2E environment setup remain.

## Todo

- [x] Split Builder pure/server tests from browser/component tests.
- [x] Migrate Builder DOM tests from jsdom to Vitest Browser Mode.
- [x] Migrate affected SDK React DOM tests to the shared Browser Mode setup.
- [x] Remove jsdom annotations, dependencies, and compatibility setup.
- [x] Replace ordinary manual E2E contexts with the isolated `page` fixture.
- [x] Keep explicit contexts only for concurrent identities and `beforeAll` setup.
- [x] Preserve setup authentication isolation from workflow pages.
- [x] Add bounded Playwright action and navigation timeouts.
- [x] Preserve first-attempt E2E failure evidence across CI retry.
- [x] Reject empty E2E selections before expensive environment setup.
- [x] Parallelize repository checks into bounded CI shards.
- [x] Isolate parallel E2E setup data across workers.
- [x] Remove the color contrast sweep bottleneck.
- [x] Update E2E contributor documentation.
- [x] Review product documentation impact; no user-facing product docs are affected.
- [x] Run the full repository tests and typechecks.
- [x] Run formatting, lint, package-boundary, generated-file, and fixture checks.
- [x] Complete final correctness and simplification review.

## Verification

- `pnpm --filter @webstudio-is/builder test`
- `pnpm -r typecheck`
- `pnpm lint`
- `pnpm check:package-boundaries`
- `pnpm check:generated-api`
- `pnpm check:generated-docs`
- `pnpm fixtures`
- Design-system: 222 tests pass locally.
- Builder: all seven Vitest projects pass locally.
- The formerly slow E2E shard passes locally in 1m55s.
- Full database-backed Builder E2E execution passes in CI.
- All CI jobs pass; the slowest completed in 4m58s.
- No fixture inputs/outputs or generated files changed.

Ref #6270


packages/design-system/package.json
```diff
@@ -83,8 +83,10 @@
     "./colors": {
       "webstudio": "./src/colors.ts"
     },
-    "./colors.css": "./src/colors/colors.css",
-    "./test-setup": "./src/test-setup.ts"
+    "./toast": {
```
kof: why did you add this entry point? that shouldn't be necessary

packages/design-system/package.json
```diff
@@ -83,8 +83,10 @@
...
+      "webstudio": "./src/components/toast.tsx"
+    },
+    "./colors.css": "./src/colors/colors.css"
```
kof, line 86: why did you add this entry point? that shouldn't be necessary

packages/sdk-components-react/src/time.browser.test.ts
```diff
@@ -150,13 +143,17 @@ test("hydrates visitor timezone without mismatch and updates after hydration", a
...
   document.body.appendChild(container);
 
-  process.env.TZ = "UTC";
   container.innerHTML = renderToStaticMarkup(ui);
   expect(container.textContent).toBe("17:45");
 
   const consoleError = vi.spyOn(console, "error").mockImplementation(() => {});
 
-  process.env.TZ = "Europe/Berlin";
+  vi.spyOn(Intl.DateTimeFormat.prototype, "resolvedOptions").mockReturnValue({
```
kof: why are we testing native Intl api exactly?

packages/design-system/package.json
kof: Removed it. The queue test now runs with the other browser tests and imports the normal Design System barrel, so a test-driven `./toast` public entry point is no longer needed.

packages/sdk-components-react/src/time.browser.test.ts
kof: This stubs the browser timezone boundary to test our `timeZone="visitor"` hydration/update behavior; it does not assert `Intl` itself. Browser tests cannot switch `process.env.TZ` like JSDOM did. I extracted it to `mockVisitorTimeZone()` so that intent is explicit and verified the tests fail when the boundary returns UTC.

packages/design-system/package.json
kof: `./colors.css` already existed on `main`; it only appeared changed because the adjacent exports changed. The package diff now leaves this entry point unchanged and removes the unnecessary `./toast` entry.

apps/builder/dev/backend.sh
```diff
@@ -25,7 +25,11 @@ builder_backend_migrations_fingerprint() {
         -type f \( -name migration.sql -o -name migration.ts \) -print
       printf '%s\n' \
         apps/builder/dev/backend.sh \
-        apps/builder/docker-compose.yaml
+        apps/builder/docker-compose.yaml \
+        apps/builder/docker-compose.e2e.yaml \
+        apps/builder/docker-compose.e2e-shard.yaml \
+        apps/builder/e2e/postgres.Dockerfile \
+        apps/builder/e2e/postgres-init.sql
```
kof, line 32: this entire list doesn't look very maintainable

PR #6307, opened 2026-09-01
feat: Make Content Block templates the MDX authoring interface

### LINKED ISSUE #6306: Resolve Content Block MDX through live Markdown and JSX templates
## Problem

Content Block MDX currently treats ordinary Markdown and explicit template references differently. Markdown-authored elements do not consistently inherit matching templates, while named custom instances require the verbose `<ws.element ws:name="…">` form. Arbitrary JSX component names are rejected instead of resolving to Content Block templates.

This prevents authors from using natural MDX while designers retain control of every rendered element through the Content Block's Templates subtree. It also means an authored element that was initially unresolved does not necessarily begin using a matching template when that template is added later.

## Desired outcome

Treat a Content Block's Templates subtree as the live rendering vocabulary for its connected MDX source:

- Standard Markdown syntax resolves to the matching semantic template whenever one is available.
- JSX such as `<Card />` resolves to the uniquely named `Card` template.
- Authored MDX remains the source of truth and is not rewritten merely because template availability changes.
- Template resolution is dynamic: adding a matching template later updates existing authored instances automatically.
- Missing or ambiguous JSX templates remain preserved in source and produce location-aware Builder and publication diagnostics.
- Content Blocks start with templates for every standard MDX element supported by the content engine.

## Product scope

- Support standard Markdown and constrained JSX as two syntaxes within the same MDX document.
- Define deterministic matching between Markdown node kinds and standard templates.
- Define deterministic matching between JSX component names and uniquely named custom templates.
- Preserve current restrictions against executable MDX, imports, spreads, and unsafe expressions.
- Keep resolved instances editable through the existing Content mode capabilities and serialize edits back to MDX.
- Re-resolve connected content when relevant templates are added, renamed, removed, or changed.
- Keep Builder canvas, source-editor diagnostics, MCP/CLI operations, published SSR, and static export consistent.
- Preserve unresolved source without coercing it to another element or deleting its subtree.
- Document the supported MDX syntax, template matching, diagnostics, and fallback behavior.

## Constraints

- Do not execute MDX or arbitrary JSX.
- Do not require generated IDs or write template bindings into the MDX source.
- Template matching must be deterministic and scoped to the owning Content Block.
- Existing connected MDX using `ws.element` must remain backward compatible.
- One MDX asset may be used by Content Blocks with different template sets; diagnostics and rendering must remain scoped to each block.
- Standard semantic and accessibility behavior must remain valid when no template exists.

## Acceptance criteria

- Markdown headings, paragraphs, lists, links, images, code, tables, and every other supported standard MDX element use their matching Content Block templates when present.
- `<Card />` uses the uniquely named `Card` template and supports valid Content-mode props and children.
- A missing or ambiguous JSX component produces an inline, source-ranged diagnostic and a Builder placeholder without deleting or rewriting its MDX.
- Adding a previously missing matching template causes existing Markdown or JSX instances to use it without changing the MDX file.
- Renaming or removing a template immediately re-resolves affected content and reports unresolved references.
- Default Content Blocks expose templates for every standard MDX element supported by the content engine.
- Builder, published rendering, MCP/CLI inspection, and serialization agree on resolution and diagnostics.
- Existing `<ws.element ws:name="…">` sources remain supported.
- Regression coverage includes parsing, resolution, late template addition, removal/rename, duplicate names, props, children, serialization, publication, and multi-block template scopes.
- User-facing documentation is reviewed and updated.

## Related

- Follow-up to #6124.

kof:
Closes #6306

## Outcome

Content Block templates are now the public component model for MDX authoring.

- Standard Markdown resolves through matching Content Block templates.
- PascalCase JSX resolves to a named template first, then to a registered component.
- Adding or changing a template immediately affects matching MDX, including content authored before that template existed.
- Every supported standard MDX element is present in new Content Blocks by default.
- Template `name` is a stable PascalCase JSX identifier. `label` remains optional display text only.
- Lowercase JSX represents native HTML and SVG without exposing `ws.element` in authored syntax.
- Unknown, ambiguous, and unsafe JSX stays in source, produces a source-ranged Builder diagnostic, and is omitted from published output.
- Frontmatter images preserve exact `$ref` values and use the Asset description for `alt` text.
- Keyboard and mouse insertion from the slash menu remove the trigger slash and preserve the selected template.

## Public MDX contract

- Markdown is canonical for standard prose.
- Lowercase JSX creates native elements: `<section><p>Text</p></section>`.
- PascalCase JSX uses templates and registered components: `<PromotionCard />`, `<Heading tag="h1">Title</Heading>`, and `<YouTube src="https://youtu.be/KI5JHpzBK1s" />`.
- Component-name collisions use stable qualified identifiers such as `Checkbox` and `RadixCheckbox`.
- A self-closing template keeps its designed default subtree. An explicitly paired empty template clears it.
- Static quoted values and bare booleans are supported.
- Imports, exports, expressions, spreads, functions, executable JavaScript, unsafe attributes, `<$.*>`, and authored `<ws.element>` are rejected.
- `ws` remains only for compiler primitives with no React component.

## Implementation

- [x] Add stable, unique template names and keep them separate from labels.
- [x] Parse and validate Markdown and JSX through their ASTs with precise source ranges.
- [x] Share component identity between template JSX, Content MDX, Builder materialization, reload, copy, and publication.
- [x] Resolve templates live after add, edit, rename, reorder, removal, undo/redo, collaboration updates, and reload.
- [x] Preserve authored props, children, Asset references, template defaults, and the exact slash-menu template choice.
- [x] Make one typed MDX semantic inventory drive default templates and slash-menu eligibility.
- [x] Move component-dependent template assembly into the component registry.
- [x] Replace production `$`, `animation`, and `radix` JSX namespaces with direct component imports.
- [x] Keep only explicit compiler primitives in `ws` template JSX.
- [x] Migrate repository templates, fixtures, stories, examples, MCP guidance, CLI guidance, and generated outputs.
- [x] Add Builder unit and Playwright coverage for authoring, diagnostics, insertion, synchronization, reload, and publication.
- [x] Review and update the authoritative Content Block and generated MCP documentation.

## Compatibility

- Persisted project data continues to use namespaced component IDs and `ws:element` for generic HTML elements. No project-data migration is required.
- Legacy `<ws.element ws:tag="…">`, `<ws.element ws:name="…">`, `<$.*>`, `<animation.*>`, and `<radix.*>` input remains readable during the compatibility window.
- Current serializers and repository-owned sources emit only canonical Markdown, lowercase intrinsic JSX, or PascalCase identifiers.
- Untouched user MDX is not rewritten automatically. Legacy syntax is canonicalized when its owning source is edited.
- Deferred follow-up: remove legacy namespace readers and proxy support only after the canonical writer is deployed and existing projects are backfilled. This is intentionally outside this PR.

## Verification

- [x] CI package tests, typechecks, lint, package boundaries, generated assets, migrations, and size checks pass.
- [x] All six Builder E2E shards pass.
- [x] Fixture checks pass on Ubuntu and Windows.
- [x] Targeted local Content Engine, Template, SDK, registry, Project Build, migration, Builder, component, CLI, and Playwright tests pass.
- [x] Runtime contracts, CLI docs, and MCP docs were regenerated from their owning sources.
- [x] Local fixture generation completed without an unexpected fixture diff.
- [x] Documentation impact reviewed and included in this PR.
- [x] Review and approve the intentional List visual snapshot change.
- [x] Run the complete agent evaluation suite after explicit approval.
- [x] Run the network fixture link/sync pipeline against the matching PR deployment after explicit permission.

## Known visual change

The generated List story now targets its actual default `ul` element instead of `ol`, so the default list styling appears in the visual snapshot. The visual change was reviewed and the approved visual check passes.


packages/cli/evaluations/high-impact/agent-runner.ts
```diff
@@ -103,6 +107,16 @@ const markdownBlogUploadInput = JSON.stringify({
   assetsDir: ".webstudio/assets",
 });
 
+const fontAssetsUploadInput = JSON.stringify({
```
kof: looks sus

packages/content-engine/src/mdx.ts
```diff
@@ -577,6 +608,9 @@ const setHastData = (node: SyntaxTreeNode, data: Record<string, unknown>) => {
   node.data = { ...(isRecord(node.data) ? node.data : {}), ...data };
 };
 
+export const isMdxTemplateComponentName = (name: unknown): name is string =>
+  typeof name === "string" && /^[A-Z][A-Za-z0-9_$]*$/.test(name);
```
kof: is that regex really necessary?

kof: Fixed in a014a477af. The handwritten JavaScript-identifier regex is gone. Template names now use `estree-util-is-identifier-name`, followed by the product's PascalCase first-character rule. Tests cover Unicode identifiers and invalid lowercase/symbol forms.

packages/cli/evaluations/high-impact/agent-runner.ts
kof: Fixed in a014a477af. I removed the broad recursive `/token/i` blacklist. The result is constructed from a fixed shape, and we now positively validate provider, model, tool, and check identifiers plus the artifact size. A regression test covers the valid `designTokenApplied` check name.

PR #6329, opened 2026-09-04
fix: Paste in Firefox navigator

### LINKED ISSUE #6288: Paste does not work in Firefox navigator
## Problem

Copy and paste in the Builder navigator does not work in Firefox. The user reports that copying succeeds, but pasting does nothing, both with keyboard shortcuts and the Webstudio context menu.

## Reproduction

1. Open a Webstudio project in Firefox.
2. Copy an element, text box, or slot from the navigator.
3. Try to paste it with `Ctrl+C / Ctrl+V` or the Webstudio context menu.

## Expected behavior

The copied item is pasted into the selected location.

## Actual behavior

Nothing is pasted. The same workflow worked a few weeks earlier. In Vivaldi, keyboard copy/paste works, while the context menu asks for browser permission before it works.

## User impact

Firefox users cannot duplicate content through copy and paste in the navigator. The existing duplicate action still works.

## Discord report

https://discord.com/channels/955905230107738152/1541321397823996015/1541446581281103914

anmoluiux:
## Summary

- Read every clipboard mime type before the first await in both paste handlers.
- Call the paste handler registered by the active mode from `emitPaste` instead of dispatching a synthetic clipboard event.
- Add regression coverage for both paths: a clipboard event that disconnects after dispatch, and a Firefox-shaped `ClipboardEvent`.

## Root cause

Two independent Firefox behaviors, one per entry point.

Keyboard paste: the plugin loop re-read `event.clipboardData` on every iteration. Firefox disconnects a paste event's `DataTransfer` as soon as the handler yields, so once `pageText` — first in the list, and async — awaited, every later plugin read an empty string and `instanceText` never received the payload. Chrome keeps the data readable across microtasks, which is why this only surfaced in Firefox. Introduced in #5798, which put `pageText` at the head of the plugin list.

Context menu paste: `emitPaste` dispatched `new ClipboardEvent("paste", { clipboardData })`. Gecko's `ClipboardEventInit` has no `clipboardData` member, so the event arrived with a non-null but empty `clipboardData`. `validateClipboardEvent` only rejects a null one, so the paste silently did nothing — no toast, no error.

Measured in Firefox 134 against Chrome and Edge 152: `getData` after one microtask returns the payload in Chromium and an empty string in Firefox; a synthetic event's `getData` likewise returns the payload in Chromium and an empty string in Firefox.

No behavior change in Chromium. Every plugin returns `pasteIgnored` before its first await, so reading all mime types up front yields the same input the loop received before.

## Verification

- `pnpm --filter @webstudio-is/builder exec vitest run app/shared/copy-paste app/builder/shared/asset-manager app/shared/commands-emitter.test.ts` — 335 tests passed.
- `pnpm --filter @webstudio-is/builder typecheck` — passed.
- `pnpm exec oxlint --deny-warnings apps/builder/app/shared/copy-paste/` — passed.
- `pnpm exec oxfmt --check apps/builder/app/shared/copy-paste/copy-paste.ts apps/builder/app/shared/copy-paste/copy-paste.test.ts` — passed.
- Reverted each fix in turn against the new tests: 13 failures for the keyboard path, 2 for the context menu, all passing again once restored.
- e2e runs Chromium only, and the previous test `DataTransfer` never disconnected, so neither path was covered before.

Closes #6288

### DIFF
--- a/apps/builder/app/shared/copy-paste/copy-paste.test.ts
+++ b/apps/builder/app/shared/copy-paste/copy-paste.test.ts
@@ -156,18 +156,6 @@ class TestDataTransfer {
   }
 }
 
-class TestClipboardEvent extends Event {
-  clipboardData?: DataTransfer;
-
-  constructor(
-    type: string,
-    options: EventInit & { clipboardData?: DataTransfer }
-  ) {
-    super(type, options);
-    this.clipboardData = options.clipboardData;
-  }
-}
-
 class FirefoxClipboardEvent extends Event {
   clipboardData = new TestDataTransfer() as unknown as DataTransfer;
 
@@ -1248,6 +1236,7 @@ test("copies selected instance to clipboard", async () => {
 
 test("copies multi-selected instances to clipboard and emits paste for every root", async () => {
   resetStores();
+  setupPage();
   const abortController = new AbortController();
   initCopyPaste({ signal: abortController.signal });
   let clipboardText = "";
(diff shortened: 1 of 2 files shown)


kof:
Merged, thank you!

Issue #4364, opened 2024-11-04
Universally link any part of builder


TrySound:
We need to support linking any part of builder. This will unlock following use cases
- sharing exact place between users
- links in error messages will point to exact place of error
- any other diagnostics
- command panel will be more generalized

We need to store following pieces
- instance selector
- style source id
- style state
- page id
- breakpoint id

Additionally, diagnostics will require
- style panel/settings panel/page settings panel
- style section/property name

PR #6342, opened 2026-09-08
feat: Copy links to instances from the context menu

### LINKED ISSUE #4610: Link to instance
As a builder I want to be able to share a link to a specific instance, not just a page.  For example sometimes I need to get help with a resource config and I can't link to that specific instance where resource is defined.

kof:
Right-click an instance in the Navigator and choose **Copy link to instance** to share its full Builder URL. Opening it restores the page and the same instance occurrence, including descendants of shared Slots. The shared menu also exposes the action on the canvas.

Copying and opening a link use the same existing `getDeepLinkedInstanceSelection` function, moved unchanged from `use-switch-page.ts` into `shared/instance-utils/link.ts`, alongside URL creation and the existing instance URL parser. It validates the instance path and page access. The context menu and pre-publish “Show element” links both call `getInstanceLink(instanceSelector)`. It lives in `shared/instance-utils/link.ts` and owns validation and the complete URL setup: project, page, full selector, origin, share token, editor mode, and safe mode, using the existing `builderUrl`. There is no separate link-permission implementation.

The link keeps the current share access and editor mode, clears unrelated scroll anchors, and shows clipboard success or failure. Body is supported; Global Root and multiple selections are unavailable.

Instance-only navigation now preserves loaded Builder data: the route comparison recognizes the `instance` parameter as well as the older `instanceId`. A functional regression test confirms that changing the target does not change the data URL comparison, while changing share access does.

MCP guidance explains how to extract the target ID from the link for `inspect-instance`, verify the project and shared Slot context, and avoid guessing when an instance no longer exists. This does not add a URL-taking MCP endpoint.

1. [x] Add the menu action and reuse complete instance-link construction with pre-publish links.
2. [x] Share existing instance-link validation between copying and opening links.
3. [x] Update the Navigator and MCP guidance; regenerate CLI documentation.
4. [x] Extend the existing functional test to round-trip a shared Slot URL, including share access and editor mode. Verify it fails without the instance parameter and passes after restoration.
5. [x] Remove the React/clipboard test harness and duplicate permission-test matrix. Confirm no fixture inputs or outputs are affected.

Validation: 28 focused Builder tests, Builder typecheck, repository lint, and `git diff --check` pass on this revision. The earlier CLI documentation checks passed and their inputs are unchanged. An earlier GPT-5.4 mini subagent at low reasoning passed a focused native MCP test against commit 072a8dc534: given only a link and text-edit request, it changed the correct heading among three identical-looking headings, leaving all other fixture records unchanged. It used four MCP calls with broad instance/text reads. That agent test used an isolated local fixture; it was not a production or shared-Slot agent test. Screenshots and the full agent evaluation suite were not run.

Closes #4610
Ref #4364


apps/builder/app/builder/shared/instance-context-menu.tsx
```diff
@@ -128,8 +129,34 @@ const getMenuPermissions = ({
   };
 };
 
+const getInstanceLink = ({
```
kof: seems like a duplicate, we must have this function somewhere else already

apps/builder/app/builder/shared/instance-context-menu.tsx
```diff
@@ -174,6 +201,32 @@ export const MenuItems = () => {
           />
         </ContextMenuItemRightSlot>
       </ContextMenuItem>
+      <ContextMenuItem
+        disabled={
+          selectedInstanceSelectors.length !== 1 ||
+          instanceSelector === undefined ||
+          instanceSelector[0] === ROOT_INSTANCE_ID ||
+          page === undefined
```
kof: should be a function and also check if we have it somewhere else too

apps/builder/app/builder/shared/instance-context-menu.tsx
kof: Removed getInstanceLink and reused the existing builderPath helper from shared/router-utils. The copied URL now uses the current page, selection, auth token, and editor mode, with the current Builder origin. Clipboard tests cover stale address-bar state, share access, Body, and shared Slot/MCP lookup. Pushed in 072a8dc534.

apps/builder/app/builder/shared/instance-context-menu.tsx
kof: Moved the condition into the existing getMenuPermissions function as canCopyLink. Both the menu item and its handler use that result. Added checks for Body, Global Root, multiple or missing selections, and a missing page across design, content, and preview modes. All 27 focused tests, Builder typecheck, and lint pass. Pushed in 072a8dc534.

kof:
Tested commit 072a8dc534 with a fresh GPT-5.4 mini subagent at low reasoning through native Webstudio MCP. Passed.

The agent received only the link and the requested heading text, without an ID-extraction recipe or access to repository source. The isolated fixture had three headings with identical labels and text across two pages. It used list-instances → list-texts → update-text → list-texts. The committed write targeted the linked instance, and an independent full before/after state comparison confirmed every other record was unchanged.

It succeeded with four MCP calls, but used broad lists rather than the documented direct inspect-instance lookup. This was a local fixture-backed MCP test, not a production or shared-Slot agent test. Two earlier attempts failed during harness setup; the successful run used corrected source-package loading and connector isolation. The full agent evaluation suite was not run.

kof:
Reworked the reuse in 141a8cbef6. Copying a link and opening it now call the same existing getDeepLinkedInstanceSelection function, moved unchanged into a shared utility. Removed the separate canCopyLink checks and used builderUrl directly, matching Share and pre-publish links.

Removed the 178-line React/clipboard test harness and the extra permission matrix. The existing shared Slot test now builds a URL and resolves it back to the same occurrence. All 18 focused tests, typecheck, and lint pass. This revision removes 191 lines overall.

